PR #27, opened 2018-11-26
added comment for alternative suggested by Jianlin

pruan-rht:


lib/environment.rb
```diff
@@ -372,6 +372,10 @@ def local_storage_provisioner_project
     end
 
     # return nil is no proxy enabled, else proxy value
+    # TODO: as alternative that's not OCP version dependent, but require parsing 
+    # YAML, we can do:
+    #  grep HTTP_PROXY /etc/origin/master/master-config.yaml
```
akostadinov: FYI master-config.yaml will be gone in 4.0

PR #28, opened 2018-11-27
Update the dockerfile for containerized tests running BushSlicer.

cjryan:
This update will allow a docker container to run bushslicer tests in ci-operator

### DIFF
--- a/tools/jenkins/slaves/Dockerfile.centos
+++ b/tools/jenkins/slaves/Dockerfile.centos
@@ -11,8 +11,6 @@ RUN [ -d $HOME/bushslicer/private/config/ca ] && \
     cp -v *.pem /etc/pki/ca-trust/source/anchors/ && \
     update-ca-trust extract || \
     exit 1 ) || :
-# Remove any instance of the private repository
-RUN rm -rf $HOME/bushslicer/private*
 
 RUN set -x && \
     SCL_BASE_PKGS="centos-release-scl scl-utils-build" && \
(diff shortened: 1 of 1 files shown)


tools/jenkins/slaves/Dockerfile.centos
```diff
@@ -11,6 +11,8 @@ RUN [ -d $HOME/bushslicer/private/config/ca ] && \
     cp -v *.pem /etc/pki/ca-trust/source/anchors/ && \
     update-ca-trust extract || \
     exit 1 ) || :
+# Remove any instance of the private repository
+RUN rm -rf $HOME/bushslicer/private*
```
akostadinov: now I remember we better also remove tierN

tools/jenkins/slaves/Dockerfile.centos
```diff
@@ -40,7 +42,7 @@ RUN scl enable rh-ror50 $HOME/bushslicer/tools/hack_bundle.rb
 
 RUN rpm -qa && \
     yum clean all -y && \
-    rm -rf $HOME/bushslicer /var/cache/yum /tmp/*
+    /var/cache/yum /tmp/*
```
akostadinov: where did `rm -rf` go? I think it makes more sense though to remove tierN and private here instead of earlier. Put things together should be fine IMO.

tools/jenkins/slaves/Dockerfile.centos
```diff
@@ -40,9 +38,10 @@ RUN scl enable rh-ror50 $HOME/bushslicer/tools/hack_bundle.rb
 #   security concern because of exposing VPN details
 # RUN sed -i -e 's/^\s\+curl\s/  curl -Sk /' /usr/local/bin/run-jnlp-client
 
+# Cleanup and remove any instance of the private repository
```
akostadinov: I htink we can remove this line, a little confusing on top of many things while it describes only one thing. Command is clear by itself IMO.

tools/jenkins/slaves/Dockerfile.centos
```diff
@@ -38,9 +38,10 @@ RUN scl enable rh-ror50 $HOME/bushslicer/tools/hack_bundle.rb
...
 RUN rpm -qa && \
     yum clean all -y && \
-    rm -rf $HOME/bushslicer /var/cache/yum /tmp/*
+    rm -rf /var/cache/yum /tmp/* $HOME/bushslicer/private* $HOME/bushslicer/tierN
```
akostadinov: `$HOME` needs to be quoted to make sure there are no spaces in it. Unlikely especially in this context, but better safe than sorry.

```
rm -rf /var/cache/yum /tmp/* "$HOME"/bushslicer/private* "$HOME"/bushslicer/tierN
```

I'm a little paranoid but I believe this one time is justified :innocent:

PR #29, opened 2018-11-27
allow multiple AWS configs

akostadinov:
Needs testing, do not merge yet.


akostadinov:
Tested with flexy build 53631.

PR #30, opened 2018-11-27
Init Dockerfiles for openshift-ci

droslean:
Initial Dockerfile to be used in the openshift-ci.

ref [#2213](https://github.com/openshift/release/pull/2213)

### DIFF
--- a/tools/openshift-ci/Dockerfile.centos
+++ b/tools/openshift-ci/Dockerfile.centos
@@ -1,13 +1,13 @@
 # The standard name for this image is aosqe/bushslicer-base
 
-FROM openshift/jenkins-slave-base-centos7
+FROM centos
 
 LABEL vendor="Red Hat inc."
 LABEL maintainer="AOS QE Team"
 
 USER root
 
-ADD . $HOME/bushslicer
+ADD . /bushslicer/
 
 RUN set -x && \
     SCL_BASE_PKGS="centos-release-scl scl-utils-build" && \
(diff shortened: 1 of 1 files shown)


akostadinov:
This can be closed now, correct @cjryan ?

droslean:
@akostadinov I am using that image for the tests. I think we can review it and merge it.

cjryan:
@akostadinov I believe so, yes. @droslean aside from leaving the bushslicer repo in the container and adding a WORKDIR [1], are there any major differences in the existing Dockerfile.centos with the file above that you'd need? I think it's ok to close this PR, if that's fine with you. Thanks!

[1] https://github.com/openshift/BushSlicer/blob/master/tools/jenkins/slaves/Dockerfile.centos#L58

droslean:
@akostadinov @cjryan The ci can't use this [image](https://github.com/openshift/BushSlicer/blob/master/tools/jenkins/slaves/Dockerfile.centos)

So that's why I created a new one. As you can notice here, this image is configured as a root user and not as `1001`. That is important otherwise our ci job won't be able to clone the repo in the image.

cjryan:
@droslean I think the dockerfile for the "upstream" image is:
https://github.com/openshift/jenkins/blob/master/slave-base/Dockerfile

cjryan:
@droslean  about the root/1001 user, I'm not entirely sure how that would work; in local ci-operator jobs, the 1001 user previously worked.

Happy to defer to the common consensus here though.

cjryan:
There's also been a request in a related ci-operator job to also re-base off of a vanilla centos image:
https://github.com/openshift/release/pull/2236#discussion_r237491117

droslean:
@cjryan there is no need to limit the container in users or give extra permissions in the folder/files that ci-operator needs. Also, the base image is based on `FROM openshift/origin`. 

Why the image that the tests will run needs to be base on the openshift/origin image?

cjryan:
@droslean you raise a great point; it's worth investigating if we can slim this down. I think the real issue is, we're using one dockerfile for 2 separate things - the current dockerfile in the repo is used as a jenkins slave image, so it is worthwhile to have the dependencies for jenkins jobs. However, now that we're investigating ci-operator use cases, we can eliminate the need for that jenkins-base-image and other dependencies, and can potentially eliminate bloat. It would be worth having 2 separate files here most likely, as the original intent of this PR was meant to do - one for jenkins jobs, one for ci-operator jobs. @akostadinov, what do you think?

droslean:
@cjryan @akostadinov I replace the base image with just `centos`. Still, the image is too big `1.9GB`.
I noticed in the script `install_os_deps.sh` that you install some, I can say, unnecessary stuff.  For example, all those python dependencies, or `firefox` or xorg server etc. 

All those are needed for the tests?

akostadinov:
@droslean , we support the jenkins Dockerfile because it is used for our jenkins slaves. In this way we support a single image instead of two - potentially breaking one of the use cases. This was wrt Java stuff.

I don't think we care about root vs 1001 because we run jenkins slaves as random users anyway, thus user selection can be whatever prow needs.

wrt size - these are not small tests. It is true that now we want to only run smoke tests but I understood idea is to run more if possible. We have lots of tests that require different things. Ruby and gem deps alone I guess account for a good size. Browsers, some command line tools are another thing.

One quick option to reduce size further is remove chrome installation (keep only firefox).

Would you explain why size of image is important? I think supporting a single Dockerfile is easier. Layer on top of pure centos vs jenkins image should be equal.

akostadinov:
Ok, I think final agreement is to remove chrome (leaving only firefox) as a minimal testing image. We can adjust going forward if needed.

akostadinov:
Thank you.

PR #31, opened 2018-11-28
enable metering route

pruan-rht:


features/step_definitions/metering.rb
```diff
@@ -125,3 +125,40 @@
     raise "report '#{report_type}' for project '#{namespace}' not found after #{seconds} seconds"
   end
 end
+
+Given /^I enable route for#{OPT_QUOTED} metering service$/ do | metering_name |
+  metering_name ||= "openshift-metering"
+  htpasswd = `htpasswd -nb -s #{user.name} #{user.password}`.strip
```
akostadinov: this will fail for token users (although I discourage such users usage anyway). Maybe use just some random user/pass and put into clipboard?

features/step_definitions/metering.rb
```diff
@@ -125,3 +125,40 @@
...
+  cookie_seed = `openssl rand -base64 32 | head -c32; echo`.strip
```
akostadinov: to get a random base64 string you can use `rand_str(length, :hex)`

features/step_definitions/metering.rb
akostadinov: now I see this is running local command which is not nice, need proper string generated by ruby

lib/openshift/route.rb
```diff
@@ -106,5 +106,13 @@ def dns(by: nil, user: nil, cached: true, quiet: false)
       return raw_resource(user: user, cached: cached, quiet: quiet).
         dig("spec", "host")
     end
+    # check for ingress route only
+    def ready?(user: nil, quiet: false)
```
akostadinov: cached can be true here like other methods. Often times you get object and call different methods. No need to treat `ready?` differently

lib/openshift/route.rb
```diff
@@ -106,5 +106,13 @@ def dns(by: nil, user: nil, cached: true, quiet: false)
...
+      status = raw_resource(user: user, cached: false, quiet: quiet)['status']
+      ready = status.dig('ingress').last['conditions'].any? do |con|
```
akostadinov: dig only needed for multiple elements, otherwise `['something']` works the same an is nicer. You may want to use `.&` instead of `.` though to avoid exceptions when `ingress` element is missing.

features/step_definitions/metering.rb
akostadinov: See format of passwd file [1]. Perhaps we can use OpenSSL gem to generate one of these. I can check this next week if you don't figure it out.

[1] https://httpd.apache.org/docs/trunk/misc/password_encryptions.html

akostadinov: This is how OpenSSh hashing can be used:
https://docs.ruby-lang.org/en/2.3.0/OpenSSL/HMAC.html

akostadinov: wow, now I notice there is an example how to create a SHA1 password in ruby in the apache docs file I wrote above, so it should be trivial to do

pruan-rht: Do you prefer the openssl approach or what's in the apache doc?

pruan-rht: @akostadinov, perhaps add this to `rand_str` method, although that would require adding an optional parameter for passing in the `password`, WDYT?

pruan-rht: on 2nd thought, that probably does not belong in that method.... perhaps a new one call encrypt?

pruan-rht:
@akostadinov, PR updated to address your comments.  PTAL.  Thanks!

lib/base_helper.rb
```diff
@@ -111,6 +113,9 @@ def rand_str(length=8, compat=:nospace_sane)
         return result
       end
 
+      def sh1_password(username: nil, password: nil)
```
akostadinov: can we name it sha1_htpasswd

pruan-rht: sure, done.

lib/base_helper.rb
```diff
@@ -111,6 +113,9 @@ def rand_str(length=8, compat=:nospace_sane)
         return result
       end
 
+      def sha1_password(username: nil, password: nil)
+        return "#{username}:{SHA}" + Base64.encode64(Digest::SHA1.digest(password)).strip()
+      end
```
akostadinov: Missed this first review, sorry. I'd only request this moved to `lib/ssl.rb`.

pruan-rht: This is not openssl related, so I'm not sure if it belongs in this module

pruan-rht: @akostadinov, I ran it again and don't see any `deprecated` warning from the [cli run](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/239/consoleFull).  Let me know if you insist on using OpenSSL instead of Digest ruby gem.

PR #32, opened 2018-11-28
added a class for metering CRD, and a post installation check to make…

pruan-rht:
… sure we have metering CRD.  Also, fix a minor logic issue.  TBH, I think we can assume the metering CRD exist if the metering expected metering pods are created successfully.


features/step_definitions/logging_metrics.rb
```diff
@@ -681,6 +681,7 @@
           # for metering we need to create the
           @result = user.cli_exec(:create, f: "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/metering/default-storageclass-values.yaml")
           step %Q/all metering related pods are running in the "#{target_proj}" project/
+          step %Q/I wait for the "openshift-metering" metering to appear/
```
akostadinov: does it make sense here to call `metering service has been installed successfully` instead of repeating the check in that other step?

akostadinov: ok, I see the step will also install metering if missing, ignore

PR #35, opened 2018-11-30
Log error is enough when failed to create pv

liangxia:
@akostadinov @pruan-rht Trying to add some negative coverage to persistent volume. Logger is enough, raise will fail the negative scenarios.


akostadinov:
But then all other scenarios that should fail? What happens to them? We need to create another step, shouldn't be hard to reuse one step in the other.

akostadinov:
@liangxia ?

liangxia:
@akostadinov  The other scenarios should still work. If they succeed, then no change/impact. If they fail, then fail 2 or 3 steps later, with fail info logged.

PR #37, opened 2018-12-04
Added bash script to run as part of aos-azure ci-operator.

cjryan:
This will allow for the replacement of the ci-operator entrypoint command.


droslean:
How much time all those test scenarios need to complete?

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -0,0 +1,187 @@
+#!/bin/sh
+
+export BUSHSLICER_DEFAULT_ENVIRONMENT="$TEST_ENVIRONMENT"
+export OPENSHIFT_ENV_$(echo "$TEST_ENVIRONMENT" | awk '{print toupper($0)}')_HOSTS="${TEST_CLUSTER}:etcd:master:node"
+export TESTENV="BUSHSLICER_V3"
+export BUSHSLICER_CONFIG=$TEST_CONFIG
+
```
kwoodson: I believe we need the MANAGER_USERS var:
```
export OPENSHIFT_ENV_ONLINE_STARTER_USER_MANAGER_USERS=":${TEST_TOKEN}"
```

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -0,0 +1,187 @@
+#!/bin/sh
+
+export BUSHSLICER_DEFAULT_ENVIRONMENT="${BUSHSLICER_TEST_ENVIRONMENT}"
+export OPENSHIFT_ENV_$(echo "${BUSHSLICER_TEST_ENVIRONMENT}" | awk '{print toupper($0)}')_HOSTS="${BUSHSLICER_TEST_CLUSTER}:etcd:master:node"
+export OPENSHIFT_ENV_$(echo "${BUSHSLICER_TEST_ENVIRONMENT}" | awk '{print toupper($0)}')_USER_MANAGER_USERS="${BUSHSLICER_TEST_TOKEN}"
```
kwoodson: Should this be `":${BUSHSLICER_TEST_TOKEN}"`

cjryan: good catch, thank you!

cjryan:
@droslean historically a run like this would take less than 2 hours. Just kicked off the first run of these tests, will know concretely shortly. Thanks!

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -0,0 +1,202 @@
+#!/bin/sh
```
akostadinov: please use `#!/bin/bash` because bash specific syntax is used.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -0,0 +1,202 @@
...
+  features/cli/configmap.feature:115 \
+  features/cli/configmap.feature:179 \
+  features/cli/event.feature:92 \
+  features/containers/containers.feature:7 \
+  features/cli/job.feature:192 \
+  features/networking/pod.feature:42 \
+  features/cli/policy.feature:408 \
+  features/images/jenkins.feature:382 \
+  features/web/configmap.feature:63 \
+  features/cli/oc_set_deployment_hook.feature:1
```
akostadinov: we need to generate this list on the fly based on case_id tag. Otherwise changes to the files can easily change lines of scenarios thus messing up the run completely.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -0,0 +1,202 @@
+#!/bin/sh
+
+[[ -z "$BUSHSLICER_TEST_ENVIRONMENT" ]] && { echo "BUSHSLICER_TEST_ENVIRONMENT not set"; exit 1; }
```
droslean: Can we get this from the KUBECONFIG?

akostadinov: This is actually a symbolic name. So we can hardcode it to `origin` probably.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -0,0 +1,202 @@
...
+[[ -z "$BUSHSLICER_TEST_CLUSTER" ]] && { echo "BUSHSLICER_TEST_CLUSTER not set"; exit 1; }
```
droslean: sry, I meant this env var. `BUSHSLICER_TEST_ENVIRONMENT`. Can we get this from `KUBECONFIG`?

droslean:
We also need to add the option in the script to update the authentication type of the cluster.
```console
oc patch clusterversions.config.openshift.io/version -p '{"spec":{"overrides": [{"kind": "Deployment","name": "openshift-cluster-kube-apiserver-operator","unmanaged": true}]}}' --type merge;
oc patch kubeapiserveroperatorconfig instance -p '{"spec":{"unsupportedConfigOverrides": {"oauthConfig": {"identityProviders": [{"challenge": true,"login": true,"name": "anypassword","provider": {"apiVersion": "v1","kind": "AllowAllPasswordIdentityProvider"}}],"masterCA": "/etc/kubernetes/static-pod-resources/configmaps/kubelet-serving-ca/ca-bundle.crt","masterPublicURL": "https://$cluster_url:6443","masterURL": "https://$cluster_url:6443"}}}}' --type merge;
         
```

tools/openshift-ci/verification_tests_ci_entrypoint.sh
akostadinov: btw for running tests we need cli tools. They will be missing in this image. Maybe we need to base on `origin` after all, not plain `centos`. Otherwise we need to install `oc` by ourselves.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
cjryan: test environment here is important - origin would do different things than say online_starter - for example, lack of ssh access. A hardcode here should probably be avoided

cjryan: The other option would be to set the other values (like web_console_url, admin_creds, hosts_type, etc) via env var if we do hardcode the value here

tools/openshift-ci/verification_tests_ci_entrypoint.sh
cjryan: Will investigate getting this value from KUBECONFIG

tools/openshift-ci/verification_tests_ci_entrypoint.sh
cjryan: agreed; spoke with @kwoodson about this very issue yesterday. We may want to add this functionality as a post-merge action, to automatically add testcases here once merged (if they're dedicated related tests)

tools/openshift-ci/verification_tests_ci_entrypoint.sh
droslean: Where origin-cli is based. Do we need to add the centos repositories, etc?

kargakis: Here is its [Dockerfile](https://github.com/openshift/origin/blob/master/images/cli/Dockerfile). It goes all the way back to [`origin-source`](https://github.com/openshift/origin/blob/master/images/source/Dockerfile) which is centos-based.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
akostadinov: @cjryan this is only symbolic name for the env. The actual question is about next line it seems.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
akostadinov: @cjryan what do you mena "post-merge action"? What I request is to call a script before running the tests that will generate this command line instead of having these file:num hardcoded. The script would be based on hardcoded list of test case IDs, that we (QE) will update regularly.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
stevekuznetsov: Appropriate thing to do is to build your test image in the `ci-operator` config for your repo, and add in the `oc` from origin-cli using image source

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+export TESTENV="BUSHSLICER_V3"
+export BUSHSLICER_CONFIG="${BUSHSLICER_TEST_CONFIG}"
+unset BUSHSLICER_DEBUG_AFTER_FAIL
+
+# Generate the cucumber file list from the case id lookup table
+# set Internal Field Separator for the file to the newline char
+IFS=$'\n'
+for i in $(cat < "case_id_lookup_table.txt"); do
+  COMMAND_STRING=$COMMAND_STRING$(echo " ";echo $i | awk -F '|' '{print $2}')
+done
```
akostadinov: this should all a script that will search feature files for `case_id` tags and generate the file:lineno

cjryan: Now that this PR has merged, we can probably add in the logic here: https://github.com/openshift/verification-tests/pull/52

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+
+export BUSHSLICER_DEFAULT_ENVIRONMENT="$BUSHSLICER_TEST_ENVIRONMENT"
+[[ -z "$BUSHSLICER_DEFAULT_ENVIRONMENT" ]] && { echo "BUSHSLICER_DEFAULT_ENVIRONMENT not set"; exit 1; }
+
+export OPENSHIFT_ENV_$(echo "$BUSHSLICER_TEST_ENVIRONMENT" | awk '{print toupper($0)}')_HOSTS="${BUSHSLICER_TEST_CLUSTER}:etcd:master:node"
+export OPENSHIFT_ENV_$(echo "$BUSHSLICER_TEST_ENVIRONMENT" | awk '{print toupper($0)}')_USER_MANAGER_USERS=:"${BUSHSLICER_TEST_TOKEN}"
+export OPENSHIFT_ENV_$(echo "$BUSHSLICER_TEST_ENVIRONMENT" | awk '{print toupper($0)}')_WEB_CONSOLE_URL=https://${BUSHSLICER_TEST_CLUSTER}/console
+
+export TESTENV="BUSHSLICER_V3"
+export BUSHSLICER_CONFIG="${BUSHSLICER_TEST_CONFIG}"
```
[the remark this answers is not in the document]
cjryan: This is where we would put in our config overrides, for example, marionette driver, etc. This is being supplied higher up in the ci operator, which is why it's being set as an ENV var here.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
akostadinov: yes

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+export BUSHSLICER_DEFAULT_ENVIRONMENT="$BUSHSLICER_TEST_ENVIRONMENT"
+[[ -z "$BUSHSLICER_DEFAULT_ENVIRONMENT" ]] && { echo "BUSHSLICER_DEFAULT_ENVIRONMENT not set"; exit 1; }
+
+export OPENSHIFT_ENV_$(echo "$BUSHSLICER_TEST_ENVIRONMENT" | awk '{print toupper($0)}')_HOSTS="${BUSHSLICER_TEST_CLUSTER}:etcd:master:node"
+export OPENSHIFT_ENV_$(echo "$BUSHSLICER_TEST_ENVIRONMENT" | awk '{print toupper($0)}')_USER_MANAGER_USERS=:"${BUSHSLICER_TEST_TOKEN}"
+export OPENSHIFT_ENV_$(echo "$BUSHSLICER_TEST_ENVIRONMENT" | awk '{print toupper($0)}')_WEB_CONSOLE_URL=https://${BUSHSLICER_TEST_CLUSTER}/console
+
+export TESTENV="BUSHSLICER_V3"
```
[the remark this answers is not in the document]
cjryan: This is quite possibly a legacy variable that I've carried over. If it's not necessary, we can delete it.

tools/openshift-ci/case_id_lookup_table.txt
```diff
@@ -0,0 +1,218 @@
+OCP-14582|features/admin/scheduler.feature:5
+OCP-12121|features/build/buildconfig.feature:5
+OCP-10667|features/build/buildconfig.feature:26
```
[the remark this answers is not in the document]
cjryan: Will update this file in accordance with the newly merged https://github.com/openshift/verification-tests/pull/52 , thanks

tools/openshift-ci/verification_tests_ci_entrypoint.sh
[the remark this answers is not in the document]
akostadinov: ok, please delete

tools/openshift-ci/case_id_lookup_table.txt
```diff
@@ -0,0 +1,33 @@
+OCP-9567
```
akostadinov: we spoke about file name like `cases_ded_4_0.txt`. Something that will show cluster type and version as selection depends on these.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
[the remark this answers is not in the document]
akostadinov: why don't we rename variable in template instead of copying variables here? In any case it can live forthe time being, but I fear nobody goes back to fix.

tools/openshift-ci/cases_ded_azure_3_11.txt
```diff
@@ -0,0 +1,33 @@
+OCP-9567
```
akostadinov: Shouldn't file name be cases_ded_azure_4_0.txt ? Why filter 3.11 cases?

tools/openshift-ci/verification_tests_ci_entrypoint.sh
akostadinov: @cjryan , seems like `oc` binary was handled by #49. 

The other thing is generating HOSTS spec. Do we need to do anything more for generating hosts here?

kwoodson:
@mgahagan73, This was the work that Chris was working on.  Mind taking a look?

akostadinov:
Yes! I want to wrap that up but want to make sure any existing use case is properly handled.

akostadinov:
I think nobody is using master branch ATM so I'm going to merge and will accommodate any existing use cases when needed.

PR #38, opened 2018-12-05
update webauto libs and remove ureliable ff legacy support

akostadinov:
Seeingsome failures in smoke runs related to webdriver-browser communication. I think updating web gems can help this. Also ff legacy is unreliable and we don't use it anymore. It doesn't work with the new gems.


akostadinov:
I have tested locally with `features/tierN/web/hooks.feature:224`. Want to hear any possible objections before I rebuild jenkins slave images.

wjiangjay:
LGTM

PR #41, opened 2018-12-07
Rename to consistent with repo name

liangxia:
I hope we will not do repo rename any more.

@akostadinov @pruan-rht @openshift/team-qe  

I'm trying to test the changes, but it requires runner-v3 parameters changes at the same time. So I skip the test/change. Maybe do that after a basic review is done here ?

How these changes are made.
```bash
# find all ruby files and replace BushSlicer to VerificationTests
Files=`find . -name "*.rb" -exec grep -l -- 'BushSlicer' {} \;`
sed -i 's/BushSlicer/VerificationTests/g' $Files
git diff
git add .

# all BushSlicer:: replace with VerificationTests::
Files=`grep -lr -- 'BushSlicer::' *`
sed -i 's/BushSlicer::/VerificationTests::/g' $Files
git diff
git add .

# all BUSHSLICER replaced with VERIFICATION_TESTS
Files=`grep -lr BUSHSLICER *`
sed -i 's/BUSHSLICER/VERIFICATION_TESTS/g' $Files
git diff
git add .

# remaining BushSlicer to verification-tests
Files=`grep -lr -- 'BushSlicer' *`
sed -i 's/BushSlicer/verification-tests/g' $Files
git diff
git add .

# remaining bushslicer...
vim `grep -rli BushSlicer * `
git diff
git add .
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -7,8 +7,8 @@ More info you can find in link:doc/overview.adoc[Overview document].
 === Create local dev environment
 
 . Install git if not installed already, install ruby 2.4+ if not installed
-. git clone git@github.com:openshift/bushslicer.git
-. cd bushslicer
+. git clone git@github.com:openshift/verification-tests.git
+. cd verification-tests
 . tools/install_os_deps.sh # need sudo; you may need to login again to terminal if ruby was installed via rvm during this phase
 . tools/hack_bundle.rb # normal user
 . install OKD or OpenShift Container Platform client tools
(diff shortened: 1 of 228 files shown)


pruan-rht:
According to Clayton's original comment, it was only the public repo name he objects to, he is ok with the Bushslicer reference within the codes.  I would vote for no change and reject this PR.

akostadinov:
My bigger concern is that other teams also rely on these variables. As you said also Jenkins and prow CI jobs (although prow scripts should be part of this repo so maybe not an issue).

If we are going to change we need to discuss what exactly to change to. VERIFICATION_TESTS looks too long to me. I hoped we can have some kind of real project name. VerificationTests as a module looks too generic.

liangxia:
How about shifthat ?

liangxia:
So is everyone happy with `VTESTS` ?

akostadinov:
Maybe OSVTests (OpenShift verification tests) or OKDTests? I'm voting for OKDTests.

@smarterclayton maybe you can chime in as well to say how we should name our module?

smarterclayton:
It’s fine to call them openshift tests.  I don’t have a strong preference

liangxia:
No longer needed. Closing.

PR #45, opened 2018-12-11
replace set_pair_value by set_textarea_value

xiaocwan:
This is only for 4.0
@yapei Please review, thanks.

### DIFF
--- a/lib/rules/web/admin_console/4.0/secrets.xyaml
+++ b/lib/rules/web/admin_console/4.0/secrets.xyaml
@@ -40,8 +40,8 @@ create_generic_secret:
     if_param: key 
     ref: set_pair_key
   action:
-    if_param: value
-    ref: set_pair_value
+    if_param: textarea_value
+    ref: set_textarea_value
   action:
     if_param: one_more
     ref: click_add_key_value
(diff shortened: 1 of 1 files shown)


lib/rules/web/admin_console/4.0/secrets.xyaml
```diff
@@ -40,8 +40,8 @@ create_generic_secret:
     if_param: key 
     ref: set_pair_key
   action:
-    if_param: value
-    ref: set_pair_value
+    if_param: textarea_value
+    ref: set_textarea_value
```
yapei: To not break your tests for `OCP-20800`, you also need update `set_textarea_value` rules to always send keys to the last `textarea`

xiaocwan: @yapei Never mind, I checked the features and OCP-20800 is the only case needs update, also there is another rule `click_one_dropdown_action` updating a param. So soner or later the OCP-20800 needs to patch a fix PR, so I update this one together this time, don't worry there's no big change.

yapei: ok for me

akostadinov: @yapei so you want PR merged?

yapei: @akostadinov  no problem, let's merge it

PR #48, opened 2018-12-13
rescheuled-report and report functionality has been combined into report

pruan-rht:


features/step_definitions/metering.rb
```diff
@@ -42,41 +42,52 @@
...
   end
+  opts[:start_time] ||= default_start_time
+  opts[:end_time] ||= default_end_time
+  opts[:run_immediately] = to_bool(opts[:run_immediately])
+  opts[:grace_period] ||= nil
+  opts[:schedule] ||= opts[:schedule]
+  opts[:format] ||= default_format
+  opts[:metadata_name] ||= name
+
+  unless opts[:use_existing_report] == 'true'
```
akostadinov: can't we detect whether report already exists automatically? Or is the idea that we don't want to use report that already exists by mistake. i.e. it already exists but we do not expect it to exist.

features/step_definitions/metering.rb
```diff
@@ -42,41 +42,52 @@
...
   # report object confirm it's ready to be used, before querying, we need to enable proxy on the host
   host.exec('oc proxy', background: true)
   step %Q/I perform the :view_metering_report rest request with:/, table(%{
     | project_name  | #{project.name}  |
     | name          | #{name}          |
-    | report_format | #{report_format} |
+    | report_format | #{opts[:format]} |
     })
-  if report_format != 'json' and report_format != 'yaml'
+  binding.pry
```
akostadinov: pry

lib/openshift/flakes/conditions.rb
```diff
@@ -0,0 +1,40 @@
...
+  class Conditions
+
+    attr_reader :struct
+    private :struct
+
+    def initialize(struct)
+      @struct = struct
+    end
+
+    module ExportMethods
```
akostadinov: I don't think we need this pattern for classes. For modules that we want to include or use methods from them directly it makes sense. But I don't see why we need to have `ExportMethods` module inside the class.

lib/openshift/flakes/conditions.rb
```diff
@@ -0,0 +1,40 @@
+module BushSlicer
+
+  # this class should help with parsing report conditions
+  class Conditions
```
akostadinov: this class appears to represent a single condition so a name `Condition` (without `s`) would make more sense. If it is specific for a particular object type, then it might be called `ReportCondition`. But it looks like generic so if I am not mistaken `Condition` should be fine.

lib/openshift/report.rb
```diff
@@ -1,46 +1,69 @@
...
+    end
+
+    def running?(user: nil, cached: false, quiet: false)
+      # the reason 'Scheduled' is an initial state what we
+      conditions.any? { |c| c.type == 'Running' and c.reason != 'Scheduled' }
     end
 
     # wait until the stauts of the Report becomes 'Finished'
-    def wait_till_finished(user: nil, quiet: false)
+    def wait_till_finished(user: nil, cached: false, quiet: false)
```
akostadinov: in a wait method the `cached` argument does not make sense to me

lib/openshift/report.rb
```diff
@@ -1,46 +1,69 @@
...
+      seconds = 60 # for PVs it can take longer than 30s
+      success = wait_for(seconds) do
+        finished?(user:user, cached: cached, quiet: quiet)
```
akostadinov: cached should be hardcoded to `false` here. Otherwise we can enter an infinite loop (at least until timeout is reached)

akostadinov: same applies to wait_till_running

lib/openshift/report.rb
```diff
@@ -1,46 +1,69 @@
...
-    # b. reportingStart
-    # c. reportingEnd
-    # d. runImmediately
-    def self.generate_yaml(query_type: self.name, start_time: nil, end_time: nil, run_now: "true")
-      # span the whole year if user didn't give a range
-      start_time ||= "#{Time.now.year}" + "-01-01T00:00:00Z"
-      end_time ||= "#{Time.now.year}" + "-12-30T23:59:59Z"
-      run_immediately ||= "true"
+    def finished?(user: nil, cached: false, quiet: false)
+      ## TODO: make this backwardcompatible
```
akostadinov: backward compatible to what?

features/step_definitions/metering.rb
pruan-rht: There two different types of reports

1. run-once report, with the parameter `runImmediately` set, which is a one and done. 
2. scheduled report which is generated from a range of time in a set interval specified by the user

The report resource created persists in the namespace.  This option is to give the user an option to reuse the object...in short, yes, we could detect the report already there.  We could change the behavior to default to delete the existing object unless the existing report is a scheduled report that has not reached the end of life.  WDYT?

lib/openshift/report.rb
pruan-rht: For older version of `operator-metering` the finished condition represented by a field `phase` which is gone for newer version.  It's just a nice to have, but since `metering` is still categorized as `beta`, backward compatibility is probably low priority.

pruan-rht:
@akostadinov, all the comments have been addressed with the new patch.  PTAL.  Thanks!

features/step_definitions/metering.rb
akostadinov: automatincally handling this makes more sense to me. I'm frustrated by too many params.

pruan-rht: @akostadinov there are situations in which we want to query using the same report object to save time (https://github.com/openshift/cucushift/blob/master/metering/reports.feature#L16), so I think we can leave it as is, which is if user does not specify the option `user_existing_report`, it will generate a new report object.

PR #49, opened 2018-12-13
Change base to centos.

droslean:
We already copy the oc client in the runtime. There is no need to base on `origin-cli`

https://github.com/openshift/release/blob/master/ci-operator/templates/openshift/installer/cluster-launch-installer-src.yaml#L64-L72
https://github.com/openshift/release/blob/master/ci-operator/templates/openshift/installer/cluster-launch-installer-src.yaml#L106

### DIFF
--- a/tools/openshift-ci/Dockerfile.centos
+++ b/tools/openshift-ci/Dockerfile.centos
@@ -1,11 +1,11 @@
-FROM openshift/origin-cli
+FROM centos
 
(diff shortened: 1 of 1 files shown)


akostadinov:
How do we install cli binaries? If it works with these changes I can merge. But have you tested it?

akostadinov:
I see description now. Ok.

PR #51, opened 2018-12-15
added installation of metering using shell script that comes with the…

pruan-rht:
… official operator-metering repo

### DIFF
--- a/features/step_definitions/admin.rb
+++ b/features/step_definitions/admin.rb
@@ -23,7 +23,7 @@ Given /^a secret is created for admin kubeconfig in current project$/ do
   end
 end
 
-Given /^I save the project name hosting #{QUOTED} resource named #{QUOTED} to#{OPT_QUOTED} clipboard/ do |resource, res_name, cb_name|
+Given /^I save the project hosting #{QUOTED} resource named #{QUOTED} to#{OPT_QUOTED} clipboard/ do |resource, res_name, cb_name|
   ensure_admin_tagged
 
   cb_name ||= :namespace
@@ -36,7 +36,7 @@ Given /^I save the project name hosting #{QUOTED} resource named #{QUOTED} to#{O
       logger.info("Can't find resource matching name '#{res_name} in any projects")
       cb[cb_name] = nil
     else
-      cb[cb_name] = res.first['metadata']['namespace']
+      cb[cb_name] = project(res.first['metadata']['namespace'])
     end
   end
 end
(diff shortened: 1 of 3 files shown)


features/step_definitions/admin.rb
```diff
@@ -22,3 +22,22 @@
...
   end
 end
+
+Given /^I save the project name hosting #{QUOTED} resource named #{QUOTED} to#{OPT_QUOTED} clipboard/ do |resource, res_name, cb_name|
+  ensure_admin_tagged
+
+  cb_name ||= :namespace
+  @result = admin.cli_exec(:get, all_namespaces: true, resource: resource, o: 'yaml')
+  if @result[:parsed].nil?
+    cb[cb_name] = nil
```
akostadinov: this looks like a big issue and we better fail. At least output a warning like on line 36

features/step_definitions/admin.rb
```diff
@@ -22,3 +22,22 @@
...
+  else
+    res = @result[:parsed]['items'].select { |i| i['metadata']['name']==res_name }
+    if res.count.zero?
+      logger.info("Can't find resource matching name '#{res_name} in any projects")
+      cb[cb_name] = nil
+    else
+      cb[cb_name] = res.first['metadata']['namespace']
```
akostadinov: Why not set current project to that one?

akostadinov: Other than that LGTM.

pruan-rht:
@akostadinov PR updated by adding info

features/step_definitions/admin.rb
akostadinov: In other steps, when we say we do something with pods, services, etc, we also set current object as the last used one. Such that one can use step `I delete ... in the project` instead of specifying name. That's why my suggestion is to call `cb[cb_name] = project(res.first['metadata']['namespace'])`. That will set current project to the found project as well we have the project object in the clipboard which is usually more useful than string name.
Or are current use cases obstructed by such arrangement?

features/step_definitions/admin.rb
```diff
@@ -22,3 +22,23 @@
     raise "cannot create secret, see log"
   end
 end
+
+Given /^I save the project name hosting #{QUOTED} resource named #{QUOTED} to#{OPT_QUOTED} clipboard/ do |resource, res_name, cb_name|
```
akostadinov: if you agree with me about project vs project name `s/name//`

features/step_definitions/admin.rb
pruan-rht: @akostadinov, oh I misunderstood your original comment.  I've updated the PR now to store the project object instead of the project name string.  PTAL.  Thanks!

PR #52, opened 2018-12-19
Updated case_id_splitter utility to return a simple list of filenames…

cjryan:
… and line numbers given a list of caseids.

given a file exists with caseids on each newline, like: 
```
cat file.txt
OCP-12121
OCP-11543
OCP-12606
```
Then the following command can be used to perform a verification-test run using cucumber:
```
cucumber -p junit $(./tools/case_id_splitter.rb list_features $(<file.txt))
```

### DIFF
--- a/tools/case_id_splitter.rb
+++ b/tools/case_id_splitter.rb
@@ -153,9 +153,9 @@ module BushSlicer
     # @param hash of location data, with the key as the case_id, and the value as the file
     # location and line numbers, separately.
     private def generate_feature_list(cases_loc_hash)
-      complete_feature_list = ""
+      complete_feature_list = []
       cases_loc_hash.each do |key, value|
-        complete_feature_list = complete_feature_list + "#{value[0]}:#{value[1].to_s} "
+        complete_feature_list << "#{value[0]}:#{value[1].to_s}"
       end
       return complete_feature_list
     end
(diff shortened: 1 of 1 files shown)


tools/case_id_splitter.rb
```diff
@@ -39,6 +39,23 @@ def run
         end
       end
 
+      command :list_features do |c|
```
akostadinov: I'd name this `:"file-line"` and then see next line comment

tools/case_id_splitter.rb
```diff
@@ -39,6 +39,23 @@ def run
...
+        c.syntax = "#{$0} list_features [options]"
```
akostadinov: `"#{$0} list_features CASE-123 ... [options]"`

tools/case_id_splitter.rb
```diff
@@ -39,6 +39,23 @@ def run
         end
       end
 
+      command :file-line do |c|
```
akostadinov: does this work? I thought it needs to be quoted like in my comment

tools/case_id_splitter.rb
```diff
@@ -39,6 +39,23 @@ def run
...
+        c.syntax = "#{$0} file-line [options]"
```
akostadinov: "#{$0} file-line CASE-123 ... [options]"

PR #53, opened 2018-12-19
Fail ealier if package missing

liangxia:
@akostadinov @pruan-rht 

Idea coming from "[aos-devel] Dockerfile owners: this is better than rpm -V"


akostadinov:
This is very good. There are conflicts though. Also could you update centos dockerfile as well?

liangxia:
Rebased, and centos dockerfile updated.

pruan-rht:
Lgtm

PR #56, opened 2018-12-28
Make csi driver deployment as a case so that can be run as CVP

weherdh:
@liangxia @chao007 Please help to take a look. Thanks

### DIFF
--- a/features/storage/csi.feature
+++ b/features/storage/csi.feature
@@ -1,0 +1,35 @@
+Feature: CSI testing related feature
+
+  # @author wehe@redhat.com
+  # @case_id OCP-21804
+  @admin
+  @destructive
+  Scenario: Deploy a cinder csi driver and test 
+    Given I deploy "cinder" driver using csi
+    And I create storage class for "cinder" csi driver
+    And I checked "cinder" csi driver is running
+    And I have a project
+    When I create a dynamic pvc from "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/storage/misc/pvc.json" replacing paths:
+      | ["metadata"]["name"]                                                   | pvc-<%= project.name %> |
+      | ["metadata"]["annotations"]["volume.beta.kubernetes.io/storage-class"] | cinder-csi              |
+    Then the step should succeed
+    And the "pvc-<%= project.name %>" PVC becomes :bound
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/storage/misc/pod.yaml" replacing paths:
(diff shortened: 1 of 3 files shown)


chao007:
Is it better to move your new update to ../feature/storage ？

liangxia:
@weherdh  Files in folder cluster_configuration is used to configure the clusters, so scenarios for specific cases should not be placed here.
Also for CSI, driver pod is running is not enough to validate the scenario. We should at least create one pvc/pod from the CSI storage class.

weherdh:
Hope this basic test process can pass from CVP...

features/storage/csi.feature
```diff
@@ -0,0 +1,35 @@
+Feature: CSI testing related feature
+
+  # @author wehe@redhat.com
+  # @case_id OCP-21804
+  @admin
+  @destructive
+  Scenario: Deploy a cinder csi driver and test 
+    Given I deploy "cinder" driver using csi
+    And I create storage class for "cinder" csi driver
+    And I checked "cinder" csi driver is running
```
akostadinov: the same code exists in "Enable cinder csi driver and create storage class for cinder driver" scenarion in `features/cluster_configuration/csi.feature`. It appears that this performs some cluster configuration and does **not** revert it back after scenarion.
My expectation is that first `Enable cinder csi driver and create storage class for cinder driver` would be run on a cluster and then any scenarios that need "cinder csi driver" will be run. Otherwise a second run of scenario that tries to install cinder csi would fail.

Please let me know if I miss something.

akostadinov: either you put in cluster-configration and use as part of cluster setup. Or you need to add clean-up logic to `I deploy "cinder" driver using csi`. It should be added automatically, not separate step. is it hard?

akostadinov: bump

akostadinov: This still relevant?

weherdh: @akostadinov Sorry, so busy these days, our 4.0 automation nearly crashed...without ssh access to nodes, a lots of cases have failed. I am totally forget this PR for long time...For CVP testing, this is not supposed to run more than twice? I am not sure, if it really need a cleanup logic, I will try to figure out how. BTW, Alex, for the new 4.0 regression automation test failed, where should I contact you to discuss the solutions? Open a JIRA issue to track?

akostadinov: jira, irc, email, I take test execution issues as a very high prio

weherdh: Added a registered cleanup-up step.

PR #58, opened 2019-01-03
refactor metering steps

pruan-rht:


lib/openshift/report.rb
```diff
@@ -10,7 +10,17 @@ class Report < ProjectResource
...
+        rr.map { |cond| Condition.new cond }
+      else
+        # on occasions in which metering is just installed and we call to
+        # generate report immediately, it sometime won't be ready, put in a
+        # short sleep and try again
+        sleep 10
+        logger.info("*** retry ****")
+        rr = raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'conditions')
+        rr.map { |cond| Condition.new cond }
+      end
```
akostadinov: instead of implementing such logic, can we have a step `I wait until ... report is ready` or something

pruan-rht: @akostadinov, I've took out that hack and added some changes to the expose route portion (which didn't work due to indentation issue).  PTAL.  Thanks!

akostadinov: ack, the only change I'd request is to return an empty array `[]` instead of `false`. `false` is a very different data type than what method usually returns.

akostadinov: you can check for `conditions.empty?` if you want to know whether we have such or not.

pruan-rht: @akostadinov updated.  PTAL.  Thanks!

PR #61, opened 2019-01-04
fix usage with URL vars file

akostadinov:


akostadinov:
tested to work, merging to avoid CI failures

PR #66, opened 2019-01-09
auto OCP-19667 Check Builds

yapei:
Rule files for #6900 in cucushift


lib/rules/web/admin_console/4.0/common_ui_elements.xyaml
```diff
@@ -62,6 +62,14 @@ uncheck_checkbox:
     selector:
       xpath: //input[contains(@type,'checkbox')]
     op: click
+check_link_and_text:
+  element:
+    selector:
+      xpath: //a[@class='co-resource-link__resource-name' and text()='<text>' and @href='<link_url>']
```
yanpzhan: Do we need use parameter for @class? eg @class='classname'

yapei: I think we don't, even I'm thinking if we need remove `@class='co-resource-link__resource-name'` from xpath expression, it turns out to check only an `//a` element containing desired `text` and `link`, WDYT?

yanpzhan: Agree, class is not needed.

yapei: Updated

lib/rules/web/admin_console/4.0/common_ui_elements.xyaml
```diff
@@ -62,6 +62,14 @@ uncheck_checkbox:
     selector:
       xpath: //input[contains(@type,'checkbox')]
     op: click
+check_link_and_text:
+  element:
+    selector:
+      xpath: //a[text()='<text>' and @href='<link_url>']
+check_link:
```
yapei: This will be removed to reduce line of rule files

yanpzhan:
LGTM now, @pruan-rht , could you pls help to merge the pr if no objection?

PR #73, opened 2019-01-17
selectable floating IP pools

akostadinov:
Now one can optionally select network for floating IPs.


akostadinov:
tested internally

PR #77, opened 2019-01-23
Fixes for some outputs compatible issues

stuartchuan:
Fixes for some outputs compatible issues.


pruan-rht:
LGTM

akostadinov:
`get` client command always better to replace with method calls. But since this is an improvement from before I will merge.

PR #78, opened 2019-01-25
added support to install metering using OLM

pruan-rht:


features/step_definitions/metering.rb
```diff
@@ -225,3 +224,47 @@
...
 end
+
+# install metering via OLM,
+Given /^the#{OPT_QUOTED} metering service is installed(?: to $QUOTED)? using OLM$/ do | metering_ns |
+  ensure_admin_tagged
+  # 1. create the metering namespace
+  metering_ns ||= "openshift-metering"
+  project(metering_ns)
+  step %Q/I switch to cluster admin pseudo user/
+  @result = user.cli_exec(:create_namespace, name: metering_ns)
```
akostadinov: you don't need to switch to cluster admin. You can do `admin.cli_exec(...)`.

features/step_definitions/metering.rb
```diff
@@ -225,3 +224,47 @@
...
+
+  step %Q/a pod becomes ready with labels:/, table(%{
+    | olm.catalogSource=metering |
+  })
+  @result = user.cli_exec(:apply, f: metering_crd, n: project.name)
+  raise "OLM install of metering failed" unless @result[:success]
+end
+
+# XXX: currently OLM uninstall is TBD, we uninstall by removing the namespace
+Given /^the#{OPT_QUOTED} metering service is uninstalled using OLM$/ do | metering_ns |
```
akostadinov: I don't see this step registered in cleanup. Or did I miss that?

pruan-rht: @akostadinov, the general usage from `metering` is that you install it once and run the various test.  We shouldn't be installing and uninstalling it for every test feature.  If we want to uninstall `metering`, the user should call it explicitly

features/step_definitions/metering.rb
pruan-rht: @akostadinov, but for step on line 250, the user needs to be admin, so I think we can just do it here.

features/step_definitions/metering.rb
```diff
@@ -225,3 +224,47 @@
...
+Given /^the#{OPT_QUOTED} metering service is installed(?: to $QUOTED)? using OLM$/ do | metering_ns |
+  ensure_admin_tagged
+  # 1. create the metering namespace
+  metering_ns ||= "openshift-metering"
+  project(metering_ns)
+  step %Q/I switch to cluster admin pseudo user/
+  @result = user.cli_exec(:create_namespace, name: metering_ns)
+  # prep for OLM need to define configs via YAML
+  # 1. metering-catalogsourceconfig.yaml
+  catalog_source_config = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/metering/OLM_examples/metering-catalogsourceconfig.yaml"
```
yapei: Hi @pruan-rht  Why don't we use test files from `https://github.com/operator-framework/operator-metering/tree/master/manifests`?

pruan-rht: @yapei because the OLM install is still very experimental, so I don't want to put those config into the official metering github repo.

pruan-rht:
@akostadinov, bump....

features/step_definitions/metering.rb
akostadinov: It is preferable to have things pre-installed on cluster install instead of whenever a test runs. If we keep it like this, then making step `ensure_destructive_tagged` is needed to avoid installing this at the same time by two scenarios.

Another issue is in case you want at some point to insall with a different configuration. How would you handle it?

features/step_definitions/metering.rb
akostadinov: It is preferable to have things pre-installed on cluster install instead of whenever a test runs. If we keep it like this, then making step ensure_destructive_tagged is needed to avoid installing this at the same time by two scenarios.

Another issue is in case you want at some point to insall with a different configuration. How would you handle it?

pruan-rht: good point, I'll add that tag.  WRT the configuration...the OLM install is still in its infancy, so this step will definitely need to be modified.  How about default to these for now and perhaps add a table option later to let user overwrite the defaults?

akostadinov: we need to make sure to tidy things up after GA

pruan-rht: ok, will do.  TBH, metering tests are generally run after the initial installation is done unless the testcase is installation specific.

pruan-rht: @akostadinov PR updated with tags and added more sanity check.  PTAL.  Thanks!

features/step_definitions/metering.rb
```diff
@@ -261,10 +265,27 @@
...
 end
+
+Given /^all datasources for#{OPT_QUOTED} project are imported$/ do | proj_name |
+  project ||= project(proj_name) if proj_name
+  data_sources  = BushSlicer::ReportDataSource.list(user: user, project: project)
+  seconds = 60
+  import_list = []
+  success = wait_for(seconds) {
+    data_sources.each do |ds|
+      import_time = ds.props.dig(:raw, 'status', 'prometheusMetricImportStatus', 'earliestImportedMetricTime')
```
akostadinov: extract this as a DataSource method

features/step_definitions/metering.rb
```diff
@@ -261,10 +265,27 @@
...
+
   metering_ns ||= "openshift-metering"
   step %Q/I switch to cluster admin pseudo user/
   project("openshift-marketplace")
   step %Q/I ensure "metering" catalogsourceconfig is deleted/
   project(metering_ns)
   step %Q/I ensure "#{metering_ns}" project is deleted/
 end
+
+Given /^all datasources for#{OPT_QUOTED} project are imported$/ do | proj_name |
```
akostadinov: maybe `report datasources` will speak better

pruan-rht: the alias that we call from `oc` is `datasources`...even though the YAML report `ReportDataSources`.  Perhaps I can talk with Chance about that.

features/step_definitions/metering.rb
```diff
@@ -225,3 +226,66 @@
...
+  project ||= project(proj_name) if proj_name
+  data_sources  = BushSlicer::ReportDataSource.list(user: user, project: project)
+  seconds = 60
+  import_list = []
+  success = wait_for(seconds) {
+    data_sources.each do |ds|
+      import_time = ds.props.dig(:raw, 'status', 'prometheusMetricImportStatus', 'earliestImportedMetricTime')
+      import_list << import_time.nil?
+    end
+    import_list.count(false) == 0
```
akostadinov: bt wyou can replace whole body like:
```
data_sources.any? { |ds| ds.last_imported }
```

pruan-rht:
@akostadinov, PR updated.  PTAL.  Thanks!

features/step_definitions/metering.rb
```diff
@@ -50,7 +50,7 @@
   step %Q/all metering related pods are running in the "#{cb.metering_namespace.name}" project/
   step %Q/I wait for the "#{metering_name}" metering to appear/
   # added check for datasource to make sure promethues imports are working
-  step %Q/all datasources for "<%= cb.metering_namespace.name %>" project are imported/
+  step %Q/all reportdatasources are importing from Prometheus/
```
akostadinov: Didn't we agree on a neutral name here? Also why one word `reportdatasources`?

akostadinov: anyway, merging, will let you make nicer if you wish

pruan-rht: I found out that datasources is actually a shorthand for reportdatasources.  I decided to use the full name to be more explicit.

PR #81, opened 2019-02-12
selectable floating IP pools

akostadinov:


akostadinov:
tested with install build 56157

PR #83, opened 2019-02-15
support admin kubeconfig from URL

akostadinov:


akostadinov:
@xxia FYI BUSHSLICER_CONFIG can be:
```
environments:
  ocp4:
    admin_creds_spec: "https://example.com/job/Launch%20Environment%20Flexy/56402/artifact/workdir/cucushift-oc40-standard-slave-k26jp-0/install-dir/auth/kubeconfig"
```

For local runs it can be easier:
```
export OPENSHIFT_ENV_OCP4_ADMIN_CREDS_SPEC="https://example.com/job/Launch%20Environment%20Flexy/56402/artifact/workdir/cucushift-oc40-standard-slave-k26jp-0/install-dir/auth/kubeconfig"
```

xingxingxia:
Test passes, thank you :+1:

lib/admin_credentials.rb
```diff
@@ -43,22 +61,32 @@ def get
...
+
+  class AutoKubeconfigCredentials < AdminCredentials
+    def get
+      case opts[:spec]
+      when nil
+        MasterOsAdminCredentials.new(env, **opts).get
+      when %r{://}
+        URLKubeconfigCredentials.new(env, **opts).get
+      else
+        raise "unknown gredentials specification: #{opts[:spec]}"
```
pruan-rht: minor typo should be `credentials` I think...

xingxingxia:
I added short usage of the url spec and :bastion role in the description stanza for Runner-v3 params HOSTS and BUSHSLICER_CONFIG respectively so auto builders can know.

lib/admin_credentials.rb
wjiangjay: typo from Dvorak keyboard layout?

akostadinov: Not really, more likely just being tired when typing that.

wjiangjay: I make this guess as in Dvorak keyboard layout, "g" is the neighbor of "c". And thanks for make this happen for admin tag cases

PR #84, opened 2019-02-20
fix oapi, which has become obsolete in 4.0

pruan-rht:
`/oapi` is obsoleted in 4.0 . Anything newer than 3.5 should use the groupified resources `/apis/user.openshift.io/v1/users/~`

### DIFF
--- a/lib/rest_openshift.rb
+++ b/lib/rest_openshift.rb
@@ -6,7 +6,7 @@ module BushSlicer
       extend Helper
 
       def self.populate(path, base_opts, opts)
-        populate_common("/oapi/<oapi_version>", path, base_opts, opts)
+        populate_common("/apis/user.openshift.io/<oapi_version>", path, base_opts, opts)
       end
 
       class << self
@@ -52,11 +52,22 @@ module BushSlicer
       end
 
       def self.get_user(base_opts, opts)
+        base_opts_org = base_opts.dup
         populate("/users/<username>", base_opts, opts)
-        return perform(**base_opts, method: "GET") { |res|
+        res = perform(**base_opts, method: "GET") { |res|
           res[:props][:name] = res[:parsed]["metadata"]["name"]
           res[:props][:uid] = res[:parsed]["metadata"]["uid"]
         }
+        unless res[:success]
+          # failed with new url, fall back to old url path
+          populate_common("/oapi/<oapi_version>", "/users/<username>", base_opts_org, opts)
+          res = perform(**base_opts_org, method: "GET") { |res|
+            res[:props][:name] = res[:parsed]["metadata"]["name"]
+            res[:props][:uid] = res[:parsed]["metadata"]["uid"]
+          }
+        end
+
+        return res
       end
 
       # this usually creates a project in fact


pruan-rht:
FYI `/oapi` was removed about a month ago: [openshift/origin#21782](https://github.com/openshift/origin/pull/21782) . Also, I've tested the new path it works up until 3.6, before which we have to use the `/oapi` path

wjiangjay:
Actually we still have some hotfix will be backported to the version under 3.4, for this part, how to deal with?

pruan-rht:
@wjiangjay, can you specify what kind of hotfixes you are referring to?  IMO 3.4 is barely touched now so we can consider that separately.

lib/rest_openshift.rb
```diff
@@ -6,7 +6,7 @@ module OpenShift
       extend Helper
 
       def self.populate(path, base_opts, opts)
-        populate_common("/oapi/<oapi_version>", path, base_opts, opts)
+        populate_common("/apis/user.openshift.io/<oapi_version>", path, base_opts, opts)
```
akostadinov: `::populate` method is used by all calls. And I believe `user.openshift.io` is only for the user subsystem and will not work for `::list_projects` for example.

I think that we need to drop this openshift/kube api calls separation because now openshift calls are not a separate endpoint, just extensions to kube API. And we can handle all calls in same way.

akostadinov:
@pruan-rht , @wjiangjay , maybe we can create a branch that will be dedicated to 3.5 and older so that we are not held back by such old versions. WDYT?

akostadinov:
Or shall we create a 3.x branch so that we design things according to 4.x in master from now on? Or do we still think expansion on 3.11 test cases is still ongoing?

@openshift/team-qe , opinions?

pruan-rht:
@akostadinov I vote for 4.0 branch since it's so drastically different from 3.x.  This way we don't need to monkey-patch all of these

akostadinov:
My vote is the same, but now we make `v3` branch again and keep master at `v4`. Easier to create pull requesrts on the command line. I still type `-b v3` sometimes in the hurry :)

pruan-rht:
@akostadinov I'm fine with using `v4` == `master`

wabouhamad:
@akostadinov having master at v4 sounds good to me

lib/rest_openshift.rb
pruan-rht: @akostadinov do you mean get ride of `lib/rest_openshift.rb` and merge it into `lib/rest_kubernetes.rb`

wjiangjay:
@pruan-rht @akostadinov The hotfix often means some CVEs(this need smoke or acceptance testing of versions from 3.1 to 3.11).

I am also fine with master is for v4

pruan-rht:
@wjiangjay if that's the case, then it's better IMO to create a separate branch for `3.x` so we don't interrupt CVE testing.

pruan-rht:
closing this PR in favor of https://github.com/openshift/verification-tests/pull/88

PR #86, opened 2019-02-20
generating host list with flags and small fixes

akostadinov:


pruan-rht:
LGTM.  I'll merge it.

PR #87, opened 2019-02-24
fix issue with rest request step because User needs to be APIAccessor…

pruan-rht:
… instead of User ojbect

### DIFF
--- a/features/step_definitions/common.rb
+++ b/features/step_definitions/common.rb
@@ -19,7 +19,7 @@ Then /^the step should have timed out$/ do
 end
 
 When /^I perform the :([a-z_]*?) rest request$/ do |yaml_key|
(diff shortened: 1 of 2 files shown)


pruan-rht:
this applies to `v3` branch

features/step_definitions/common.rb
```diff
@@ -19,7 +19,7 @@
 end
 
 When /^I perform the :([a-z_]*?) rest request$/ do |yaml_key|
-  @result = env.rest_request_executor.exec(user: user, req: yaml_key.to_sym)
+  @result = user.rest_request(yaml_key.to_sym)
```
akostadinov: good

akostadinov: sehr gut

pruan-rht: @akostadinov...had to do Google Translate this...now I know a German phrase. :)

PR #88, opened 2019-02-25
get rest api to work with 4.0 api structure, also took out oapi refer…

pruan-rht:
…ences since it has become obsolete.   According to the [upstream documentation](https://docs.okd.io/latest/rest_api/) there are still differences between `k8s` and `openshift` in term of base URL.  Introduce `type` to use to build the url to fetch from.

### DIFF
--- a/lib/rest_openshift.rb
+++ b/lib/rest_openshift.rb
@@ -5,8 +5,8 @@ module BushSlicer
     module OpenShift
       extend Helper
 
-      def self.populate(path, base_opts, opts)
-        populate_common("/oapi/<oapi_version>", path, base_opts, opts)
+      def self.populate(path, base_opts, opts, type)
+        populate_common("/apis/#{type}.openshift.io/<api_version>", path, base_opts, opts)
       end
 
       class << self
@@ -21,7 +21,7 @@ module BushSlicer
       #   "gitVersion": "v3.3.0.27",
       #   "buildDate": "2016-08-29T14:44:33Z"
       # }
-      def self.version(base_opts, opts)
+      def self.version(base_opts, opts, type="version")
         populate_common("/version", "/openshift", base_opts, opts)
         return perform(**base_opts, method: "GET") { |res|
           res[:props][:openshift] = res[:parsed]["gitVersion"]
@@ -31,28 +31,28 @@ module BushSlicer
         }
       end
 
-      def self.delete_oauthaccesstoken(base_opts, opts)
-        populate("/oauthaccesstokens/<token_to_delete>", base_opts, opts)
+      def self.delete_oauthaccesstoken(base_opts, opts, type="oauth")
+        populate("/oauthaccesstokens/<token_to_delete>", base_opts, opts, type)
         return perform(**base_opts, method: "DELETE")
       end
 
-      def self.list_projects(base_opts, opts)
-        populate("/projects", base_opts, opts)
+      def self.list_projects(base_opts, opts, type="project")
+        populate("/projects", base_opts, opts, type)
         return perform(**base_opts, method: "GET")
       end
 
-      def self.delete_project(base_opts, opts)
-        populate("/projects/<project_name>", base_opts, opts)
+      def self.delete_project(base_opts, opts, type="project")
+        populate("/projects/<project_name>", base_opts, opts, type)
         return perform(**base_opts, method: "DELETE")
       end
 
-      def self.get_project(base_opts, opts)
-        populate("/projects/<project_name>", base_opts, opts)
+      def self.get_project(base_opts, opts, type="project")
+        populate("/projects/<project_name>", base_opts, opts, type)
         return perform(**base_opts, method: "GET")
       end
 
-      def self.get_user(base_opts, opts)
-        populate("/users/<username>", base_opts, opts)
+      def self.get_user(base_opts, opts, type="user")
+        populate("/users/<username>", base_opts, opts, type)
         return perform(**base_opts, method: "GET") { |res|
           res[:props][:name] = res[:parsed]["metadata"]["name"]
           res[:props][:uid] = res[:parsed]["metadata"]["uid"]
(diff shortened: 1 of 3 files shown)


pruan-rht:
Please note that this fix applies to `v4` (`master`) branch.  Another separate fix is in https://github.com/openshift/verification-tests/pull/87

akostadinov:
I have created a v3 branch but already merged your other PR. And it was filed against `master`. So now there might be a little mess.

pruan-rht:
@akostadinov thanks...I think it should be ok, since we need to maintain both v3 and v4 per your discussion.  #87 should still apply to `master` branch.  Let me just file another one for the `v3` branch

pruan-rht:
@akostadinov ok, I've resolved the conflicts

lib/rest_openshift.rb
```diff
@@ -5,8 +5,8 @@ module Rest
     module OpenShift
       extend Helper
 
-      def self.populate(path, base_opts, opts)
-        populate_common("/oapi/<oapi_version>", path, base_opts, opts)
+      def self.populate(path, base_opts, opts, type)
+        populate_common("/apis/#{type}.openshift.io/<api_version>", path, base_opts, opts)
```
akostadinov: I thought that having one `populate` method makes more sense now. i.e. move the populate method to `rest.rb` and work for all APIs. as `type` the whole `something.openshift.io` or `something.k8s.io` can be specified.

But in the kube version I see `#{path}.k8s.io` vs `#{type}.openshift.io` here. Why is this difference? Why for k8s we need a separate parameter? Would be much nicer if we can unify these so that any new APIs added in the future can be supported.

akostadinov: In k8s path and type always match while in OpenShift - not?

lib/rest_openshift.rb
```diff
@@ -31,75 +31,73 @@ def self.version(base_opts, opts)
...
-      def self.delete_oauthaccesstoken(base_opts, opts)
-        populate("/oauthaccesstokens/<token_to_delete>", base_opts, opts)
+      def self.delete_oauthaccesstoken(base_opts, opts, type="oauth")
+        populate("/oauthaccesstokens/<token_to_delete>", base_opts, opts, type)
         return perform(**base_opts, method: "DELETE")
       end
 
-      def self.list_projects(base_opts, opts)
-        populate("/projects", base_opts, opts)
+      def self.list_projects(base_opts, opts, type="project")
```
akostadinov: I think that `type` here should not be a parameter but hardcoded when calling `populate` method. Why would we ever want to specify a different type when calling `list_projects`?

lib/rest_openshift.rb
pruan-rht: That's a mistake, I'll fix that.

lib/rest_openshift.rb
pruan-rht: yeah, we can put it here or at the `populate` method as you suggested

lib/rest_openshift.rb
pruan-rht: @akostadinov, so the choice is having user to put just the `type` vs `something.openshift.io` or `something.k8s.io`.  I was going for the shorter string the user needs to write.  Also, the separation of `openshift` and `k8s`.  It's kind of a grey area...I just went with the shorter things to write.

pruan-rht: @akostadinov, I've updated the PR...although I'm sticking with the separate `populate` method to save some writing.

akostadinov: `populate` method is only an internal helper so not much user writing stuff. I was thinking for merging k8s and OpenShift calls in same rules and use the full identifyer such that if we need to call something that isn't `k8s.io` or `openshift.io` we can just add without adding additional rules files.

Let me know your thought about this. Maybe it's unlikely to have such new APIs as well perhaps not a bad idea to have rules separation.

lib/rest_openshift.rb
akostadinov: Ok, please remove this parameter `type` from here as well other individual calls. `populate` looks good.

See my other comment. I would leave it to you about merging k8s and openshift calls in same place.

pruan-rht: @akostadinov ok, I fixed that now.

lib/rest_openshift.rb
pruan-rht: I'm in favor of keep it separate if it's okay with you.

lib/rest_openshift.rb
```diff
@@ -110,42 +108,26 @@ def self.rollback_deploy(base_opts, opts)
...
+        populate("/namespaces/<project_name>/deploymentconfigrollbacks", base_opts, opts, type="apps")
         return Http.request(**base_opts, method: "POST")
       end
 
-      def self.get_subresources_oapi(base_opts, opts)
-        populate("/namespaces/<project_name>/<resource_type>/<resource_name>/status", base_opts, opts)
-        return Http.request(**base_opts, method: "GET")
-      end
-
-      def self.post_role_oapi(base_opts, opts)
```
stuartchuan: @pruan-rht any replacement for this method?

pruan-rht: @stuartchuan, for 4.0, oapi has been removed, that's why I took out this method

lib/rest_openshift.rb
akostadinov: Using `api_version` here is not correct. In fact all these `something.openshift.io` APIs may use different API version. But since it is less broken than before, we can merge as is to get something working. We need more investigation to figure out what might be an approach that makes sense.

/lgtm

pruan-rht: What do you mean by `may use different API version`?  Shouldn't they all be using the same api version?

akostadinov: well initial assumpion was that there will be one global API version like 1, 2, 3. Now they have multiple `something.k8s.io` and `somehting.openshift.io` APIs and each of them can have a different independent API level. Having a single parameter for this will not work.

On the other hand it can be overridden for each call so maybe will work. As long as it works now, I'm fine.

PR #91, opened 2019-02-27
Ocp 19810

anuragthehatter:
Gherkin Scenario for SDN test case **OCP-19810**. 
The aim of the test is to make sure conntrack rule for UDP traffic should be removed when the pod for NodePort service is deleted.

A pod on schedule able host node is created under a project (via a json template) which is listening to udp traffic. The pod will be exposed on node ip via a nodeport with protocol UDP . A client pod will send udp traffic to that node ip and exposed port to make sure that a conntrack entry is created. When the host pod gets deleted, rc will create a similar new pod with udp listener. We send udp traffic to new pod and check the conntrack entries to make sure deleted pod entry should not exist in conntrack table but the new pod one.  

Already discussed with Bo Meng offline. Made some changes as per his suggestions and hence opening PR finally.

**Passed Logs for reference**: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/248/console

### DIFF
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -203,13 +203,12 @@ Feature: Pod related networking scenarios
    Then the step should succeed
    Given a pod becomes ready with labels:
      | name=udp-pods |
-   And evaluation of `pod.name` is stored in the :host_pod1_name clipboard
-   And evaluation of `pod.ip` is stored in the :host_pod1_ip clipboard
+   And evaluation of `pod` is stored in the :host_pod1 clipboard
    And evaluation of `host_subnet(cb.nodes[0].name).ip` is stored in the :node_ip clipboard
    
    When I run the :expose client command with:
      | resource      | pod                      |
-     | resource_name | <%= cb.host_pod1_name %> |
+     | resource_name | <%= cb.host_pod1.name %> |
      | type          | NodePort                 |
      | port          | 8080                     |
      | protocol      | UDP                      |
@@ -226,13 +225,12 @@ Feature: Pod related networking scenarios
    Then the step should succeed
    Given a pod becomes ready with labels:
      | name=hello-pod |
-   And evaluation of `pod.name` is stored in the :client_pod clipboard
-   And evaluation of `host_subnet(cb.nodes[1].name).ip` is stored in the :client_ip clipboard
+   And evaluation of `pod` is stored in the :client_pod clipboard
    
    # 'yes' command will send a character "h" continously for 3 seconds to /dev/udp on listener where the node is listening for udp traffic on exposed nodeport. The 3 seconds mechanism will create an Assured
    #  entry which will give us enough time to validate upcoming steps 
    When I run the :exec background client command with:
-     | pod              | <%= cb.client_pod %>                                  |
+     | pod              | <%= cb.client_pod.name %>                             |
      | oc_opts_end      |                                                       | 
      | exec_command     | bash                                                  |
      | exec_command_arg | -c                                                    |
@@ -244,19 +242,19 @@ Feature: Pod related networking scenarios
      | conntrack -L \| grep <%= cb.nodeport %> | 
    Then the step should succeed 
    And the output should contain: 
-     |<%= cb.host_pod1_ip %>|
+     |<%= cb.host_pod1.ip %>|
    
    When I run the :delete client command with:
      | object_type       | pod                      |
-     | object_name_or_id | <%= cb.host_pod1_name %> | 
+     | object_name_or_id | <%= cb.host_pod1.name %> | 
    Given a pod becomes ready with labels:
      | name=udp-pods |
-   And evaluation of `pod.ip` is stored in the :host_pod2_ip clipboard
+   And evaluation of `pod` is stored in the :host_pod2 clipboard
    
    # 'yes' command will send a character "h" continously for 3 seconds to /dev/udp on listener where the node is listening for udp traffic on exposed nodeport. The 3 seconds mechanism will create an Assured
    #  entry which will give us enough time to validate upcoming steps 
    When I run the :exec background client command with:
-     | pod              | <%= cb.client_pod %>                                  |
+     | pod              | <%= cb.client_pod.name %>                             |
      | oc_opts_end      |                                                       |
      | exec_command     | bash                                                  |
      | exec_command_arg | -c                                                    |
@@ -266,6 +264,6 @@ Feature: Pod related networking scenarios
    
    When I run commands on the host:
      | conntrack -L \| grep <%= cb.nodeport %> |
-   Then the output should contain "<%= cb.host_pod2_ip %>"
-   And the output should not contain "<%= cb.host_pod1_ip %>"    
+   Then the output should contain "<%= cb.host_pod2.ip %>"
+   And the output should not contain "<%= cb.host_pod1.ip %>"    
 
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -194,6 +194,7 @@ Feature: Pod related networking scenarios
   # @case_id OCP-19810
   @admin
   Scenario: Conntrack rule for UDP traffic should be removed when the pod for NodePort service deleted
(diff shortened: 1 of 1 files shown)


anuragthehatter:
@bmeng @zhaozhanqi

zhaozhanqi:
@anuragthehatter 
Great. BTW, could you help try if this can be also passed on 4.0 env.

anuragthehatter:
> @anuragthehatter
> Great. BTW, could you help try if this can be also passed on 4.0 env.
Sure @zhaozhanqi

features/networking/pod.feature
```diff
@@ -190,3 +190,82 @@ Feature: Pod related networking scenarios
...
+   And evaluation of `host_subnet(cb.nodes[1].name).ip` is stored in the :client_ip clipboard
+   
+   # 'yes' command will send a character "h" continously for 3 seconds to /dev/udp on listener where the node is listening for udp traffic on exposed nodeport. The 3 seconds mechanism will create an Assured
+   #  entry which will give us enough time to validate upcoming steps 
+   When I run the :exec background client command with:
+     | pod              | <%= cb.client_pod %> |
+     | oc_opts_end      |    |
+     | exec_command     | bash |
+     | exec_command_arg | -c |
+     | exec_command_arg | yes "h">/dev/udp/<%= cb.node_ip %>/<%= cb.nodeport %> |
```
akostadinov: I don't think `bash` supports making connections in Red Hat linux.

In any case, as Peter said, this will not work with internal IPs. You can probably test one node by running some command on another node.

Additionally relying on local shell commands is veeery discouraged. Either use pure ruby code, run commands in pods, or ocmmands on nodes. In that order of preference.

features/networking/pod.feature
```diff
@@ -190,3 +190,82 @@ Feature: Pod related networking scenarios
...
+   Given I store the schedulable nodes in the :nodes clipboard
+   Given I use the "<%= cb.nodes[0].name %>" node
+   And I have a project
+   
+   When I run the :create client command with:
+     | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/pod_with_udp_port_listener.json |
+   Then the step should succeed
+   Given a pod becomes ready with labels:
+     | name=udp-pods |
+   And evaluation of `pod.name` is stored in the :host_pod1_name clipboard
```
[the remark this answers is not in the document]
anuragthehatter: Sure @pruan-rht that make sense. Thanks!

features/networking/pod.feature
```diff
@@ -190,3 +190,82 @@ Feature: Pod related networking scenarios
...
+   And evaluation of `pod.ip` is stored in the :host_pod1_ip clipboard
+   And evaluation of `host_subnet(cb.nodes[0].name).ip` is stored in the :node_ip clipboard
+   
+   When I run the :expose client command with:
+     | resource          | pod |
```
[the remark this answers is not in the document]
anuragthehatter: Will do, Thanks!

features/networking/pod.feature
```diff
@@ -190,3 +190,82 @@ Feature: Pod related networking scenarios
     Then the step should succeed
     And the output should not contain "ingress_policing_rate: 1953"
 
+  # @author anusaxen@redhat.com
+  # @case_id OCP-19810
```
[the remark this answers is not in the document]
anuragthehatter: > I doubt that as it's written will work with 4.0 since it's accessing the hosts...which in 4.0 the external IP is deprecated.

I agree, this might need some tweak to support 4.0. Will look into that

features/networking/pod.feature
```diff
@@ -190,3 +190,82 @@ Feature: Pod related networking scenarios
...
+   
+   When I run the :expose client command with:
+     | resource          | pod |
+     | resource_name     | <%= cb.host_pod1_name %> |
+     | type              | NodePort |
+     | port              | 8080 |
+     | protocol          | UDP |
+   Then the step should succeed
+   
+   When I run the :get client command with:
```
pruan-rht: it's preferable to use ruby-code and class support instead of calling cli directly.  You can replace line 128-222 by one line like 
`And evaluation of `service(pod.name).node_port(port: 8080)` is stored in the :nodeport clipboard` or just call

anuragthehatter: Thanks Peter for suggestion. Yep seems like i can also do this way. I will check on that

features/networking/pod.feature
anuragthehatter: I agree Alex. For 4.0, this would need some tweak.
Regarding Bash support in RHEL, i am executing this command on pod. 
So when i did this:
oc exec pod_name -- sh -c yes "h">/dev/udp/node_ip/udp_port
**error**: cant find /dev/udp/node_ip/udp_port
And when i did "bash" above it took it. I guess the reusable "hello-pod" docker image might have built on "FROM docker.io/centos:7"" version which supports sh but conntrack entries showed that connections were honored at the listener side (RHEL) using bash

The pod, as per my experiment didn't support "sh" but "bash". I was in favor of "sh" though. I am not sure if sh is favoured then in that case i can create a dockerfile/image using "FROM docker.io/centos:7"

anuragthehatter: reusable "hello-pod" docker image might **NOT** have

features/networking/pod.feature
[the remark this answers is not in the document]
bmeng: Hi @pruan-rht 
What is the criteria that we should put the case in public or tierN?

features/networking/pod.feature
bmeng: Yeah, the aos-pod-for-ping image is built on top of alpine, which may not be functional as centos.
But as the bash is executed inside the container and we can get the expected result.
I am not sure what is the problem though.

features/networking/pod.feature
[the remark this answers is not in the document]
pruan-rht: @bmeng all private tests should be under there according to the [README](https://github.com/openshift/verification-tests/blob/3506c0863277bc75445526ea92062e052c8cd84a/README.adoc#running-tests-from-outside-this-repository)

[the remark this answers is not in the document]
bmeng: Ok. Thanks.

I do not think this case should be private testing. Let's keep it here.

features/networking/pod.feature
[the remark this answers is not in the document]
anuragthehatter: Done.

features/networking/pod.feature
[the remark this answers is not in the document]
anuragthehatter: Done

features/networking/pod.feature
pruan-rht: Since this test won't work for OCP >= 4.0, perhaps we can add a statement to fail early in the OCP version does not meet the criteria

anuragthehatter: Sounds good to me, Peter. Testing is going on in a parallel though on 4.0 but no luck yet. Investigation is on..

pruan-rht: @anuragthehatter, ok, that's why I suggested to gate the testcase by checking on the ocp version of the current master.  You can also change Polarion to specify that only OCP <= 3.11 applies to this test case.

anuragthehatter: Thanks for Info,Peter. I will discuss with @bmeng today eve and will conclude the appropriate step

anuragthehatter: @pruan-rht I am going to gate the test case by checking 'oc version' on master as discussed with @bmeng. 'oc version' doesn't support json output on 3.xx so i am going to try grep/awk to fetch that value and the abort if value doesn't match 3.xx. Thanks

akostadinov: Master branch now only supports 4.0+. So you can update things to work with 4.0. the `v3` branch is for 3.x

anuragthehatter:
Closing this PR as master branch now points to 4.x and the this scenario is gated to work on 3.x version. Opened https://github.com/openshift/verification-tests/pull/108 instead

PR #93, opened 2019-02-28
Add authentication endpoint for 4.x

stuartchuan:
Add authentication endpoint for 4.x

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -164,6 +164,20 @@ module BushSlicer
       return @admin_console_url
     end
 
+    def authentication_url
+      unless @authentication_url
+        if opts[:authentication_url]
+          @authentication_url= opts[:authentication_url]
+        else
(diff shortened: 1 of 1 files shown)


lib/environment.rb
```diff
@@ -164,6 +164,20 @@ def admin_console_url
...
 
+    def authentication_url
+      unless @authentication_url
+        if opts[:authentication_url]
+          @authentication_url= opts[:authentication_url]
+        else
+          authenticationproject = Project.new(name: "openshift-authentication", env: self)
+          authenticationservice = Service.new(name: "openshift-authentication", project: authenticationproject )
+          authenticationroute = Route.new(name: "openshift-authentication", project: authenticationproject, service: authenticationservice)
+          @authentication_url = "https://" + authenticationroute.dns(by: admin)
```
akostadinov: can't we figure out auth URL by trying to login to API? Sounds like better than requiring admin access.

PR #94, opened 2019-03-01
added support when user specify a local kubeconfig location

pruan-rht:
similiar to #83, but accept kubeconfig that's stored locally.  This will be helpful in case the URL is not accesible.  You can set environment variable like so
`export OPENSHIFT_ENV_OCP4_ADMIN_CREDS_SPEC=~/.kube/config`

### DIFF
--- a/lib/admin_credentials.rb
+++ b/lib/admin_credentials.rb
@@ -77,23 +77,26 @@ module BushSlicer
     end
   end
 
+  # expect user env to set simliar to below (prepend with `file:///`)
+  # export OPENSHIFT_ENV_OCP4_ADMIN_CREDS_SPEC=file:///~/.kube/config
   class LocalKubeconfigCredentials < AdminCredentials
     def get
-      raise "kubeconfig does not exists" unless File.exists? opts[:spec]
-      config = File.open(opts[:spec]).read
+      path = File.expand_path(opts[:spec].split("file:///")[1])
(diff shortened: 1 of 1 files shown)


akostadinov:
Can we specify file with the `file:///..` URL? We need to handle differently than the `http` url of course. Or maybe switch to using the `open` call insteald of `Http.get`.

pruan-rht:
@akostadinov do you mean change `URLKubeconfigCredentials` to accept file format on top of http URL?

pruan-rht:
@akostadinov PR updated, PTAL.  Thanks!

akostadinov:
I thought that you can update `URLKubeconfigCredentials` actually to also work with `file:///` urls.

pruan-rht:
@akostadinov I prefer to have it separate since local file is not a URL

akostadinov:
`file:///somehting` is a URL. If we have just a path, then this is not a URL (although one can argue it is a relative URL).
But if we use plain path, then how do we distinguish between different kubeconfigs in the `AutoKubeconfigCredentials`?

So my take is that either `LocalKubeconfigCredentials` should work with plain paths or if the path is in the URL form `file:///`, then implement in `URLKubeconfigCredentials`(because it is a URL with the `file` schema).

pruan-rht:
@akostadinov, ok, I've updated the URL method to look for `file:///` scheme.  PTAL.  Thanks!

akostadinov:
Thank you!

PR #95, opened 2019-03-04
Auto case OCP-22515

stuartchuan:
OCP-22515 will replace OCP-10593 in 4.x

### DIFF
--- a/features/cli/create.feature
+++ b/features/cli/create.feature
@@ -1,35 +1,5 @@
 Feature: creating 'apps' with CLI
 
-  # @author wsun@redhat.com
-  # @case_id OCP-10593
-  Scenario: Could not create any context in non-existent project
-    Given I create a new application with:
-      | docker image | openshift/ruby-20-centos7~https://github.com/openshift/ruby-hello-world |
-      | name         | myapp          |
-      | n            | noproject      |
-    Then the step should fail
-    Then the output should match "User "<%=@user.name%>" cannot create imagestream.* in (project|the namespace) "noproject""
-    Given I run the :create client command with:
-      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/pods/hello-pod.json |
-      | n | noproject |
-    Then the step should fail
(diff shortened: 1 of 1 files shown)


akostadinov:
can we remove OCP-10593 then? `v3` branch is for 3.x.

features/step_definitions/oauth_cluster.rb
```diff
@@ -0,0 +1,30 @@
+# setup tier_down to restore scc after scenario end;
+Given /^oauth cluster is restored after scenario$/ do |cluster|
```
akostadinov: btw this step is not used in the scenario. It is preferable to add new steps when they are used instead together with unrelated scenarios.

features/cli/create.feature
```diff
@@ -167,3 +167,32 @@ Feature: creating 'apps' with CLI
...
+    Given I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift/origin/master/examples/image-streams/image-streams-centos7.json |
+      | n | noproject |
+    Then the step should fail
+    Then the output should match "User "<%=@user.name%>" cannot create resource "imagestreams".* in the namespace "noproject""
+    Given I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift/origin/master/examples/sample-app/application-template-stibuild.json |
+      | n | noproject |
+    Then the step should fail
+    Then the output should match "User "<%=@user.name%>" cannot create resource "templates".* in the namespace "noproject""
```
akostadinov: all these tests can happily be unit tests. This will be better to avoid breakage due to human readable strings in the future. Otherwise looks good and can merge if you wish them here.

features/step_definitions/oauth_cluster.rb
stuartchuan: Sorry, my mistake to add this step.

features/cli/create.feature
stuartchuan: Yes, should be removed.

PR #96, opened 2019-03-04
update the project delete step

stuartchuan:
Update project step for 4.x
```
Given the "<%= cb.prj_name %>" project is deleted                                      # features/step_definitions/project.rb:192
      project.project.openshift.io "56m0m" deleted
      
      [07:09:11] INFO> Exit Status: 0
      error getting project '56m0m' existence: 
      STDERR:
      Error from server (Forbidden): projects.project.openshift.io "56m0m" is forbidden: User "pm1" cannot get resource "projects" in API group "project.openshift.io" in the namespace "56m0m"
       (RuntimeError)
  ../redhat/verification-tests/lib/openshift/project.rb:27:in `visible?'
```

### DIFF
--- a/lib/openshift/project.rb
+++ b/lib/openshift/project.rb
@@ -21,7 +21,7 @@ module BushSlicer
         return true
       else
         case  result[:response]
-        when /cannot get project/, /not found/
+        when /Error from server \(Forbidden\)/
           return false
         else
           raise "error getting project '#{name}' existence: #{result[:response]}"


lib/openshift/project.rb
```diff
@@ -21,7 +21,7 @@ def visible?(user: nil, result: {}, quiet: false)
         return true
       else
         case  result[:response]
-        when /cannot get project/, /not found/
+        when /cannot get project/, /not found/, /Error from server \(Forbidden\)/
```
akostadinov: we can remove old messages because this branch is for v4.0+, no need to keep old baggage

stuartchuan: Yes, removed.

lib/openshift/project.rb
```diff
@@ -21,7 +21,7 @@ def visible?(user: nil, result: {}, quiet: false)
         return true
       else
         case  result[:response]
-        when /cannot get project/, /not found/
+        when /Error from server \(Forbidden\)/
```
akostadinov: wh ydo we need these parentheses here btw?

PR #97, opened 2019-03-05
Fix check_doc_get_started_with_cli and check_dim_resource_list

xiaocwan:
@yanpzhan Could you help to review? Thanks!

### DIFF
--- a/lib/rules/web/admin_console/4.0/status.xyaml
+++ b/lib/rules/web/admin_console/4.0/status.xyaml
@@ -24,6 +24,6 @@ check_getting_started:
 check_dim_resource_list:
   elements:
   - selector:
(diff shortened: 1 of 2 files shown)


lib/rules/web/admin_console/4.0/status.xyaml
```diff
@@ -24,6 +24,6 @@ check_getting_started:
 check_dim_resource_list:
   elements:
   - selector:
-      xpath: //div[@class='co-disabled']//button[contains(@id,'create')]
```
xiaocwan: Create button has changed to drop-down button list for some resources.

yanpzhan:
The fix LGTM.

PR #98, opened 2019-03-06
Modify jenkins steps to adjust 4.0 change

xiuwang:
[jenkins log](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/38703/console)
@openshift/devexp-qe please help review

### DIFF
--- a/features/step_definitions/images/jenkins.rb
+++ b/features/step_definitions/images/jenkins.rb
@@ -1,11 +1,5 @@
 #Use this step definition file to create image-specific steps
 
-#This empty step will be removed once we can start updating the existing
-#jenkins cases to use the new OAuth method; this way there is no
-#'missing step' error here.
-Given /^I save the jenkins password of dc #{QUOTED} into the#{OPT_SYM} clipboard$/ do
-end
-
 Given /^I have an?( ephemeral| persistent)? jenkins v#{NUMBER} application(?: from #{QUOTED})?$/ do |type, version, cust_templ|
 
   if cust_templ && type
@@ -30,15 +24,14 @@ Given /^I have an?( ephemeral| persistent)? jenkins v#{NUMBER} application(?: fr
     source = "| template | jenkins-#{type.strip} |"
   end
 
-  lt34 = env.version_lt("3.4", user: user)
   last_startup_check = monotonic_seconds
 
-  # On v3.4 and above, if we can login to OpenShift via password, then
+  # If we can login to OpenShift via password, then
   #   we can use SSO login.
-  # On 3.3 and earlier, or if we only know user token, then we must use
+  # If we only know user token, then we must use
   #   non-SSO login.
   # Installation step also installs based on this.
-  if lt34 || !user.password?
+  if !user.password?
     step %Q/I run the :new_app client command with:/, table(%{
       #{source}
       | p | ENABLE_OAUTH=false                          |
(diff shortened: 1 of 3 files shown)


xiuwang:
When get server version of 4.0 above cluster, the returned result is "Server Version: version.Info{Major:"1", Minor:"12+".....}", so I add one step "Given /^I have a jenkins server with above 40 cluster$/ do" .
After enable any openshift oauth, we get two choose when login openshift websole, the kubeadmin account and common account. So I add  steps "jenkins_multi_oauth_login" and "Given /^I log into jenkins with above 40 cluster$"

akostadinov:
We should rather update Environment#get_version method work using the 4.x call instead of the old one. Then use existing step /^the master version ([<>=]=?) #{QUOTED}$/

wrt the login step,what is the need of this new step?

akostadinov:
Master branch is only for 4.0 cluster version. Can we update existing step instead?

xiuwang:
@akostadinov since this pr [1], the  version for 4.0 is kubernetes version now, could we use client version instead ?

xiuwang:
For the login step,need add a step to choose the auth type since there are two auth type(kubeadmin and other one auth type) at least.

akostadinov:
I need to check how we can do things. Maybe we can use `clusterversion` instead. WDYT?

xiuwang:
@akostadinov  `get clusterversion` need admin role, IMO, it's not a best choice.
Could we hardcode the version to 4.0 [here](https://github.com/openshift/verification-tests/blob/master/lib/environment.rb#L244) ?  I will try launch a 3.2 version to see if version/openshift could be obtained, later will add more comment.

xiuwang:
Both 3.1 and 3.2 version get 403 forbidden when access version/openshift url.

xiuwang:
We have a cmd `openshift version` that common user could use, this works in all 3.x and 4.0 cluster. 

3.1
# openshift version 
openshift v3.1.1.11-9-g44fe9ba
kubernetes v1.1.0-origin-1107-g4c8e6f4
etcd 2.1.2

4.0
$openshift  version
openshift v4.0.2

akostadinov:
I think that `openshift version` would return version of binary instead of version of server. Am I wrong?

xiuwang:
Alex, you are right, `openshift version` returns the binary version, and this cmd has been deprecated from latest 4.1 package.
IMO, the devel said the way in bug might be  https://bugzilla.redhat.com/show_bug.cgi?id=1658957#c11 .
Meantime ,confirmed with yapei,  `<yapei> xiuwang, no, clusterversion info is shown on Administration -> Cluster Settings page on web console which is invisible for non cluster-admin users`.

akostadinov:
I'm trying to reach directly to the developer to see what he meant by getting api server version.

xiuwang:
[3.11 passed log](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39292/console)
[4.0 passed log](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39299/console[)

features/step_definitions/images/jenkins.rb
```diff
@@ -102,11 +103,16 @@
 #   non-SSO login.
 # Installation step also installs based on this.
 Given /^I log in to jenkins$/ do
-  if env.version_ge("3.4", user: user) && user.password?
+  if env.version_ge("3.4", user: user) && env.version_le("3.11", user: user) && user.password?
```
akostadinov: in this branch we should only support 4.x, conditions for older cluster verisions better be gone. WDYT?

akostadinov: not sure what CVP is. In any case, master branch not suitable to run stuff on 3.x

xiuwang: CVP is [container verification pipeline tool][https://docs.engineering.redhat.com/pages/viewpage.action?spaceKey=CVP&title=Container+Verification+Pipeline+E2E+Documentation#ContainerVerificationPipelineE2EDocumentation-ProductTests) which  anli is taking development.
I will remove the conditions for older cluster verisions later.

xiuwang:
@akostadinov Removed the conditions for older cluster verisions, and test case [passed](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39356/consoleFull).

akostadinov:
/lgtm

Thank you!

PR #100, opened 2019-03-06
proxy command should not ask

akostadinov:


xingxingxia:
@akostadinov sorry for delayed response to [the issue](https://projects.engineering.redhat.com/browse/OPENSHIFTQ-12621).
Verified it works now, [see pass job](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/38920/console). Thanks!

Seems not all guys are aware of /b/, thus fyi @openshift/team-qe :
Pls prefix each internal master/node with /b/ when :bastion is used, not just set `api.....qe.devcluster.openshift.com:master:node:lb`.
Or a simpler way, just copy host.spec from Flexy job which already has prefix /b/.
With this, cases like below can pass:
```
    And I select a random node's host
```

zhaozhanqi:
Great work.  thanks.

PR #103, opened 2019-03-07
iSCSI initiator should only setup on nodes

liangxia:
@akostadinov  @jhou1  @chao007  @weherdh  @qinpingli 

iSCSI initiator should only apply on nodes.


liangxia:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=1686290

akostadinov:
looks nice!

PR #106, opened 2019-03-08
fix issue when YAML uses anchoring

akostadinov:
@gpei FYI

Issue was when we use ancoring in YAML and then merge them. When anchoring is
uses, the same Hash object can be present in multiple places within the
parsed Hash object. Then merging the Hashes in place results in changing
Hash in all places while only in one part of the tree it needs to be merged.

Now we always non-destructive merge hashes to avoid this situation.


gpei:
Thanks, built a 3.11 cluster with flexy job on aws today, it's working well.

PR #107, opened 2019-03-08
updated owners file to remove cjryan

pruan-rht:


pruan-rht:
minor update with no functionality changes...just owner auditing.  I'll just merge it myself

PR #109, opened 2019-03-12
update for case 10717

zhouying7780:
@xingxingxia  Please help review , have pass locally.


zhouying7780:
@pruan-rht  Please help review , thanks.

PR #112, opened 2019-03-13
make jira create issue work and allow running on command line

akostadinov:
`issue.fetch` always results in `bad request`. Also I don't think we need this fetch to begin with so I remove it. If anybody needs, then it can be called later.

/assign @pruan-rht

### DIFF
--- a/lib/jira_rht.rb
+++ b/lib/jira_rht.rb
@@ -24,6 +24,7 @@ module BushSlicer
         }
       end
       unless @options[:password]
+        require 'io/console'
         STDERR.puts "JIRA Password: "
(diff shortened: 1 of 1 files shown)


akostadinov:
FYI sumoheavy/jira-ruby#175

pruan-rht:
interesting...#111 & #112 is the same?

pruan-rht:
closing this one as #111 is the same it looks like, pls reopen if otherwise.

akostadinov:
they are for different branches ;)

akostadinov:
Wouldn't it be nice if we had a command `/backport branch` and have PR automatically applied to respective branch in case of no conflicts?

PR #120, opened 2019-03-18
support for test case queries

akostadinov:
/assign @pruan-rht

### DIFF
--- a/lib/polarshift/request.rb
+++ b/lib/polarshift/request.rb
@@ -174,6 +174,54 @@ module BushSlicer
         }
       end
 
+      def query_test_cases(project_id:,
+                     case_query: nil,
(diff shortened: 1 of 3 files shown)


lib/polarshift/request.rb
```diff
@@ -174,6 +174,54 @@ def create_run_smart(timeout: 360, **opts)
...
+          method: :post,
+          url: "#{base_url}project/#{project_id}/test-cases/query",
+          payload: create_opts,
+          raise_on_error: false,
+          **common_opts
+        )
+      end
+
+
+      def query_test_cases_smart(timeout: 360, **opts)
```
pruan-rht: @akostadinov 360 seconds seems too long...perhaps lowering that.

Issue #118, opened 2019-03-18
lib/rest_kubernetes.rb argument number


xingxingxia:
https://github.com/openshift/verification-tests/pull/88/ is a great PR. Found an issue related to it: `self.populate` adds 4th argument `type` in method def but leaves all callings still only 3 arguments without 4th argument `type`. This makes below failure, like in [the log](http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2019/03/15/07:09:31/PDB_take_effective_with_percentage_number_with_beta1/console.html):
```
wrong number of arguments (given 3, expected 4)
ArgumentError
/home/jenkins/workspace/Runner-v3/lib/rest_kubernetes.rb:8:in `populate'
/home/jenkins/workspace/Runner-v3/lib/rest_kubernetes.rb:79:in `create_pod_eviction'
/home/jenkins/workspace/Runner-v3/lib/rest.rb:101:in `delegate_rest_request'
/home/jenkins/workspace/Runner-v3/lib/rest.rb:27:in `exec'
/home/jenkins/workspace/Runner-v3/lib/api_accessor.rb:67:in `rest_request'
/home/jenkins/workspace/Runner-v3/lib/api_accessor_owner.rb:28:in `public_send'
/home/jenkins/workspace/Runner-v3/lib/api_accessor_owner.rb:28:in `block (2 levels) in '
/home/jenkins/workspace/Runner-v3/features/step_definitions/common.rb:26:in `/^I perform the :([a-z_]*?) rest request with:$/'
features/tierN/admin/pod.feature:480:in `When I perform the :create_pod_eviction rest request with:'
```
The api's are diverse. @akostadinov @pruan-rht any idea?

pruan-rht:
@akostadinov @xingxingxia addressed by PR https://github.com/openshift/verification-tests/pull/121

akostadinov:
I have provided a possible solution as a comment to #121

PR #121, opened 2019-03-18
revert changes to populate method, perhaps we need nother version of …

pruan-rht:
…the populate for 'apis/k8s.xxx' calls.  For reference https://github.com/openshift/verification-tests/issues/118

### DIFF
--- a/lib/rest_kubernetes.rb
+++ b/lib/rest_kubernetes.rb
@@ -5,8 +5,8 @@ module BushSlicer
     module Kubernetes
       extend Helper
 
-      def self.populate(path, base_opts, opts, type)
-        populate_common("/apis/#{type}.k8s.io/<api_version>", path, base_opts, opts)
+      def self.populate(path, base_opts, opts)
+        populate_common("/api/<api_version>", path, base_opts, opts)
       end
 
       class << self


lib/rest_kubernetes.rb
```diff
@@ -5,8 +5,8 @@ module Rest
     module Kubernetes
       extend Helper
 
-      def self.populate(path, base_opts, opts, type)
-        populate_common("/apis/#{type}.k8s.io/<api_version>", path, base_opts, opts)
+      def self.populate(path, base_opts, opts)
+        populate_common("/api/<api_version>", path, base_opts, opts)
```
akostadinov: How about
```
def self.populate(path, base_opts, opts, type: nil)
  type_str = type ? "/#{type}.k8s.io" : ""
  populate_common("/apis#{type_str}/<api_version>", path, base_opts, opts)
```

I will apply this patch now to get things fixed. But for future resolution.

pruan-rht: sounds good.  It should be `type_str = type ? "s/#{type}.k8s.io" : ""` though because the other format is `apis`

xingxingxia:
Thanks for the PR and above new solution.
Will test the case mentioned with failure log in https://github.com/openshift/verification-tests/issues/118 top comment. @zhouying7780 could you please help the test when available?

PR #123, opened 2019-03-19
handle differnt api url depending on the type of operation

pruan-rht:
addresses comment in issue https://github.com/openshift/verification-tests/issues/118

### DIFF
--- a/lib/rest_kubernetes.rb
+++ b/lib/rest_kubernetes.rb
@@ -5,8 +5,10 @@ module BushSlicer
     module Kubernetes
       extend Helper
 
-      def self.populate(path, base_opts, opts)
-        populate_common("/api/<api_version>", path, base_opts, opts)
+      # the api url can either be /api/<api_version> or /apis/#{type}.k8s.io/<api_version> depending on the operation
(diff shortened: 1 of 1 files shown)


lib/rest_kubernetes.rb
```diff
@@ -5,8 +5,10 @@ module Rest
     module Kubernetes
       extend Helper
 
-      def self.populate(path, base_opts, opts)
-        populate_common("/api/<api_version>", path, base_opts, opts)
+      # the api url can either be /api/<api_version> or /apis/#{type}.k8s.io/<api_version> depending on the operation
+      def self.populate(path, base_opts, opts, type: nil)
+        type_str = type ? "s/#{type}.k8s.io" : ""
```
akostadinov: looks like unintended `s` needs to be removed

pruan-rht: @akostadinov, no, the `s` should be there according to https://docs.okd.io/latest/rest_api/

akostadinov: my bad, missed that

PR #124, opened 2019-03-20
Step for cluster crd backup restore

stuartchuan:
Add step for cluster crd backup restore.

### DIFF
--- a/features/step_definitions/crd.rb
+++ b/features/step_definitions/crd.rb
@@ -1,11 +1,11 @@
 Given /^the #{QUOTED} #{QUOTED} CRD is recreated after scenario$/ do |name, crd|
   ensure_admin_tagged
-  allowed_crds = ["oauth", "kubeapiserver"]
+  allowed_crds = ["kubeapiserver"]
 
-  if allowed_crds.included? 'crd'
-    cb.cluster_resource_to_recreate = crd(name)
+  if allowed_crds.include? crd
+    cb.cluster_resource_to_recreate = self.send(crd, name)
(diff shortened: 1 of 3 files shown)


features/step_definitions/crd.rb
```diff
@@ -0,0 +1,24 @@
...
+      raise "Could not get #{crd}: #{name}"
+    end
+
+    _admin = admin
+    teardown_add {
+      admin.cli_exec(:delete, object_type: crd, object_name_or_id: name)
+
+      @result = _admin.cli_exec(:create, f: "-", _stdin: origin_crd)
+      raise "Cannot restore #{crd}: #{name}" unless @result[:success]
+    }
```
akostadinov: can we, instead of the step body, delegate to `hidden recreate cluster resource( after scenario)`?
```
  cb.cluster_resource_to_recreate = self.send(crd, name)
  step 'hidden recreate cluster resource after scenario'
```

akostadinov: This assumes we create placeholder objects for the CRDs in question but it should be very easy. See example `lib/openshift/identity.rb`.

features/step_definitions/crd.rb
```diff
@@ -0,0 +1,24 @@
...
+  else
+    raise "No such #{crd} CRD in the cluster"
```
akostadinov: error message better be somehting like: #{crd} not supported by this step, or something along these lines.

stuartchuan:
Please help review, only add kubeapiserver CRD for the step now.

PR #125, opened 2019-03-20
Get metering report via rest

pruan-rht:


pruan-rht:
@akostadinov PTAL, thanks!

features/step_definitions/metering.rb
```diff
@@ -167,32 +163,45 @@
...
-          authProxy:
-            enabled: true
-            htpasswdData: |
-              #{htpasswd}
-            cookieSeed: "#{cookie_seed}"
-            subjectAccessReviewEnabled: true
-            delegateURLsEnabled: true
+  # create the route only if one does not exists (currently it's hardcoded
+  # in the chart file charts/reporting-operator/values.yaml)
+  unless route('metering').exists?
```
akostadinov: I think that you need to be in the project where this route lives. Will be more reliable to specify project in the `route` call.

features/step_definitions/metering.rb
```diff
@@ -167,32 +163,45 @@
...
+    org_user = user
+    if env.is_admin? user
+      # admin user doesn't have name and password properties, we switch to a user
+      step %Q/the first user is cluster-admin/
+      step %Q/I switch to the first user/
+    end
+    htpasswd = BushSlicer::SSL.sha1_htpasswd(username: user.name, password: user.password)
```
akostadinov: I think that
1. username and password here don't need to match the username and password of a user
1. you can user1 username and password without making it cluster admin; i.e. `user(0).name`, `user(0).password` but then submit the yaml as the cluster admin user.

Preferable to use fake username and password instead of putting actual user/pass

features/step_definitions/metering.rb
pruan-rht: ok, done.

features/step_definitions/metering.rb
pruan-rht: @akostadinov ok, done.  PTAL.  Thanks!

PR #126, opened 2019-03-22
Ocp 22704

anuragthehatter:
Gherkin Scenario for SDN test case OCP-22704. Separate operator.feature is created to accommodate future network operator related tests.
The aim of this test is to make sure that the clusteroperator should be able to reflect the network operator version corresponding to the OCP version. OCP version is checked first. Network operator version is checked when its "Available" value is "True". Both version are compared to make sure they are same.
@bmeng @zhaozhanqi @pruan-rht @akostadinov 
Passing logs: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/289/console

### DIFF
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -12,4 +12,4 @@ Feature: Operator related networking scenarios
     #Making sure that network operator AVAILABLE status value is True
     Then the expression should be true> cb.operator_status["status"]=="True"
     #Confirm whether network operator version matches with ocp version
-    And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)==true
+    And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -1,0 +1,27 @@
+Feature: Operator related networking scenarios
+  
+  # @auther anusaxen@redhat.com
+  # @case_id OCP-22704
+  @admin
+  Scenario: The clusteroperator should be able to reflect the network operator version corresponding to the OCP version
+    Given I use the "openshift-sdn" project
+    #Getting OCP version
+    When I run the :get admin command with:
+      | resource | clusterversion                                |
+      | output   | jsonpath='{.items[*].status.desired.version}' |
+    Then the step should succeed
+    And evaluation of `@result[:response]` is stored in the :ocp_version clipboard
+
+    #Making sure that network operator AVAILABLE value is True 
+    When I run the :get admin command with:
+      | resource | clusteroperators.config.openshift.io/network                   |
+      | output   | jsonpath='{.status.conditions[?(@.type=="Available")].status}' |
+    Then the step should succeed
+    And the output should contain "True"
+
+    #Make sure that network operator version matches with ocp version
+    When I run the :get admin command with:
+      | resource | clusteroperators.config.openshift.io/network |
+      | output   | jsonpath='{.status.versions[*].version}'     |
+    Then the step should succeed
+    And the output should equal "<%= cb.ocp_version %>"
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -14,4 +14,4 @@ Feature: Operator related networking scenarios
 
     Given evaluation of `cluster_operator('network').versions` is stored in the :operator_version clipboard
     #Confirm whether network operator version matches with ocp version`
-    Then the expression should be true> cb.operator_version="<%= cb.ocp_version%>"
+    Then the expression should be true> cb.operator_version=cb.ocp_version
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -4,7 +4,7 @@ Feature: Operator related networking scenarios
   # @case_id OCP-22704
   @admin
   Scenario: The clusteroperator should be able to reflect the network operator version corresponding to the OCP version
-    Given I use the "openshift-sdn" project
+    Given I have a project
     #Getting OCP version
     When I run the :get admin command with:
       | resource | clusterversion                                |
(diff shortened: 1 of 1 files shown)


features/networking/operator.feature
```diff
@@ -0,0 +1,27 @@
+Feature: Operator related networking scenarios
+  
+  # @auther anusaxen@redhat.com
+  # @case_id OCP-22704
+  @admin
+  Scenario: The clusteroperator should be able to reflect the network operator version corresponding to the OCP version
+    Given I have a project
+    #Getting OCP version
+    When I run the :get admin command with:
```
pruan-rht: you don't need to be admin to run these commands

pruan-rht: Also, we prefer to avoid doing steps like `I run the :get command with` but with a more OO approach in getting these properties.

features/networking/operator.feature
```diff
@@ -0,0 +1,27 @@
+Feature: Operator related networking scenarios
+  
+  # @auther anusaxen@redhat.com
+  # @case_id OCP-22704
+  @admin
+  Scenario: The clusteroperator should be able to reflect the network operator version corresponding to the OCP version
+    Given I have a project
```
pruan-rht: you don't need to create a project

features/networking/operator.feature
pruan-rht: @anuragthehatter, you can shorten this sceanrio once #127 is merged.  For example, you can replace line 9-13 with the following: `The evaluation of `cluster_version('version').version` is stored in the :ocp_version clipboard`

features/networking/operator.feature
anuragthehatter: Agree, Peter. We don't need that step, guess i am accustomed to perform all operations in separate project unless a specific one is required.

features/networking/operator.feature
anuragthehatter: Right, Peter,  admin is not necessarily required. Regarding #127, yep that will be useful and i similar one would be also useful if designed for operator version as there are numerous operators we can check the version for. I can look at designing that step definition.

anuragthehatter: Oh i noticed #126 would also support clusteroperator if merged. i will wait. Thanks!

pruan-rht: @anuragthehatter, my mistake, it seems that both `oc get clusterversion` and `oc get clusteroperator'  do need to be executed as admin user.

anuragthehatter: Yep, i assumed that Peter. Thanks!

anuragthehatter:
Hello folks, I have modified the whole scenaio oo way. Thanks to @pruan-rht helper #127 and suggestions. Please take a loo at my latest commit and let me know if any more suggestions. Thanks!
Passing log: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/296/console

features/networking/operator.feature
```diff
@@ -14,4 +14,4 @@ Feature: Operator related networking scenarios
 
     Given evaluation of `cluster_operator('network').versions` is stored in the :operator_version clipboard
     #Confirm whether network operator version matches with ocp version`
-    Then the expression should be true> cb.operator_version="<%= cb.ocp_version%>"
+    Then the expression should be true> cb.operator_version=cb.ocp_version
```
pruan-rht: @anuragthehatter I think this is incorrect...assignment will be true....I think you meant to do a comparsion which should be `==`

anuragthehatter: Agree, @pruan-rht. My bad i have used '==' in another epression above this but seems like a typo in this 2nd expression. Good catch. Thanks!

anuragthehatter:
Will wait for #131 to be merged. Thanks!

anuragthehatter:
Works good now post incorporating #131 ,PTAL. Thanks, Peter.
Logs: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/306/console

bmeng:
LGTM, thanks

features/networking/operator.feature
```diff
@@ -11,7 +11,5 @@ Feature: Operator related networking scenarios
     And evaluation of `cluster_operator('network').condition(type: 'Available')` is stored in the :operator_status clipboard
     #Making sure that network operator AVAILABLE status value is True
     Then the expression should be true> cb.operator_status["status"]=="True"
-
-    Given evaluation of `cluster_operator('network').versions` is stored in the :operator_version clipboard
-    #Confirm whether network operator version matches with ocp version`
-    Then the expression should be true> cb.operator_version=cb.ocp_version
+    #Confirm whether network operator version matches with ocp version
+    And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)==true
```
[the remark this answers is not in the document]
anuragthehatter: @pruan-rht `cluster_operator('network').versions` is not present in latest commit as versions_exists def suffice that check. Regarding `And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)`, i believe i verified that guess i had some check issues. I might be wrong but will check again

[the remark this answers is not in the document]
pruan-rht: @anuragthehatter both versions work, but `cluster_operator('network').version_exists?(version: cb.ocp_version)` is enough since the `#version_exists?` returns a boolean already.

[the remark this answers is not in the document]
anuragthehatter: Yep, confirmed again. `==true` is unnecessary, removed that. Thanks!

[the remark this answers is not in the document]
pruan-rht: @anuragthehatter thanks for the update...I will merge it now.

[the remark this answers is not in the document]
anuragthehatter: Agree, Peter. `.versions` returns version in an array while version_exists? returns boolean. Both would be useful based on what the usecase is.

PR #127, opened 2019-03-23
added api support for clusteroperator and clusterversion for 4.x

pruan-rht:
helper for #126

### DIFF
--- a/lib/openshift/cluster_operator.rb
+++ b/lib/openshift/cluster_operator.rb
@@ -1,0 +1,22 @@
+module BushSlicer
+  class ClusterOperator < ClusterResource
+    RESOURCE = "clusteroperators"
+    # return the an array of versions for a particular clusteroperator
+    def versions(user: nil, cached: true, quiet: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
+      rr.dig('status', 'versions').map { |v| v['version'] }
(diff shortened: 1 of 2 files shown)


lib/openshift/cluster_operator.rb
```diff
@@ -0,0 +1,22 @@
+module BushSlicer
+  class ClusterOperator < ClusterResource
+    RESOURCE = "clusteroperators"
```
akostadinov: maybe we need a way to specify namespace of this resource, e.g. "k8s.io" but it is goot for the time being, just a thought

lib/openshift/cluster_operator.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    def conditions(user: nil, cached: true, quiet: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
+      rr.dig('status', 'conditions')
+    end
+
+    # returns a hash simliar to
+    # {"lastTransitionTime"=>2019-03-23 07:33:34 UTC, "status"=>"False", "type"=>"Progressing"}
+    def condition(type: type, user: nil, cached: true, quiet: false)
```
akostadinov: looks like these methods can go to `Resource` but ok to be here for the time  being

PR #131, opened 2019-03-27
moved methods to lib/resource and added a version check

pruan-rht:
addresses https://github.com/openshift/verification-tests/pull/127#discussion_r268840474.  @akostadinov  PTAL.

### DIFF
--- a/lib/openshift/cluster_operator.rb
+++ b/lib/openshift/cluster_operator.rb
@@ -7,7 +7,7 @@ module BushSlicer
       rr.dig('status', 'versions').map { |v| v['version'] }
     end
     # given a target_version return true if has matching verison else false
-    def has_version?(version: nil, user: nil, cached: true, quiet: false)
+    def version_exists?(version: nil, user: nil, cached: true, quiet: false)
       res = self.versions.select { |v| v == version }
       res.count != 0
     end
--- a/lib/openshift/cluster_operator.rb
+++ b/lib/openshift/cluster_operator.rb
@@ -6,17 +6,10 @@ module BushSlicer
       rr = raw_resource(user: user, cached: cached, quiet: quiet)
       rr.dig('status', 'versions').map { |v| v['version'] }
     end
-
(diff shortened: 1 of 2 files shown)


lib/openshift/cluster_operator.rb
```diff
@@ -6,17 +6,10 @@ def versions(user: nil, cached: true, quiet: false)
...
-      rr.dig('status', 'conditions')
-    end
-
-    # returns a hash simliar to
-    # {"lastTransitionTime"=>2019-03-23 07:33:34 UTC, "status"=>"False", "type"=>"Progressing"}
-    def condition(type: type, user: nil, cached: true, quiet: false)
-      res = self.conditions.select { |c| c['type'] == type }
-      res.first
+    # given a target_version return true if has matching verison else false
+    def has_version?(version: nil, user: nil, cached: true, quiet: false)
```
akostadinov: `is_version?` sounds better to me, wdyt? Has versions like owning the version which a little bit makes me confused.

lib/openshift/resource.rb
```diff
@@ -279,6 +279,18 @@ def status?(user: nil, status:, quiet: false, cached: false)
...
+      rr.dig('status', 'conditions')
+    end
+
+    # returns a hash simliar to
+    # {"lastTransitionTime"=>2019-03-23 07:33:34 UTC, "status"=>"False", "type"=>"Progressing"}
+    def condition(type: nil, user: nil, cached: true, quiet: false)
+      res = self.conditions.select { |c| c['type'] == type }
+      res.first
+    end
+
```
akostadinov: good

lib/openshift/cluster_operator.rb
pruan-rht: @akostadinov, the use case from @anuragthehatter is that he wants to check if a version exists.. perhaps `exists`?

pruan-rht: @akostadinov  ok, I changed it to `version_exists?`

PR #132, opened 2019-03-28
Skip node InitatorName config for iscsi as no need in 4.0

liangxia:
@chao007  @weherdh  Please take a look.

More detail in https://bugzilla.redhat.com/show_bug.cgi?id=1687722#c7

### DIFF
--- a/features/step_definitions/helper_services.rb
+++ b/features/step_definitions/helper_services.rb
@@ -473,7 +473,6 @@ Given /^I have a iSCSI setup in the environment$/ do
   else
     env.node_hosts.each do |host|
       setup_commands = [
-        "echo 'InitiatorName=iqn.2016-04.test.com:test.img' > /etc/iscsi/initiatorname.iscsi",
         "sed -i '/^node.session.auth./'d  /etc/iscsi/iscsid.conf",
         "cat >> /etc/iscsi/iscsid.conf << EOF\n" +
           "node.session.auth.authmethod = CHAP\n" +


weherdh:
LGTM, thanks.

pruan-rht:
@liangxia, if the PR is checkin in master, we can just remove the check all together since we have the v3 branch, WDYT @akostadinov?

features/step_definitions/helper_services.rb
```diff
@@ -471,20 +471,22 @@
...
-          "node.session.auth.authmethod = CHAP\n" +
-          "node.session.auth.username = 5f84cec2\n" +
-          "node.session.auth.password = b0d324e9\n" +
-          "EOF\n",
-        "systemctl enable iscsid",
-        "systemctl restart iscsid"
-      ]
-      res = host.exec_admin(*setup_commands)
-      raise "iSCSI initiator setup commands error" unless res[:success]
+    if env.version_cmp("4.0", user: user) < 0
```
akostadinov: As Peter said, master branch is only for 4.x.

Just FYI, in version comparisons better to use `env.version_gt/ge/lt/le/eq` methods instead of `env.version_cmp`. They are more easily readable what they mean. Now it is hard to tell whether condition is for 4.0+ of 3.x

liangxia:
Removed the version check and unnecessary config steps.

liangxia:
@pruan-rht @akostadinov

PR #135, opened 2019-03-29
AWS improvements

akostadinov:
/assign @pruan-rht

### DIFF
--- a/lib/launchers/amz.rb
+++ b/lib/launchers/amz.rb
@@ -47,13 +47,26 @@ module BushSlicer
           awscred["AWSSecretKey"]
         )
       }) )
-      client = Aws::EC2::Client.new
-      @ec2 = Aws::EC2::Resource.new(client: client)
+    end
 
+    private def client_ec2
+      @client_ec2 ||= Aws::EC2::Client.new
+    end
+
+    private def client_s3
+      @client_s3 ||= Aws::S3::Client.new
     end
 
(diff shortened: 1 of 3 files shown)


lib/launchers/amz.rb
```diff
@@ -47,9 +47,22 @@ def initialize(access_key: nil, secret_key: nil, service_name: nil)
...
-      client = Aws::EC2::Client.new
-      @ec2 = Aws::EC2::Resource.new(client: client)
+    end
+
+    private def ec2
+      @ec2 ||= Aws::EC2::Resource.new(client: ec2_client)
+    end
+
+    private def ec2_client
+      @clien ||= Aws::EC2::Client.new
```
pruan-rht: @akostadinov is this used anywhere?  Also why named it `@clien` instead of `@client`

akostadinov: it must be `@client`. Typo. The `ec2_client` **is** used. the variable is for caching the object instead of creating many times.

akostadinov: Fixed now, thanks a lot for catching this. Actually bigger issue was that bot s3 and ec2 used the same caching variable :/

akostadinov: Fixed in both pull requests.

pruan-rht: cool, thanks.  the rest looks good!

PR #137, opened 2019-03-31
modify the rules for web login

yuwang-RH:
This PR is trying to fix the problem of the web console logging. The web console logging will fail as some logging workflow change. You can find the failed log as bellow(log1).
The log for this PR is as bellow, log2.
log1:https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39419/console
log2:https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39414/console
@yasun1 could you help review this PR? Thanks

### DIFF
--- a/lib/rules/web/console/base/login.xyaml
+++ b/lib/rules/web/console/base/login.xyaml
@@ -11,10 +11,11 @@ login_sequence:
     ref:
     - click_pre_login_button
     - pre_login_buttons_gone
+    - login_page_loaded
     if_element:
       selector: &pre_login_buttons
         # we just choose the first button per OPENSHIFTQ-12062
-        css: ".login-redhat:last-of-type, .idp:last-of-type"
+        css: ".login-redhat:first-of-type, .idp:first-of-type"
       timeout: 1
   action:
     ref: login_console
@@ -50,28 +51,32 @@ login_rhsso:
         id: password
       op: send_keys <password>
       type: input
+  elements:
     - selector:
         id: kc-login
       type: input
       op: click
+
 login_console:
   elements:
     - selector:
-        id: username
+        id: inputUsername
       op: send_keys <username>
       type: input
     - selector:
-        id: login-show-step2
-      op: click
-      type: button
-    - selector:
-        id: password
+        id: inputPassword
       op: send_keys <password>
       type: input
+  scripts:
+  - command: return document.getElementsByName("csrf").length > 0
+    expect_result: true
+  cookies:
+  - name: csrf
+    expect_result: true
   elements:
     - selector:
-        id: kc-login
-      type: input
+        text: Log In
+      type: button
       op: click
 
 login_token:
@@ -95,7 +100,7 @@ login_token:
 login_page_loaded:
   element:
     selector:
-      css: "#password, #inputPassword, .login-redhat, .idp, .navigation-container"
+      css: "#kc-form-login, #password, #inputPassword, .login-redhat, .idp, .navigation-container"
     timeout: 30
 
 verify_logged_in:
--- a/lib/rules/web/console/base/login.xyaml
+++ b/lib/rules/web/console/base/login.xyaml
@@ -11,7 +11,6 @@ login_sequence:
     ref:
     - click_pre_login_button
     - pre_login_buttons_gone
-    - login_page_loaded
     if_element:
       selector: &pre_login_buttons
         # we just choose the first button per OPENSHIFTQ-12062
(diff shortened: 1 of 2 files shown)


lib/rules/web/console/base/login.xyaml
```diff
@@ -11,21 +11,12 @@ login_sequence:
     ref:
     - click_pre_login_button
     - pre_login_buttons_gone
-    - login_page_loaded
```
yasun1: The rule login_page_loaded is used to make sure the required elements are loaded successfully. I am not sure what will be caused by simply removing it. @akostadinov cloud you give some advices?

akostadinov: Exactly, that action should make sure that login prompt has appeared. In case it is not working well now, then better fix it than remove. Also IIRC sometimes we need first click a button to reach login prompt, this also checks for it.

If it is removed, it is possible that we open the page and try to set username/password before the text boxes appeared (or button to click appeared).

lib/rules/web/console/base/login.xyaml
```diff
@@ -43,36 +42,36 @@ login_rhsso:
...
-  - command: return document.getElementsByName("csrf").length > 0
-    expect_result: true
-  cookies:
-  - name: csrf
-    expect_result: true
   elements:
     - selector:
-        text: Log In
-      type: button
+        id: kc-login
```
akostadinov: it looks like the issue with `login_page_loaded` is with this change here. I think you can add this id as `#kc-login` to the list under `login_page_loaded` rule and you would get things working without removing the check.

yuwang-RH:
@yasun1 I updated this PR based the comments from you and Alex. 
1 change the pre_login_buttons CSS selector, from the last-of-type to the first-of-type
2 change the visible_text of the login_rhsso action selector
3 update the login_rhsso action
4 add one login_page_loaded selector, #kc-form-login
The log of the jenkins job is  as bellow:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39492/console

Please help review this code change. Thanks

lib/rules/web/console/base/login.xyaml
```diff
@@ -15,7 +15,7 @@ login_sequence:
     if_element:
       selector: &pre_login_buttons
         # we just choose the first button per OPENSHIFTQ-12062
-        css: ".login-redhat:last-of-type, .idp:last-of-type"
+        css: ".login-redhat:first-of-type, .idp:first-of-type"
```
akostadinov: It didn't work with last of type? Usually last of type is the inner-most element so must be somehow safer to use last-of-type if it works.

akostadinov: Other than that LGTM.

yasun1: @akostadinov In the past, you and me made a discussion and decided to use the first-of-type. But it was changed to last-of-type by PR https://github.com/openshift/verification-tests/pull/57.  Our several cases are based on choosing the first button. @yanpzhan @yapei any reasons that change it to last-of-type?

akostadinov: @yasun1 , hmm, very nice that you found the change. I don't know why it changed, whether it was discussed or just an oversight. I initially thought there was some issue with nesting elements but maybe not. So I'm ok with switching back to first-of-type. Thank you.

yapei:
@yuwang-RH  Have you tested your changes against OCP clusters?

lib/rules/web/console/base/login.xyaml
yapei: @yasun1  Since the first IDP choice for OCP 4.x clusters is for kube:admin, we don't want to user `kube:admin` for testing so changed to the last one

yasun1: @yapei so the last-of-type should be a required pre-condition for OCP 4.X cluster, correct? Compared the effected cases, I think ours are much less than yours.  And for this PR has been merged, you can withdraw the first-of-type to the last-of-type in a new PR, and notice us to update our cases. Thanks.
@akostadinov maybe still need bother you merge a new PR again. :)

yasun1:
@yapei We do not take OCP clusters. But if we make the change on the common part, we insist on asking you the affection. That is why I ping you in the above comments.

lib/rules/web/console/base/login.xyaml
akostadinov: Sure thing. I guess we can also select second element or third element if we wish. IIRC it was possible but needs to be checked.

PR #138, opened 2019-04-03
HostPath for 4.x

liangxia:
@weherdh  @chao007  PTAL

Jenkins log, Runner-v3/39454


weherdh:
I think it is better to create a new one since what this does is different with before, but if this is only for 4.0 I think it is ok. LG

liangxia:
@wehe master branch is for 4.x, while we have v3 branch for 3.x

PR #141, opened 2019-04-03
4.x clean up on node need sudo

liangxia:
@akostadinov @pruan-rht  
Issue described in cucushift mail list with subject "4.x clean up on nodes need sudo"

### DIFF
--- a/lib/host.rb
+++ b/lib/host.rb
@@ -524,10 +524,12 @@ module BushSlicer
       end
       if opts[:raw]
         exec_raw("rm #{r} -f -- #{file}", opts)
+        exec_raw("sudo rm #{r} -f -- #{file}", opts)
         opts[:quiet] = true
         res = exec_raw("ls -d -- #{file}", opts)
       else
         exec("rm #{r} -f -- #{file}", opts)
+        exec("sudo rm #{r} -f -- #{file}", opts)
         opts[:quiet] = true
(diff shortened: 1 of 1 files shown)


lib/host.rb
```diff
@@ -524,10 +524,12 @@ def delete(file, opts={})
       end
       if opts[:raw]
         exec_raw("rm #{r} -f -- #{file}", opts)
```
pruan-rht: @liangxia why not remove the old lines?

liangxia: @pruan  We may have some situations where sudo are not configured.

pruan-rht: @liangxia if `sudo` is not configured, then this test won't work, if I read the description `clean up on node need sudo` correctly....

akostadinov: The low-level method should not assume user wants `sudo`. But the step
where deletion is done, should know. My suggestion is to fix the step(s).

Lets alter the steps necessary. Do you know where the delete step is registered as a clean-up for this failure?

liangxia: You mean this one ?
`the "/etc/origin/hostpath/<%= project.name %>" path is recursively removed on the host after scenario`

akostadinov: This one also is a little bit low level. I mean to update this step so it does remove files as admin on the host where needed. Or do you think too many places need update?

PR #144, opened 2019-04-03
Add instructions for downloading and installing Web Driver

mgahagan73:
This PR adds instructions for download and installing the web driver for both Firefox and Chrome. Several test cases require a web driver otherwise they fail and this requirement was not documented anywhere previously. Instructions should be identical for all branches.


pruan-rht:
LGTM, @mgahagan73 can you pls submit another PR against the `master` branch with the same changes?

mgahagan73:
Submitted identical PR for Master branch here:
https://github.com/openshift/verification-tests/pull/146

PR #145, opened 2019-04-03
Added support for cluster upgrade and check status

pruan-rht:


features/step_definitions/installer.rb
```diff
@@ -0,0 +1,15 @@
+## installer related steps such as upgrade
+#
+Given /^I upgrade my cluster to:$/ do | table |
+  ensure_admin_tagged
+  _admin = admin
+  upgrade_timeout = 2 * 60 * 60  # 2 hrs.
+  opts = opts_array_to_hash(table.raw)
+  opts[:force] = true
+  target_version = opts[:to_image]
```
anuragthehatter: @pruan-rht Thanks for creating this rb. I am wondering how 'to_image' opt will get its value? The nightly releases will keep getting pruned and we can't define a static value for 'to_image' in scenarios. Thinking...

pruan-rht: @anuragthehatter it's getting it from environment variable `UPGRADE_TARGET_VERSION` which is set by the user...perhaps I should clarify that in the comment section.

anuragthehatter: Gotha. Sure, commnt would be helpful. Thanks!

lib/openshift/cluster_version.rb
```diff
@@ -1,19 +1,70 @@
 module BushSlicer
   class ClusterVersion < ClusterResource
     RESOURCE = "clusterversions"
-    def channel(user: nil, cached: true, quiet: false)
+    def channel(user: nil, cached: false, quiet: false)
```
akostadinov: I prefer to keep `cached: true` as a convention everywhere and use `cached: false` calls when needed.

lib/openshift/cluster_version.rb
```diff
@@ -1,19 +1,70 @@
...
+    def history(user: nil, cached: false, quiet: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
+      rr.dig('status', 'history')
+    end
+
+    # return a list of history matching the state we want to filter on
+    # query_hash is the parameter we want to filter on,
+    def history_matching(query_hash: nil , user: nil, cached: true, quiet: false)
+      key = query_hash.keys.first
+      value = query_hash[key]
```
akostadinov: Can we instead have `key` and `value` parameters to the method instead of this `query_hash` parameter that makes less clear what should it contain?

lib/openshift/cluster_version.rb
```diff
@@ -1,19 +1,70 @@
...
+      self.history.select { |h| h[key] == value }
+    end
+
+    def image(user: nil, cached: true, quiet: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
+      rr.dig('status', 'desired', 'image')
+    end
+
+    # get the last entry of the spec.history which should be where the original
+    def image_base_url(user: nil, cached: true, quiet: false)
```
akostadinov: if I understand comment correctly, perhaps `initial_image_url` would be better

lib/openshift/cluster_version.rb
```diff
@@ -1,19 +1,70 @@
...
+      res = self.history_matching(query_hash: {'version' => target_version})
+      if res.count > 0
+        logger.debug("MSG: #{res.first['message']}")
+        res.first['state'] == 'Completed'
+      else
+        false
+      end
+    end
+
+    def wait_for_upgrade_completion(target_version:, user:nil, cached: false, quiet: false, upgrade_timeout: 2*60*60)
```
akostadinov: `s/upgrade_timeout/timeout/`, it is clear what is the timeout about in this context

Can we live without `target_version` parameter? Can we figure out by ourselves whether there is a pending update ATM or not?

lib/openshift/cluster_version.rb
```diff
@@ -1,19 +1,70 @@
...
+      wait_for(upgrade_timeout) {
+        upgrade_completed?(target_version: target_version)
+      }
+    end
+    # extract the percentage done from the message field from spec.conditions
+    # default to return nil if no upgarde is being done.
+    def completed_percentage(user: nil, cached: true, quiet: false)
+      progressing_cond = self.condition(type: 'Progressing')
```
akostadinov: I guess if there is no condition type `Progressing`, then any updates have finished, no?

features/test/new_api.feature
```diff
@@ -0,0 +1,14 @@
...
+
+  @admin
+  Scenario: clusterversion apis
+    And evaluation of `cluster_version('version').completed_percentage` is stored in the :upgrade_completion clipboard
+    And evaluation of `cluster_version('version').history` is stored in the :history clipboard
+    And evaluation of `cluster_version('version').history_matching(query_hash: {'version' => cluster_version('version').version})` is stored in the :history_match clipboard
+    And evaluation of `cluster_version('version').upgrade_completed?(target_version: ENV['UPGRADE_TARGET_VERSION'])` is stored in the :upgrade_completion clipboard
+    And evaluation of `cluster_version('version').wait_for_upgrade_completion(target_version: ENV['UPGRADE_TARGET_VERSION'], upgrade_timeout: 10)` is stored in the :upgrade_status clipboard
+
+@admin
```
akostadinov: indentation

features/step_definitions/installer.rb
```diff
@@ -0,0 +1,24 @@
+## installer related steps such as upgrade
+#
+# NOTE: it is expected that user set the environment variable UPGRADE_TARGET_VERSION prior to running this step
```
akostadinov: I don't see this variable used in the step, only in scenarios. This makes sense to me. But this note here would be confusing. Better remove that note and add list of step table parameters that need to be added instead.

lib/openshift/cluster_version.rb
pruan-rht: having this `target_version` parameter allows user to force upgrade to anything image he wants.

lib/openshift/cluster_version.rb
pruan-rht: All of the other conditions are present, it's the extra flag that tells if it's occurring or not.  For example ```  conditions:
  - lastTransitionTime: 2019-04-04T21:45:20Z
    message: Done applying 4.0.0-0.nightly-2019-04-03-202419
    status: "True"
    type: Available
  - lastTransitionTime: 2019-04-04T21:33:16Z
    status: "False"
    type: Failing
  - lastTransitionTime: 2019-04-04T21:45:20Z
    message: Cluster version is 4.0.0-0.nightly-2019-04-03-202419
    status: "False"
    type: Progressing
  - lastTransitionTime: 2019-04-04T21:33:16Z
    message: 'Unable to retrieve available updates: unknown version 4.0.0-0.nightly-2019-04-03-202419'
    reason: RemoteFailed
    status: "False"
    type: RetrievedUpdates```

lib/openshift/cluster_version.rb
akostadinov: The point is that we know whether an upgrade is in progress or not (per your other comment). Why do we need to ask user specify a version?

akostadinov: IMO we wait for upgrade to finish, then we can check what version we are running.

lib/openshift/cluster_version.rb
pruan-rht: done

lib/openshift/cluster_version.rb
pruan-rht: done

lib/openshift/cluster_version.rb
pruan-rht: Having this parameter will allow the user to check if a `target_version` upgrade status from the history.  It's possible to have multiple upgrades

lib/openshift/cluster_version.rb
pruan-rht: done

lib/openshift/cluster_version.rb
akostadinov: So we get a new condition by each change such that if we upgrade again, then we will still see the above conditions + new ones?

pruan-rht: @akostadinov FYI, during upgrade, https://github.com/openshift-qe/output_references/blob/master/clusterversion/clusterversion_during_upgarde.yml
and after upgarde https://github.com/openshift-qe/output_references/blob/master/clusterversion/upgarde_complete.yml . The `conditions` looks like always have the same amount of element.

akostadinov: but history sounds like will include more in case you do another upgrade? Is it possible to try?

pruan-rht: @akostadinov I'm not sure what you want me to explore here.

lib/openshift/cluster_version.rb
pruan-rht: With regards to `upgrade_timeout`.  I think I'll just take that parameter out since the wait method already has 2hrs as default.

lib/openshift/cluster_version.rb
pruan-rht: Anyway, the idea for this method is to track the current upgrade percentage, we may want to track that because I've seen cases in which the number fluctuates instead of going up only.  The percentage information will only be displayed during upgrade and in the `condition` section only.

pruan-rht:
@akostadinov, I've updated the PR, PTAL.  Thx!

lib/openshift/cluster_version.rb
akostadinov: ack, rename or remove works for me

akostadinov: other than that lgtm

lib/openshift/cluster_version.rb
```diff
@@ -6,14 +6,63 @@ def channel(user: nil, cached: true, quiet: false)
...
     end
 
-    def upstream(user: nil, cached: true, quiet: false)
+    def upstream(user: nil, cached: false, quiet: false)
       rr = raw_resource(user: user, cached: cached, quiet: quiet)
       rr.dig('spec', 'upstream')
     end
 
-    def version(user: nil, cached: true, quiet: false)
+    def version(user: nil, cached: false, quiet: false)
```
akostadinov: here and a couple of mroe places where still `cached: false`

pruan-rht: @akostadinov, I've updated the default to true for all calls except `upgrade_completed` and `wait_for_upgrade_completion` since I've been bitten when the `cache: true` for default for those two methods.  PTAL and merge if there are no more issues.

akostadinov: `wait_for_upgrade_completion` should **not** have the `cached` option at all. Wait method should call `upgrade_completed(..., cached: false)`. And `upgrade_completed` should default to `cached: true`. Generic workflow is that you *getter* methods will returned cached values when available. So if you call multiple methods you do not perform multiple calls to API. Point in time checks. When you want updated data, you need to call a getter method with `cached: false`.

lib/openshift/cluster_version.rb
```diff
@@ -15,5 +15,55 @@ def version(user: nil, cached: true, quiet: false)
...
+      res = self.history_matching(key: 'version', value: version)
+      if res.count > 0
+        logger.debug("MSG: #{res.first['message']}")
+        res.first['state'] == 'Completed'
+      else
+        false
+      end
+    end
+
+    def wait_for_upgrade_completion(version:, user: nil, cached: false, quiet: false, upgrade_timeout: 2*60*60)
```
akostadinov: did we talk about `s/upgrade_timeout/timeout/` or remove that parameter altogether?

lib/openshift/cluster_version.rb
pruan-rht: @akostadinov, I get it now... ok, PR updated, PTAL.  Thanks!

lib/openshift/cluster_version.rb
pruan-rht: I changed it to `timeout`

akostadinov:
Awesome, thanks a lot!

lib/openshift/cluster_version.rb
pruan-rht: @akostadinov similarly, should `cached` set to true for this method? https://github.com/openshift/verification-tests/blob/master/lib/openshift/resource.rb#L282

PR #148, opened 2019-04-04
[networking] initiate the multus related testing automation

bmeng:
testing log: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39584/console

@zhaozhanqi @anuragthehatter PTAL

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -631,3 +631,13 @@ Given /^an IP echo service is setup on the master node and the ip is stored in t
     raise "Failed to delete the docker container." unless @result[:success]
   }
 end
+
+
+Given /^the multus is enabled on the cluster$/ do
+  ensure_admin_tagged
+
+  desired_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:desired]
+  available_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:available]
+
+  raise "Multus is not running correctly!" unless desired_multus_replicas == available_multus_replicas && available_multus_replicas != 0
+end
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -1,0 +1,57 @@
+Feature: Multus-CNI related scenarios
+
+  # @author bmeng@redhat.com
+  # @case_id OCP-21151
+  @admin
+  Scenario: Create pods with multus-cni - macvlan bridge mode
+    # Make sure that the multus is enabled
+    Given the master version >= "4.0"
+    And the multus is enabled on the cluster
+
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml |
+      | n | <%= project.name %> |
+    Then the step should succeed
+
+    # Create the first pod which consumes the macvlan custom resource
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-bridge.yaml |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | name=macvlan-bridge-pod |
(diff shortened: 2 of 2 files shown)


features/step_definitions/networking.rb
```diff
@@ -631,3 +631,24 @@
     raise "Failed to delete the docker container." unless @result[:success]
   }
 end
+
+
+Given /^the multus is enabled on the cluster$/ do
+  ensure_admin_tagged
+
+  @result = admin.cli_exec(:get, resource: "daemonset", resource_name: "multus", n: "openshift-multus", template: "{{.status.desiredNumberScheduled}}")
```
pruan-rht: @bmeng, we can leverage off the existing `daemon_set` class to get these parameters which can shorten this step to 
```
  org_user = user
  org_project = project if @project
  project('openshift-multus')
  desired = daemon_set('multus').replica_counters[:desired]
  available = daemon_set('multus').replica_counters[:available]
  raise "Multus is not running correctly!" unless available == desired
  @user = org_user
  @project = org_project if org_project
```

pruan-rht: @bmeng the only downside is that we have to take care of project and user switching within the step.  It looks cleaner IMO, WDYT @akostadinov?

akostadinov: I don't see the need for `I switch to cluster admin pseudo user` so restoring user shouldn'e be needed.
```
desired = daemon_set('multus').replica_counters(user: admin)[:desired]
```

Restoring project is not a big deal.

pruan-rht: @akostadinov, accessing this `openshift-multus` project needs admin privilege, so I think we need to switch the user since we can't guarantee the user has admin privilege prior to calling the method.

pruan-rht: hmm...actually, I may be wrong about the privilege...need to check it later.

akostadinov: that's why I suggest `daemon_set('multus').replica_counters(user: admin)[:desired]`. This means query as cluster admin.

pruan-rht: ah, I didn't see that suggestion earlier... yeah, that looks cleaner and we can remove the switching part which results in only 3 lines (1 if we cramp everything inline)
```
desired = daemon_set('multus').replica_counters[:desired]
available = daemon_set('multus').replica_counters[:available]
raise "Multus is not running correctly!" unless available == desired
```
sweet!

bmeng:
@pruan-rht @akostadinov I have updated the new step to make it more simpler. Please take a look. Thanks.

features/step_definitions/networking.rb
```diff
@@ -631,3 +631,13 @@
...
 end
+
+
+Given /^the multus is enabled on the cluster$/ do
+  ensure_admin_tagged
+
+  desiredMultusReplicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:desired]
+  availableMultusReplicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:available]
+
+  raise "Multus is not running correctly!" unless desiredMultusReplicas == availableMultusReplicas
```
zhaozhanqi: How about add availableMultusReplicas != 0. eg:
`availableMultusReplicas != 0 && desiredMultusReplicas == availableMultusReplicas`

bmeng: Yeah, I think it is a good point, I have updated the step to check the available number is not 0.

Thanks.

features/step_definitions/networking.rb
```diff
@@ -631,3 +631,13 @@
     raise "Failed to delete the docker container." unless @result[:success]
   }
 end
+
+
+Given /^the multus is enabled on the cluster$/ do
+  ensure_admin_tagged
+
+  desiredMultusReplicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:desired]
```
pruan-rht: @bmeng looks good, one minor issue is that it looks like you are ready to jump into Golang already :), but for ruby the naming convention for variables are snake case (https://github.com/rubocop-hq/ruby-style-guide#snake-case-symbols-methods-vars)

bmeng: Aha, yeah... read golang code more recently.

Have updated the names. Thanks.

PR #150, opened 2019-04-09
[networking] automated more macvlan cases

bmeng:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39708/console

Two more cases for macvlan with different modes.

@zhaozhanqi @anuragthehatter PTAL. Thanks.

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -55,3 +55,118 @@ Feature: Multus-CNI related scenarios
       | curl | --connect-timeout | 5 | <%= cb.pod1_multus_ip %>:8080 |
     Then the step should succeed
     And the output should contain "Hello OpenShift"
+
+
+  # @author bmeng@redhat.com
+  # @case_id OCP-21489
+  @admin
+  Scenario: Create pods with multus-cni - macvlan private mode
+    # Make sure that the multus is enabled
+    Given the master version >= "4.0"
+    And the multus is enabled on the cluster
+
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    When I run the :create admin command with:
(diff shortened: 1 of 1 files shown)


features/networking/multus.feature
```diff
@@ -55,3 +55,118 @@ Feature: Multus-CNI related scenarios
...
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-private.yaml" replacing paths:
+      | ["spec"]["nodeName"] | "<%= cb.pod_node %>" |
+    Then the step should succeed
+    And 2 pods become ready with labels:
+      | name=macvlan-private-pod |
+    And evaluation of `pod(-1).name` is stored in the :pod2 clipboard
+
+    # Try to access both the cluster ip and macvlan ip on pod1 from pod2
+    When I execute on the "<%= cb.pod2 %>" pod:
+      | curl | --connect-timeout | 5 | <%= cb.pod1_sdn_ip %>:8080 |
```
pruan-rht: @bmeng, what's the reason for setting the specific timeout period to 5 seconds?

anuragthehatter: I guess he don't want curl to wait forever for response which might block the automation resources but limiting curling time to 5 seconds. I guess curl sometimes go in forever waiting state. If he don't get curl response in 5 seconds the subsequent steps will eventually fail anyhow

pruan-rht: is 5 seconds enough?

bmeng: Yeah, just like @anuragthehatter said, we just want it exit in a time limit if the target cannot be reached via the curl, unless it will wait for a quite long time.

And if the access succeeded, the curl request should be finished in less than 0.1 second.

pruan-rht: ok.  LGTM.

PR #151, opened 2019-04-10
add case OCP-21690 for tsb

zihantang-rh:
Add case OCP-21690 for tsb v3.x
Remove OCP-15486 which is automated using many steps from UI.

Here's the log
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39716/console

@chengzhang1016 @akostadinov @pruan-rht please help review and merge, thanks.

### DIFF
--- a/features/service_catalog_tsb/tsb.feature
+++ b/features/service_catalog_tsb/tsb.feature
@@ -1,52 +1,15 @@
 Feature: Template service broker related features
 
-  # @author xiuwang@redhat.com
-  # @case_id OCP-15486
-  Scenario: Deprovision serviceinstance when multiple servicebinding existed 
-    Given the master version >= "3.7"
-    And I have a project
-    When I run the :goto_home_page web console action
-    Then the step should succeed
-    When I perform the :provision_serviceclass_with_binding_on_homepage web console action with:
-      | primary_catagory | Databases          |
-      | sub_catagory     | MySQL              |
-      | service_item     | MySQL (Ephemeral)  |
-    Then the step should succeed
-    When I get project serviceinstances as JSON
-    And evaluation of `service_instance.name` is stored in the :svcinstancename clipboard
-    Given I wait for the "<%= cb.svcinstancename %>" service_instance to become ready up to 180 seconds
-    And a pod becomes ready with labels:
-      | name=mysql |
-    When I get project servicebinding as JSON
-    And evaluation of `service_binding.name` is stored in the :svcbindingname clipboard
-    When I run the :delete client command with:
-      | object_type       | servicebinding           |
-      | object_name_or_id | <%= cb.svcbindingname %> |
-    Then the step should succeed
-    And I wait for the resource "servicebinding" named "<%= cb.svcbindingname %>" to disappear
-    When I perform the :create_binding_on_overview_page web console action with:
-      | project_name  | <%= project.name %> |
-      | resource_name | MySQL               |
-    Then the step should succeed
-    When I perform the :create_binding_on_overview_page web console action with:
-      | project_name  | <%= project.name %> |
-      | resource_name | MySQL               |
-    Then the step should succeed
-    Given I wait for all servicebindings in the project to become ready
-    Given I get project secrets
-    And the output should match 4 times:
-      | mysql.*Opaque | 
-    When I perform the :goto_overview_page web console action with:
-      | project_name  | <%= project.name %> |
-    Then the step should succeed
-    When I perform the :delete_serviceinstance_on_overview_page web console action with:
-      | resource_name | MySQL (Ephemeral)   |
-      | project_name  | <%= project.name %> |
-    Then the step should succeed
-    And I wait for the resource "serviceinstance" named "<%= cb.svcinstancename %>" to disappear within 300 seconds
-    Given I get project secrets
-    And the output should not contain "mysql"
-    Given I get project templateinstances
-    And the output should match:
-      | No resources found |
-
+  # @author zitang@redhat.com
+  # @case_id OCP-21690
+  @admin
+  Scenario: [CVP] Clusterserviceclass and clusterserviceplan of templateinstance were created 
+    Given admin checks that the "template-service-broker" cluster_service_broker exists
+    When I run the :get client command with:
+      | resource | clusterserviceclass                                                       |
+      | o        | custom-columns=NAME:.metadata.name,EXTERNAL\ NAME:.spec.externalName,BROKER:.spec.clusterServiceBrokerName |
+    Then the output should contain "template-service-broker"
+    When I run the :get client command with:
+      | resource | clusterserviceplan                                                      |
+      | o        | custom-columns=NAME:.metadata.name,EXTERNAL\ NAME:.spec.externalName,BROKER:.spec.clusterServiceBrokerName |
+    Then the output should contain "template-service-broker"
--- a/features/service_catalog_tsb/tsb.feature
+++ b/features/service_catalog_tsb/tsb.feature
@@ -10,5 +10,4 @@ Feature: Template service broker related features
       | o        | custom-columns=NAME:.metadata.name,EXTERNAL\ NAME:.spec.externalName,BROKER:.spec.clusterServiceBrokerName |
     Then the output should contain "template-service-broker"
     Given cluster service classes are indexed by external name in the :csc clipboard
-    Then the expression should be true> ! cb.csc.values.find {|c| c.cluster_svc_broker_name == "template-service-broker"}.nil?
-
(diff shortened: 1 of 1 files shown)


chengzhang1016:
@pruan-rht The design of TC OCP-15486 is not well for us, broker team always focus on backend testing, TC design should design from backend, but not UI. We will re-design and automate this TC in later. 
Furthermore, check points of UI will be covered by web console sub-team.

features/service_catalog_tsb/tsb.feature
```diff
@@ -1,52 +1,15 @@
...
-    Given I get project templateinstances
-    And the output should match:
-      | No resources found |
-
+  # @author zitang@redhat.com
+  # @case_id OCP-21690
+  @admin
+  Scenario: [CVP] Clusterserviceclass and clusterserviceplan of templateinstance were created 
+    Given admin checks that the "template-service-broker" cluster_service_broker exists
+    When I run the :get client command with:
```
pruan-rht: @zihantang-rh try to use the existing class methods to get information instead of using `:get` client call directly.

pruan-rht: @zihantang-rh I've submitted a PR #153 which should help you to shorten the scenario.  You will use it like the following:
```
Given evaluation of `BushSlicer::ClusterServiceClass.list(user: admin).first` is stored in the :entry clipboard
Then the expression should be true> cluster_service_class(cb.entry.name).cluster_svc_broker_name == 'template-service-broker'
Then the expression should be true> cluster_service_plan(cb.entry.name).cluster_svc_broker_name == 'template-service-broker'
```

zihantang-rh: In this scenario, I'm not sure which clusterserviceclass is from `template-service-broker`,
````
# oc get clusterserviceclass
NAME                                   EXTERNAL-NAME                                       BROKER                    AGE
615e40bd-5c02-11e9-bd52-0e715c043546   mariadb-ephemeral                                   template-service-broker   6h
...
6aeae718-5c03-11e9-bd52-0e715c043546   registry-console                                    template-service-broker   6h
....
7b8ec55310e26eda60bb681ef9dbf066       quay-postgresql-apb                                 ansible-service-broker    40m

```
So I can not use `ClusterServiceClass.list(user:admin).first` to get the exact one. 
Just to check there will be one class from `template-service-broker` is enough .

akostadinov: Maybe 
```
Given cluster service classes are indexed by external name in the clipboard
Then the expression should be true> ! cb.csc.values.find {|c| c.external_name == "template-service-broker"}.empty?
```
While not much easier to read, it makes more clear what we are testing. And that is 
1. we have a csc with external name `template-service-broker` 
2. it has at least one plan 

This is what I understand, correct me if I'm wrong.

pruan-rht: @akostadinov, nice...didn't noticed the method `cluster service classes are indexed by external name...` , that works but the testcase is also checking on the `clusterserviceplan` object.

akostadinov: I assume all plans of a class have same external name. But I don't know much about these objects.

pruan-rht: @zihantang-rh, please confirm if Alex's suggestion works for you.

zihantang-rh:
```
Then the expression should be true> ! cb.csc.values.find {|c| c.cluster_svc_broker_name == "template-service-broker"}.nil?
```
this step works for me.
I will try clusterserviceplan later.

pruan-rht: @zihantang-rh, Alex's question is if `all plans of a class have the same external name`.  If that's the case, then we don't need to check the `clusterserviceplan` since checking `cluserserviceclass` would've been enough.

zihantang-rh: @pruan-rht this case is for 'template-service-broker', all plans have same external name but different `name`, `clusterserviceplan` and `clusterserviceclass` are different objects. so need double check. I'll copy a step like `Given cluster service plans are indexed by external name in the clipboard` later

zihantang-rh: I tried to add a step for clusterserviceplan like：
````
# indexed listing
Given /^cluster service plans are indexed by name in the#{OPT_SYM} clipboard$/ do | cb_name |
  cb_name ||= :csc_list
  cb[cb_name] = CucuShift::ClusterServicePlan.list(user: admin).map { |csp|
    [csp.name, csp]
  }.to_h
  cache_resources *cb[cb_name].values
end
````
but when running, got the following error after it indexed some plans.
```  
    stack level too deep (SystemStackError)
      /home/zitang/project/cucushift/lib/openshift/cluster_resource.rb:23:in `default_user'
      /home/zitang/project/cucushift/lib/openshift/resource.rb:65:in `get'
      /home/zitang/project/cucushift/lib/openshift/resource.rb:43:in `exists?'
```
there're more than 200 plans from tsb.
I so I still want to use `get` to check clusterserviceplan

**Comment by zihantang-rh** [comment]
Add new log:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39853/console

File: features/service_catalog_tsb/tsb.feature (hunk 1 above)
Comment (akostadinov) [reply features/service_catalog_tsb/tsb.feature]: if multiple plans have same external name, then indexing will not work well.
You rather needs something like:
```
cb.csc[my_ext_name].plans.first.external_name == my_ext_name
```

Comment (zihantang-rh) [reply features/service_catalog_tsb/tsb.feature]: for plans, I didn't use external name to index, I use `name` (unique) to index.
`Given /^cluster service plans are indexed by name in the#{OPT_SYM} clipboard$/`

Comment (akostadinov) [reply features/service_catalog_tsb/tsb.feature]: Still I don't see why do we need to index them by name. Did you try to run check as I suggested?

Comment (akostadinov) [reply features/service_catalog_tsb/tsb.feature]: btw need a longer stack trace to see where issue comes from

Comment (zihantang-rh) [reply features/service_catalog_tsb/tsb.feature]: The checking for clusterserviceplan in this case, is just need to check there's at least 1 plan from `template-service-broker`.
As suggested, If not using `get` method, It might be checked as clusterserviceclass like
````
    Given cluster service classes are indexed by external name in the :csc clipboard
    Then the expression should be true> ! cb.csc.values.find {|c| c.cluster_svc_broker_name == "template-service-broker"}.nil?
````
since plan's external name is the same, so I use name to index to store. then checking the `cluster_svc_broker_name`.
but `stack error` happens in the middle of `Given cluster service plans are indexed by name in the :csp clipboard` (some stored, some have not)

btw, there are 260+ clusterserviceplan and the data is more detailed than clusterserviceclass, I prefer to directly use `get` methord to check.
```
      | resource | clusterserviceplan                                                      |
      | o        | custom-columns=NAME:.metadata.name,EXTERNAL\ NAME:.spec.externalName,BROKER:.spec.clusterServiceBrokerName |
    Then the output should contain "template-service-broker"
```

Comment (akostadinov) [reply features/service_catalog_tsb/tsb.feature]: That's why I suggested `cb.csc[my_ext_name].plans.first.external_name == my_ext_name` without adding any further steps. We have a step that works, we can find plans based on the classes. There is no need to index plans by name.

But I think in this case getting this absolutely the best way is not so productive so will merge as is.

File: features/service_catalog_tsb/tsb.feature [hunk 2]
```diff
@@ -1,52 +1,14 @@
...
+      | resource | clusterserviceplan                                                      |
+      | o        | custom-columns=NAME:.metadata.name,EXTERNAL\ NAME:.spec.externalName,BROKER:.spec.clusterServiceBrokerName |
+    Then the output should contain "template-service-broker"
+    Given cluster service classes are indexed by external name in the :csc clipboard
+    Then the expression should be true> ! cb.csc.values.find {|c| c.cluster_svc_broker_name == "template-service-broker"}.nil?
```
akostadinov: Actually I'd edit this as it is double negation to:
```
cb.csc.values.find {|c| c.cluster_svc_broker_name == "template-service-broker"}
```

features/service_catalog_tsb/tsb.feature
akostadinov: I see you have edited PR anyway. Either case it can be merged, thank you.

akostadinov: wait, where did `#cluster_svc_broker_name` method come from?

zihantang-rh: Peter implemented this in PR https://github.com/openshift/verification-tests/pull/153

akostadinov: got it, I didn't see it locally and thought something was missing

PR #153, opened 2019-04-10
added methods to get clusterSerivceBrokerName

pruan-rht:
helps #151


akostadinov:
v3 only?

/lgtm

pruan-rht:
@akostadinov, yes, this object is not present in 4.x

pruan-rht:
@akostadinov, yes, this object is not present in 4.x

PR #155, opened 2019-04-11
change condition check to default to cached: false

pruan-rht:
@akostadinov, WDYT?

### DIFF
--- a/lib/openshift/resource.rb
+++ b/lib/openshift/resource.rb
@@ -286,8 +286,8 @@ module BushSlicer
 
     # returns a hash simliar to
     # {"lastTransitionTime"=>2019-03-23 07:33:34 UTC, "status"=>"False", "type"=>"Progressing"}
-    def condition(type: nil, user: nil, cached: true, quiet: false)
-      res = self.conditions.select { |c| c['type'] == type }
+    def condition(type: nil, user: nil, cached: false, quiet: false)
+      res = self.conditions(user: user, cached: cached, quiet: quiet).select { |c| c['type'] == type }
       res.first
     end
 
--- a/lib/openshift/resource.rb
+++ b/lib/openshift/resource.rb
@@ -286,7 +286,7 @@ module BushSlicer
 
     # returns a hash simliar to
     # {"lastTransitionTime"=>2019-03-23 07:33:34 UTC, "status"=>"False", "type"=>"Progressing"}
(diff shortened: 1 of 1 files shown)


lib/openshift/resource.rb
```diff
@@ -286,8 +286,8 @@ def conditions(user: nil, cached: true, quiet: false)
 
     # returns a hash simliar to
     # {"lastTransitionTime"=>2019-03-23 07:33:34 UTC, "status"=>"False", "type"=>"Progressing"}
-    def condition(type: nil, user: nil, cached: true, quiet: false)
-      res = self.conditions.select { |c| c['type'] == type }
+    def condition(type: nil, user: nil, cached: false, quiet: false)
```
akostadinov: can we plese default to `true`?

akostadinov: if we rely on defaults in called methods we end up with missing parameter passing as it seems the case ;)

pruan-rht: @akostadinov  I prefer to have it `false` because it’s a dynamic value…don’t want to assume the caller of the method to set that value.  For example https://github.com/openshift/verification-tests/blob/master/lib/openshift/resource.rb#L261

pruan-rht: @akostadinov as an actual example, @anuragthehatter was using the method and was getting staled data currently.

akostadinov: everything is dynamyc. Most everything is dynamic. He must remember that values are cached unless enforced.

pruan-rht: true, some are more dynamic than others. :)

akostadinov: See, in the beginning tried to differentiate and at the end it was a mess. pls take as a requirement that values are cached by default. Wait methods should hardcode `cached: false`.

akostadinov: usually makes scenario inline checks shorter also

pruan-rht: updated...PTAL... thx!

akostadinov: thank you!

anuragthehatter: Thanks @pruan-rht @akostadinov. Enforcing cached wherever needed sounds good to me. Defaulted to true is okay.

PR #156, opened 2019-04-12
Added Scenario for OCP-22706

anuragthehatter:
This is negative usecase to make sure that network operator version is still reflected post bad network config 
Ref: OCP-22706
Test logs: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/325/console
@bmeng @zhaozhanqi PTAL

### DIFF
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -17,6 +17,7 @@ Feature: Operator related networking scenarios
   # @auther anusaxen@redhat.com
   # @case_id OCP-22706
   @admin
+  @destructive
   Scenario: The clusteroperator should be able to reflect the correct version field post bad network operator config
 
     Given the master version >= "4.0"
@@ -38,8 +39,14 @@ Feature: Operator related networking scenarios
     And evaluation of `cluster_operator('network').condition(type: 'Failing',cached: false)` is stored in the :failing_status_post_patch clipboard
     Then the expression should be true> cb.failing_status_post_patch["status"]=="True"
     And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)
-
-    #Fixing bad patch to make cluster healthy before exiting the test
+    
+    #Registering clean-up steps to move networkType back to OpenShiftSDN and to check Failing status is False before test exits
+    Given I register clean-up steps:
+    """
+    20 seconds have passed
+    evaluation of `cluster_operator('network').condition(type: 'Failing',cached: false)` is stored in the :failing_status clipboard
+    the expression should be true> cb.failing_status["status"]=="False"
+    """
     Given I register clean-up steps:
     """
     When I run the :patch admin command with:
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -13,3 +13,39 @@ Feature: Operator related networking scenarios
     Then the expression should be true> cb.operator_status["status"]=="True"
     #Confirm whether network operator version matches with ocp version
     And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)
+
+  # @auther anusaxen@redhat.com
+  # @case_id OCP-22706
+  @admin
+  Scenario: The clusteroperator should be able to reflect the correct version field post bad network operator config
+
+    Given the master version >= "4.0"
+    #Getting OCP version
+    Given evaluation of `cluster_version('version').version` is stored in the :ocp_version clipboard
+    #Making sure that operator is not Failing before proceesing further steps
+    And evaluation of `cluster_operator('network').condition(type: 'Failing')` is stored in the :failing_status_before_patch clipboard
+    Then the expression should be true> cb.failing_status_before_patch["status"]=="False"
+    #Editing networks.config.openshift.io cluster to reflect bad config like changing networktype from OpenShiftSDN to OpenShift
+    When I run the :patch admin command with:
+      | resource      | networks.config.openshift.io         |
+      | resource_name | cluster                              |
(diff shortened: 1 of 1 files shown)


features/networking/operator.feature
```diff
@@ -13,3 +13,39 @@ Feature: Operator related networking scenarios
     Then the expression should be true> cb.operator_status["status"]=="True"
     #Confirm whether network operator version matches with ocp version
     And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)
+
+  # @auther anusaxen@redhat.com
+  # @case_id OCP-22706
+  @admin
```
zhaozhanqi: this case is destructive, so we should add @destructive

features/networking/operator.feature
```diff
@@ -13,3 +13,39 @@ Feature: Operator related networking scenarios
...
+    Then the step should succeed
+
+    #Normally it takes 5-10 seconds for network config update to reconcile across the cluster but taking 20 seconds wait to make sure that Failing status becomes True post bad patch
+    Given 20 seconds have passed
+    And evaluation of `cluster_operator('network').condition(type: 'Failing',cached: false)` is stored in the :failing_status_post_patch clipboard
+    Then the expression should be true> cb.failing_status_post_patch["status"]=="True"
+    And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)
+
+    #Fixing bad patch to make cluster healthy before exiting the test
+    Given I register clean-up steps:
```
bmeng: Ususally, we will register the clean up steps right after the step which needs to be cleaned. And it will also be executed in the end of scenario.

features/networking/operator.feature
```diff
@@ -13,3 +13,39 @@ Feature: Operator related networking scenarios
...
+      | p             | {"spec":{"networkType":"OpenShift"}} |
+      | type          | merge                                |
+    Then the step should succeed
+
+    #Normally it takes 5-10 seconds for network config update to reconcile across the cluster but taking 20 seconds wait to make sure that Failing status becomes True post bad patch
+    Given 20 seconds have passed
+    And evaluation of `cluster_operator('network').condition(type: 'Failing',cached: false)` is stored in the :failing_status_post_patch clipboard
+    Then the expression should be true> cb.failing_status_post_patch["status"]=="True"
+    And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)
+
```
bmeng: I think we also need to make sure that the Failing status back to False in the end of scenario.

features/networking/operator.feature
bmeng: please also add `@destructive` tag here, since the modify network operator config may bring the cluster into problem status for a short while.

bmeng: +1

anuragthehatter: Agree. Thanks for pointing out that

anuragthehatter: Done!

features/networking/operator.feature
anuragthehatter: Thanks! done.

features/networking/operator.feature
anuragthehatter: Bo, apparently the cleanup steps right after step won't work in this scenario, (the bad patch is applied first and then Failing status is expected true. We want same seq to bring cluster back to normal - good patch then failing false) I have appended them in the end. LMK in my latest commit. If i applied cleanup post steps, the failing status will be checked first and patch will transpire post that.

bmeng: These two clean up blocks can be merged into one. Like
```
When I run the :patch admin command with:
Then the step should succeed
20 seconds have passed
evaluation ......
the expression ......
```

bmeng: Re the original comments, the clean-up step will always be executed in the end of the scenario. No matter you registered it right after the change or in the current position.
But usually, we register the clean-up steps right after the steps which make the change. So it is easy to know the what will be cleaned up by the clean-up step. Especially when the scenario has lots of steps.

Eg,
```
When I change some node config
Then the step succeeded
And I restart the node service
I register clean-up steps:
"""
Revert the change of the node config
The step succeeded
I restart the node service
"""
I do some testing based on the changed node config
```

The testing in the last line will also be tested with the changed node config, and the node config will be changed back once the whole scenario finished.

anuragthehatter: Understand that @bmeng but combining both clean-up steps didn't work for me https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/329/console. @pruan-rht any thoughts why combined clean-up steps doesn't work?

anuragthehatter: Okay so it seems like cleanup steps doesn't like linespaces. Something new i learnt :). I just fixed that in latest commit. Thanks!

bmeng: I think you still did not put the clean up step to the `right after` the changes. For this case you should put the clean-up steps before:
```
#Normally it takes 5-10 seconds for network config update to reconcile across the cluster but taking 20 seconds wait to make sure that Failing status becomes True post bad patch
```
Another reason I can say that we need to register the clean up step right after the changes is, if the following checks failed, we still be able to invoke the clean up step to fix the cluster.

akostadinov: basically registering clean-up steps at the end of scenario makes little practical sense (although sometimes it could - to make sure execution stops altogether)

bmeng:
LGTM

PR #157, opened 2019-04-12
update no project return step for 4.x

stuartchuan:
Update no project return step for 4.x

### DIFF
--- a/features/step_definitions/resources.rb
+++ b/features/step_definitions/resources.rb
@@ -174,7 +174,7 @@ Given /^I wait for the resource "(.+)" named "(.+)" to disappear(?: within (\d+)
     res = user.cli_exec(:get, **opts)
     case res[:response]
     # the resource has terminated which means we are done waiting.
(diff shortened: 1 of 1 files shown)


features/step_definitions/resources.rb
```diff
@@ -174,7 +174,7 @@
     res = user.cli_exec(:get, **opts)
     case res[:response]
     # the resource has terminated which means we are done waiting.
-    when /cannot get project/, /not found/, /No resources found/
+    when /Error from server (Forbidden): projects.project.openshift.io/, /is forbidden/
```
akostadinov: if we run this step as cluster admin, then it will likely return `not found` or `No resources found` or something. Better to still cover that use case. Also this step is used not only for projects but other other objects. So maybe only `cannot get project` needs to be changed to `Error from server (Forbidden): projects.project.openshift.io` and the others kept intact.

stuartchuan: Updated as comments.

PR #161, opened 2019-04-16
[networking] automated multus cases for multiple cni plugins

bmeng:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39878/console

Add a new step to run network related command on the sdn pod (since the sdn pod is running as hostnetwork).

Automated some more multus test cases.

@zhaozhanqi @anuragthehatter PTAL

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -174,6 +174,7 @@ Feature: Multus-CNI related scenarios
   # @author bmeng@redhat.com
   # @case_id OCP-21853
   @admin
+  @destructive
   Scenario: Create pods with multus-cni - host-device
     # Make sure that the multus is enabled
     Given the master version >= "4.0"
@@ -181,34 +182,14 @@ Feature: Multus-CNI related scenarios
     And I select a random node's host
     And evaluation of `node.name` is stored in the :target_node clipboard
 
-    # Add a host-device which will be used by the pod later
-    Given I run the :get admin command with:
-      | resource      | pod                                 |
-      | n             | openshift-sdn                       |
-      | fieldSelector | spec.nodeName=<%= cb.target_node %> |
-      | l             | app=sdn                             |
-      | o             | jsonpath={.items[0].metadata.name}  |
-    Then the step should succeed
-    And evaluation of `@result[:response]` is stored in the :target_sdnpod clipboard
-
-    Given I run the :exec admin command with:
-      | pod              | <%= cb.target_sdnpod %>                              |
-      | n                | openshift-sdn                                        |
-      | oc_opts_end      |                                                      |
-      | exec_command     | sh                                                   |
-      | exec_command_arg | -c                                                   |
-      | exec_command_arg | ip link add eth08 link eth0 type macvlan mode bridge |
+    # Prepare the net link on the node which will be attached to the pod
+    When I run network command on the "<%= cb.target_node %>" node's sdn pod:
+       | sh | -c | ip link add eth08 link eth0 type macvlan mode bridge |
     Then the step should succeed
     Given I register clean-up steps:
     """
-    I run the :exec admin command with:
-      | pod              | <%= cb.target_sdnpod %> |
-      | n                | openshift-sdn           |
-      | oc_opts_end      |                         |
-      | exec_command     | sh                      |
-      | exec_command_arg | -c                      |
-      | exec_command_arg | ip link del eth08       |
-    the step should succeed
+    I run network command on the "<%= cb.target_node %>" node's sdn pod:
+       | sh | -c | ip link del eth08 |
     """
 
     # Create the net-attach-def via cluster admin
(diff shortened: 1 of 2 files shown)


features/networking/multus.feature
```diff
@@ -255,12 +231,168 @@ Feature: Multus-CNI related scenarios
...
+    """
+
+    # Create the pod which consumes two hostdev custom resources
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/2interface-hostdevice-hostdevice.yaml" replacing paths:
+      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | host-device, host-device-2 |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | name=two-host-device-pod |
+    Given I register clean-up steps:
```
zhaozhanqi: here delete the pod means: it's high priority when entry the clean up flow after the scenario?

bmeng: We need to delete the pod to release the host-device to the host, and then delete the net link from the host. So delete the pod has higher priority.

features/networking/multus.feature
```diff
@@ -240,27 +223,185 @@ Feature: Multus-CNI related scenarios
...
+  Scenario: Create pods with muliple cni plugins via multus-cni - macvlan + macvlan
+    # Make sure that the multus is enabled
+    Given the master version >= "4.0"
+    And the multus is enabled on the cluster
+
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml |
+      | n | <%= project.name %> |
```
akostadinov: I assume this is a namespaced resource, otherwise we can again get name collisions.

features/networking/multus.feature
```diff
@@ -240,27 +223,185 @@ Feature: Multus-CNI related scenarios
...
+    the step should succeed
+    """
+
+    # Create the pod which consumes both hostdev and macvlan custom resources
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/2interface-macvlan-hostdevice.yaml" replacing paths:
+      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | name=macvlan-hostdevice-pod |
+    Given I register clean-up steps:
```
akostadinov: IMO clean-up step should be before pod becomes ready.

features/networking/multus.feature
```diff
@@ -240,27 +223,185 @@ Feature: Multus-CNI related scenarios
...
+    """
+    I run the :delete client command with:
+      | object_type | pod                         |
+      | l           | name=macvlan-hostdevice-pod |
+    the step should succeed
```
akostadinov: maybe add step `all existing pods die with labels:`

features/networking/multus.feature
akostadinov: Here also I think clean-up should register before we check pod is ready.

features/step_definitions/networking.rb
```diff
@@ -641,3 +641,16 @@
 
   raise "Multus is not running correctly!" unless desired_multus_replicas == available_multus_replicas && available_multus_replicas != 0
 end
+
+
+Given /^I run network command on the#{OPT_QUOTED} node's sdn pod:$/ do |node_name, table|
+  ensure_admin_tagged
+  network_cmd = table.raw
+  node_name = node_name
```
akostadinov:
```
node_name = node.name
```

features/step_definitions/networking.rb
```diff
@@ -641,3 +641,16 @@
 
   raise "Multus is not running correctly!" unless desired_multus_replicas == available_multus_replicas && available_multus_replicas != 0
 end
+
+
+Given /^I run network command on the#{OPT_QUOTED} node's sdn pod:$/ do |node_name, table|
```
akostadinov: I think we can skip word `network` from step name because I see you can run any command. Step does not seems pecific to network commands. You can happily run `ls`, `cat` or whatever from what I see.

bmeng: It is ok to use which step name for me. Just in this case, only network commands make sense since the sdn pod is using hostNetwork, means network related commands run in the pod are same as run them on the node.

features/networking/multus.feature
bmeng: Yes, it is a namespaced resource.

bmeng:
@akostadinov fixed the comments, please help check

bmeng:
@akostadinov Can you help review this? I have fixed the conflicts.

features/step_definitions/networking.rb
```diff
@@ -661,3 +661,16 @@
 
   raise "The status of condition #{type} is incorrect." unless expected_status == real_status
 end
+
+
+Given /^I run command on the#{OPT_QUOTED} node's sdn pod:$/ do |node_name, table|
+  ensure_admin_tagged
+  network_cmd = table.raw
+  node_name = node.name
```
akostadinov: ops, missed this last time. 
```
node_name ||= node.name
```

Otherwise the node name specified in step will not be used but we always use current node.

PR #163, opened 2019-04-17
clusterlogging installation via OLM

pruan-rht:
Polarion: [OCP-21311](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-21311).  @anpingli @QiaolingTang, PTAL.  Thanks!

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -1,9 +1,29 @@
 ### logging related step definitions
 # for 4.x we default logging installation is via OLM only
 
+### none configurable, just use default parameters
 Given /^logging service has been installed successfully$/ do
   ensure_destructive_tagged
+  crd_yaml = "https://raw.githubusercontent.com/operator-framework/operator-metering/master/manifests/metering-config/default.yaml"
+  step %Q/logging service is installed with:/, table(%{
+    | keep_installation | true            |
+    | crd_yaml          | <%= crd_yaml %> |
+  })
+end
 
+### configurable installation of logging
+# available parameters are:
+# :keep_installation: true/false   -- keep installation if true
+# :crd_yaml  -- http url for logging crd yaml
+# operator_group_yaml -- http url for operator group yaml
+# :catsrc_logging_yaml] -- http url for catalogsourceconfig for logging
+# :catsrc_elasticsearch_yaml] -- http url for catalogsourceconfig for ES
+# :sub_logging_yaml  -- http url for logging subscription yaml
+# :sub_elasticsearch_yaml -- http url for elastic search yaml
+#
+Given /^logging service is installed with:$/ do | table |
+  ensure_destructive_tagged
+  opts = opts_array_to_hash(table.raw)
   step %Q/I switch to cluster admin pseudo user/
   step %Q/I use the "openshift-marketplace" project/
   # first check packagemanifest exists for cluster-logging and elasticsearch-operator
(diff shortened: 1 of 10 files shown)


features/step_definitions/logging.rb
```diff
@@ -0,0 +1,128 @@
+### logging related step definitions
+# for 4.x we default logging installation is via OLM only
+
+Given /^logging service has been installed successfully$/ do
```
akostadinov: can we remove old code related to these two steps
```
there should be (\d+) (logging|metrics) services? installed
all deployer logging pods are running in the#{OPT_QUOTED} project
```

features/step_definitions/logging.rb
```diff
@@ -0,0 +1,128 @@
...
+    raise "Error creating namespace" unless @result[:success]
+  end
+  logging_ns = "openshift-logging"
+  operators_ns = "openshift-operators"
+
+  # project('openshift-logging')
+  # Create operator group in `openshift-logging` namespace
+  operator_group_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/OCP-21311/operator-group.yaml"
+  @result = admin.cli_exec(:create, f: operator_group_yaml, n: logging_ns)
+  raise "Error creating namespace" unless @result[:success]
```
akostadinov: maybe another error message

features/step_definitions/logging.rb
```diff
@@ -0,0 +1,128 @@
...
+  # create catalogsourceconfig for logging and elasticsearch
+  catsrc_logging_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/OCP-21311/csc-clusterlogigng.yaml"
+  catsrc_elasticsearch_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/OCP-21311/csc-elasticsearch.yaml"
+
+  @result = admin.cli_exec(:create, f: catsrc_logging_yaml)
+  raise "Error creating catalogsourceconfig for cluster_logging" unless @result[:success]
+  @result = admin.cli_exec(:create, f: catsrc_elasticsearch_yaml)
+  raise "Error creating catalogsourceconfig for elasticsearch" unless @result[:success]
```
akostadinov: two error messages exactly the same

features/step_definitions/logging.rb
```diff
@@ -0,0 +1,128 @@
...
+
+  @result = admin.cli_exec(:create, f: sub_logging_yaml) #, n: operators_ns)
+  raise "Error creating subscription for cluster_logging" unless @result[:success]
+  @result = admin.cli_exec(:create, f: sub_elasticsearch_yaml) #, n: operators_ns)
+  raise "Error creating subscription for elasticsearch" unless @result[:success]
+  # create instance in openshift-logging namespace
+  logging_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/example.yaml"
+  @result = admin.cli_exec(:create, f: logging_yaml, n: logging_ns)
+  # Check the resource created by OLM
+  step %Q/I wait for clusterlogging to be functional in the project/
```
akostadinov: shall we remove what we created at end of scenario? I see step is destructive tagged

lib/openshift/cluster_logging.rb
```diff
@@ -0,0 +1,107 @@
...
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'logStore')
+    end
+
+    def message(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'message')
+    end
+
+    def visualization(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'visualization')
+    end
```
akostadinov: what values do these return?

akostadinov: what I would like to is to have all low-level methods that return some raw YAML to be private. Users shouldn't parse raw structures. Then we can gradually add methods returning proper objects when needed.

Similar to how we have `QuotaLimits#memory_limit_raw` vs `QuotaLimits#memory_limit` methods. The raw method returns unprocessed structure. The non-raw method returns some more convenient object.

features/step_definitions/logging.rb
pruan-rht: For 4.x I think the file `features/step_definitions/logging_metrics.rb` can be removed.

features/step_definitions/logging.rb
pruan-rht: one is for `clusterlogging` the other is for  `elasticsearch`

features/step_definitions/logging.rb
pruan-rht: @akostadinov since with OLM install, the logging installation variations is much less than for ansible install for 3.x, I'm thinking for 4.x testing, we just install it once and leave it running.  @anpingli or @QiaolingTang should make a call on that front since they are the owner of those user stories.

features/step_definitions/logging.rb
akostadinov: yes, please remove dead code

features/step_definitions/logging.rb
pruan-rht: what do you mean?  A more explicit error message with the name of the namespace?

lib/openshift/cluster_logging.rb
pruan-rht: Those are part of the `status` return values.  Unlike `QuotaLimits`, they are not values but rather `yaml/json` structs.  You can look at it here https://github.com/openshift-qe/output_references/blob/master/logging/4.x/clusterlogging.yaml

akostadinov: I know, that's why I'm asking you to name methods `..._raw` and make `private` all that are never called outside the class. We can create deal with any external users when we have such.

features/step_definitions/logging.rb
akostadinov: this error message is same as line 19 but looks like we are not creating a namespace but a group or something

pruan-rht: oh, yeah, copy and paste error...it should say `operatorgroup` .  Good catch!

features/step_definitions/logging.rb
anpingli: @pruan-rht   It is better to split this step to cluster-logging operator ready and provide CRD xxx are provided.
Most of cases are as following
1.  Cluster logging operator is ready.
2. Provide different type of resource file or modify the current resource file
So it is very importance to have a Cluster logging operator ready step.

features/step_definitions/logging.rb
anpingli: @pruan-rht  Most of time, we have to remove the clusterlogging/instance 1). it is better to clean it each time.
We needn't to clear the clusterlogging operator/elasticsearch operator and namespace each time.

1) To clean up clusterlogging, just remove the clusterlogging/instance resource  
#oc delete clusterlogging/instance -n openshift-logging
2) To clean up clusterlogging operator, we need to unsubscribe the operator
3) To clean up elasicsearch operator, we need to unsubscribe the operator
     
     
To clean up all resource

pruan-rht: @anpingli Thanks for the feedback.  If you do recommend removing after each install, perhaps we can add an option like `keep_installation: true` to the step and default to remove all previous installations to make it simpler.

anpingli: Looks good to have a option keep_installation: true

pruan-rht:
@akostadinov PR udpated.  PTAL again.  Thanks!

features/step_definitions/logging.rb
akostadinov: I still see `namespace` here.

features/step_definitions/logging.rb
akostadinov: Do you intend to make clean-up optional? Or do we drop the idea?

features/step_definitions/logging.rb
akostadinov: I don't see old file removed. Maybe missed to commit?

lib/openshift/cluster_logging.rb
```diff
@@ -0,0 +1,107 @@
...
+      self.fluentd_pods(user: user, cached: cached, quiet: quiet)['ready']
+    end
+
+    def fluentd_failed_pods(user: nil, cached: true, quiet: false)
+      self.fluentd_pods(user: user, cached: cached, quiet: quiet)['failed']
+    end
+
+    def fluentd_notready_pods(user: nil, cached: true, quiet: false)
+      self.fluentd_pods(user: user, cached: cached, quiet: quiet)['notReady']
+    end
```
akostadinov: All `fluentd*_pods` methods can be private as far as I can tell. If they are to be public, then they better return `Pod` objects instead of some unknown structure.

lib/openshift/cluster_logging.rb
```diff
@@ -0,0 +1,107 @@
...
+      es_status_raw(user: user, cached: cached, quiet: quiet).first['pods']
+    end
+
+    def kibana_status(user: nil, cached: true, quiet: false)
+      visualization_raw(user: user, cached: cached, quiet: quiet)['kibanaStatus']
+    end
+
+    def kibana_pods(user: nil, cached: true, quiet: false)
+      self.kibana_status(user: user, cached: cached, quiet: quiet).first['pods']
+    end
```
akostadinov: all these methods also return some structure instead of objects user can more easily interact with. And not used outside the class. So `es_cluster_health`, `es_pods`, `kibana_status` and `kibana_pods` can happily be private.

features/step_definitions/logging.rb
pruan-rht: @akostadinov @anli agrees with having that as optional.

features/step_definitions/logging.rb
pruan-rht: There are some methods in there that might still be applicable.  Since there is currently no conflict, I would prefer to have the removal of the file be done in a separate PR>

pruan-rht:
@akostadinov, I believe I've addressed all of your concerns now, except removal of the file, which I think we should do in another PR because I think some of the methods there are still applicable.  PTAL again.  Thanks!

features/step_definitions/logging.rb
akostadinov: I missed that it is there, sorry.

features/step_definitions/logging.rb
akostadinov: ok, but remember to do it ;)

PR #165, opened 2019-04-17
[networking] Automate network-operator cases

bmeng:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/39963/consoleFull

@anuragthehatter @zhaozhanqi  PTAL

### DIFF
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -52,3 +52,134 @@ Feature: Operator related networking scenarios
     And evaluation of `cluster_operator('network').condition(type: 'Failing',cached: false)` is stored in the :failing_status_post_patch clipboard
     Then the expression should be true> cb.failing_status_post_patch["status"]=="True"
     And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)
+
+  # @author bmeng@redhat.com
+  # @case_id OCP-22201
+  @admin
+  Scenario: Should have a clusteroperator object created under config.openshift.io api group for network-operator
+    Given the master version >= "4.0"
+    # Check the operator object has version
+    Given the expression should be true> cluster_operator('network').versions.length > 0
+    # Check the operator object has status for Failing|Progressing|Available
+    And the expression should be true> cluster_operator('network').condition(type: 'Available')['status'] == "True"
+    And the expression should be true> cluster_operator('network').condition(type: 'Failing')['status'] == "False"
+    And the expression should be true> cluster_operator('network').condition(type: 'Progressing')['status'] == "False"
+
+
+  # @author bmeng@redhat.com
+  # @case_id OCP-22419
+  @admin
+  @destructive
+  Scenario: The clusteroperator should be able to reflect the realtime status of the network when the config has problem
+    Given the master version >= "4.0"
+    # Check that the operator is not failing
+    Given the expression should be true> cluster_operator('network').condition(type: 'Failing')['status'] == "False"
+    # Copy the value of the networktype for backup
+    When I run the :get admin command with:
+      | resource      | network.config.openshift.io |
+      | resource_name | cluster                     |
+      | template      | {{.spec.networkType}}       |
+    Then the step should succeed
+    And evaluation of `@result[:response]` is stored in the :network_type clipboard
+    # Do some modification on the network.config.openshift.io
+    When I run the :patch admin command with:
+      | resource      | network.config.openshift.io     |
+      | resource_name | cluster                         |
+      | p             | {"spec":{"networkType":"None"}} |
+      | type          | merge                           |
+    Then the step should succeed
+    Given I register clean-up steps:
+    """
+    When I run the :patch admin command with:
+      | resource      | network.config.openshift.io                       |
+      | resource_name | cluster                                           |
+      | p             | {"spec":{"networkType":"<%= cb.network_type %>"}} |
+      | type          | merge                                             |
+    Then the step should succeed
+    """
+    # Check that the operator status reflect the problem
+    Given I wait up to 10 seconds for the steps to pass:
+    """
+    Given the status of condition "Failing" for network operator is :True
+    And the status of condition "Available" for network operator is :True
+    """
+    # Change the network.config.openshift.io back
+    When I run the :patch admin command with:
+      | resource      | network.config.openshift.io                       |
+      | resource_name | cluster                                           |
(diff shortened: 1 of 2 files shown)


features/networking/operator.feature
```diff
@@ -13,3 +13,167 @@ Feature: Operator related networking scenarios
...
+      | name     | <%= cb.machineset %>        |
+      | replicas | <%= cb.original_replicas %> |
+      | n        | openshift-machine-api       |
+    Then the step should succeed
+    """
+
+    # Check that the status of Progressing is truned to True during the new node provisioning
+    Given I wait up to 360 seconds for the steps to pass:
+    """
+#    And the expression should be true> `cluster_operator('network').condition(type: 'Progressing')['status']` == "True"
```
akostadinov: Please fix indentation here (and possibly other similar places). We use 2 spaces for indentation. In Cucumber the keyword `Examples: ` is indented the same as the `Scenario` line. Tables are indented deeper by 2 characters. Multi-line strings are indented at the same level  as the step they are par of (not deeper).

We now have also in the project a [EditorConfig](http://editorconfig.org/) file. If provides an editor agnostic way to set some styling settings like indentation type and width and removal of trailing white spaces. Please check the linked website to find whether your editor is supported.

For generic code style rules and more information, more style information, please check [coding_guide.adoc](https://github.com/openshift/cucushift/blob/v3/doc/coding_guide.adoc)

For vim specific settings you can see [doc/vim.adoc](https://github.com/openshift/cucushift/blob/v3/doc/vim.adoc).

-- saved reply

features/networking/operator.feature
```diff
@@ -13,3 +13,167 @@ Feature: Operator related networking scenarios
     Then the expression should be true> cb.operator_status["status"]=="True"
     #Confirm whether network operator version matches with ocp version
     And the expression should be true> cluster_operator('network').version_exists?(version: cb.ocp_version)
+
+
+  # @author bmeng@redhat.com
+  # @case_id OCP-22201
+  @admin
+  Scenario: Should have a clusteroperator object created under config.openshift.io api group for network-operator
+    Given the master version > "3.11"
```
akostadinov: Will this work on 3.12 if it is ever released? I doubt we will ever have such version but in any case, better do as
```
    Given the master version >= "4.0"
```

features/networking/operator.feature
```diff
@@ -13,3 +13,167 @@ Feature: Operator related networking scenarios
...
+    Then the step should succeed
+    Given I register clean-up steps:
+    """
+    When I run the :patch admin command with:
+      | resource      | network.config.openshift.io                       |
+      | resource_name | cluster                                           |
+      | p             | {"spec":{"networkType":"<%= cb.network_type %>"}} |
+      | type          | merge                                             |
+    Then the step should succeed
+    """
```
akostadinov: worth writing a step `cluster network type is restored after scenario`

features/networking/operator.feature
```diff
@@ -13,3 +13,167 @@ Feature: Operator related networking scenarios
...
+      | o             | jsonpath='{.status.conditions[?(.type == "Failing")].status}'     |
+    Then the step should succeed
+    And the output should contain "True"
+    When I run the :get admin command with:
+      | resource      | clusteroperators                                                  |
+      | resource_name | network                                                           |
+      | o             | jsonpath='{.status.conditions[?(.type == "Available")].status}'   |
+    Then the step should succeed
+    And the output should contain "True"
+    """
```
akostadinov: worth writing a step `cluster network is ready` or something like this

features/networking/operator.feature
```diff
@@ -13,3 +13,167 @@ Feature: Operator related networking scenarios
...
+      | o             | jsonpath='{.status.conditions[?(.type == "Failing")].status}'     |
+    Then the step should succeed
+    And the output should contain "False"
+    When I run the :get admin command with:
+      | resource      | clusteroperators                                                  |
+      | resource_name | network                                                           |
+      | o             | jsonpath='{.status.conditions[?(.type == "Available")].status}'   |
+    Then the step should succeed
+    And the output should contain "True"
+    """
```
akostadinov: worth writing a step `cluster network is ready` or something like this

features/networking/operator.feature
```diff
@@ -13,3 +13,167 @@ Feature: Operator related networking scenarios
...
+    # Check that the operator is not progressing
+    Given the expression should be true> cluster_operator('network').condition(type: 'Progressing')['status'] == "False"
+
+    # Record the original machine replica and scale it up to number +1
+    When I run the :get admin command with:
+      | resource | machineset                         |
+      | n        | openshift-machine-api              |
+      | template | {{(index .items 0).metadata.name}} |
+    Then the step should succeed
+    And evaluation of `@result[:response]` is stored in the :machineset clipboard
```
akostadinov: `:stdout` would be safer. Even better write a `MachineSet` class and use programmatic access.

akostadinov:
The mandatory changes for me are indentation and better cluster version comparison. The others are recommendations.

features/networking/operator.feature
bmeng: Added a new step to check the status for network operator with different condition types

features/networking/operator.feature
bmeng: Actually, we do not support changing the networkType, this testing is just checking that the operator will report when problem. So I think we will not have many cases to use this method.

bmeng:
@akostadinov Made some changes, please help check. Thanks.

bmeng:
resolved conflict.

features/networking/operator.feature
```diff
@@ -52,3 +52,134 @@ Feature: Operator related networking scenarios
...
+  # @case_id OCP-22201
+  @admin
+  Scenario: Should have a clusteroperator object created under config.openshift.io api group for network-operator
+    Given the master version >= "4.0"
+    # Check the operator object has version
+    Given the expression should be true> cluster_operator('network').versions.length > 0
+    # Check the operator object has status for Failing|Progressing|Available
+    And the expression should be true> cluster_operator('network').condition(type: 'Available')['status'] == "True"
+    And the expression should be true> cluster_operator('network').condition(type: 'Failing')['status'] == "False"
+    And the expression should be true> cluster_operator('network').condition(type: 'Progressing')['status'] == "False"
```
akostadinov: You could use this expression inside the new step instead of the get commands. And then use the step here. But don't want to delay this pull request. You can make consistent later.

PR #167, opened 2019-04-22
Fix OCP-19631

xiuwang:
Pass in my local host.
@openshift/devexp-qe @akostadinov Could you help review this? 
Thanks~

### DIFF
--- a/features/cli/build.feature
+++ b/features/cli/build.feature
@@ -753,7 +753,7 @@ Feature: build 'apps' with CLI
       | Build ConfigMaps:\s+cmtest->. |
       | Build Secrets:\s+secrettest->.|
     And the "ruby-hello-world-1" build completed
-    #Bug 1669981 comment #1, the secreti will not included in the output image.
+    #Bug 1669981 comment #1, the secret will not included in the output image.
     When I run the :get client command with:
       | resource      | po                       |
       | resource_name | ruby-hello-world-1-build |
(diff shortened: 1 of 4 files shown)


wzheng1:
LGTM,thx

features/cli/build.feature
```diff
@@ -753,6 +753,14 @@ Feature: build 'apps' with CLI
       | Build ConfigMaps:\s+cmtest->. |
       | Build Secrets:\s+secrettest->.|
     And the "ruby-hello-world-1" build completed
+    #Bug 1669981 comment #1, the secreti will not included in the output image.
```
pruan-rht: typo should be `secret` instead of `secreti`

akostadinov: lgtm otherwise

features/cli/build.feature
```diff
@@ -753,6 +753,14 @@ Feature: build 'apps' with CLI
...
+    When I run the :get client command with:
+      | resource      | po                       |
+      | resource_name | ruby-hello-world-1-build |
+      | o             | yaml                     |
+    Then the output should match:
```
pruan-rht: @xiuwang should this line be `Then the output should not match:` because of your comment in line 756?

xiuwang: To adjust comment in line 756, here check secret in the build pod, not in the output image.

PR #168, opened 2019-04-22
Workaround bug 1610661

liangxia:
Automation will quit prematurely due to this issue.
Check more detail in https://bugzilla.redhat.com/show_bug.cgi?id=1610661

@zhouying7780 @xingxingxia @yasun1 @emmajiafan @chengzhang1016 @barleyer @akostadinov

### DIFF
--- a/lib/rules/cli/4.0.yaml
+++ b/lib/rules/cli/4.0.yaml
@@ -253,7 +253,7 @@
      :exec_command: <value>
      :exec_command_arg: <value>
 :delete:
-  :cmd: oc delete
+  :cmd: oc delete --wait=false
   :options:
     :all: --all
     #The following 'all' can be used with selectors, like the -l flag.


lib/rules/cli/4.0.yaml
```diff
@@ -253,7 +253,7 @@
      :exec_command: <value>
      :exec_command_arg: <value>
 :delete:
-  :cmd: oc delete
+  :cmd: oc delete --wait=false
```
xingxingxia: What such 1610661-like cases make `automation will quit prematurely`? I'd like to add the flag in the specific cases instead of global rule.

liangxia: Most of the test cases which contains the step to create PV/PVC. E.g, OCP-9638

liangxia:
Add @weherdh @chao007 who is affected.

lib/rules/cli/4.0.yaml
akostadinov: Normally I'd say that you need to add this parameter to `features/storage/storage_object_in_use_protection.feature` and it should fix all scenarios using these steps. 

On the other hand I can't `grep` any users of `wait: true` at the moment. So I can accept such change in case setting `wait: true` as a parameter would override this default of `--wait=false` here.

I'm really frustrated with https://bugzilla.redhat.com/show_bug.cgi?id=1610661 😒

liangxia: @akostadinov @xingxingxia  Check Runner-v3/40196 with more detail where/why/how it is happen.

xingxingxia: $ oc delete -h
      --force=false: ... If true, immediately remove resources from API and bypass graceful
deletion. **Note that immediate deletion of some resources may result in inconsistency or data loss and requires confirmation**.
      --wait=true: If true, wait for resources to be gone before returning. This waits for finalizers.

The affect of `--wait=false` is like `--force=true`. I'm not sure if adding `--wait=false` will affect other `oc delete` places "in terms of inconsistency or data loss", e.g. clean project in After Scenario.
Thus I still prefer to only add the parameter to feature files where needed.

liangxia: The hang happens in the "after scenario", so not sure how to work with @xingxingxia's suggestion "add the flag in the specific cases" and @akostadinov's suggestion "add this parameter to features/storage/storage_object_in_use_protection.feature"

xingxingxia: It is hooked in features/step_definitions/pv.rb _pv.delete, how about there?

liangxia: #169 for the idea. Thanks.

liangxia:
Closing as in favor of #169

PR #172, opened 2019-04-23
workaround bug 1610661

akostadinov:
@liangxia sorry, didn't have time to test this but I think it should resolve 
the issue your team is hitting.

PR #174, opened 2019-04-24
Bz1610661

liangxia:
@akostadinov @pruan-rht  
With minor change to #172 , test passed with Runner-v3/40267

CC @weherdh @chao007  as they are affected.


akostadinov:
Thank you! I want to pass `String` instead of a boolean to be sure. So I amended #172.

PR #175, opened 2019-04-24
add check deploy status for ocp-11326

zhouying7780:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/40334/console

@xingxingxia  Please help review. Only for 4.x ,so only for the master branch.

### DIFF
--- a/features/cli/deploy.feature
+++ b/features/cli/deploy.feature
@@ -756,6 +756,11 @@ Feature: deployment related features
     Then the step should succeed
     And the output should contain:
       | deployment.kubernetes.io/revision:2 |
+    When I run the :get client command with:
+      | resource      | deployment                |
+    Then the step should succeed
+    And the output should contain:
+      | READY |
 
   # @author mcurlej@redhat.com
   # @case_id OCP-10902


features/cli/deploy.feature
```diff
@@ -756,6 +756,11 @@ Feature: deployment related features
     Then the step should succeed
     And the output should contain:
       | deployment.kubernetes.io/revision:2 |
+    When I run the :get client command with:
+      | resource      | deployment                |
+    Then the step should succeed
+    And the output should contain:
+      | READY |
```
akostadinov: better use somehtinglike:
```
Then the the expression should be true> deployment("deployment-name").ready?
```

pruan-rht: keep beating me to the punch today. :)

xingxingxia: @zhouying7780 , LGTM, thanks. Like https://github.com/openshift/origin/pull/22572#discussion_r278029486 , "It is always helpful to write a PR description stating why we need this change" is always recommended.
@akostadinov this PR (and https://github.com/openshift/cucushift/pull/6973 ) are added for https://bugzilla.redhat.com/show_bug.cgi?id=1608448 (MSTR-189), so it LGTM
Sorry for delayed review due to 4.1 higher : )

akostadinov: Can we just use the method explained above instead of string check? This was the suggestion.

xingxingxia: Above mentioned https://bugzilla.redhat.com/show_bug.cgi?id=1608448 (MSTR-189) is change of **display** of output for oc get sts/deployment. Author made the PRs to track it.

akostadinov: In such case it is much better to put this test into unit tests. But if you still believe we need it here, I can merge.

liangxia:
@zhouying7780 OCP-11326 is inactive in Polarion. Should we close this PR ?

PR #176, opened 2019-04-24
Updated scripts for 4.x

wewang58:
Passed in my local host, @openshift/devexp-qe pls help to review it, thanks

### DIFF
--- a/features/cli/oc_set_build_hook.feature
+++ b/features/cli/oc_set_build_hook.feature
@@ -36,18 +36,4 @@ Feature: oc_set_build_hook
     When I run the :start_build client command with:
       | buildconfig | rails-postgresql-example |
     Then the step should succeed
-    Given the "rails-postgresql-example-3" build completes
-    When I run the :set_build_hook client command with:
-      | buildconfig | bc/rails-postgresql-example |
-      | post_commit | true                        |
-      | oc_opts_end |                             |
-      | args        | bundle                      |
-      | args        | exec                        |
-      | args        | rake                        |
-      | args        | test                        |
-    Then the step should succeed
-    When I run the :start_build client command with:
-      | buildconfig | rails-postgresql-example |
-    Then the step should succeed
-    Given the "rails-postgresql-example-4" build completes
-
+    And the "rails-postgresql-example-3" build completes
--- a/features/cli/resources.feature
+++ b/features/cli/resources.feature
@@ -14,10 +14,11 @@ Feature: resouces related scenarios
     And the output should match:
       | Name:\\s+database            |
       | Selector:\\s+name=database   |
-    When I run the :export client command with:
-      | resource     | svc           |
-      | name         | database      |
-      | o            | yaml          |
+    When I run the :get client command with:
+      | resource      | svc      |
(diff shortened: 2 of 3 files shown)


akostadinov:
lgtm

features/cli/resources.feature
```diff
@@ -14,10 +14,10 @@ Feature: resouces related scenarios
...
       | Name:\\s+database            |
       | Selector:\\s+name=database   |
-    When I run the :export client command with:
-      | resource     | svc           |
-      | name         | database      |
-      | o            | yaml          |
+    When I run the :get client command with:
+      | resource      | svc      |
+      | resource_name | database |
+      | output        | yaml     |
```
akostadinov: I wonder if using the `--export` option will be better here.

wewang58: It's not support export command in 4.x now

features/cli/oc_set_build_hook.feature
```diff
@@ -36,18 +36,4 @@ Feature: oc_set_build_hook
     When I run the :start_build client command with:
       | buildconfig | rails-postgresql-example |
     Then the step should succeed
-    Given the "rails-postgresql-example-3" build completes
-    When I run the :set_build_hook client command with:
```
xiuwang: If the set hook feature is removed from 4.x?

xiuwang: If the set hook feature is removed from 4.x?

features/cli/resources.feature
akostadinov:
```
🐚 oc get -h | grep export
      --export=false: If true, use 'export' for the resources.  Exported resources are stripped of cluster-specific information.
```

wewang58: Oh, I misunderstood  you, I thought "oc export xxx", I will update it later.

features/cli/oc_set_build_hook.feature
wewang58: No, just set hook commond did not support with args options in 4.x. so remove the part. there's a bug: https://bugzilla.redhat.com/show_bug.cgi?id=1670000

features/cli/resources.feature
wewang58: @akostadinov  Updated it with export option, pls review it, thanks.

features/cli/oc_set_build_hook.feature
akostadinov: bug shows verified. Do we still need to remove those?

features/cli/resources.feature
akostadinov: Thank you!

features/cli/oc_set_build_hook.feature
wewang58: Hi @akostadinov you can see https://github.com/openshift/origin/pull/21880/files#diff-77d633c62fb7fa312b6e7edff1743c7cL2942 in pr https://github.com/openshift/origin/pull/21389/files, codes already remove it, so scripts removed it too.

wewang58:
@openshift/devexp-qe  I think it can be merge now.

PR #182, opened 2019-04-27
Update metering.rb

pruan-rht:
we need to wait until the exposed metering route is accessible first.

### DIFF
--- a/features/step_definitions/metering.rb
+++ b/features/step_definitions/metering.rb
@@ -313,6 +313,7 @@ When /^I perform the GET metering rest request with:$/ do | table |
     @result[:parsed] = YAML.load(@result[:response])
   end
 end
+
 Given /^I wait(?: up to ([0-9]+) seconds)? for metering route to be accessible$/ do | seconds |
   seconds = seconds.to_i unless seconds.nil?
   seconds ||= 60
--- a/features/step_definitions/metering.rb
+++ b/features/step_definitions/metering.rb
@@ -199,12 +199,12 @@ BASE_TEMPLATE
     # switch back to original user
     @user = org_user if org_user
   end
+  step %Q/I wait for metering route to be accessible/
 end
 
 # XXX: should we check metering route exists first prior to patching?
 Given /^I disable route for#{OPT_QUOTED} metering service$/ do | metering_name |
-
-  metering_name ||= metering.name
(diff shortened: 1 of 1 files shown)


features/step_definitions/metering.rb
```diff
@@ -302,12 +302,24 @@
...
+  url_path ||= opts[:custom_url]
+  url_path ||= "/api/v1/reports/get?name=#{report_name}&namespace=#{cb.metering_namespace.name}&format=json"
+
   report_query_url = route.dns + url_path
   @result = BushSlicer::Http.request(url: report_query_url, **https_opts, method: 'GET')
   if @result[:success]
     @result[:parsed] = YAML.load(@result[:response])
   end
 end
+Given /^I wait(?: up to ([0-9]+) seconds)? for metering route to be accessible$/ do | seconds |
```
akostadinov: one empty line would be nice here

features/step_definitions/metering.rb
```diff
@@ -302,12 +302,24 @@
...
+  seconds = seconds.to_i unless seconds.nil?
+  seconds ||= 60
+  wait_for(seconds) {
+    step %Q/I perform the GET metering rest request with:/, table(%{
+      | custom_url | /healthy |
+    })
+    @result[:success]
+  }
```
akostadinov: there is nothing to make this step fail on timeout. Or do you expect next steps to fail? It is fine as long as error message clearly indicates the route failed to become accessible in time.

pruan-rht: @akostadinov I added this step to make sure the route is accessible before doing REST query.  Do you want me to add an error message within this step if `@result[:success]` is false?

features/step_definitions/metering.rb
pruan-rht: updated.

pruan-rht:
@akostadinov, PR update, PTAL.  Thanks!

features/step_definitions/metering.rb
akostadinov: If you want step to fail, then you need to raise an exception. Presently it will wait but on timeout it will just continue without warning.

pruan-rht: @akostadinov ok, raising error instead of relying on the query step to fail.  PTAL.  Thanks!

PR #190, opened 2019-05-07
Added bash script to run as part of aos-azure ci-operator. (#37)

akostadinov:


akostadinov:
same as #37 but fot 3.x

PR #192, opened 2019-05-09
update logging CRD to the correct URL

pruan-rht:


pruan-rht:
I'll merge this one since w/o it the existing step won't work.  I've verified it locally.

PR #194, opened 2019-05-10
Fix scenarios

anuragthehatter:
Fixing OCP-16710 to support 3.11 and 4.1. Broke every time during 3.11 or 4.x regression
Pass logs on 3.11 - https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/338/console
Pass logs on 4.1 - https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/339/console
@bmeng @zhaozhanqi

### DIFF
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -81,7 +81,6 @@ Feature: SDN related networking scenarios
     And evaluation of `@result[:parsed]['items'][0]['metadata']['name']` is stored in the :sdn_pod clipboard
  
     Given the cluster network plugin type and version and stored in the clipboard 
-    
     #Changing plugin version to some arbitary value ff
     When I run command on the "<%= cb.node_name %>" node's sdn pod: 
       | ovs-ofctl| -O | openflow13 | mod-flows | br0 | table=253, actions=note:<%= cb.net_plugin[:type] %>.ff |
@@ -108,8 +107,7 @@ Feature: SDN related networking scenarios
 
     #Changing plugin type to some arbitary value 99
     When I run command on the "<%= cb.node_name %>" node's sdn pod: 
-	    | ovs-ofctl| -O | openflow13 | mod-flows | br0 | table=253, actions=note:99.<%= cb.net_plugin[:version] %> |
-
+      | ovs-ofctl| -O | openflow13 | mod-flows | br0 | table=253, actions=note:99.<%= cb.net_plugin[:version] %> |
     Then the step should succeed
     #Expecting sdn pod to be restarted due to plugin type value change
     And I wait up to 60 seconds for the steps to pass:
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -66,7 +66,6 @@ Feature: SDN related networking scenarios
   @admin
   @destructive
   Scenario: SDN will detect the version and plugin type mismatch in openflow and restart node automatically
-    Given the master version >= "3.11"
     Given I select a random node's host
     And evaluation of `node.name` is stored in the :node_name clipboard
 
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -99,7 +99,7 @@ Feature: SDN related networking scenarios
       | Starting openshift-sdn network plugin         |
     
     """
-    # Expecting sdn pod to come back to default version the cluser was on
+    # Expecting sdn pod to come back to default version the cluser was on initially
     Given I wait up to 60 seconds for the steps to pass:
     """
     When I run command on the "<%= cb.node_name %>" node's sdn pod: 
@@ -124,7 +124,7 @@ Feature: SDN related networking scenarios
       | Starting openshift-sdn network plugin         |
     
     """
-    # Expecting sdn pod to come back to default type the clustr was on
+    # Expecting sdn pod to come back to default type the cluster was on initially
     Given I wait up to 60 seconds for the steps to pass:
     """
     When I run command on the "<%= cb.node_name %>" node's sdn pod: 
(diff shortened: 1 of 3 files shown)


features/networking/sdn.feature
```diff
@@ -66,53 +66,69 @@ Feature: SDN related networking scenarios
...
-    Then the step should succeed
-    Then the output should contain "<%= cb.net_plugin[:type] %>.<%= cb.net_plugin[:version] %>"
-    """
-    And the node service is verified
-    And the node network is verified
-
-    When I run the ovs commands on the host:
-      | ovs-ofctl -O openflow13 mod-flows br0 "table=253, actions=note:<%= cb.net_plugin[:type] %>.ff" |
-    Then the step should succeed
+    Given the master version >= "3.11"
```
zhaozhanqi: @anuragthehatter 
Thanks your effort.
 BTW, that's mean this case would cannot be tested before 3.11 after the fixed?

anuragthehatter: Actually this can be also tested on 3.10 as well so it can be changed as >="3.10". As starting 3.10, sdn ovs services are containerized while pre 3.10 sdn/ovs services were running on the node. So i am not sure at this point if this case also need to be supported on 3.9 and down or depends how many testing requests we get on 3.9 or before to find out if its worth the efforts and time. If that's the case i need to spend good time on figuring out like if version less than 3.10 focus on journalctl logs else sdn pod logs and otehr stuff etc or a separate scenario needs to be created for version less than 3.10 as whole scenario will be changed - no pods, no oc logs but just commands on host and journalctl on node. It seems difficult but not impossible to make this one scenario works for all version. Thanks! I guess this calls for a discussion.

features/step_definitions/networking.rb
```diff
@@ -437,12 +437,18 @@
...
   ensure_admin_tagged
   _host = node.host
-
-  step %Q/I run the ovs commands on the host:/, table([[
-    "ovs-ofctl dump-flows br0 -O openflow13 | grep table=253 | sed -n -e 's/^.*note://p'"
-  ]])
-  if @result[:success]
-    of_note = @result[:response].chomp
+  
+  if env.version_lt("3.11", user: user)
```
akostadinov: master branch is only for 4.x so conditions for 3.x are not needed. If yo want to fix 3.x, then `v3` branch needs to be updated.

anuragthehatter: @akostadinov Yea, Thanks for pointing out that. Right. I will re-look at it

features/networking/sdn.feature
akostadinov: Anything that you want to run on 3.x must go as a separate PR to the `v3` branch.. just FYI

anuragthehatter: @akostadinov Please help me understand bit more on it. Say i have a scenario which says "Given master version >=3.7" which means that this scenario can run on 3.x and 4.x both. Where would you add the scenario, in master branch or v3?

akostadinov: @anuragthehatter - both

anuragthehatter: @akostadinov make sense, thanks. Gonna have to work on this PR again, currently engaged in high priority stuff

anuragthehatter:
As @akostadinov suggested to keep changes separate for separate branches. networking.rb is fixed accordingly and gating criteria is removed. Created #201 relevant to v3 branch.  Please help merging it. As per recent testing, there is lot of work to be involved to make this work on >=3.7 and <3.10 which comparatively get less testing request as compare to 3.10,3.11 and 4.x branch. So i suggest we merge it to avoid manual investigation in demanding branches during regression and needless to say, we can always keep working on it to support max branches in future
@bmeng @zhaozhanqi

features/networking/sdn.feature
```diff
@@ -66,53 +66,68 @@ Feature: SDN related networking scenarios
...
+
+    #Capturing the node's corresponding sdn pod to execute ovs commands on it
+    When I run the :get admin command with:
+       | resource      | pods                              |
+       | fieldSelector | spec.nodeName=<%= cb.node_name %> |
+       | namespace     | openshift-sdn                     |
+       | l             | app=sdn                           |
+       | output        | json                              |
+    Then the step should succeed
+    And evaluation of `@result[:parsed]['items'][0]['metadata']['name']` is stored in the :sdn_pod clipboard
```
akostadinov: In #206 I made step `I run command on the "<%= cb.node_name %>" node's sdn pod:` also cache the pod. Could you:
* rebase this pull request
* remove the `get` and `clipboard` steps above
* instead of `cb.sdn_pod` in scenario use `pod.name`

This will make scenario shorter and nicer.

akostadinov: Other than that looks good.

anuragthehatter: Sounds good, I will look at that. Thanks!

anuragthehatter: Done! Thanks for #206 @akostadinov

features/networking/sdn.feature
```diff
@@ -69,27 +69,18 @@ Feature: SDN related networking scenarios
...
-    Then the step should succeed
-    And evaluation of `@result[:parsed]['items'][0]['metadata']['name']` is stored in the :sdn_pod clipboard
- 
     #Below step will save plugin type and version value in net_plugin variable
     Given the cluster network plugin type and version and stored in the clipboard 
     #Changing plugin version to some arbitary value ff
     When I run command on the "<%= cb.node_name %>" node's sdn pod: 
       | ovs-ofctl| -O | openflow13 | mod-flows | br0 | table=253, actions=note:<%= cb.net_plugin[:type] %>.ff |
     Then the step should succeed
+    And evaluation of `pod` is stored in the clipboard
```
akostadinov: This is not needed. Step `I run command on the "<%= cb.node_name %>" node's sdn pod:` already caches that pod. And in fact `pod` does not refer to the clipboard. `cb.pod` is what above step creates. So the line can just be removed.

PR #196, opened 2019-05-13
update to the default interface of node for multus

zhaozhanqi:
pod will cannot be created when the node do not have the 'eth0' interface. so update it to the default interface.

pass logs:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/40826/consoleFull

@bmeng @anuragthehatter

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -7,12 +7,14 @@ Feature: Multus-CNI related scenarios
     # Make sure that the multus is enabled
     Given the master version >= "4.0"
     And the multus is enabled on the cluster
-
+    Given the default interface on nodes is stored in the :default_interface clipboard
+    And evaluation of `node.name` is stored in the :target_node clipboard    
     # Create the net-attach-def via cluster admin
     Given I have a project
-    When I run the :create admin command with:
-      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml |
-      | n | <%= project.name %> |
+    When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml" replacing paths:
+      | ["metadata"]["namespace"] | <%= project.name %> |    
+      | ["spec"]["config"]["eth0"]| <%= cb.default_interface %> |
+      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |      
     Then the step should succeed
 
     # Create the first pod which consumes the macvlan custom resource
@@ -64,12 +66,14 @@ Feature: Multus-CNI related scenarios
     # Make sure that the multus is enabled
     Given the master version >= "4.0"
     And the multus is enabled on the cluster
-
+    Given the default interface on nodes is stored in the :default_interface clipboard
+    And evaluation of `node.name` is stored in the :target_node clipboard    
     # Create the net-attach-def via cluster admin
     Given I have a project
-    When I run the :create admin command with:
-      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-private.yaml |
-      | n | <%= project.name %> |
+    When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-private.yaml" replacing paths:
+      | ["metadata"]["namespace"] | <%= project.name %> |
+      | ["spec"]["config"]["eth0"]| <%= cb.default_interface %> |
+      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |      
     Then the step should succeed
 
     # Create the first pod which consumes the macvlan custom resource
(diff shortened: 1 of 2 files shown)


features/networking/multus.feature
```diff
@@ -7,12 +7,12 @@ Feature: Multus-CNI related scenarios
...
-
+    Given the default interface on the node is stored in the :default_interface clipboard
     # Create the net-attach-def via cluster admin
     Given I have a project
-    When I run the :create admin command with:
-      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml |
-      | n | <%= project.name %> |
+    When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml" replacing paths:
+      | ["metadata"]["namespace"] | <%= project.name %> |    
+      | ["spec"]["config"]["eth0"]| <%= cb.default_interface %> |
```
bmeng: Is the replacement correct here?
It looks a little weird to me. 

The contents under the spec.config should be treated as a single string.

zhaozhanqi: yes, see logs:
 [07:24:49] INFO> {"apiVersion":"k8s.cni.cncf.io/v1","kind":"NetworkAttachmentDefinition","metadata":{"name":"macvlan-bridge","namespace":"jivvv"},"spec":{"config":"{ \"cniVersion\": \"0.3.0\", \"type\": \"macvlan\", \"master\": \"ens3\", \"mode\": \"bridge\", \"ipam\": { \"type\": \"host-local\", \"subnet\": \"10.1.1.0/24\", \"rangeStart\": \"10.1.1.100\", \"rangeEnd\": \"10.1.1.200\", \"routes\": [ { \"dst\": \"0.0.0.0/0\" } ], \"gateway\": \"10.1.1.1\" } }"}}

bmeng: Woo, that is amazing!

anuragthehatter: Yep, the logs a lot. ens3 , seems like it grabbed the public cluster default interface. LGTM

features/step_definitions/networking.rb
```diff
@@ -674,3 +674,13 @@
   @result = sdn_pod.exec(network_cmd, as: admin)
   raise "Failed to execute network command!" unless @result[:success]
 end
+
+Given /^the default interface on the node is stored in the#{OPT_SYM} clipboard$/ do |cb_name|
+  ensure_admin_tagged
+  hosts = step "I select a random node's host"
```
akostadinov: if the step name reads `default interface on the node` that means current node. And here we select a random one. I'd suggest either replace with `node.host` or change step to read something like `the default interface on nodes is stored ...`. 

This assumes all nodes will have same default interface. I'm not sure this is a safe assumption tbh.

zhaozhanqi: ok, updated to 'on nodes', thanks

bmeng: On nodes does not fix anything. 
If we select a random host, we need to make sure the pod will be created on the selected node.

And this will make the cases select the random node twice at least in case OCP-21855

zhaozhanqi: I think the pod also can be scheduled to the matched node. for example, if the default interface is 'eth0' for all nodes. then it can be scheduled on any node.  it does have to the selected node.
for OCP-21855,  this case need the selected node. 
I deleted the step `And I select a random node's host` for OCP-21855 and move `store the target node` step.
please help check again. thanks

bmeng: Yes, the old steps assume that all the nodes' default interface is named `eth0`, and this is true for IPI on AWS with coreOS node. And at this point, the pod scheduled to which node is ok.

But for the baremetal cluster, the interface name maybe different on different nodes or different distros. So if we get the default interface's name, we need to make sure that the test pod is also scheduled on the same node.

zhaozhanqi: hmm, I think the pod can find the matched node to be scheduled.
eg.  I have two nodes here
 node1   eth0
 node2   eth2
steps:
1. select node2 and get the interface is eth2
2. replace the eth0 to 'eth2' in NetworkAttachmentDefinitions
3.  Create the pod and this pod can be scheduled to node2

bmeng: So, the question here is why the pod will be scheduled to the selected node in step3? It is unreasonable IMO.

zhaozhanqi: I think it's should be a bug if it cannot be scheduled to the selected node.
since I don't think normal user need to add node-selector for the pod and it also don't know which node it should be used.

akostadinov: This was exactly my question - whether we expect default interface to ever differ between different nodes.

So in case we expect that to sometimes differ, then we need to select a schedulable node and then bind pod to node and get that specific node default interface.

Maybe create one dummy pod and get node from it. That would ensure the node is able to schedule pods without too much troubles (taints/whatever). WDYT?

zhaozhanqi: I just tried many times. the pod can find the matched node to be scheduled.

bmeng: So that means the scheduler will put the pod on the node which have the correct nic which can hold the macvlan conf. 
This might be true, I did not aware of this before.

zhaozhanqi: sorry, I'm wrong. I tried more time again. the pod will not be scheduled the pod to node according to the nic. I can update the script to make sure the pod using the selected node later.
but I still think it's a bug here if the cluster have multi-node with different NIC 
WDYT?

bmeng: Ok, that sounds more reasonable.
The cni config is not part of the cluster things. So I think the scheduler will not check the value there.

The order should be 1) scheduler put the pod on the node which matches the other thing  2) the pod is trying to create the secondary interface which defined in cni config

zhaozhanqi: updated the steps to make use selected node
`      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |`

logs:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/41049/consoleFull

bmeng:
LGTM

PR #198, opened 2019-05-15
allow ruby expressions in template

akostadinov:


akostadinov:
@jianlinliu , please try running with my branch before we merge

akostadinov:
tested to work

PR #199, opened 2019-05-20
No need to create storage class for OCP-10447

liangxia:
@chao007 @weherdh 

Tested the changes on OSO Starter, and it passed.

### DIFF
--- a/features/cli/policy.feature
+++ b/features/cli/policy.feature
@@ -165,38 +165,36 @@ Feature: change the policy of user/service account
 
   # @author chaoyang@redhat.com
   # @case_id OCP-10447
-  @admin
   Scenario: Basic user could not get deeper storageclass object info
     Given I have a project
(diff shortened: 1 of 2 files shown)


chao007:
LGTM

liangxia:
@pruan-rht  @akostadinov

features/cli/policy.feature
```diff
@@ -165,38 +165,36 @@ Feature: change the policy of user/service account
 
   # @author chaoyang@redhat.com
   # @case_id OCP-10447
-  @admin
```
akostadinov: I have removed tag `admin` from Polarion. Make sure to update `v3` branch as well because it will be a discrepancy in this way which can cause failures running smoke tests.

liangxia: @akostadinov PR for v3, https://github.com/openshift/verification-tests/pull/204

PR #200, opened 2019-05-20
added check for clusteroperators after upgrade to make sure no DEGRAD…

pruan-rht:
…ED condition exists

### DIFF
--- a/features/step_definitions/operators.rb
+++ b/features/step_definitions/operators.rb
@@ -8,10 +8,11 @@ Given /^all clusteroperators reached version #{QUOTED} successfully$/ do |versio
     # True        False         False
     conditions = co.conditions
     expected = {"Degraded"=>"False", "Progressing"=>"False", "Available"=>"True"}
-    conditions.each do |k, v|
+    conditions.each do |c|
       # only care about the `expected`, don't compare otherwise
-      if expected.keys.include? k
-        raise "Failed for condition #{k}" unless expected[k] == v['status']
+      if expected.keys.include? c['type']
+        expected_status = expected[c['type']]
+        raise "Failed for condition #{c['type']}, expected: #{expected_status}, got: #{c['status']}" unless expected_status == c['status']
       end
     end
   end
--- a/features/step_definitions/operators.rb
+++ b/features/step_definitions/operators.rb
@@ -1,0 +1,20 @@
+# operators related helper steps
+Given /^all clusteroperators reached version #{QUOTED} successfully$/ do |version|
+  ensure_admin_tagged
+  clusteroperators = BushSlicer::ClusterOperator.list(user: admin)
+  clusteroperators.each do | co |
+    raise "version does not match #{version}" unless co.version_exists?(version: version)
+    # AVAILABLE   PROGRESSING   DEGRADED
+    # True        False         False
+    conditions = co.conditions
+    expected = {"Degraded"=>"False", "Progressing"=>"False", "Available"=>"True"}
+    conditions.each do |k, v|
+      # only care about the `expected`, don't compare otherwise
+      if expected.keys.include? k
+        raise "Failed for condition #{k}" unless expected[k] == v['status']
+      end
+    end
+  end
+end
+
+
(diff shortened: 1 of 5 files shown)


lib/openshift/cluster_operator.rb
```diff
@@ -4,12 +4,22 @@ class ClusterOperator < ClusterResource
...
+
+    # return Hash of conditions for easier access
+    def conditions(user: nil, cached: true, quiet: false)
+      conditions = raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'conditions')
+      cond_hash = {}
+      conditions.each do |con|
+        cond_hash[con['type']] = con
+      end
+      cond_hash
+    end
```
akostadinov: This method can go to `resource.rb` IMO.

features/step_definitions/operators.rb
```diff
@@ -0,0 +1,20 @@
+# operators related helper steps
+Given /^all clusteroperators reached version #{QUOTED} successfully$/ do |version|
+  ensure_admin_tagged
+  clusteroperators = BushSlicer::ClusterOperator.list(user: admin)
+  clusteroperators.each do | co |
+    raise "version does not match #{version}" unless co.version_exists?(version: version)
```
akostadinov: What if one upgrades only single cluster operator? Should it just check all operators are non-degraded and then separate step to test particular operator is at version `X`? Or how does it work?

pruan-rht: @akostadinov this method for cluster-wide upgrade check.   I've not tried a single clusteroperator upgarde...not sure if that's possible (need to check).   As the method name indicates it's checking `all clusteropeators`.

pruan-rht: @akostadinov found out that the only way to update an individual `clusteroperator` is to create a release image with only one operator that was updated... so for non-ART folks, it's not possible to do that.  It's probably a good to check all of them anyway.

lib/openshift/cluster_operator.rb
pruan-rht: @akostadinov should've checked it earlier...there's already one [there](https://github.com/openshift/verification-tests/blob/master/lib/openshift/resource.rb#L283), but it just return the raw array of conditions.

features/step_definitions/operators.rb
akostadinov: You sure we never get operators with different versions? Hard to believe for me. Can anybody else comment on such probability?

lib/openshift/cluster_operator.rb
akostadinov: You can create a method to return to you based on type. It is a big anti-pattern to override a method changing its contract (in this case return data structure). See [Liskov principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

akostadinov: e.g. `oc.conditions_by_type("my_type", user: admin)`

features/step_definitions/operators.rb
pruan-rht: @akostadinov Trevor states that `anyone can create a new release image; you just need a registry somewhere to push it to`.  For our upgrade workflow, we just do a cluster-wide upgrade and not individual clusteroperators.

lib/openshift/cluster_operator.rb
pruan-rht: @akostadinov apparently there's also an existing method `[condition](https://github.com/openshift/verification-tests/blob/master/lib/openshift/resource.rb#L290)`

pruan-rht:
@akostadinov PR updated.  PTAL.  Thanks!

lib/openshift/cluster_operator.rb
akostadinov: great find!

features/step_definitions/operators.rb
akostadinov: I'm not sure I completely agree here but we can understand better what we need once it is needed so ack.

PR #201, opened 2019-05-20
Fix scenarios v3

anuragthehatter:
This is similar to #194 but relevant to v3 branch. As @akostadinov  suggested to keep changes separate for separate branches. networking.rb is fixed accordingly and gating criteria in scenario is clear now. Please help merging it. As per recent testing, there is lot of work to be involved to make this work on >=3.7 and <3.10 which comparatively get less testing request as compare to 3.10,3.11 and 4.x branch. So i suggest we merge it to avoid manual investigation in demanding branches during regression and needless to say, we can always keep working on it to support max branches in future
@bmeng @zhaozhanqi

### DIFF
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -66,57 +66,73 @@ Feature: SDN related networking scenarios
   @admin
   @destructive
   Scenario: SDN will detect the version and plugin type mismatch in openflow and restart node automatically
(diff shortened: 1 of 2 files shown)


anuragthehatter:
Will open a new one, sorry lot of conflicts on my env

PR #203, opened 2019-05-20
OCP_16217 fixes to be merged in v3 branch

anuragthehatter:
This is similar to #194  but relevant to v3 branch. As @akostadinov suggested to keep changes separate for separate branches. networking.rb is fixed accordingly and gating criteria in scenario is clear now. Please help merging it. As per recent testing, there is lot of work to be involved to make this work on >=3.7 and <3.10 which comparatively get less testing request as compare to 3.10,3.11 and 4.x branch. So i suggest we merge it to avoid manual investigation in demanding branches during regression and needless to say, we can always keep working on it to support max branches in future
@bmeng @zhaozhanqi

### DIFF
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -70,27 +70,18 @@ Feature: SDN related networking scenarios
     Given I select a random node's host
     And evaluation of `node.name` is stored in the :node_name clipboard
 
-    #Capturing the node's corresponding sdn pod to execute ovs commands on it
-    When I run the :get admin command with:
-       | resource      | pods                              |
-       | fieldSelector | spec.nodeName=<%= cb.node_name %> |
-       | namespace     | openshift-sdn                     |
-       | l             | app=sdn                           |
-       | output        | json                              |
-    Then the step should succeed
-    And evaluation of `@result[:parsed]['items'][0]['metadata']['name']` is stored in the :sdn_pod clipboard
- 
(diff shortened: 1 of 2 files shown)


akostadinov:
Did you test this? I'm don't see step `I run command on the "<%= cb.node_name %>" node's sdn pod:` in v3 branch.

Other than that the comment for making shorter like in #194 applies here as well.

anuragthehatter:
@akostadinov Do you mean in scenario? I see it present there at line 86,105,111 and 130.

anuragthehatter:
Oh you might mean whether this will work in 3.x. Yes it does work on 3.10 and 3.11 thats why the gating is set to >=3.10. It was tested earlier and logs are present in #194

akostadinov:
I was asking where the step is defined, not where it is used.

anuragthehatter:
My bad, yep, figured that 3.x scenario was tested on master branch code hence worked there. I need "Given I run command on the node's sdn pod:" definition. I will fix this and re-run the test to make sure. And will incorporate your #206 changes as well

anuragthehatter:
Done! PTAL. Added  cache_resources <pod> in networking.rb here
Pass logs: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/345/console

PR #211, opened 2019-05-24
fix various bugs with metering automation

pruan-rht:


lib/openshift/report_data_source.rb
```diff
@@ -3,19 +3,19 @@ class ReportDataSource < ProjectResource
     RESOURCE = "reportdatasources"
 
     def last_import_time(user: nil, cached: false, quiet: false)
```
akostadinov: must be cached:true

akostadinov: all places
I think we spoke about it

PR #213, opened 2019-05-25
added new openshift objects to help another PR

pruan-rht:
potential helper for https://github.com/openshift/cucushift/pull/6991

### DIFF
--- a/lib/openshift/ingress_controller.rb
+++ b/lib/openshift/ingress_controller.rb
@@ -1,0 +1,24 @@
+module BushSlicer
+  class IngressController < ProjectResource
+    RESOURCE = "ingresscontrollers"
+
+
+    ### status related methods
+    def available_replicas(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'availableReplicas')
+    end
+
+    def domain(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'domain')
+    end
+
+    def endpoint_publishing_strategy(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'endpointPublishingStrategy', 'type')
+    end
+
+    def selector(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'selector')
+    end
+
+  end
+end
--- a/lib/openshift/ingress_controller.rb
+++ b/lib/openshift/ingress_controller.rb
@@ -1,5 +1,5 @@
 module BushSlicer
-  class IngressController < ProjectResource
+  class IngressController < PodReplicator
     RESOURCE = "ingresscontrollers"
 
 
--- a/lib/openshift/service_monitor.rb
+++ b/lib/openshift/service_monitor.rb
@@ -1,0 +1,5 @@
+module BushSlicer
+  class ServiceMonitor < ProjectResource
+    RESOURCE = "servicemonitors"
+  end
+end
--- a/lib/openshift/deployment.rb
+++ b/lib/openshift/deployment.rb
@@ -15,6 +15,7 @@ module BushSlicer
       current:   %w[status replicas].freeze,
       updated:   %w[status updatedReplicas].freeze,
       available: %w[status availableReplicas].freeze,
+      ready: %w[status readyReplicas].freeze,
     }.freeze
 
     # we define this in method_missing so alias can't fly


lib/openshift/ingress_controller.rb
```diff
@@ -0,0 +1,24 @@
+module BushSlicer
+  class IngressController < ProjectResource
+    RESOURCE = "ingresscontrollers"
+
+
+    ### status related methods
+    def available_replicas(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'availableReplicas')
+    end
```
akostadinov: Can this be a schild of `PodReplicator` like `Deployment` and have this method inheritted?

lib/openshift/namespace.rb
```diff
@@ -0,0 +1,5 @@
+module BushSlicer
+  class Namespace < ProjectResource
+    RESOURCE = "namespaces"
```
akostadinov: I don't think `namespace` is a namespaces resource. Must be `ClusterResource` as far as I can tell.

lib/openshift/service_monitor.rb
```diff
@@ -0,0 +1,5 @@
+module BushSlicer
+  class ServiceMonitor < ProjectResource
+    RESOURCE = "servicemonitors"
```
akostadinov: did you verify this is a Project resource?

pruan-rht: I'm not 100% certain,  this is all I have from `oc explain`

> KIND:     ServiceMonitor
> VERSION:  monitoring.coreos.com/v1
> 
> DESCRIPTION:
>      <empty>

lib/openshift/namespace.rb
pruan-rht: ok, i'll change that.

lib/openshift/ingress_controller.rb
pruan-rht: I thought about it, but ingress and pods are two very different kind of things...it just didn't sound right to me.

lihongan: `ingresscontrollers.operator.openshift.io` is a CustomResource, maybe it should be a child of `ClusterResource`?

lib/openshift/service_monitor.rb
akostadinov: There is an easy way to figure out whether a resource is namespaced or not:

https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#not-all-objects-are-in-a-namespace

lib/openshift/ingress_controller.rb
akostadinov: Custom vs non-custom does not affect whether resource is namespaced or not [1] .

If ingress does not replicate pods, then you are right. If replicas are pods, then it must be `PodReplicator`. What are these replicas?


[1] https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#not-all-objects-are-in-a-namespace

akostadinov: Basically if `PodReplicator` methods can be implemented for ingress, then can be a subclass.

lihongan: Yes, `ingresscontroller` can replicate pods, the operator read the spec.replicas and update `deployment`.

lib/openshift/service_monitor.rb
lihongan: `ServiceMonitor` is namespaced resource.

lib/openshift/ingress_controller.rb
pruan-rht: @lihongan thanks for the tips.  @akostadinov I'll change it to inherit from `PodReplicator` then.

lib/openshift/service_monitor.rb
pruan-rht: @akostadinov, I'm not sure what/how to use that command mentioned in your link.  Can you please specify?

pruan-rht:
@akostadinov, oh, looks like I have anticipated that your assessment about `ingress-controller` to be correct and have submitted the update already.  PTAL.

lib/openshift/service_monitor.rb
akostadinov: Just run the commands and see list. Anyway, @lihongan has confirmed apparently.

lib/openshift/ingress_controller.rb
akostadinov: Also method `available_replicas` needs to be deleted and contant REPLICA_COUNTERS to be added instead. See `lib/openshift/deployment.rb` for example. 

Make sure that `#template` and `#containers_spec` methods from `PodReplicator` also make sense for this class. If they do NOT, then this is NOT a `PodReplicator` after all.

pruan-rht: @akostadinov hmm...it looks like we can't inherit it from PodReplicator as indicated below, there are no `template` nor `container_spec`

```apiVersion: operator.openshift.io/v1
kind: IngressController
metadata:
  creationTimestamp: "2019-05-28T17:04:03Z"
  finalizers:
  - ingresscontroller.operator.openshift.io/finalizer-ingresscontroller
  generation: 1
  name: default
  namespace: openshift-ingress-operator
  resourceVersion: "14383"
  selfLink: /apis/operator.openshift.io/v1/namespaces/openshift-ingress-operator/ingresscontrollers/default
  uid: 98af0220-816a-11e9-82c1-02b3adb00b00
spec: {}
status:
  availableReplicas: 2
  conditions:
  - lastTransitionTime: "2019-05-28T17:07:59Z"
    status: "True"
    type: Available
  domain: apps.qe-pruan-05-28.qe.devcluster.openshift.com
  endpointPublishingStrategy:
    type: LoadBalancerService
  selector: ingresscontroller.operator.openshift.io/deployment-ingresscontroller=default```

Comment (pruan-rht) [reply lib/openshift/ingress_controller.rb]: @akostadinov should we put it back to what I had originally then?

Comment (akostadinov) [reply lib/openshift/ingress_controller.rb]: yes, but we may create a class `Replicable` that only handles the `replicas` stuff. Then have `PodReplicator` inherit from `Replicable` and `IngressController` also inherit from `Replicable`. Is this explanation good enough?

Comment (lihongan) [reply lib/openshift/ingress_controller.rb]: Do we have a solution for this?
FYI, the `spec` is null in the default IngressController, but in a custom IngressController it might be like below:
```
spec:
  defaultCertificate:
    name: test-certs
  domain: router-test.example.com
  replicas: 1
  nodePlacement:
    nodeSelector:
      matchLabels:
        beta.kubernetes.io/os: linux
    tolerations:
    - effect: NoSchedule
      operator: Exists
  routeSelector:
    matchLabels:
      route: router-test
  namespaceSelector:
    matchLabels:
      namespace: router-test
```

Comment (akostadinov) [reply lib/openshift/ingress_controller.rb]: It should return `0` when `spec` is `nil`. So I think it would work fine.

Comment (lihongan) [reply lib/openshift/ingress_controller.rb]: The operator will set default (hard coded) value for some parameters when `spec` is nil. 
For example, replicas will be set to 2 if not defined.
```
spec: {}
status:
  availableReplicas: 2
```

akostadinov: Then `availableReplicas` should be read from `status`, not `spec`. I don't think this is an issue.

akostadinov:
resolve conflicts, otherwise lgtm

PR #216, opened 2019-05-28
Update OCP-9940 for 4.x to not interfere with each other

liangxia:
@chao007 @weherdh 

Add storage class to nfs PV/PVC so it does not interfere with each other.

Tested on bare metal (where there are pre-created iSCSI/NFS PVs) and vSphere (where there is default storage class), both work.


chao007:
LGTM

liangxia:
@akostadinov @pruan-rht

PR #220, opened 2019-05-29
step for dns basedomain

lihongan:
helper for openshift/cucushift#6991

@akostadinov @pruan-rht pelase help review.

### DIFF
--- a/features/step_definitions/route.rb
+++ b/features/step_definitions/route.rb
@@ -131,3 +131,13 @@ Given /^all default router pods become ready$/ do
       | #{label_filter} |
     })
 end
+
+Given /^dns basedomain is stored in the#{OPT_SYM} clipboard$/ do | cb_name |
+  step %Q/I run the :get admin command with:/, table(%{
+    | resource      | dnses.config         |
+    | resource_name | cluster              |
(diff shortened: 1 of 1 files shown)


features/step_definitions/route.rb
```diff
@@ -131,3 +131,13 @@
       | #{label_filter} |
     })
 end
+
+Given /^dns basedomain is stored in the#{OPT_SYM} clipboard$/ do | cb_name |
```
akostadinov: We already have a step `I store default router subdomain in the#{OPT_SYM} clipboard` and it does NOT require admin access. Better use it. It requires you to have a project though. Shouldn't be a big deal comapred to requirement to have admin access.

lihongan: I'd like to get the basedomain here, but actually the subdomain is `"apps." + basedomain` so it is OK for me to use it. Thanks.

lihongan:
I'm closing the PR since we have another similar step.

features/step_definitions/route.rb
akostadinov: In such case it is good to create your step , but lets base it on the existing step.

PR #222, opened 2019-05-30
fix running a command without arguments during install

akostadinov:
please hold on merging before we test change with master branch


akostadinov:
#223 needs to be tested first

PR #223, opened 2019-05-30
fix running a command without arguments during install

akostadinov:


akostadinov:
@jianlinliu , this should fix use case when running single command with spaces or other special characters. Just passing single string to `Process.spawn` causes it to use shell thus shell escaping is needed. So we use the third possible syntax - where we pass a two element array such that shell is avoided like with command+arguments thus shell escaping is not needed.

akostadinov:
/assign @pruan-rht 

lets merge this one if you don't see anything bad. There are too many use cases to be completely sure everything works.

jianlinliu:
@Aleksandar Kostadinov <akostadi@redhat.com> seem like this fix introduce
some critical regression issue.
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Launch%20Environment%20Flexy/61482/console

akostadinov:
fixed with #224, thank you for catching this

akostadinov:
also #225

PR #224, opened 2019-05-31
command might be array of two element array so handle it

akostadinov:
@jianlinliu, now this should fix #223 issue you reported, thank you!


jianlinliu:
@akostadinov just talked you with new finding via IRC, seem like still not working. Once any following fix, pls let me know, I will re-test.

PR #226, opened 2019-05-31
more robust CI script

akostadinov:


akostadinov:
Looks like script as it was was not working since the very beginning so I removed all old baggage. Also made it run on any machine, not only RHEL7. Reporting location now controlled by a dedicated environment variable.

PR #228, opened 2019-06-03
Update OCP-14382 for 4.x

liangxia:
@chao007  @weherdh


akostadinov:

-- saved reply

chao007:
Do we need to update the sc annotion like this one？ 
#https://github.com/openshift/cucushift/pull/6999 #

PR #229, opened 2019-06-03
Fix OCP-12151

xingxingxia:
@zhouying7780 please help review, thanks!

### DIFF
--- a/features/cli/deploy.feature
+++ b/features/cli/deploy.feature
@@ -599,9 +599,9 @@ Feature: deployment related features
     Given I wait up to 60 seconds for the steps to pass:
     """
     When I get project pods
-    And the output should contain:
-      | DeadlineExceeded |
-      | hooks-1          |
+    And the output should match:
+      | DeadlineExceeded\|Error |
+      | hooks-1                 |
     """
 
   # @author yinzhou@redhat.com


features/cli/deploy.feature
```diff
@@ -599,9 +599,9 @@ Feature: deployment related features
     Given I wait up to 60 seconds for the steps to pass:
     """
     When I get project pods
-    And the output should contain:
-      | DeadlineExceeded |
-      | hooks-1          |
+    And the output should match:
+      | DeadlineExceeded\|Error |
```
xingxingxia: Not robust. Failure log: http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2019/05/24/08:22:44/When_the_latest_deployment_failed_auto_rollback_to_the_active_deployment/console.html
Manual observation sees the status is first shown as DeadlineExceeded then quickly changes to Error.
Pass log: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/41474/console

akostadinov: Thank you for explanation. Such things also can be added as comments above the line in scenario. Because nobody would pull out the pull request to understand why... unless we don't really care after change has been done. In this case though it might be worth adding a comment.

PR #231, opened 2019-06-03
allow report generation for scheduled reports

pruan-rht:


features/step_definitions/metering.rb
```diff
@@ -325,3 +325,70 @@
...
+#         "spec" => {
+#           "reportingStart" => opts[:start_time],
+#           "reportingEnd" => opts[:end_time],
+#           "query" => opts[:query_type],
+#           "gracePeriod" => opts[:grace_period],
+#           "runImmediately" => opts[:run_immediately],
+#           "schedule" => schedule
+#         },
+#
+Then /^I generate a metering report with:$/ do |table|
```
akostadinov: shouldn't this be `When`?

lib/openshift/report.rb
```diff
@@ -21,7 +21,7 @@ def conditions(user: nil, cached: false, quiet: false)
 
     def running?(user: nil, cached: false, quiet: false)
       # the reason 'Scheduled' is an initial state what we
-      conditions.any? { |c| c.type == 'Running' and c.reason != 'Scheduled' }
+      conditions.any? { |c| c.type == 'Running' }
```
akostadinov: shold this check that the status is `true` or something? Conditions can be true or false, no?

akostadinov: If this is not the case, then you can merge without me if that's blocking you.

features/step_definitions/metering.rb
pruan-rht: updated.

lib/openshift/report.rb
pruan-rht: @akostadinov, this is correct, we are just checking the word `Running` is there.

PR #232, opened 2019-06-06
Use attribute storageClassName to replace annotations

liangxia:
From k8s [docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class),
In the past, the annotation volume.beta.kubernetes.io/storage-class was used instead of the storageClassName attribute. This annotation is still working, however it will become fully deprecated in a future Kubernetes release.

@chao007 This should addressed your concern in https://github.com/openshift/verification-tests/pull/228#issuecomment-499010882
@pruan-rht @akostadinov


chao007:
LGTM

liangxia:
@akostadinov @pruan-rht

PR #235, opened 2019-06-11
update login flow, add IDP option

yapei:
Using the `first` IDP blocks console automation because the first IDP is `kube:admin` on OCP clusters , running automation with `kube:admin` is dangerous and not allowed in cucushift
```
     [09:15:09] INFO> Exit Status: 0
      system projects visible to user kube:admin, clean-up too dangerous (RuntimeError)
      /Users/yapei/repo/verification-tests/lib/openshift/user.rb:119:in `clean_projects'
```
So we wonder if we can add IDP option as environment variables, that will be much easier/better.  

@yasun1 Can you please review the changes to see how online cases will be affected?
FYI @yanpzhan  @SSshahan  @xiaocwan

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -164,15 +164,15 @@ module BushSlicer
       return @admin_console_url
     end
 
-    def openshift_idp
-      unless @openshift_idp
-        if opts[:openshift_idp]
-          @openshift_idp = opts[:openshift_idp]
+    def idp
+      unless @idp
+        if opts[:idp]
+          @idp = opts[:idp]
         else
-          @openshift_idp = "kube:admin"
+          @idp = ''
         end
       end
-      return @openshift_idp
+      return @idp
     end
 
     def authentication_url
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -164,6 +164,17 @@ module BushSlicer
       return @admin_console_url
     end
 
+    def openshift_idp
+      unless @openshift_idp
+        if opts[:openshift_idp]
+          @openshift_idp = opts[:openshift_idp]
+        else
+          @openshift_idp = "kube:admin"
+        end
+      end
+      return @openshift_idp
+    end
+
     def authentication_url
       unless @authentication_url
         if opts[:authentication_url]
--- a/lib/rules/web/console/base/login.xyaml
+++ b/lib/rules/web/console/base/login.xyaml
@@ -14,8 +14,10 @@ login_sequence:
     - login_page_loaded
     if_element:
       selector: &pre_login_buttons
-        # we just choose the first button per OPENSHIFTQ-12062
-        xpath: //a[text()='<idp>']
+        # when idp is specified we use the correct IDP
+        # otherwise we always use the first one
+        # a hacking way to cover our requirements
+        xpath: //a[contains(@class, 'idp<idp>') or contains(@class, 'login-redhat<idp>') or text()='<idp>'][1]
       timeout: 1
   action:
     ref: login_console
--- a/lib/rules/web/console/base/login.xyaml
+++ b/lib/rules/web/console/base/login.xyaml
@@ -15,7 +15,7 @@ login_sequence:
     if_element:
       selector: &pre_login_buttons
         # we just choose the first button per OPENSHIFTQ-12062
-        css: ".login-redhat:first-of-type, .idp:first-of-type"
+        xpath: //a[text()='<idp>']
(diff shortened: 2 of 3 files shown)


yasun1:
@yuwang-RH Please help take a look.

yapei:
[Pass log on 4.1](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/41546/console)

lib/environment.rb
```diff
@@ -164,6 +164,17 @@ def admin_console_url
       return @admin_console_url
     end
 
+    def openshift_idp
+      unless @openshift_idp
+        if opts[:openshift_idp]
+          @openshift_idp = opts[:openshift_idp]
+        else
+          @openshift_idp = "kube:admin"
```
yapei: If no `openshift_idp` is defined `kube:admin` will be used, usually the first one

lib/rules/web/console/base/login.xyaml
```diff
@@ -15,15 +15,15 @@ login_sequence:
     if_element:
       selector: &pre_login_buttons
         # we just choose the first button per OPENSHIFTQ-12062
```
pruan-rht: @yapei seems like an outdated comment, pls remove it.

lib/rules/web/console/base/login.xyaml
```diff
@@ -15,15 +15,15 @@ login_sequence:
...
-        css: ".login-redhat:first-of-type, .idp:first-of-type"
+        xpath: //a[text()='<idp>']
```
akostadinov: I am concerned about this, that it will not work in case we don't know the idp name. I wonder if we can keep pushing first button in case IDP name not specified. WDYT?

yapei: I agree to always click the first IDP when no IDP is set, how can I update the code?

akostadinov: Appears that matching based on condition is tricky [1].

Another option is for the method `Environment#openshift_idp` method to execute API calls to figure out the IDP in case such is not provided in configuration. Is this possible?

[1] https://stackoverflow.com/questions/971067/is-there-an-if-then-else-statement-in-xpath

lib/environment.rb
```diff
@@ -164,6 +164,17 @@ def admin_console_url
       return @admin_console_url
     end
 
+    def openshift_idp
```
akostadinov: I think we can rename this to just `idp`.

lib/rules/web/console/base/login.xyaml
akostadinov: Need to play with `curl` and API to see if this is visible.

lib/rules/web/console/base/login.xyaml
yapei: ok, will do

lib/environment.rb
yapei: ok, I will update

lib/rules/web/console/base/login.xyaml
yapei: I realize that maybe there's issue if we execute API calls to get IDP when it's not provided, this operation will not be allowed for online, right?

yasun1: @yapei  definitely you can not. Online ENV is owned by Ops team, I believe in future the 'admin' idp will not be exposed out.

yapei: @akostadinov  Does it mean we can't run API calls to IDP with above limitation?

yapei: @akostadinov  Can you please take a look at this?

akostadinov: It is not about discussing. It is more about trying it out.

akostadinov: maybe it will be easier if we have `xpath: //a[text()='<idp>' or contains(@class, 'idp') or contains(@class, 'login-redhat')][0]`, wdyt?

The only issue I see is that in case `idp` is specified but not matched, then it would pick up first IDP, thus potential issues with IDP name might be missed.

yapei:
@akostadinov  @pruan-rht   Can you help confirm if it can be merged? I think all issues have been addressed and it would be great if we can merge it as soon as possible to unblock console automation. Thanks

akostadinov:
Thank you!

PR #236, opened 2019-06-11
Move annotation is-default-class from beta to stable

liangxia:
@chao007


chao007:
LGTM

liangxia:
@akostadinov @pruan-rht

PR #238, opened 2019-06-11
added a couple of new openshift resource objects.  

pruan-rht:
simple additions... I'll just merge it.  Helper for https://github.com/openshift/cucushift/pull/7001


akostadinov:
/lgtm

thanks!

PR #250, opened 2019-06-22
fix OLM cluster logging installation

pruan-rht:


akostadinov:
error has nothing to do with PR. Need to see if other PRs are affected and fix issue in test system if so. Will merge this one now.

PR #253, opened 2019-06-25
add one f5 router case

lihongan:
Since f5 router is deprecated since 3.11 so I'd like to merge this to branch v3 only.
This case is for smoke testing and ose-f5-router image testing as well.
 
logs:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/41784/consoleFull

### DIFF
--- a/features/routing/f5-router.feature
+++ b/features/routing/f5-router.feature
@@ -1,0 +1,33 @@
+Feature: F5 router related scenarios
+
+  # @author hongli@redhat.com
+  # @case_id OCP-24080
+  @admin
+  @destructive
+  Scenario: the f5 router image should contains openssh-clients package
+    Given I switch to cluster admin pseudo user
+    And I use the router project
+    And default router image is stored into the :router_image clipboard
+    And default router replica count is restored after scenario
+    When I run the :scale client command with:
+      | resource | dc     |
+      | name     | router |
+      | replicas | 0      |
+    Then the step should succeed
+    Given admin ensures "f5router" dc is deleted after scenario
+    And admin ensures "f5router" service is deleted after scenario
+    And admin ensures "external-host-private-key-secret" secret is deleted after scenario
+    And I download a file from "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/routing/fake_id_rsa"
+    When I run the :oadm_router admin command with:
+      | name                      | f5router                                    |
+      | type                      | f5-router                                   |
+      | images                    | <%= cb.router_image.gsub("haproxy","f5") %> |
+      | external_host             | 10.66.144.115                               |
+      | external_host_username    | username                                    |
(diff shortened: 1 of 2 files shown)


features/routing/f5-router.feature
```diff
@@ -0,0 +1,33 @@
...
+    Then the step should succeed
+    Given admin ensures "f5router" dc is deleted after scenario
+    And admin ensures "f5router" service is deleted after scenario
+    And admin ensures "external-host-private-key-secret" secret is deleted after scenario
+    And I download a file from "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/routing/fake_id_rsa"
+    When I run the :oadm_router admin command with:
+      | name                      | f5router                                    |
+      | type                      | f5-router                                   |
+      | images                    | <%= cb.router_image.gsub("haproxy","f5") %> |
+      | external_host             | 10.66.144.115                               |
```
akostadinov: mybe use step `F5 router public IP is stored in the :vserver_ip clipboard` instead of hardcoding IP.

features/routing/f5-router.feature
```diff
@@ -0,0 +1,33 @@
...
+  @destructive
+  Scenario: the f5 router image should contains openssh-clients package
+    Given I switch to cluster admin pseudo user
+    And I use the router project
+    And default router image is stored into the :router_image clipboard
+    And default router replica count is restored after scenario
+    When I run the :scale client command with:
+      | resource | dc     |
+      | name     | router |
+      | replicas | 0      |
```
akostadinov: This is going to work only with `dc` router. I thought that we have a special step. Are we ever using non-dc routers? If so it is worth writing a step. Maybe after PR is merged.

lihongan: We are always using dc for routers (haproxy and f5) in 3.x, but using deployment since 4.1. If needed we can have a step and update related scenarios later.

features/routing/f5-router.feature
lihongan: The step `F5 router public IP is stored in the :vserver_ip clipboard` is actually a flag that indicates the manual preparation steps for F5 have been finished. This case is different from the ones we automated before. This one expects to can be run on any cluster without manual intervention and just verify the image and its package. So we can take it as CVP case.
This IP can be an "unreachable" IP address in this scenario, the f5 router pod will time out and restart after a while but we have enough time the check the openssh-clients package.

akostadinov: So the IP is irrelevant? I was concerned that we are exposing an internal IP. Is this IP meaningful at all?

features/routing/f5-router.feature
akostadinov: Got it. we can keep as is for the time being.

PR #259, opened 2019-07-02
Simplify the logic of default storage class

liangxia:
@chao007 @akostadinov @pruan-rht 

1. The logic of delete the default storage class, then create a new default storage class looks weird.
2. Default storage class can not be deleted in 4.x (Automation will quit prematurely due to failed to recover the deleted default storage class as it already exist).

### DIFF
--- a/features/storage/storage_class.feature
+++ b/features/storage/storage_class.feature
@@ -167,17 +167,8 @@ Feature: storageClass related feature
       | azure-disk  | # @case_id OCP-13490
 
   # @author lxia@redhat.com
-  @admin
-  @destructive
   Scenario Outline: New created PVC without specifying storage class use default class when only one class is marked as default
-    Given default storage class is deleted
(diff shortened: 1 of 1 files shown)


akostadinov:
This will create a discrepancy between 3.x and 4.x cases. Former will be destructive and latter non-destructive. I wonder if we should split the test management system cases for 3.x and 4.x.

Other than that LGTM. I'll merge now.

/lgtm

liangxia:
> This will create a discrepancy between 3.x and 4.x cases. Former will be destructive and latter non-destructive. I wonder if we should split the test management system cases for 3.x and 4.x.
> 

@akostadinov  PR #261 for 3.x
1. OCP 3.4 & 3.5 now end of maintenance (effective end of April, 2019).
2. We have default storage class on 3.6 (verified with OCP v3.6.173.0.164 on Openstack).

PR #260, opened 2019-07-02
fix OCP-19532

yapei:
@yanpzhan  @SSshahan  Can you please help review?
All test cases have passed locally, when no changes are requested I can submit Jenkins pass log

### DIFF
--- a/lib/openshift/project.rb
+++ b/lib/openshift/project.rb
@@ -150,6 +150,11 @@ module BushSlicer
     end
     alias_method :pods, :get_pods
 
+    def get_jobs(by:, **get_opts)
+      Job.list(user: by, project: self, **get_opts)
+    end
+    alias_method :jobs, :get_jobs
+
     def get_builds(by:, **get_opts)
       Build.list(user: by, project: self, **get_opts)
     end
--- a/lib/openshift/project.rb
+++ b/lib/openshift/project.rb
@@ -150,10 +150,9 @@ module BushSlicer
     end
     alias_method :pods, :get_pods
 
-    def get_jobs(by:, **get_opts)
+    def jobs(by:, **get_opts)
       Job.list(user: by, project: self, **get_opts)
     end
-    alias_method :jobs, :get_jobs
 
     def get_builds(by:, **get_opts)
       Build.list(user: by, project: self, **get_opts)
--- a/lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
+++ b/lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
@@ -46,7 +46,7 @@ check_resources_title_on_page:
 check_resource_name_and_icon:
   elements:
   - selector:
-      xpath: //div[@class='co-m-pane__name'][/*[contains(.,'<icon_text>')] and /*[contains(., '<resource_name>')]]
+      xpath: //div[contains(@class,'co-m-pane__name')][/*[contains(.,'<icon_text>')] and /*[contains(., '<resource_name>')]]
 check_page_contains:
   element:
     selector:
--- a/lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
+++ b/lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
@@ -50,7 +50,7 @@ check_resource_name_and_icon:
 check_page_contains:
   element:
     selector:
-      xpath: //*[contains(.,'<content>')]
+      visible_text: !ruby/regexp /<content>/i
 submit_changes:
   element:
     selector:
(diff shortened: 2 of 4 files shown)


lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
```diff
@@ -50,7 +50,7 @@ check_resource_name_and_icon:
 check_page_contains:
   element:
     selector:
-      xpath: //*[contains(.,'<content>')]
+      visible_text: !ruby/regexp /<content>/i
```
yapei: Change text matching to `regexp` to avoid failures caused by upper/lower cases

lib/rules/web/admin_console/4.1/filters.xyaml
```diff
@@ -45,14 +45,14 @@ check_results_contain_correct_type:
       css: span.co-m-resource-icon
   scripts:
     - command: |
-        var resourceicons = document.querySelectorAll('span.co-m-resource-icon');
+        var resourceicons = document.querySelectorAll('span.co-m-resource-icon.co-m-resource-<type>');
```
yapei: css changes

lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
```diff
@@ -46,11 +46,11 @@ check_resources_title_on_page:
 check_resource_name_and_icon:
   elements:
   - selector:
-      xpath: //div[@class='co-m-pane__name'][/*[contains(.,'<icon_text>')] and /*[contains(., '<resource_name>')]]
+      xpath: //div[contains(@class,'co-m-pane__name')][/*[contains(.,'<icon_text>')] and /*[contains(., '<resource_name>')]]
```
yapei: css changes

lib/openshift/project.rb
```diff
@@ -150,6 +150,11 @@ def get_pods(by:, **get_opts)
     end
     alias_method :pods, :get_pods
 
+    def get_jobs(by:, **get_opts)
```
yapei: Bring the changes in #65 back for use in OCP-19672

yanpzhan:
The fix LGTM.

lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
SSshahan: Do we really want check content  with case insensitive matching here?

lib/openshift/project.rb
yapei: @akostadinov  can you please check if I added correctly?

lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
yapei: I'm afraid yes, sometimes the text changes from `No Events -> No events` while sometimes vice visa. Using case insensitive matching can make sure the whole texts exist no matter upper or lower cases. In fact we don't care too much whether the message are in upper or lower cases

SSshahan: Make sense. And How do you think change the rule name to `check_page_match` since it is sort of fuzzy matching?

SSshahan:
LGTM, thanks!

lib/openshift/project.rb
```diff
@@ -150,6 +150,11 @@ def get_pods(by:, **get_opts)
...
+      Job.list(user: by, project: self, **get_opts)
+    end
+    alias_method :jobs, :get_jobs
```
akostadinov: where is this method used? If not used, then better not add it.

If we need it added, then `get_jobs` is not needed. Keeping some `get_something` methods for historic reasons only. They don't look good.

akostadinov:
Other than one comment, lgtm.

lib/openshift/project.rb
yapei: It's used in step `a job appears with labels` and used in [feature](https://github.com/openshift/cucushift/blob/536d14aef3434e5d1bc9179c90866007fe280893/web/admin-console/cronjobs.feature#L45)

lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
yapei: Do we need exact matching for web cases?

yapei: Ok, let me change it

lib/openshift/project.rb
yapei: also `get_jobs` is removed from `alias_method`

lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
yapei: also feature updates see #7021 in cucushift

SSshahan: LGTM now. thanks

yapei:
@akostadinov  Can you help review again?

PR #261, opened 2019-07-03
Simplify the logic of default storage class

liangxia:
CP of #259


akostadinov:
Thank you!

/lgtm

PR #263, opened 2019-07-03
Remove region/zone from storage class as 4.x does not support that

liangxia:

3.x supports parameter zone in storage class, while 4.x does not support that.
Update 4.x to clone from default storage class to get rid of the unsupported parameter.

Not touching the auto script parameters, so it can support both 3.x and 4.x (though some parameters are not used in 4.x).


akostadinov:
/lgtm

I think that you can remove this parameter if it is not supported by 4.x but also cloning from default, which afaiu is mandatory, is also fine. Thanks!

PR #264, opened 2019-07-03
Fix rules for OCP-19577

SSshahan:
@yapei  fix rules for OCP-19577 Check project page, please help to review. Thanks.

### DIFF
--- a/features/step_definitions/resources.rb
+++ b/features/step_definitions/resources.rb
@@ -174,8 +174,9 @@ Given /^I wait for the resource "(.+)" named "(.+)" to disappear(?: within (\d+)
     res = user.cli_exec(:get, **opts)
     case res[:response]
     # the resource has terminated which means we are done waiting.
-    when /Error from server (Forbidden): projects.project.openshift.io/, /not found/, /No resources found/
+    when /Error from server \((Forbidden\)): projects.project.openshift.io/, /not found/, /No resources found/
       break true
+
     end
   }
   res[:success] = success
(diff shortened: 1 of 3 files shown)


features/step_definitions/resources.rb
```diff
@@ -174,8 +174,9 @@
     res = user.cli_exec(:get, **opts)
     case res[:response]
     # the resource has terminated which means we are done waiting.
-    when /Error from server (Forbidden): projects.project.openshift.io/, /not found/, /No resources found/
+    when /Error from server \((Forbidden\)): projects.project.openshift.io/, /not found/, /No resources found/
       break true
```
SSshahan: It should escape brackets here, or it will not match always until timeout no matter if the response is expected.

lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
```diff
@@ -24,6 +24,11 @@ click_cancel_button:
   - selector:
       xpath: //button[contains(text(),'Cancel')]
     op: click
+click_button_type_on_page:
```
yapei: How about adding a general action maybe named as`click_button` which can be used for any button clicking
`//button[@type='button' and contains(.,'<button_text>')]`

yapei: I see we have many similar actions, only the button name is different. Adding a general action saves/reduces more duplicate lines.

SSshahan: This rule is not very general actually, it only can be used for those type is button.  And I found that this kind of button were not as much as we thought. I can update the action name to `click_button`

SSshahan: OK I will update

SSshahan:
@akostadinov @pruan-rht  could you help to merge?

features/step_definitions/resources.rb
akostadinov: There is no need to have extra brackets here. it can be `\(Forbidden\)`

PR #265, opened 2019-07-04
Fix OCP-10465

liangxia:

In 4.x, the error is `Error from server (Forbidden): persistentvolumes is forbidden: User "testuser-23" cannot list resource "persistentvolumes" in API group "" at the cluster scope`


liangxia:
Log, https://privatebin-it-iso.int.open.paas.redhat.com/?4f8895e2d8c9204d#gql1jN+aWG8+9s2D9AIQioWoNGIdIbBnggg1sxoSP30=

PR #268, opened 2019-07-05
auto 4.2 caseOCP-23681

yapei:


lib/rules/web/admin_console/4.1/browser_opts.yml
```diff
@@ -0,0 +1,40 @@
+---
+:scroll_strategy: 1
```
yapei: Looks like this file is missed when 4.1 rules were copied, bring it back

yapei:
@yanpzhan  @SSshahan  Can you please help review? Thanks

lib/rules/web/admin_console/4.2/actions.xyaml
```diff
@@ -0,0 +1,57 @@
+click_one_dropdown_action:
+  action: click_resource_action_button
+  action: choose_item_from_list
+click_resource_action_button:
+  elements:
+  - selector:
+      xpath: //button[contains(.,'Actions')]
```
yapei: rules is changed from `/button//span[contains(text(),‘Actions’)]` to `//button[contains(.,'Actions')]`, when using `/button//span[contains(text(),‘Actions’)]` overlap is found and element is not clickable

lib/rules/web/admin_console/4.2/storage.xyaml
```diff
@@ -0,0 +1,70 @@
...
+  - selector:
+      id: storageclass-dropdown
+    type: button
+    op: click
+    timeout: 20
+  - selector:
+      id: <storage_class_name>-link
+    type: a
+    op: click
+set_pvc_name:
```
yapei: line 24 - line 70 is newly added

akostadinov:

-- saved reply

lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
```diff
@@ -0,0 +1,301 @@
...
+  elements:
+  - selector:
+      xpath: //button[contains(text(),'Create')]
+    op: click
+click_cancel_button:
+  elements:
+  - selector:
+      xpath: //button[contains(text(),'Cancel')]
+    op: click
+click_button:
```
yapei: `click_button_on_page` is changed to `click_button` with parameter `<text>` changed to `<button_text>` @SSshahan

yapei: https://github.com/openshift/verification-tests/blob/master/lib/rules/web/admin_console/4.1/common_ui_elements.xyaml#L27

SSshahan: Do we need update this rule in 4.1?

SSshahan:
It looks good to me. Thanks.

PR #273, opened 2019-07-09
Fix openstack auth-url

liangxia:
CP #272 to v3


akostadinov:
Is this valid for 3.x?

liangxia:
> Is this valid for 3.x?

Yes, it is. We are using upshift for OCP 3.x, and upshift is using v3 auth.

PR #274, opened 2019-07-09
support to allocate floating ip from known network name

wjiangjay:
Add a helper function to allocate floating ip from a known network.
This is step one for ipi on osp automation support.

### DIFF
--- a/lib/launchers/openstack10.rb
+++ b/lib/launchers/openstack10.rb
@@ -631,24 +631,28 @@ module BushSlicer
       url = self.os_network_url + "/v2.0/floatingips/#{floating_ip_id}"
       params = {}
       fip_res = self.rest_run(url, "GET", params, self.os_token)
-      unless fip_res[:success]
-        raise "could not get floating ip \"#{floating_ip_id}\""
-      end
-      logger.warn("deleteing floating ip \"#{floating_ip_id}\"")
-      res = self.rest_run(url, "DELETE", params, self.os_token)
-      unless res[:success]
-        raise "failed to delete floating ip \'#{floating_ip_id}\""
-      end
-      1.upto(60) do
-        sleep 10
-        res = self.rest_run(url, "GET", params, self.os_token)
-        if res[:exitstatus] == 404
-          return true
-        else
-          logger.info("Wait for 10s to delete floating ip #{floating_ip_id}")
+      if fip_res[:exitstatus] == 404
+        logger.warn("the floating ip \"#{floating_ip_id}\" is already deleted")
+        return true
+      elsif fip_res[:success]
+        logger.warn("deleteing floating ip \"#{floating_ip_id}\"")
+        res = self.rest_run(url, "DELETE", params, self.os_token)
+        unless res[:success]
+          raise "failed to delete floating ip \'#{floating_ip_id}\""
         end
+        1.upto(60) do
+          sleep 10
+          res = self.rest_run(url, "GET", params, self.os_token)
+          if res[:exitstatus] == 404
+            return true
+          else
+            logger.info("Wait for 10s to delete floating ip #{floating_ip_id}")
+          end
+        end
+        raise "could not delete floating ip #{floating_ip_id}"
+      else
+        raise "we got some problem to fetch floating ip\n#{fip_res[:response]}"
       end
-      raise "could not delete floating ip #{floating_ip_id}"
     end
 
     # @param [Array<Hash>] launch_opts where each element is in the format
@@ -711,7 +715,7 @@ module BushSlicer
       end
     end
     
-    def allocate_floating_ip(network_name, reuse=true)
+    def allocate_floating_ip(network_name, reuse: true)
       network = floating_ip_networks.find { |n| n["name"] == network_name }
       unless network
         raise "could not find network #{network_name} in current tenant."
--- a/lib/launchers/openstack10.rb
+++ b/lib/launchers/openstack10.rb
@@ -688,26 +688,49 @@ module BushSlicer
     end
 
     def allocate_floating_ip(network_name)
-      request_url = self.os_network_url + "/v2.0/floatingips"
-      method = "POST"
+      fipsres = self.rest_run(self.os_network_url + "/v2.0/floatingips", "GET", {}, self.os_token)
+      if fipsres[:success]
+        fips = fipsres.dig(:parsed, "floatingips")
+      else
+        raise "Could not get all existing floating ips\n#{fipsres[:response]}"
+      end
+
       network = floating_ip_networks.find { |n| n["name"] == network_name }
       unless network
-        raise "could not find network in current tenant."
+        raise "could not find network #{network_name} in current tenant."
       end
-      network_id = network["id"]
-      payload = {
-        floatingip: {
-          floating_network_id: network_id,
-          project_id: self.os_tenant_id,
-          description: "IP allocated automatically by BushSlicer"
-        }
+
+      # filter condition is
+      # 1, floating ip is not belong to given network_name and
+      # 2, floating ip is not preserved
+      # 3, floating ip is not associated to instance
+      filtered_fips = fips.find_all { |n|
+        n["floating_network_id"] == floating_ip_networks.select { |f|
+          f["name"] == network_name
+        }.first["id"] && !n["description"][/[Pp]reserve/] &&
+        !n["fixed_ip_address"]
       }
-      res = self.rest_run(request_url, method, payload, self.os_token)
-      logger.debug(res[:response])
-      if res[:success]
-        return res.dig(:parsed, "floatingip")
+      unless filtered_fips.empty?
+        return filtered_fips.sample
       else
-        raise "could not allocate new floating ip:\n#{res[:response]}"
+        request_url = self.os_network_url + "/v2.0/floatingips"
+        method = "POST"
+  
+        network_id = network["id"]
+        payload = {
+          floatingip: {
+            floating_network_id: network_id,
+            project_id: self.os_tenant_id,
+            description: "IP allocated automatically by OpenShift verification-tests"
+          }
+        }
+        res = self.rest_run(request_url, method, payload, self.os_token)
+        logger.debug(res[:response])
+        if res[:success]
+          return res.dig(:parsed, "floatingip")
+        else
+          raise "could not allocate new floating ip:\n#{res[:response]}"
+        end
       end
     end
 
--- a/lib/launchers/openstack10.rb
+++ b/lib/launchers/openstack10.rb
@@ -687,6 +687,29 @@ module BushSlicer
       end
     end
 
+    def allocate_floating_ip(network_name)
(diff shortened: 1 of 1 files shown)


wjiangjay:
```pry
[1] pry(#<BushSlicer::DefaultWorld>)> osp = iaas_by_service("openstack_upshift")
[3] pry(#<BushSlicer::DefaultWorld>)> res = osp.allocate_floating_ip("provider_net_shared_2")
[5] pry(#<BushSlicer::DefaultWorld>)> res[:parsed]
=> {"floatingip"=>
  {"router_id"=>nil,
   "status"=>"DOWN",
   "description"=>"IP allocated automatically by BushSlicer",
   "tags"=>[],
   "tenant_id"=>"542c6ebd48bf40fa857fc245c7572e30",
   "created_at"=>"2019-07-09T06:08:00Z",
   "updated_at"=>"2019-07-09T06:08:00Z",
   "floating_network_id"=>"14c15d33-175c-424e-88ba-361a875e0c5c",
   "fixed_ip_address"=>nil,
   "floating_ip_address"=>"10.0.79.73",
   "revision_number"=>0,
   "project_id"=>"542c6ebd48bf40fa857fc245c7572e30",
   "port_id"=>nil,
   "id"=>"4196d20a-baa3-4d20-a1be-e89fbf053ce4",
   "qos_policy_id"=>nil}}
[7] pry(#<BushSlicer::DefaultWorld>)> res[:parsed]["floatingip"]["floating_ip_address"]
=> "10.0.79.73"

```

lib/launchers/openstack10.rb
```diff
@@ -687,6 +687,29 @@ def assign_ip(instance)
...
+      network = get_networks.find {|n| n["name"] == network_name}
+      unless network
+        raise "could not find network in current tenant."
+      end
+      network_id = network["id"]
+      payload = {
+        floatingip: {
+          floating_network_id: network_id,
+          project_id: self.os_tenant_id,
+          description: "IP allocated automatically by BushSlicer"
```
akostadinov: project name is no longer BushSlicer, we just didn't figure out yet how to `sed` module/class names

lib/launchers/openstack10.rb
```diff
@@ -687,6 +687,29 @@ def assign_ip(instance)
       end
     end
 
+    def allocate_floating_ip(network_name)
+      request_url = self.os_network_url + "/v2.0/floatingips"
+      method = "POST"
+      network = get_networks.find {|n| n["name"] == network_name}
```
akostadinov: If you only allocate floating IPs, then use method `#floating_ip_networks` instead of `#get_networks`. If you can allocate IPs from networks that are NOT `router:external`, then you can rename the method to `allocate_ip` instead of `allocate_floating_ip`.

lib/launchers/openstack10.rb
```diff
@@ -687,6 +687,29 @@ def assign_ip(instance)
...
+      payload = {
+        floatingip: {
+          floating_network_id: network_id,
+          project_id: self.os_tenant_id,
+          description: "IP allocated automatically by BushSlicer"
+        }
+      }
+      res = self.rest_run(request_url, method, payload, self.os_token)
+      if res[:success]
+        return res
```
akostadinov: Returning whole result is inconvenient for user. Better return `res.dig(:parsed, "floatingip")` such that user gets the information about the IP without noise.

lib/launchers/openstack10.rb
```diff
@@ -687,6 +687,29 @@ def assign_ip(instance)
...
+      end
+      network_id = network["id"]
+      payload = {
+        floatingip: {
+          floating_network_id: network_id,
+          project_id: self.os_tenant_id,
+          description: "IP allocated automatically by BushSlicer"
+        }
+      }
+      res = self.rest_run(request_url, method, payload, self.os_token)
```
akostadinov: Adding a line here would be nice:
```
logger.debug(res[:response])
```

lib/launchers/openstack10.rb
wjiangjay: ok, got it, will update later

lib/launchers/openstack10.rb
wjiangjay: ok, got, will update later.

lib/launchers/openstack10.rb
wjiangjay: got you

wjiangjay:
```pry
[1] pry(#<BushSlicer::DefaultWorld>)> osp = iaas_by_service("openstack_upshift")                                                         
[2] pry(#<BushSlicer::DefaultWorld>)> res = osp.allocate_floating_ip("provider_net_shared_2")                                            
=> {"router_id"=>nil,
 "status"=>"DOWN",
 "description"=>"IP allocated automatically by BushSlicer",
 "tags"=>[],
 "tenant_id"=>"542c6ebd48bf40fa857fc245c7572e30",
 "created_at"=>"2019-07-09T09:53:17Z",
 "updated_at"=>"2019-07-09T09:53:17Z",
 "floating_network_id"=>"14c15d33-175c-424e-88ba-361a875e0c5c",
 "fixed_ip_address"=>nil,
 "floating_ip_address"=>"10.0.79.42",
 "revision_number"=>0,
 "project_id"=>"542c6ebd48bf40fa857fc245c7572e30",
 "port_id"=>nil,
 "id"=>"4aabadd9-b866-4949-8474-d81c4301262a",
 "qos_policy_id"=>nil}
[3] pry(#<BushSlicer::DefaultWorld>)> res["floating_ip_address"]                                                                         
=> "10.0.79.42"

```

akostadinov:
How will this be used? I'm thinking that we better use existing IP if there are already allocated. But only if conflicts can be avoided, otherwise we need a system to clean-up unused allocated IPs.

lib/launchers/openstack10.rb
akostadinov: We can write it as `IP allocated automatically by OpenShift verification tests.`

wjiangjay:
@akostadinov this is for IPI on OSP, and it requires there should have a pre-provisioned floating ip within the given External Network name.
Means, it can not work if the External Network of the floating ip is different with the External Network in the install-config.yaml which we use to create the resources.

So, here I make the network name as the parameter of the function, then we can easily make a consistent name of External Network between floating ip and install-config.yaml

wjiangjay:
@akostadinov 
I make this reuse existing floating ip before create new one, but due to this, I make a complicated rules to filter existing floating ips.

lib/launchers/openstack10.rb
```diff
@@ -687,6 +687,50 @@ def assign_ip(instance)
...
+      # 1, floating ip is not belong to given network_name and
+      # 2, floating ip is not preserved
+      # 3, floating ip is not associated to instance
+      filtered_fips = fips.find_all { |n|
+        n["floating_network_id"] == network["id"] && !n["description"][/[Pp]reserve/] && !n["fixed_ip_address"]
+      }
+      unless filtered_fips.empty?
+        return filtered_fips.sample
+      else
+        request_url = self.os_network_url + "/v2.0/floatingips"
```
akostadinov: indentation from this line down is deeper

akostadinov: Other than that looks good.

One thing to consider, I tried to touch on that with my previous comment, is that if there is time between selecting and using the IP, it is possible to have a conflict between two installations.

e.g. 
1. install_1 selects IP
2. install_2 selects same IP
3. both installations fight at some point for the IP

wjiangjay: Good point, it will, and only one of two install procedure will success. 
So any suggestion to make a lock?

akostadinov: the other option would be to always allocate a new one but deallocate daily old IPs.

wjiangjay:
For the ipi on osp, will always use `reuse=false` to avoid conflicts.

```
[5] pry(#<BushSlicer::DefaultWorld>)> osp.allocate_floating_ip("provider_net_shared_2")
=> {"router_id"=>nil,
 "status"=>"DOWN",
 "description"=>"",
 "tags"=>[],
 "tenant_id"=>"542c6ebd48bf40fa857fc245c7572e30",
 "created_at"=>"2019-07-10T16:13:08Z",
 "updated_at"=>"2019-07-10T16:13:08Z",
 "floating_network_id"=>"14c15d33-175c-424e-88ba-361a875e0c5c",
 "fixed_ip_address"=>nil,
 "floating_ip_address"=>"10.0.78.26",
 "revision_number"=>0,
 "project_id"=>"542c6ebd48bf40fa857fc245c7572e30",
 "port_id"=>nil,
 "id"=>"9a819ff2-8058-46f4-a6e5-1556c61b6763",
 "qos_policy_id"=>nil}
[6] pry(#<BushSlicer::DefaultWorld>)> osp.allocate_floating_ip("provider_net_shared_2", reuse=false)
=> {"router_id"=>nil,
 "status"=>"DOWN",
 "description"=>"IP allocated automatically by OpenShift verification-tests",
 "tags"=>[],
 "tenant_id"=>"542c6ebd48bf40fa857fc245c7572e30",
 "created_at"=>"2019-07-11T08:17:48Z",
 "updated_at"=>"2019-07-11T08:17:48Z",
 "floating_network_id"=>"14c15d33-175c-424e-88ba-361a875e0c5c",
 "fixed_ip_address"=>nil,
 "floating_ip_address"=>"10.0.79.65",
 "revision_number"=>0,
 "project_id"=>"542c6ebd48bf40fa857fc245c7572e30",
 "port_id"=>nil,
 "id"=>"d1c50720-819e-4cc0-94fc-5f4822a42b9b",
 "qos_policy_id"=>nil}
```

lib/launchers/openstack10.rb
```diff
@@ -686,6 +710,54 @@ def assign_ip(instance)
         raise "could not associate a floating ip:\n#{res[:response]}"
       end
     end
+    
+    def allocate_floating_ip(network_name, reuse=true)
```
akostadinov: better to use `reuse: true`. Then when you call the method it will be `allocate_floating_ip("my-network-name", reuse: false)` for example. In this way reader knows easier what this second parameter means.

lib/launchers/openstack10.rb
```diff
@@ -627,6 +627,30 @@ def delete_instance(instance_name)
...
+        sleep 10
+        res = self.rest_run(url, "GET", params, self.os_token)
+        if res[:exitstatus] == 404
+          return true
+        else
+          logger.info("Wait for 10s to delete floating ip #{floating_ip_id}")
+        end
+      end
+      raise "could not delete floating ip #{floating_ip_id}"
+    end
```
akostadinov: I think that if IP doesn't exist to begin with, there is no need to fail. IMO if first check for IP it shows 404, then just return without failure.

lib/launchers/openstack10.rb
wjiangjay: make sense

lib/launchers/openstack10.rb
wjiangjay: Should be resolved now.

PR #278, opened 2019-07-11
added related objects support in cluster_operator class

pruan-rht:
helper for https://github.com/openshift/cucushift/pull/7026

### DIFF
--- a/lib/openshift/cluster_operator.rb
+++ b/lib/openshift/cluster_operator.rb
@@ -1,3 +1,5 @@
+require 'openshift/flakes/related_object'
+
 module BushSlicer
   class ClusterOperator < ClusterResource
     RESOURCE = "clusteroperators"
@@ -23,8 +25,8 @@ module BushSlicer
     #     name: gp2
     #     resource: storageclasses
     def related_objects(user: nil, cached: true, quiet: false)
-      rr = raw_resource(user: user, cached: cached, quiet: quiet)
-      rr.dig('status', 'relatedObjects')
+      ro_raw = raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'relatedObjects')
+      ro_raw.map { |ro|  RelatedObject.new ro }
     end
 
     # @param filter_by [String] the hash key to filter by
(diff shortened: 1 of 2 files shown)


lib/openshift/cluster_operator.rb
```diff
@@ -11,5 +11,28 @@ def version_exists?(version:, user: nil, cached: true, quiet: false)
...
+    #   - group: config.openshift.io
+    #     name: cluster
+    #     resource: infrastructures
+    #   - group: storage.k8s.io
+    #     name: gp2
+    #     resource: storageclasses
+    def related_objects(user: nil, cached: true, quiet: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
+      rr.dig('status', 'relatedObjects')
+    end
```
akostadinov: Could you make this return objects from the OpenShift classes instead of the raw map?

See my comment:
https://github.com/openshift/cucushift/pull/7026#issuecomment-509797154

pruan-rht: @akostadinov I've added a related_object class under `lib/openshift/flakes` to represent the hash.  PTAL again.  Thanks!

akostadinov: Actually the related objects are not a special kind of object but:

Resource name "namespaces" of kind "openshift-cluster-storage-operator",
resource name "cluster" of kind "config.openshift.io"
etc.

So we need a way to generate these objects when they are already defined, or a fallback object otherwise. If I can't explain clear I can take a look later to do myself or at least pseudo code to make more clear. Please let me know.

lib/openshift/flakes/related_object.rb
```diff
@@ -0,0 +1,27 @@
+module BushSlicer
+
+  # this class should help with parsing status.relatedObjects
+  # https://github.com/openshift-qe/output_references/blob/master/clusteroperators/service-ca-operator.yaml
+  class RelatedObject
+
+    attr_reader :struct
+    private :struct
+
+    def initialize(struct)
```
akostadinov, line 10: I would suggest to check `lib/openshift/flakes/image_ref.rb`. This class can convert the object to actual resource. Or fail if such resource class is not defined already.

lib/openshift/cluster_operator.rb
akostadinov: bump, what to discuss this?

PR #279, opened 2019-07-11
Split 13914 mapping the e2e test scripts

xiuwang:
Passed in the local.

### DIFF
--- a/features/build/buildlogic.feature
+++ b/features/build/buildlogic.feature
@@ -258,65 +258,173 @@ Feature: buildlogic.feature
 
   # @author xiuwang@redhat.com
   # @case_id OCP-13914
+  # Split 13914 mapping the e2e test scripts
   Scenario: Prune old builds automaticly
+    #Should prune completed builds based on the successfulBuildsHistoryLimit setting	
     Given I have a project
-    When I run the :new_app client command with:
-      | image_stream | ruby                                          |
-      | code         | https://github.com/openshift/ruby-hello-world |
+    When I run the :new_build client command with:
+      | D    | FROM busybox\nRUN touch /php-file |
+      | name | myphp                             |
     Then the step should succeed
-    When I run the :get client command with:
-      | resource      | buildconfig      |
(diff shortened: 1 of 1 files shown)


features/build/buildlogic.feature
```diff
@@ -258,65 +258,173 @@ Feature: buildlogic.feature
...
+    When I run the :new_build client command with:
+      | app_repo | https://github.com/openshift/ruby-hello-world.git |
+    Then the step should succeed
+    When I run the :describe client command with:
+      | resource | buildconfig      |
+      | name     | ruby-hello-world |
+    Then the step should succeed
+    Then the output should contain:
+      | Builds History Limit |
+      | Successful:	5    |
```
wewang58: better to align

wewang58: other lgtm

features/build/buildlogic.feature
```diff
@@ -258,65 +258,173 @@ Feature: buildlogic.feature
...
+      | Git.*Cancelled |
     Then the output should not contain:
-      |ruby-hello-world-1|
+      | rails-postgresql-example-1 |
     """
-    Given I run the steps 3 times:
+
+  # @author xiuwang@redhat.com
+  # @case_id OCP-24155
+  # Split 13914 mapping the e2e test scripts
```
akostadinov: Do we need this comment and other similar ones? After the split we have separate test cases and nobody would care what the original one was. Or not?

akostadinov: otherwise lg, thanks

xiuwang: Right, nobody cares the original one. I will remove these comments

xiuwang: Only keep comment for the original case OCP-13914.

wzheng1:
LGTM, thanks!

akostadinov:
I believe commit message is enough for anynody that wants to understand the change. So I removed last comment and I'm merging. Let me know if comment is important to you.

Thank you!

PR #280, opened 2019-07-11
Add steps for 4.x registry

wzheng1:
Since OCP internal registry has huge change, so add two steps for it. @openshift/devexp-qe Please review, thanks!

### DIFF
--- a/features/step_definitions/registry.rb
+++ b/features/step_definitions/registry.rb
@@ -371,3 +371,22 @@ Given /^I attempt the registry route based on API url and store it in the#{OPT_S
   cb_name ||= :registry_route
   cb[cb_name] = api_hostname.gsub("api","registry")
 end
+
+# Generate default route
+Given /^I enable image-registry default route$/ do
+  ensure_admin_tagged
+  step 'I run the :patch admin command with:', table(%{
+      | resource      | configs.imageregistry.operator.openshift.io |
+      | resource_name | cluster                                     |
+      | p             | {"spec":{"defaultRoute":true}}              |
+      | type          | merge                                       |
+  })
+  step %Q/the step should succeed/
+end
+
+Given /^default image registry route is stored in the#{OPT_SYM} clipboard$/ do |cb_name| 
+  org_proj_name = project(generate: false).name rescue nil
+  cb_name ||= :registry_route
+  cb[cb_name] = route('default-route', service('default-route',project('openshift-image-registry'))).dns(by: admin)
+  project(org_proj_name) if org_proj_name
+end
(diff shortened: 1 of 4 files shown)


lib/rules/cli/3.3.yaml
```diff
@@ -1245,6 +1245,7 @@
   :cmd: oc adm prune images
   :options:
     :confirm: --confirm=<value>
+    :ignore_invalid_refs: --ignore-invalid-refs=<value>
```
xiuwang: if better to add in  4.x yaml?

features/step_definitions/registry.rb
```diff
@@ -371,3 +371,22 @@
...
+  step 'I run the :patch client command with:', table(%{
+      | resource      | configs.imageregistry.operator.openshift.io |
+      | resource_name | cluster                                     |
+      | p             | {"spec":{"defaultRoute":true}}              |
+      | type          | merge                                       |
+  })
+  step %Q/the step should succeed/
+end
+
+Given /^default (default-route) route is stored in the#{OPT_SYM} clipboard$/ do |route_name, cb_name|
```
wewang58: IMO, (default-route)  can  replace with #{QUOTED}

features/step_definitions/registry.rb
```diff
@@ -371,3 +371,22 @@
   cb_name ||= :registry_route
   cb[cb_name] = api_hostname.gsub("api","registry")
 end
+
+# Generate default route
+Given /^I enable image-registry default route$/ do
+  ensure_admin_tagged
```
akostadinov: is not this a configuration change that can easily interfere with other tests? In such case it needs to be `ensure_destructive_tagged` even if it will not break the cluster. Also registering clean-up to revert to original state must be part of the step.

features/step_definitions/registry.rb
akostadinov: it is presenty hardcoding route name to `default-route`. It is not clear from step name that it is getting route from registry namespace. I would suggest a step definition:
```
Given /^default image registry route is stored in the#{OPT_SYM} clipboard$/ do |cb_name|
```
then you can hardcode route name and project name inside the step body.

lib/rules/cli/3.3.yaml
wzheng1: This is an option since 3.x.

features/step_definitions/registry.rb
wzheng1: This step doesn't make any image-registry pod restart and won't influence ongoing tests. If the route is enabled at the first place, it doesn't matter this command executed again, usually, user needs this route enabled.

features/step_definitions/registry.rb
wzheng1: Good idea, I have changed, thanks!

PR #282, opened 2019-07-12
fix indentation and took out debug statement

pruan-rht:
fixes #281


akostadinov:
/lgtm

my bad, I noticed in original PR and forgot

PR #283, opened 2019-07-15
ocp 23543 - iptables binary and rules on sdn containers should be the same as host

anuragthehatter:
This was a test case created for SDN-341.
Logs: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/410/console
Not sure if :239 is right way to define sudo or if there is a step which can facilitate that. Pleas advise
Thanks.
@zhaozhanqi

### DIFF
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -229,7 +229,7 @@ Feature: SDN related networking scenarios
     Then the step should succeed
     And evaluation of `@result[:response]` is stored in the :iptables_version_host clipboard
     
-    When I run command on the  node's sdn pod:
+    When I run command on the node's sdn pod:
       | iptables-save --version |
     Then the step should succeed
     And evaluation of `@result[:response]` is stored in the :iptables_version_pod clipboard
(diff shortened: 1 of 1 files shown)


zhaozhanqi:
LGTM,thanks

akostadinov:
If this is a smoke test, then tag as `@smoke`. Also will iptables change if simultaneously running scenarios create projects, pods, etc? If so, then scenario needs to also be tagges as `@destructive`. I'll merge though and you can add tags according to answers of the above questions.

/lgtm

anuragthehatter:
> If this is a smoke test, then tag as `@smoke`. Also will iptables change if simultaneously running scenarios create projects, pods, etc? If so, then scenario needs to also be tagges as `@destructive`. I'll merge though and you can add tags according to answers of the above questions.
> 
> /lgtm

Thanks for checking @akostadinov. Came back from PTO today. So this case is just checking iptables version, not changing iptables or so. So not a destructive one and smoke as well.

PR #292, opened 2019-07-22
Fix OCP-12855

wewang58:
Now, in 4.x, no command: "oc expor", so fix the case. @openshift/devexp-qe please help to review it, thanks


wzheng1:
LGTM

PR #295, opened 2019-07-24
auto OCP-9767

lihongan:
This case is for CVP in 3.x release and just ensure no issue with ipfailover image and the pod can be `Running`.
@zhaozhanqi Please help review. Thanks.

log: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/42475/consoleFull

### DIFF
--- a/features/routing/ipfailover.feature
+++ b/features/routing/ipfailover.feature
@@ -1,0 +1,30 @@
+Feature: Testing ipfailover scenarios
+
+  # @author hongli@redhat.com
+  # @case_id OCP-9767
+  @admin
+  @destructive
+  Scenario: Configure a highly available network service
+    Given the cluster is running on OpenStack
+    And I switch to cluster admin pseudo user
+    And I use the "default" project
+    And default router image is stored into the :router_image clipboard
+    And SCC "privileged" is added to the "ipfailover" service account
+    And admin ensures "ipf-9767" dc is deleted after scenario
+
+    When I run the :oadm_ipfailover admin command with:
+      | name        | ipf-9767                                                              |
+      | images      | <%= cb.router_image.gsub("haproxy-router","keepalived-ipfailover") %> |
+      | virtual_ips | 192.168.1.1                                                           |
+      | watch_port  | 53                                                                    |
+    Then a pod becomes ready with labels:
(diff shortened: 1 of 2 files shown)


features/routing/ipfailover.feature
```diff
@@ -0,0 +1,29 @@
...
+
+  # @author hongli@redhat.com
+  # @case_id OCP-9767
+  @admin
+  Scenario: Configure a highly available network service
+    Given the cluster is running on OpenStack
+    And I switch to cluster admin pseudo user
+    And I use the "default" project
+    And default router image is stored into the :router_image clipboard
+    And SCC "privileged" is added to the "ipfailover" service account
```
akostadinov: why do we need this? Should users update this service account to use ipfailover? Can't things work OOB somehow?

I'm also concerned about other scenarios relying on this service account having default permissions to collide if executed in parallel.

lihongan: The ”ipfailover" is the default service account when creating IP Failover configuration, and I'm sure it is used by IP Failover feature only.
The SA need sufficient privileges to create ipfailover, or it will prompt user to grant. See below:
```
# oc adm ipfailover -h
<---snip--->
Options:
      --service-account='ipfailover': Name of the service account to use to run the ipfailover pod.
<---snip--->
# oc adm ipfailover ipfailover --virtual-ips="1.2.3.4,5.6.7.8"
error: ipfailover could not be created; service account "ipfailover" does not have sufficient privileges, grant access with: oc adm policy add-scc-to-user privileged -z ipfailover
```

zhaozhanqi:
LGTM

features/routing/ipfailover.feature
akostadinov: It is a global service account. If you change it, then another test case may want to use it at the same time and we need to make test case destructive.

As a user I'd expect default SA to have enough permissions. If dev team insists it should NOT have such by default, then perhaps we can create a special temporary service account and use it instead of the default one, to avoid collision between test scenarios.

WDYT?

lihongan: That make sense. I marked it as destructive. Thanks.

PR #296, opened 2019-07-25
Add more time for oauth restore step

stuartchuan:
Add more time for oauth restore step, in case of it not finish when next scenario started.

### DIFF
--- a/features/step_definitions/oauth.rb
+++ b/features/step_definitions/oauth.rb
@@ -14,5 +14,6 @@ Given /^the #{QUOTED} oauth CRD is restored after scenario$/ do |name|
     opts = {resource: 'oauth', resource_name: name, p: patch_json, type: 'merge' }
     @result = _admin.cli_exec(:patch, **opts)
     raise "Cannot restore OAuth: #{name}" unless @result[:success]
+    sleep 60
   }
 end


features/step_definitions/oauth.rb
```diff
@@ -14,5 +14,6 @@
     opts = {resource: 'oauth', resource_name: name, p: patch_json, type: 'merge' }
     @result = _admin.cli_exec(:patch, **opts)
     raise "Cannot restore OAuth: #{name}" unless @result[:success]
+    sleep 60
```
pruan-rht: @stuartchuan instead of a hard-coded wait, is there another way to verify `Oauth` is restored properly?

akostadinov: Can we instead of hardcode waiting, check when it becomes usable? That means repeat an operation that will succeed only after the CRD works properly? WDYT?

akostadinov: hah, didn't see Peter's comment

stuartchuan: Here is the idea come from, https://github.com/openshift/cluster-authentication-operator/blob/master/README.md, the last step. @akostadinov

akostadinov: In this case trying to login is even more crap than hardcoding wait :/
I guess we leave it as is for the time being.. unless anybody has some better idea.

PR #299, opened 2019-07-26
some fixes

yapei:
Pass log in [cucushift PR 7046](https://github.com/openshift/cucushift/pull/7046)

### DIFF
--- a/lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
+++ b/lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
@@ -110,12 +110,12 @@ click_secondary_menu:
 click_dropdown_field:
   elements:
   - selector:
-      xpath: //label[(.='<dropdown_field>')]/following-sibling::*//button[contains(@class, 'dropdown__toggle')]
+      xpath: //*[(.='<dropdown_field>')]/following-sibling::*//button[contains(@class, 'dropdown__toggle')]
     op: click
 click_one_item:
   elements:
   - selector:
-      xpath: //*[@id='<dropdown_item>-link']
+      xpath: //li//*[@id='<dropdown_item>-link']
     op: click
 choose_dropdown_item:
   action: click_dropdown_field
--- a/lib/rules/web/admin_console/4.2/environments.xyaml
+++ b/lib/rules/web/admin_console/4.2/environments.xyaml
@@ -19,16 +19,16 @@ add_env_var_name:
 add_env_var_pair_value:
   elements:
   - selector:
-      xpath: //div[@class='co-m-pane__body-group']//span[contains(normalize-space(text()),'Select a resource')]
+      xpath: (//input[@value='<env_var_name>']/parent::*/following-sibling::*[contains(@class,'pairs-list__value-pair-field')]//button)[1]
     op: click
   - selector:
-      xpath: //div[@class='co-m-pane__body-group']//span[contains(.,'<env_source_name>')]
+      xpath: //li//*[text()='<env_source_name>']
     op: click
   - selector:
-      xpath: //div[@class='co-m-pane__body-group']//span[contains(normalize-space(text()),'Select a key')]
+      xpath: (//input[@value='<env_var_name>']/parent::*/following-sibling::*[contains(@class,'pairs-list__value-pair-field')]//button)[2]
     op: click
   - selector:
-      xpath: //div[@class='co-m-pane__body-group']//a[contains(.,'<env_source_key>')]
+      xpath: //li//*[text()='<env_source_key>']
     op: click
 add_env_var_value:
   element:
@@ -38,7 +38,7 @@ add_env_var_value:
 choose_container:
   elements:
   - selector:
-      xpath: //div[@class='dropdown']//span[@class='co-resource-item']
+      xpath: //*[(.='Container:')]/following-sibling::*//button[contains(@class, 'dropdown__toggle')]
     op: click
   action: choose_item_from_list
 update_env_vars:
--- a/lib/rules/web/admin_console/4.2/filters.xyaml
+++ b/lib/rules/web/admin_console/4.2/filters.xyaml
@@ -1,7 +1,7 @@
 set_filter_strings:
   elements:
   - selector: &filter_text_ele
-      css: input.text-filter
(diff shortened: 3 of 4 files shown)


lib/rules/web/admin_console/4.2/secrets.xyaml
```diff
@@ -73,7 +73,7 @@ choose_authentication_type:
       xpath: //*[@class='co-create-secret__dropdown']
     op: click
   - selector:
-      xpath: //a[text()='<auth_type>']
```
yapei: changes targeted to OCP-20020

lib/rules/web/admin_console/4.2/filters.xyaml
```diff
@@ -1,7 +1,7 @@
 set_filter_strings:
   elements:
   - selector: &filter_text_ele
-      css: input.text-filter
```
yapei: Changes targeted for `Scenario: Check resource events`

lib/rules/web/admin_console/4.2/environments.xyaml
```diff
@@ -19,16 +19,16 @@ add_env_var_name:
 add_env_var_pair_value:
   elements:
   - selector:
-      xpath: //div[@class='co-m-pane__body-group']//span[contains(normalize-space(text()),'Select a resource')]
```
yapei: Changes targeted for `Scenario: Check environment editor for init container`

lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
```diff
@@ -110,12 +110,12 @@ click_secondary_menu:
 click_dropdown_field:
   elements:
   - selector:
-      xpath: //label[(.='<dropdown_field>')]/following-sibling::*//button[contains(@class, 'dropdown__toggle')]
```
yapei: common changes

yapei:
@yanpzhan  Please help review, thanks

lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
```diff
@@ -110,12 +110,12 @@ click_secondary_menu:
...
+      xpath: //*[(.='<dropdown_field>')]/following-sibling::*//button[contains(@class, 'dropdown__toggle')]
```
akostadinov: don't we know type of this element? If we don't, then fine but perhaps better to select the proper element type.

lib/rules/web/admin_console/4.2/environments.xyaml
```diff
@@ -38,7 +38,7 @@ add_env_var_value:
 choose_container:
   elements:
   - selector:
-      xpath: //div[@class='dropdown']//span[@class='co-resource-item']
+      xpath: //*[(.='Container:')]/following-sibling::*//button[contains(@class, 'dropdown__toggle')]
```
akostadinov: same here

lib/rules/web/admin_console/4.2/filters.xyaml
```diff
@@ -1,7 +1,7 @@
...
+      css: input[class*='text-filter']
```
akostadinov: could you explain how this selector differs from the previous one?

lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
yapei: Tried to remember why I changed this but with no luck : (, I just remembered that it seems some elements are not in `//label`

yapei: I remembered why I changed from `//label` to `//*`,as we can see from the following changes in `choose_container`, it use exactly the same xpath expression with `click_dropdown_field`, in that case the element 
```
<div class="co-toolbar__item">Container:</div>
```

lib/rules/web/admin_console/4.2/environments.xyaml
yapei: same as above, now the element is defined as `//div` but I'm afraid someday it will change to `//label` or something else

lib/rules/web/admin_console/4.2/filters.xyaml
yapei: `input.text-filter`: match input, has attribute class, and the one of the class value **is exactly equal** to text-filter
`input[class*='text-filter']`: match input, has attribute class, and class value **contains** text-filter
And here's how the element is defined:
```
<input autocapitalize="none" class="pf-c-form-control co-text-filter" placeholder="Filter Events by name or message..." tabindex="0" type="text" value="">
```

yapei: References: http://xahlee.info/js/css_selector_syntax.html

yanpzhan:
The fix looks good  for me.

lib/rules/web/admin_console/4.2/filters.xyaml
akostadinov: bump, this is most interesting one, lgtm otherwise

akostadinov: What I see that `*=` means that it will match a class of `something text-filter anotherthing` as well `whaever-text-filter-whatever`.

akostadinov: Why is this needed? What class we can't match with original version?

yapei: Comment added in https://github.com/openshift/verification-tests/pull/299#discussion_r308069274

yapei:
```
<input autocapitalize="none" class="pf-c-form-control co-text-filter" placeholder="Filter Events by name or message..." tabindex="0" type="text" value="">
```

yapei:
```
$$("input[class*='text-filter']")
Array [ input.pf-c-form-control.co-text-filter
 ]

$$('input.text-filter')
Array []
```

yapei: > What I see that `*=` means that it will match a class of `something text-filter anotherthing` as well `whaever-text-filter-whatever`.

Yes

PR #308, opened 2019-08-02
Samples operator finalizer

xiuwang:


lib/base_helper.rb
```diff
@@ -385,12 +386,17 @@ def find_follow(*paths)
         end
       end
 
+      def compare_time(t1, t2)
```
akostadinov: compare usually returns `0`, `1`, or `-1`. Also looking at the logic, a better name would be `compare_str_time`.
Instead of `1` or `-1` also just integer value should also be acceptable. You can `return tp2 - tp1` at the end.

features/step_definitions/is.rb
```diff
@@ -46,3 +46,27 @@
...
+
+  @result = admin.cli_exec(:get, resource: crd_type, resource_name: crd_name, o: 'yaml')
+  ts1 = @result[:parsed]['metadata']['creationTimestamp']
+  
+  success = wait_for(300) { 
+    @result = admin.cli_exec(:delete, object_type: crd_type, object_name_or_id: crd_name)
+  }
+  unless success
+    raise "The crd #{crd_type} can't be deleted."
+  end
```
akostadinov: If you want resource to be gone, you better use step `/^(I|admin) ensures? #{QUOTED} (\w+) is deleted(?: from the#{OPT_QUOTED} project)?( after scenario)?$/`. i.e. `admin ensures "#{crd_name}" #{crd_type} is deleted`

features/step_definitions/is.rb
```diff
@@ -46,3 +46,27 @@
...
+
+  @result = custom_resource_definition(crd_type).wait_to_appear(_user, 60)
+  @result = admin.cli_exec(:get, resource: crd_type, resource_name: crd_name, o: 'yaml')
+  ts2 = @result[:parsed]['metadata']['creationTimestamp']
+  
+  unless compare_time(ts1, ts2)
```
akostadinov: `unless compare_str_time(ts1, ts2) > 0`

features/step_definitions/is.rb
```diff
@@ -46,3 +46,27 @@
...
+    raise "The crd #{crd_type} doesn't be recreated. "
```
akostadinov: `was not recreated`

features/step_definitions/is.rb
```diff
@@ -46,3 +46,27 @@
     raise "ImageStreamTag #{istag_name} never created"
   end
 end
+
+Given /^the #{QUOTED} #{QUOTED} CRD was finalized$/ do |crd_type, crd_name|
```
akostadinov: what does it mean `finalized`? Could you submit scenario where this step is used so it is more obvious whether the step name is meaningful?

akostadinov: There are other possible simplifications but it is hard to suggest them now before usage of the step becomes clearer.

features/step_definitions/is.rb
```diff
@@ -46,3 +46,27 @@
...
+  ts1 = @result[:parsed]['metadata']['creationTimestamp']
+  
+  success = wait_for(300) { 
+    @result = admin.cli_exec(:delete, object_type: crd_type, object_name_or_id: crd_name)
+  }
+  unless success
+    raise "The crd #{crd_type} can't be deleted."
+  end
+
+  @result = custom_resource_definition(crd_type).wait_to_appear(_user, 60)
```
akostadinov: we removed resource name `cluster`, type `config.samples.operator.openshift.io`. Why would the whole CRD `config.samples.operator.openshift.io` disappear so we need to wait for it?

Did you intend to remove the whole CRD or only `cluster` resource of it?

akostadinov:
Looking at https://github.com/openshift/cucushift/pull/7064 it looks like when the resource `cluster` of type `config.samples.operator.openshift.io` is deleted, then it must become re-created.

I don't think that all CRD resources will get re-created when they are deleted. So such step is only limited to particular resources that are managed by operators and get re-created once they are manually deleted.

In my opinion it makes sense to make these deletions and stuff in scenario so it is more clear what is actulally being done.

features/step_definitions/is.rb
xiuwang: I want a scenario that the resource could be recreated after destroy.

features/step_definitions/is.rb
xiuwang: only `cluster` resource, this step is not correct, need update.

lib/base_helper.rb
xiuwang: Thanks, will update.

features/step_definitions/is.rb
xiuwang: My resource is `config.samples.operator.openshift.io`, the shotest name must be `config.samples`.
So I couldn't add it as  an  instance variable under lib/openshift dir. Do you have better idea to use the step here?

xiuwang:
@akostadinov here is my instance https://github.com/openshift/cucushift/pull/7064 .
I only want to make sure the `config.samples.operator cluster` is recreated, but all crds. that's step is not suitable there.

features/step_definitions/is.rb
akostadinov: maybe we can try using `_` in place of dot and see if rest of the logic works correctly

xiuwang: Not work...
 `oc delete config_samples_operator_openshift_io cluster --wait=false --config=/home/wxj/workdir/dhcp-140-124-wxj/ocp4_admin.kubeconfig
      
      STDERR:
      error: the server doesn't have a resource type "config_samples_operator_openshift_io" `

akostadinov: I thought that you mean you tried to add this as a resource in verification tests. Ok, let me try doing by myself and see how such resources can be handled.

xiuwang: Yeah, I tried to add a resources under lib/openshift, like below. Unlucky met the above error.
config_samples_operator.rb
`require 'openshift/cluster_resource'

module BushSlicer
  class ConfigSamplesOperator < ClusterResource
    RESOURCE = 'config_samples_operator_openshift_io'
  end 
end`

akostadinov: Please check if #329 does the job for you.

akostadinov: `admin ensures "cluster" config_samples_operator_openshift_io is deleted`

xiuwang:
@akostadinov Could you help review it again? Thanks

PR #312, opened 2019-08-06
Add step to record image-registry generation number

wzheng1:
@openshift/devexp-qe Please help to review, thanks!


features/step_definitions/registry.rb
```diff
@@ -385,8 +386,18 @@
...
 Given /^default image registry route is stored in the#{OPT_SYM} clipboard$/ do |cb_name| 
+  ensure_admin_tagged
+  ensure_destructive_tagged
   org_proj_name = project(generate: false).name rescue nil
   cb_name ||= :registry_route
   cb[cb_name] = route('default-route', service('default-route',project('openshift-image-registry'))).dns(by: admin)
   project(org_proj_name) if org_proj_name
 end
+
+Given /^current generation number of image-registry deployment is stored into#{OPT_SYM} clipboard$/ do |cb_name|
```
xiuwang: Could we make this more suitable for other resources?

wzheng1: Good suggestion!

wzheng1: Done, please check.

xiuwang:
LGTM, thanks

features/step_definitions/registry.rb
```diff
@@ -385,8 +386,18 @@
...
 Given /^default image registry route is stored in the#{OPT_SYM} clipboard$/ do |cb_name| 
+  ensure_admin_tagged
+  ensure_destructive_tagged
   org_proj_name = project(generate: false).name rescue nil
   cb_name ||= :registry_route
   cb[cb_name] = route('default-route', service('default-route',project('openshift-image-registry'))).dns(by: admin)
   project(org_proj_name) if org_proj_name
 end
+
+Given /^current generation number of#{OPT_QUOTED} deployment under #{QUOTED} is stored into#{OPT_SYM} clipboard$/ do |name,namespace,cb_name|
```
pruan-rht: @wzheng1, it looks like the name of the deployment is required?  If so then  `#{OPT_QUOTED}` should be `#{QUOTED}`.

pruan-rht: @wzheng1 perhaps we should add a method to `lib/openshift/deployment.rb` to support getting the `generation` field.

pruan-rht: @wzheng1 added helper with this PR https://github.com/openshift/verification-tests/pull/315

wzheng1: @pruan-rht Thanks for the pr, it simplifies my definition!

features/step_definitions/registry.rb
```diff
@@ -385,8 +386,16 @@
...
+  ensure_admin_tagged
+  ensure_destructive_tagged
   org_proj_name = project(generate: false).name rescue nil
   cb_name ||= :registry_route
   cb[cb_name] = route('default-route', service('default-route',project('openshift-image-registry'))).dns(by: admin)
   project(org_proj_name) if org_proj_name
 end
+
+Given /^current generation number of#{OPT_QUOTED} deployment is stored into#{OPT_SYM} clipboard$/ do |name, cb_name|
+  ensure_admin_tagged
```
akostadinov: getting `deployment` info should not require `admin` tag. Maybe for some deployments. But we will already be admins.

features/step_definitions/registry.rb
```diff
@@ -385,8 +386,16 @@
 end
 
 Given /^default image registry route is stored in the#{OPT_SYM} clipboard$/ do |cb_name| 
+  ensure_admin_tagged
+  ensure_destructive_tagged
```
akostadinov: I don't think this operation is destructive or can interfere with other operations. Can we remove the `ensure_destructive` call?

features/step_definitions/registry.rb
wzheng1: Yeah, you are correct. Tag is removed.

features/step_definitions/registry.rb
wzheng1: Yeah, you are correct. Tag is removed.

PR #313, opened 2019-08-06
Remove SystemGroup and SystemUser definition

stuartchuan:
SystemGroup and SystemUser not used in 4.x rolebinding and clusterrolebinding kind field, remove them.


lib/openshift/role_binding.rb
```diff
@@ -21,9 +21,7 @@ class RoleBinding < ProjectResource
     # - kind: ServiceAccount
     #   name: deployer
     #   namespace: user1-test-prj
-    # - kind: SystemUser
     #   name: system:admin
-    # - kind: SystemGroup
```
akostadinov: I would suggest to change `SystemGroup/User` to `Group/User` instead of removing the lines. Otherwise the comment becomes confusing.

akostadinov:
> SystemGroup and SystemUser not used in 4.x rolebinding and clusterrolebinding kind field, remove them.

Are they not used in rolebindings or not used at all anymore? Would be LGTM in case not used at all anymore.

stuartchuan:
Yes, they are not used in rolebindings also in 4.x.

PR #318, opened 2019-08-07
Add  step to update operator operand managementstate

wewang58:
@openshift/devexp-qe pls help to review it, thx

### DIFF
--- a/features/step_definitions/operator.rb
+++ b/features/step_definitions/operator.rb
@@ -1,0 +1,11 @@
+Given /^Admin updated the operator crd "([^"]*)" managementstate operand to (Managed|Removed|Unmanaged)$/ do |cluster_operator, manage_type|
+  ensure_admin_tagged
+  ensure_destructive_tagged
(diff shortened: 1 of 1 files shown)


features/step_definitions/operator.rb
```diff
@@ -0,0 +1,11 @@
+Given /^I updated the operator crd "([^"]*)" managementstate operand to (Managed|Removed|Unmanaged)$/ do |cluster_operator, manage_type|
```
xiuwang: s/I/admin ?

features/step_definitions/operator.rb
```diff
@@ -0,0 +1,11 @@
...
+  ensure_admin_tagged
+  ensure_destructive_tagged
+  step %Q/I run the :patch client command with:/, table(%{
```
xiuwang: Should not be 'admin command'?

wewang58: Yes, should admin, updated

xiuwang:
LGTM

PR #319, opened 2019-08-07
auto OCP-21772 check user starter guide on projects

yapei:
@yanpzhan  @SSshahan   Please help review, thanks

### DIFF
--- a/lib/rules/web/admin_console/4.2/status.xyaml
+++ b/lib/rules/web/admin_console/4.2/status.xyaml
@@ -27,3 +27,13 @@ check_dim_resource_list:
       xpath: //div[@class='co-disabled']//button[contains(.,'Create')]
   - selector:
       xpath: //div[@class='co-disabled']//div[contains(.,'No <resource_singular>s Found')]
+check_user_starter_guide_message:
+  elements:
+  - selector:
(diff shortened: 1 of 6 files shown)


lib/rules/web/admin_console/4.2/status.xyaml
```diff
@@ -27,3 +27,13 @@ check_dim_resource_list:
...
+check_user_starter_guide_message:
+  elements:
+  - selector:
+      visible_text: !ruby/regexp /Welcome to OpenShift/i
+  - selector:
+      visible_text: !ruby/regexp /Create a project for your application/i
+check_get_started_message_when_no_resources:
+  elements:
+  - selector:
+      visible_text: !ruby/regexp /No Workloads Found/i
```
SSshahan: Curiously, why we didn't use the existing `check_page_contains` rule?

yapei: For current 4.2 implementation, only `No Workloads Found` message and can be replaced with 
```
check_get_started_message_when_no_resources:
  params:
    content: No Workloads Found
  action: check_page_match
```

yapei: Is this what you mean?

SSshahan:
Thanks for updates. LGTM

PR #320, opened 2019-08-07
WIP admin console must read browser_opts.yml

akostadinov:
@yapei, please see if this will get you browser.yml respected. Order in which rules are specified is important to find the additional opts.


yapei:
Thanks @akostadinov  I confirmed that with this code change, `browser_opts.yml` can be loaded. I met other errors http://pastebin.test.redhat.com/787040

yapei:
@akostadinov  Can it be merged?

PR #323, opened 2019-08-12
Fix rules for OCP-19669: landing page and empty resources detail page

xiaocwan:
Fix between 4.2 and earlier:
1) Use the rules `goto_landing_page_for_empty_project` and `check_message_and_doc_for_new_user` to check the first page when normal user login.
2) On each resource detail page, 4.2 has removed Create button

Fix rules: https://github.com/openshift/cucushift/pull/7075
@yapei Pls review, thanks!

### DIFF
--- a/lib/rules/web/admin_console/3.11/goto.xyaml
+++ b/lib/rules/web/admin_console/3.11/goto.xyaml
@@ -5,8 +5,6 @@ goto_project_status:
   url: status/ns/<project>
 goto_resource_page_under_default_project:
   url: k8s/ns/default/<resource_url_name>
-goto_landing_page_for_empty_project:
-  action: goto_project_status
 
 # builds
 goto_one_buildconfig_page:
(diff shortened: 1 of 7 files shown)


lib/rules/web/admin_console/3.11/goto.xyaml
```diff
@@ -5,6 +5,8 @@ goto_project_status:
   url: status/ns/<project>
 goto_resource_page_under_default_project:
   url: k8s/ns/default/<resource_url_name>
+goto_landing_page_for_empty_project:
```
yapei: Can you explain why we need action `goto_landing_page_for_empty_project`? I assume that after user is logged in, he is expected to redirected to the correct landing page by default already, why do we need this action again?

xiaocwan: The landing page is different between 4.1 and 4.2, that's why automation failed.

xiaocwan: @yapei Dropped, let landing by logging.

yapei:
LGTM, thanks for update

PR #325, opened 2019-08-14
Fix scenario dynamic provisioning

liangxia:

### DIFF
--- a/features/storage/dynamic_provisioning.feature
+++ b/features/storage/dynamic_provisioning.feature
@@ -3,32 +3,19 @@ Feature: Dynamic provisioning
   # @author lxia@redhat.com
   @admin
   Scenario Outline: dynamic provisioning
-    Given a 5 characters random string of type :dns is stored into the :proj_name clipboard
-    When I run the :oadm_new_project admin command with:
-      | project_name  | <%= cb.proj_name %>         |
-      | node_selector | <%= cb.proj_name %>=dynamic |
-      | admin         | <%= user.name %>            |
-    Then the step should succeed
-
-    Given I store the ready and schedulable nodes in the :nodes clipboard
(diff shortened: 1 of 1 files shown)


liangxia:
Adding @jhou1 @qinpingli to review the PR as @chao007 said she have no time (busy on CSI stuff) to do the review.

qinpingli:
@liangxia Could you add some comments for why do this change?
Seems in the OCP4.x the sc.volumeBindingMode is not "Immediate", it's just for 4.2 or both for 4.1?
Can we try to get the sc.volumeBindingMode and update the PVC binding mode with the sc.volumeBindingMode?

liangxia:
> @liangxia Could you add some comments for why do this change?
> Seems in the OCP4.x the sc.volumeBindingMode is not "Immediate", it's just for 4.2 or both for 4.1?
> Can we try to get the sc.volumeBindingMode and update the PVC binding mode with the sc.volumeBindingMode?

@qinpingli 
The volumeBindingMode in storage class varies from platform to platform, both in 4.1 and 4.2.
To simplify the automation, we just postpone PVC status check (Check PVC status after pod is ready), so it works both for volumeBindingMode Immediate and WaitForFirstConsumer.

Also, we can get the node info where the pod is scheduled by `pod.node_name`, so store node info and labelling nodes is not need now.

Also a minor change from mypvc1/mypod1 to mypvc/mypod since we only create one pvc/pod.

qinpingli:
@liangxia Ok, do we have cases about the volumeBindingMode? If we have, then the change LGTM.

jhou1:
LGTM

liangxia:
> @liangxia Ok, do we have cases about the volumeBindingMode? If we have, then the change LGTM.

Yes, they are covered in cucushift/blob/master/storage/storage_class_options.feature

PR #326, opened 2019-08-15
added support for machineautoscaler

pruan-rht:


pruan-rht:
@akostadinov PTAL, thanks!

lib/openshift/node.rb
```diff
@@ -73,6 +73,12 @@ def schedulable?(user: nil, cached: true, quiet: false)
       return !spec['unschedulable']
     end
 
+    def is_worker?(user: nil, cached: true, quiet: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
+      current_config = rr.dig('metadata', 'annotations', 'machineconfiguration.openshift.io/currentConfig')
+      current_config.start_with? 'rendered-worker'
```
akostadinov: is this role documented? I wonder how reliable such check is.

features/step_definitions/machine_set.rb
```diff
@@ -0,0 +1,9 @@
+# machineset supporting steps
+#
+Given /^I store all machinesets to the#{OPT_SYM} clipboard$/ do |cb_name|
+  cb_name ||= :machinesets
+  @result = admin.cli_exec(:get, resource: 'machineset', n: 'openshift-machine-api', o: 'yaml')
+  cb[cb_name] = @result[:parsed]["items"].map { |i| i['metadata']['name']}
```
akostadinov: Can we store machineset objects here instead of string names? `MachineSet.list(...)`?

features/step_definitions/helper_services.rb
```diff
@@ -714,3 +714,33 @@
...
 end
+
+# example of machine auto-scaler https://github.com/openshift-qe/output_references/blob/master/autoscale/machine-autoscaler.yaml
+Given /^I enable autoscaling for my cluster$/ do
+  name_place_holder = ""
+  base_template_hash = {
+    "apiVersion" => "autoscaling.openshift.io/v1beta1",
+    "kind" => "MachineAutoscaler",
+    "metadata" => {
+      "name" => name_place_holder,
```
akostadinov: we don't need a separate variable for a placeholder. You could skip setting that value and set it afterwards. Or set to hardcoded value `nil` or empty string `""`. Any special need to use a variable?

lib/openshift/node.rb
pruan-rht: yeah, I debated about that...an alternative is to check for the existence of the label `node-role.kubernetes.io/master=` for a node that's a `master` or  `node-role.kubernetes.io/worker=`.  That's probably more reliable.

akostadinov: +1 to use `node-role`

features/step_definitions/helper_services.rb
pruan-rht: the name has to match the node

akostadinov: I'm not arguing against that. Just line 742 will still work if you remove lines 720 and 725. Or am I missing something?

pruan-rht: @akostadinov ah ok, I was not clear on what you mean...yeah, I can do that...

features/step_definitions/machine_set.rb
pruan-rht: Sure, but that would mean we have to save project context and restore it back

pruan-rht:
@akostadinov I have address all of your comments, PTAL again, thanks!

lib/openshift/node.rb
```diff
@@ -75,8 +75,8 @@ def schedulable?(user: nil, cached: true, quiet: false)
 
     def is_worker?(user: nil, cached: true, quiet: false)
       rr = raw_resource(user: user, cached: cached, quiet: quiet)
-      current_config = rr.dig('metadata', 'annotations', 'machineconfiguration.openshift.io/currentConfig')
-      current_config.start_with? 'rendered-worker'
+      res = rr.dig('metadata', 'labels', 'node-role.kubernetes.io/worker')
+      res.nil? ? false : true
```
akostadinov: `! res.nil?`

akostadinov: `return ! res.nil?`

or

`return !!res`

pruan-rht: @akostadinov updated... PTAL.  thx!

features/step_definitions/helper_services.rb
```diff
@@ -714,3 +714,34 @@
   step 'the step should succeed'
   step 'the pod named "cluster-capacity" becomes ready'
 end
+# https://docs.openshift.com/container-platform/4.1/machine_management/applying-autoscaling.html
+# two steps, 1. create clusterautoscaler, follow by machineautoscaler
+# example of machine auto-scaler https://github.com/openshift-qe/output_references/blob/master/autoscale/machine-autoscaler.yaml
+Given /^I enable autoscaling for my cluster$/ do
```
akostadinov: Sorry, I missed this earlier.

Is this a destructive step? It needs to check for the destructive tag.
Also I don't see clean-up logic.

pruan-rht: @akostadinov it's more of an enhancement step to ensure that it will automatically spin up more resources as needed.  I didn't put any clean up because it's meant to be permanent for the life-time of the cluster...run once and forget about it kind of deal.

akostadinov: Then it needs to be inside `features/step_definitions/cluster_configuration/` directory. That directory we use for such kind of steps.

pruan-rht: @akostadinov ok, updated

PR #327, opened 2019-08-15
Default storage class can not be deleted in 4.x

liangxia:
The step `default storage class is deleted` will recover the storage class in after hooks. As default storage class will be automatically recovered by cluster-storage-operator, the after hook will fail, which will cause auto execution quit prematurely.

@chao007 @qinpingli @jhou1 @akostadinov @pruan-rht

### DIFF
--- a/features/step_definitions/cluster_configuration/csi.rb
+++ b/features/step_definitions/cluster_configuration/csi.rb
@@ -118,7 +118,7 @@ Given /^I create (default )?storage class for #{QUOTED} csi driver$/ do |default
   filepath = @result[:abs_path]
 
   if default_sc
-    step %Q/default storage class is deleted/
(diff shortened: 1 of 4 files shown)


jhou1:
@liangxia Will the operator recreate a default StorageClass after we patch the default one to non-default? If that is not happening, then it will not disrupt the test, and LGTM.

liangxia:
> @liangxia Will the operator recreate a default StorageClass after we patch the default one to non-default? If that is not happening, then it will not disrupt the test, and LGTM.

@jhou1 The created storage class can be made non-default by editing its annotation but cannot be deleted so long as the operator runs. From https://github.com/openshift/cluster-storage-operator/#cluster-storage-operator

jhou1:
LGTM

PR #329, opened 2019-08-18
support resources with dots in their Kind field

akostadinov:
@xiuwang, please test whether this will help #308

you can refer to the name "cluster" of kind "config.samples.operator.openshift.io" by `config_samples_operator_openshift_io("cluster")`. And let me know!

@pruan-rht, notice file URL change here

### DIFF
--- a/lib/admin_credentials.rb
+++ b/lib/admin_credentials.rb
@@ -68,12 +68,12 @@ module BushSlicer
       end
     end
   end
- 
+
   class URLKubeconfigCredentials < AdminCredentials
     def get
       # export OPENSHIFT_ENV_OCP4_ADMIN_CREDS_SPEC=file:///~/.kube/config
       if opts[:spec].start_with? 'file:///'
-        path = File.expand_path(opts[:spec].split("file:///")[1])
+        path = File.expand_path(opts[:spec].split("file://")[1])
         raise "kubeconfig does not exists" unless File.exists? File.expand_path(path)
         config = File.open(path).read
       else
@@ -84,7 +84,7 @@ module BushSlicer
       return accessor_from_kubeconfig(config)
     end
   end
-  
+
   class AutoKubeconfigCredentials < AdminCredentials
     def get
       case opts[:spec]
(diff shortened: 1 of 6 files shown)


xiuwang:
Alex, Thanks.
I will test this pr.

lib/admin_credentials.rb
```diff
@@ -68,12 +68,12 @@ def get
...
     end
   end
- 
+
   class URLKubeconfigCredentials < AdminCredentials
     def get
       # export OPENSHIFT_ENV_OCP4_ADMIN_CREDS_SPEC=file:///~/.kube/config
       if opts[:spec].start_with? 'file:///'
-        path = File.expand_path(opts[:spec].split("file:///")[1])
+        path = File.expand_path(opts[:spec].split("file://")[1])
```
pruan-rht: @akostadinov any reason to change the local `file:`` directive?

pruan-rht: https://en.wikipedia.org/wiki/File_URI_scheme

akostadinov: Yes. According to that page, if you put a URL `file:///tmp/test.kubeconfig` then you expect an absolute PATH `/tmp/test.kubeconfig`. Without the change it is treated as the relative `tmp/test.kubeconfig` PATH.

If we want to allow relative paths, we need to also update line 75 to make it also allow `file://` (only to slashes) although this is not allowed by specification. WDYT?

pruan-rht: @akostadinov I think we should stick with a scheme that's valid specification. I don't think relative paths are that big of a deal.

akostadinov: I thought you needed them, that's why I suggested we can support such. In any case my change I believe is according to spec.

pruan-rht: @akostadinov ok, I have no objections then.

pruan-rht: LGTM, once @xiuwang has verified this PR works, then we can merge it.

xiuwang: @akostadinov @pruan-rht Sorry for laying down the steps, the PR works, you can merge it now
Thanks～

PR #330, opened 2019-08-20
add step to support extracting IDP creds

pruan-rht:


features/step_definitions/oauth.rb
```diff
@@ -17,3 +17,11 @@
     sleep 60
   }
 end
+
+# extract the secret given a htpasswd spec name
+Given /^the secret for #{QUOTED} htpasswd is stored in the#{OPT_SYM} clipboard$/ do |name, cb_name|
+  cb_name ||= :htpasswd_secret
+  generated_htpasswd_name = o_auth('cluster').generated_htpasswd_name(name: name)
+  step %Q/I use the "openshift-config" project/
```
akostadinov: Do we need to change current project? Can we avoid this?

lib/openshift/o_auth.rb
```diff
@@ -0,0 +1,28 @@
+module BushSlicer
+  class OAuth < ClusterResource
+    RESOURCE = "oauth"
```
akostadinov: shouldn't this be `oauths`?

lib/openshift/o_auth.rb
```diff
@@ -0,0 +1,28 @@
...
+    # pls refer https://github.com/openshift-qe/output_refrences/tree/master/oauth for output example
+
+    # return a list of Hash with the user defined name mapping to the system generated name
```
akostadinov: try to follow yardoc spec with `@return`. Can we return just a `Hash` where key is identity provider name and value is secret name?

akostadinov: then you will not need `generated_htpasswd_name` method. You can just do `htpasswds(...)["idp_name"]` to get secret name (if provider exists).

features/step_definitions/oauth.rb
pruan-rht: looks like we can, I'll take that line out

lib/openshift/o_auth.rb
pruan-rht: ok, I can change that

lib/openshift/o_auth.rb
pruan-rht: @akostadinov the use case from @yapei is that we need to find the generated name first before calling the `secret` object

akostadinov: I mean instead of current example return value to return:
```
{
  "flexy-htpasswd-provider"=>"htpass-secret",
  "htpasswd"=>"htpasswd-2nhtj",
  "htpasswd-foo"=>"htpasswd-rtc5q",
  "htpasswd-bar"=>"htpasswd-8zlqs",
  "foobar"=>"htpasswd-txc9w"
}
```
I see no benefit in returning an array of hashes regardless of the use case. Or can you give an example where the other data structure provides benefits?

pruan-rht: @akostadinov, thanks for the clarification.  I've updated the PR.  PTAL.  Thanks!

lib/openshift/o_auth.rb
```diff
@@ -1,27 +1,23 @@
...
-
-    # return the generated name given a user provided name
-    def generated_htpasswd_name(name: nil, user: nil, cached: true, quiet: false)
-      htpasswds = self.htpasswds(user: user, cached: cached, quiet: quiet)
-
-      res = htpasswds.select { |k| k.values.first  if k.keys.first == name}
-      raise "No matching htpasswd name '#{name}' found!" if res.count == 0
-      generated_name = res.first[name]
+      # covert array of hashes to just Hash
+      passwds.inject(:merge)
```
akostadinov: nice!

PR #331, opened 2019-08-22
Fixes for verification_tests_ci_entrypoint script

mgahagan73:
This PR fixes a couple of issues with tools/openshift-ci/verification_tests_ci_entrypoint.sh. Tested on a container built from the dockerfile under openshift-ci/

### DIFF
--- a/tools/common.sh
+++ b/tools/common.sh
@@ -11,6 +11,11 @@ function setup_git()
         git config --global user.email "$USER@redhat.com"
 }
 
+function is_ruby_proper_version()
+{
+    ruby -e 'exit Gem::Version.new("2.3") <= Gem::Version.new(RUBY_VERSION)'
+}
+
 function install_rvm_if_ruby_is_outdated()
 {
     if ! ruby -e 'exit Gem::Version.new("2.3") <= Gem::Version.new(RUBY_VERSION)' ; then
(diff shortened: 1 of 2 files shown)


tools/common.sh
```diff
@@ -11,6 +11,11 @@ function setup_git()
         git config --global user.email "$USER@redhat.com"
 }
 
+function is_ruby_proper_version()
+{
+    ruby -e 'exit Gem::Version.new("2.3") <= Gem::Version.new(RUBY_VERSION)'
+}
+
```
akostadinov: I think line 21 and the RVM ruby version also need updating like in b980a4c5.

tools/openshift-ci/verification_tests_ci_entrypoint.sh
```diff
@@ -17,4 +17,4 @@ elif ! is_ruby_proper_version; then
   echo "Ruby not proper version, see tool/common.sh#is_ruby_proper_version"
 fi
 
-"$TOOLS_HOME"/case_id_splitter.rb file-line $cases | xargs -d '\n' $cucumber_command
+/usr/bin/scl enable rh-git29 rh-ror50 -- "$TOOLS_HOME"/case_id_splitter.rb file-line $cases | xargs -d '\n' $cucumber_command
```
akostadinov: I prefer to have this handles in the `if` condition line 14 so that we can support RHEL8 executors too.

akostadinov:
Thank you for the update! I think I see what issues you hit and will update accordingly.

akostadinov:
@mgahagan73 , could you check if #333 works for you?

PR #340, opened 2019-08-29
modify steps to deploy logging operators

QiaolingTang:
@anpingli @pruan-rht PTAL.
In this PR, I added some scripts to parse different log collectors, and I changed the files we use to create subscriptions to make it could be run in OCP 4.2 cluster.

Here is the log: http://pastebin.test.redhat.com/792801

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -38,12 +38,12 @@ Given /^logging operators are installed with:$/ do | table |
   # Create namespace
   unless project('openshift-logging').exists?
     # oc create -f https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/OCP-21311/namespace.yaml
-    namespace_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/4.2/01_clo_ns.yaml"
+    namespace_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/01_clo_ns.yaml"
     @result = admin.cli_exec(:create, f: namespace_yaml)
     raise "Error creating namespace" unless @result[:success]
   end
   unless project('openshift-operators-redhat').exists?
-    eso_namespace_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/4.2/01_eo-project.yaml"
+    eso_namespace_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/01_eo-project.yaml"
     @result = admin.cli_exec(:create, f: eso_namespace_yaml)
     raise "Error creating namespace" unless @result[:success]
   end
@@ -57,17 +57,17 @@ Given /^logging operators are installed with:$/ do | table |
 
   # project('openshift-logging')
   # Create operator group in `openshift-logging` namespace
-  opts[:clo_operator_group_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/4.2/02_clo_og.yaml"
+  opts[:clo_operator_group_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/02_clo_og.yaml"
   @result = admin.cli_exec(:create, f: opts[:clo_operator_group_yaml])
   raise "Error creating operatorgroup" unless @result[:success]
 
   # Create operator group in `openshift-operators-redhat` namespace
-  opts[:eo_operator_group_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/4.2/02_eo-og.yaml"
+  opts[:eo_operator_group_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/02_eo-og.yaml"
   @result = admin.cli_exec(:create, f: opts[:eo_operator_group_yaml])
   raise "Error creating operatorgroup" unless @result[:success]
 
   # create RBAC object in `openshift-operators-redhat` namespace
-  opts[:operator_group_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/4.2/03_eo-rbac.yaml"
+  opts[:operator_group_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/03_eo-rbac.yaml"
   @result = admin.cli_exec(:create, f: opts[:operator_group_yaml])
   raise "Error creating operatorgroup" unless @result[:success]
 
@@ -75,25 +75,24 @@ Given /^logging operators are installed with:$/ do | table |
   if cb.ocp_cluster_version.include? "4.1."
     # create catalogsourceconfig for cluster-logging-operator and elasticsearch-operator
     opts[:catsrc_logging_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/4.1/03_clo_csc.yaml"
-    opts[:catsrc_elasticsearch_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/4.1/03_eo-csc.yaml"
+    opts[:catsrc_elasticsearch_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/4.1/04_eo-csc.yaml"
     @result = admin.cli_exec(:create, f: opts[:catsrc_logging_yaml])
     raise "Error creating catalogsourceconfig for cluster_logging" unless @result[:success]
     @result = admin.cli_exec(:create, f: opts[:catsrc_elasticsearch_yaml])
     raise "Error creating catalogsourceconfig for elasticsearch" unless @result[:success]
+
     # create subscription in `openshift-logging` and "openshift-operators-redhat" namespace:
     opts[:sub_logging_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/4.1/04_clo_sub.yaml"
     opts[:sub_elasticsearch_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/4.1/05_eo-sub.yaml"
-
     @result = admin.cli_exec(:create, f: opts[:sub_logging_yaml])
     raise "Error creating subscription for cluster_logging" unless @result[:success]
     @result = admin.cli_exec(:create, f: opts[:sub_elasticsearch_yaml])
     raise "Error creating subscription for elasticsearch" unless @result[:success]
 
-  elsif cb.ocp_cluster_version.include? "4.2."
+  else
     # create subscription in `openshift-logging` and "openshift-operators-redhat" namespace:
     opts[:sub_logging_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/4.2/03_clo_sub.yaml"
     opts[:sub_elasticsearch_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/4.2/04_eo-sub.yaml"
-
     @result = admin.cli_exec(:create, f: opts[:sub_logging_yaml])
     raise "Error creating subscription for cluster_logging" unless @result[:success]
     @result = admin.cli_exec(:create, f: opts[:sub_elasticsearch_yaml])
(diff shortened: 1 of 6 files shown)


features/step_definitions/logging.rb
```diff
@@ -34,12 +35,12 @@
...
   unless project('openshift-logging').exists?
     # oc create -f https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/OCP-21311/namespace.yaml
-    namespace_yaml = "https://raw.githubusercontent.com/openshift/origin-aggregated-logging/master/openshift/ci-operator/build-image/openshift-logging-namespace.yaml"
+    namespace_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/4.2/01_clo_ns.yaml"
     @result = admin.cli_exec(:create, f: namespace_yaml)
     raise "Error creating namespace" unless @result[:success]
   end
   unless project('openshift-operators-redhat').exists?
-    eso_namespace_yaml = "https://raw.githubusercontent.com/openshift/origin-aggregated-logging/master/openshift/ci-operator/build-image/openshift-operators-redhat-namespace.yaml"
+    eso_namespace_yaml = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/4.2/01_eo-project.yaml"
```
anpingli: @QiaolingTang  @pruan-rht  as 4.1 and 4.2 have different resource files. I am thinking how to use a common step to deploy the operators.
@pruan-rht  Could we get the clusterversion in cucushift?

features/step_definitions/logging.rb
```diff
@@ -53,41 +54,51 @@
...
-  # create catalogsourceconfig for logging and elasticsearch
-  opts[:catsrc_logging_yaml] ||= "https://raw.githubusercontent.com/openshift/origin-aggregated-logging/master/openshift/ci-operator/build-image/cluster-logging-catalogsourceconfig.yaml"
-  opts[:catsrc_elasticsearch_yaml] ||= "https://raw.githubusercontent.com/openshift/origin-aggregated-logging/master/openshift/ci-operator/build-image/elasticsearch-catalogsourceconfig.yaml"
-
+  # Create operator group in `openshift-operators-redhat` namespace
+  opts[:eo_operator_group_yaml] ||= "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eleasticsearch/deploy_via_olm/4.2/02_eo-og.yaml"
+  @result = admin.cli_exec(:create, f: opts[:eo_operator_group_yaml])
+  raise "Error creating operatorgroup" unless @result[:success]
 
-  @result = admin.cli_exec(:create, f: opts[:catsrc_logging_yaml])
```
anpingli: @QiaolingTang  it is not right to remove catsrc. we still neet to support 4.1. I am think if we can pull all resources in one yaml file.

features/step_definitions/logging.rb
pruan-rht: @anpingli to get the clusterverison, the user has to be a cluster-admin

anpingli: @pruan-rht  Is there a global variable to return clusterversion.

pruan-rht: Dev is aware of the issue and promised to fix it so a normal user can get that informaiton. 
 Currently, if the user is a clusteradmin, then make the call `cluster_version('version').verison` would do return that information.

pruan-rht: Since the installation require `admin` privilege, we can use the call above to decide which yaml to use

lib/openshift/cluster_logging.rb
```diff
@@ -53,7 +76,7 @@ class ClusterLogging < ClusterResource
     end
 
     private def es_cluster_health(user: nil, cached: true, quiet: false)
-      es_status_raw(user: user, cached: cached, quiet: quiet).first['clusterHealth']
+      es_status_raw(user: user, cached: cached, quiet: quiet).first['cluster']['status']
```
pruan-rht: @QiaolingTang this change seem to work for 4.2 but for version <= 4.1?

QiaolingTang: @pruan-rht  It works for 4.2, and I don't think it will work for 4.1.

features/step_definitions/logging.rb
QiaolingTang: Thanks Anping and Peter, I updated my PR.

anpingli:
@pruan-rht  Have you any idea about deploy operator using webconsole? Are there some operatorhub api in cucushift?

lib/openshift/cluster_logging.rb
akostadinov: Ok, instead of hardcoding to `['cluster']['status']`, you can check whether `['clusterHealth']` exist and use `['cluster']['status']` when needed.

lib/openshift/cluster_logging.rb
```diff
@@ -69,11 +93,22 @@ class ClusterLogging < ClusterResource
...
+
     def fluentd_ready?(user: nil, cached: true, quiet: false)
       fluentd_nodes = fluentd_status_raw(user: user, cached: cached, quiet: quiet)['nodes'].keys.sort
       fluentd_nodes == fluentd_ready_pods && fluentd_failed_pods.count == 0 && fluentd_notready_pods.count == 0
     end
 
+    def wait_until_rsyslog_is_ready(user: nil, quiet: false, timeout: 5*60)
+      wait_for(timeout) {
+        rsyslog_ready?(user: user, cached: false, quiet: quiet)
+      }
```
akostadinov: should this fail when timeout is reached. Presently it will just exit.

lib/openshift/cluster_logging.rb
```diff
@@ -23,15 +23,38 @@ class ClusterLogging < ClusterResource
...
+    private def rsyslog_ready_pods(user: nil, cached: true, quiet: false)
+      rsyslog_pods(user: user, cached: cached, quiet: quiet)['ready']
+    end
+
+    private def rsyslog_failed_pods(user: nil, cached: true, quiet: false)
+      rsyslog_pods(user: user, cached: cached, quiet: quiet)['failed']
+    end
+
+    private def rsyslog_notready_pods(user: nil, cached: true, quiet: false)
+      rsyslog_pods(user: user, cached: cached, quiet: quiet)['notReady']
```
akostadinov: for the methods above, better create `Pod` objects and return them instead of returning raw hash. Given methods are `privete` though, we can keep as is also.

lib/openshift/cluster_logging.rb
QiaolingTang: I'll add steps to raise error in logging.rb when timeout is reached.

lib/openshift/cluster_logging.rb
QiaolingTang: Thanks @akostadinov , I updated this file per your advice, PTAL.

lib/openshift/cluster_logging.rb
QiaolingTang: @akostadinov I added a statement to raise error when timeout is reached in logging.rb. And the message would be:
 ```  Failed to wait for rsyslog to be ready (RuntimeError)
      /home/qitang/github/verification-tests/features/step_definitions/logging.rb:135:in `/^I wait for clusterlogging(?: named "(.+)")? with "(.+?)" log collector to be functional in the(?: "(.+?)")? project$/'
      /home/qitang/github/verification-tests/features/step_definitions/logging.rb:255:in `/^I create clusterlogging instance with:$/'
      features/logging/test.feature:28:in `Given I create clusterlogging instance with:'
```

**Comment by QiaolingTang** [comment]
@akostadinov @pruan-rht  I have a question: for logging test, we have to deploy operators first, but we don't need to deploy operators in every scenario, that means once the operators deployed successfully, we don't need to delete or redeploy them before the job finished. When we trigger auto job, how could I ensure that the steps or scenario to deploy operators could be executed at the beginning of the job ? Do we have something like `before scenario` or `before suit` in verification-tests?

File: lib/openshift/cluster_logging.rb (hunk 4 above)
Comment (akostadinov) [reply lib/openshift/cluster_logging.rb]: sorry, can't see these in here

**Comment by akostadinov** [comment]
If you need cluster with particular configuration to run tests, then preferably you install after cluster is created. In the past we discussed whether we need logging with single or multiple configurations, initially we thought that we can go with single. But then it turned out dofferent tests needed different logging config.

Why is it now different that we thing same operators can serve all tests?

I would suggest to use step `logging service has been installed successfully` to make sure logging operators are installed. And make any test destructive that does not use the default logging config. How you restore the default logging config - I have no idea.

Did I get your question or am I missing the point?

**Comment by akostadinov** [comment]
I think I get the idea behind operators. Maybe I didn't understand your initial question. Do you want to install particular operator configuration for a number of tests to save time? In such case, you can have before/after for a whole feature file. Please let me know if this is what you want and I will provide more details.

Also please check whether my other 2 comments.

**Comment by QiaolingTang** [comment]
Yes, I want to ensure my 2 operators are installed successfully before all the test scenarios running. I will try to add `before` hooks in my feature files. Thanks.

File: lib/openshift/cluster_logging.rb (hunk 4 above)
Comment (QiaolingTang) [reply lib/openshift/cluster_logging.rb]: Please look at: https://github.com/openshift/verification-tests/pull/340/files#diff-0f0202f69f619cc471255d933a81861cR139-R140 and https://github.com/openshift/verification-tests/pull/340/files#diff-0f0202f69f619cc471255d933a81861cR145-R146

**Comment by akostadinov** [comment]
@QiaolingTang, try this one: https://stackoverflow.com/a/31120238/520567

**Comment by QiaolingTang** [comment]
@akostadinov I created a file named `clusterlogging_env.rb` under `features/support/` directory, and I use tag `@clusterlogging` as an identifier of loggging 4.x scenarios. Thanks a lot for the help.

File: lib/openshift/cluster_logging.rb (hunk 5 above)
Comment (QiaolingTang) [reply lib/openshift/cluster_logging.rb]: The output here is in json/yaml struct, and I think it's hard to return `Pod` objects. To check the rsyslog or fluentd pods ready or not, we use` rsyslog_nodes == rsyslog_ready_pods && rsyslog_failed_pods.count == 0 && rsyslog_notready_pods.count == 0`, so we only need the `count` here.

Example of status info:
```  
  status:
    collection:
      logs:
        fluentdStatus:
          daemonSet: fluentd
          nodes:
            fluentd-7qqdw: ip-10-0-162-127.eu-west-3.compute.internal
            fluentd-554s2: ip-10-0-158-63.eu-west-3.compute.internal
            fluentd-bd86w: ip-10-0-134-228.eu-west-3.compute.internal
            fluentd-m5kq4: ip-10-0-165-224.eu-west-3.compute.internal
            fluentd-mqt4h: ip-10-0-128-209.eu-west-3.compute.internal
            fluentd-wvr6f: ip-10-0-156-139.eu-west-3.compute.internal
          pods:
            failed: []
            notReady: []
            ready:
            - fluentd-554s2
            - fluentd-7qqdw
            - fluentd-bd86w
            - fluentd-m5kq4
            - fluentd-mqt4h
            - fluentd-wvr6f
```

**Comment by QiaolingTang** [comment]
@akostadinov @anpingli @pruan-rht  I've made some changes, could you please take a look again? Thanks a lot.

File: lib/openshift/cluster_logging.rb (hunk 4 above)
Comment (akostadinov) [reply lib/openshift/cluster_logging.rb]: What I request is to make this method fail in case rsyslog never does not become ready. It is needed and it is less reliable to ask all users of the method to do a check after calling it. When you do it here, all users of the method will fail when waiting is not successful.

File: lib/openshift/cluster_logging.rb (hunk 5 above)
Comment (akostadinov) [reply lib/openshift/cluster_logging.rb]: It is not hard to return `Pod` objects. It will be something like:
```
rsyslog_pods(user: user, cached: cached, quiet: quiet)['notReady'].map { |name|
  Pod.new(name: name, project: project)
}
```

But if we don't need the pods, then you can rename the methods to
```
rsyslog_notready_pod_names
rsyslog_ready_pod_names
...
```

Comment (akostadinov) [reply lib/openshift/cluster_logging.rb]: For the above suggestion to work, you'd have to fix how `ClusterLogging` resource is defined. It is presently `ClusterLogging < ClusterResource` but it is in fact a namespaces resource (it is within a namespace), thus it needs to be `ClusterLogging < ProjectResource`. @pruan-rht, please watch this in case anything relies on the wrong inheritance.

Comment (akostadinov) [reply lib/openshift/cluster_logging.rb]: YAML as a reference http://pastebin.test.redhat.com/799528

File: lib/openshift/cluster_logging.rb [hunk 6]
```diff
@@ -53,7 +76,12 @@ class ClusterLogging < ClusterResource
     end
 
     private def es_cluster_health(user: nil, cached: true, quiet: false)
-      es_status_raw(user: user, cached: cached, quiet: quiet).first['clusterHealth']
+      @result = es_status_raw(user: user, cached: cached, quiet: quiet).first['clusterHealth']
+      if @result
+        es_status_raw(user: user, cached: cached, quiet: quiet).first['clusterHealth']
+      else
+        es_status_raw(user: user, cached: cached, quiet: quiet).first['cluster']['status']
+      end
```
akostadinov: Instance variables should not be used here. A simpler way to do this would be:
```
private def es_cluster_health(user: nil, cached: true, quiet: false)
  es_status_raw(user: user, cached: cached, quiet: quiet).first['clusterHealth'] ||
    es_status_raw(user: user, cached: true, quiet: quiet).first['cluster']['status']
```

This will also avoid an extra API call (notice `cached: true` on second line.

lib/openshift/cluster_logging.rb
akostadinov: I mean somehting like:
```
    def wait_until_rsyslog_is_ready(user: nil, quiet: false, timeout: 5*60)
      success = wait_for(timeout) {
        rsyslog_ready?(user: user, cached: false, quiet: quiet)
      }
      unless success
        raise "rsyslog did not become ready within #{timeout} seconds"
    end
```

QiaolingTang:
@akostadinov  I updated my PR per your comments, please review it again. Thanks a lot.

akostadinov:
@pruan-rht , if nothing on your end, pls merge. I didn't check in details the steps logic.

PR #347, opened 2019-09-05
helper for https://github.com/openshift/cucushift/pull/7083

pruan-rht:
@yanpzhan, this PR adds route spec property support.  Example usage is `route('grafana').spec.host`.  There's a similar PR (#346 by @yapei).  Both PRs accomplishes the same end result (different ways of calling methods) but with different internal implementation.  @akostadinov, can you PTAL both and see which one you prefer?  Thanks!

### DIFF
--- a/lib/openshift/flakes/route_spec.rb
+++ b/lib/openshift/flakes/route_spec.rb
@@ -15,6 +15,10 @@ module BushSlicer
       return struct['host']
     end
 
+    def path
+      struct['path']
+    def
+
     def target_port
       return struct.dig('port', 'targetPort')
     end
--- a/lib/openshift/flakes/route_spec.rb
+++ b/lib/openshift/flakes/route_spec.rb
@@ -4,8 +4,6 @@ module BushSlicer
 
   # this class should help with parsing route specifications
   class RouteSpec
-    include Common::Helper
-
     attr_reader :struct
     private :struct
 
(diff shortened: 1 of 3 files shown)


yanpzhan:
@pruan-rht  Thanks for your help, I will have a try with that locally. Since both pr346 and pr347 meet my request, after the final chosen pr merge I will update my scenario.

yanpzhan:
@pruan-rht  The pr works for my scenario, thanks very much!

lib/openshift/flakes/route_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      def tls_insecureEdgeTerminationPolicy
+        return struct.dig('tls', 'insecureEdgeTerminationPolicy')
+      end
+
+      def tls_termination
+        return struct.dig('tls', 'termination')
+      end
+
+      def to_kind
```
yapei: maybe `target_kind` instead of `to_kind`?

lib/openshift/flakes/route_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        return struct.dig('to', 'name')
+      end
+
+      def to_weight
+        return struct.dig('to', 'weight')
+      end
+
+      def wildcardPolicy
+        return struct['wildcardPolicy']
+      end
```
yapei: We need `path` as well to parse `spec.path`

lib/openshift/flakes/route_spec.rb
yapei: Anyway they still looks good if we don't change

lib/openshift/flakes/route_spec.rb
yanpzhan: It's also be better If ca info could be added:  tls_certificate, tls_key, tls_caCertificate, tls_destinationCACertificate

pruan-rht: @yapei @yanpzhan do you have an example output I can look at because from the route that I've seen, I don't seen those elements.

yanpzhan: @pruan-rht Here is a example route yaml you can refer to: http://pastebin.test.redhat.com/797253

lib/openshift/flakes/route_spec.rb
akostadinov: we can probably alias the methods, wdyt?

akostadinov:
I thought this is another PR that I promissed to update. It is not that one but since I did already, @yapei , please check whether this version works for you.

lib/openshift/flakes/route_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    def target_weight
+      return struct.dig('to', 'weight')
+    end
+
+    def wildcard_policy
+      return struct['wildcardPolicy']
+    end
+  end
```
yapei: Looks like we missed `path` ?

yapei:
@akostadinov   @pruan-rht  We also need `path`
```
kind: Route
apiVersion: route.openshift.io/v1
metadata:
  name: reenroute
  namespace: prozyp
  selfLink: /apis/route.openshift.io/v1/namespaces/prozyp/routes/reenroute
  uid: ba12b684-d82b-11e9-8b92-0a580a81001f
  resourceVersion: '49104'
  creationTimestamp: '2019-09-16T02:43:12Z'
spec:
  host: retest.example.com
  subdomain: ''
  path: /test
  to:
    kind: Service
    name: service-secure
    weight: 100
  port:
    targetPort: https
.....
```

PR #348, opened 2019-09-06
Fix some cases

wewang58:


wewang58:
@openshift/devexp-qe pls help to review it, thanks

PR #351, opened 2019-09-10
Check if node is a master

xiuwang:
@openshift/devexp-qe this pr is used in https://github.com/openshift/cucushift/pull/7092

### DIFF
--- a/lib/openshift/node.rb
+++ b/lib/openshift/node.rb
@@ -79,6 +79,12 @@ module BushSlicer
       return ! res.nil?
     end
 
+    def is_master?(user: nil, cached: true, quiet: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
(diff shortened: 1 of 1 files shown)


wewang58:
LGTM

lib/openshift/node.rb
```diff
@@ -79,6 +79,12 @@ def is_worker?(user: nil, cached: true, quiet: false)
       return ! res.nil?
     end
 
+    def is_master?(user: nil, cached: true, quiet: false)
```
pruan-rht: @xiuwang there's already a `is_worker?` method (https://github.com/openshift/verification-tests/blob/master/lib/openshift/node.rb#L76).  @akostadinov, do you think we can reuse that method?  This is with the assumption that there will only be two types of host (if not a worker, then it's a master)...

PR #353, opened 2019-09-11
auto OCP-21085

SSshahan:
@yapei pls help to review. Thanks

PR #354, opened 2019-09-11
Add 4.2 rules for OCP-20849

xiaocwan:
Add 4.2 rules for OCP-20849
@yapei Please help to review the 4.2 rules. 4.1 is quite different from 4.2. So if it looks good,  IMO it's better to merge this one first to avoid conflict between 4.2 PRs. Also I'm continue to update 4.1 rules.

### DIFF
--- a/lib/rules/web/admin_console/4.2/deployment.xyaml
+++ b/lib/rules/web/admin_console/4.2/deployment.xyaml
@@ -39,14 +39,9 @@ search_nonexisted_image_and_check_message:
 search_and_deploy_image:
   action: search_image
   action: click_create_button
-delete_dc:
-  action: click_one_dropdown_action
-  params:
-    button_text: No Deployment Configs Found
-  action: check_page_contains
 delete_dc_with_dependency_objects:
-  action: click_one_dropdown_action
   params:
+    item: Delete Deployment Config
     button_text: Delete
+  action: click_one_dropdown_action
   action: click_button
-
--- a/lib/rules/web/admin_console/4.2/deployment.xyaml
+++ b/lib/rules/web/admin_console/4.2/deployment.xyaml
@@ -30,3 +30,23 @@ search_image:
   action: clear_input_value
   action: set_input_value
   action: click_search_button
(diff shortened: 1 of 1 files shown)


xiaocwan:
Rules about search image are pending on https://github.com/openshift/verification-tests/pull/353

lib/rules/web/admin_console/4.2/deployment.xyaml
```diff
@@ -30,3 +30,31 @@ search_image:
   action: clear_input_value
   action: set_input_value
   action: click_search_button
+deploy_search_nonexisted_image:
+  params:
+    search_content: non-existed
+    content: Could not load image metadata
+  action: search_image
+  action: check_page_contains
+deploy_search_existed_image:
```
yapei: I think this can be totally replaced with `search_image` it's not necessary to check the message for the positive testing

lib/rules/web/admin_console/4.2/deployment.xyaml
```diff
@@ -30,3 +30,31 @@ search_image:
   action: clear_input_value
   action: set_input_value
   action: click_search_button
+deploy_search_nonexisted_image:
```
yapei: Since we didn't deploy app using this step, so I would call it `search_nonexisted_image_and_check_message`

lib/rules/web/admin_console/4.2/deployment.xyaml
```diff
@@ -30,3 +30,23 @@ search_image:
...
+search_nonexisted_image_and_check_message:
+  params:
+    search_content: non-existed
+    content: Could not load image metadata
+  action: search_image
+  action: check_page_contains
+search_and_deploy_image:
+  action: search_image
+  action: click_create_button
+delete_dc:
```
yapei: Seems it's not used, help remove if it's not needed

yapei:
LGTM, thanks very much

xiaocwan:
@akostadinov Thanks!

PR #360, opened 2019-09-19
Update ocp-13509, 15017, 13507, 15465

huiran0826:
These are some updates for cases:13509, 15017, 13507, 15465.
13507,15017 was updated because original way checking OVS rules from node  failed in 4.x. Instead of that, check OVS rules from SDN pod.

15465 need add @admin tag before the case.

13509 use yahoo.com to replace www.yahoo.com, sometimes nslookup of www.yahoo.com got less than 3 ip addresses that cause the case failing. Moreover, the default dns-egresspolicy2.json match yahoo.com and we don't need to change it anymore.

logs:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/44103/console
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/44104/console

Any comments will be welcomed, thanks!
@zhaozhanqi @anuragthehatter @lihongan

### DIFF
--- a/features/networking/egress-ingress.feature
+++ b/features/networking/egress-ingress.feature
@@ -109,21 +109,18 @@ Feature: Egress-ingress related networking scenarios
     And evaluation of `project.name` is stored in the :proj1 clipboard
  
     # Create egress policy in project-1
-    And evaluation of `BushSlicer::Common::Net.dns_lookup("yahoo.com")` is stored in the :yahoo_ip clipboard
-    When I download a file from "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/egress-ingress/dns-egresspolicy1.json"
-    And I replace lines in "dns-egresspolicy1.json":
-      | 98.138.0.0/16 | <%= cb.yahoo_ip %>/32 |
     And I run the :create admin command with:
-      | f | dns-egresspolicy1.json |
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/egress-ingress/dns-egresspolicy1.json |
       | n | <%= cb.proj1 %> |
     Then the step should succeed
  
     # Check egress rule added in openflow
     Given I select a random node's host
-    When I run the ovs commands on the host:
-       | ovs-ofctl dump-flows br0 -O openflow13 \| grep <%= cb.yahoo_ip %> |
-    And the output should contain 1 times:
-      | actions=drop |
+    And evaluation of `node.name` is stored in the :node_name clipboard
+    When I run command on the "<%= cb.node_name %>" node's sdn pod:
+      | ovs-ofctl| -O | openflow13 | dump-flows | br0 |
+    And the output should contain:
+      | nw_dst=98.138.0.0/16 actions=drop |
 
     # Check egress policy can be deleted
     When I run the :delete admin command with:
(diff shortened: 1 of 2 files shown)


features/networking/egress-ingress.feature
```diff
@@ -257,13 +255,14 @@ Feature: Egress-ingress related networking scenarios
...
-      | ovs-ofctl dump-flows br0 -O openflow13 \| grep table=101 |
+    When I run command on the "<%= cb.node_name %>" node's sdn pod:
+      | ovs-ofctl| -O | openflow13 | dump-flows | br0 | table=101 |
     And the output should contain:
       | actions=drop |
     """
     # Check ping from pod
-    When I execute on the pod:
+    Given I use the "<%= cb.proj1 %>" project
+	When I execute on the "<%= cb.pod1 %>" pod:
```
zhaozhanqi: please add indentation

zhaozhanqi:
Others LGTM, Thanks for updating this..

features/networking/egress-ingress.feature
huiran0826: Updated it, thanks for the comment!

lihongan:
LGTM, thanks.
If possible, we should avoid to `run commands on the host` in 4.x.

features/networking/egress-ip.feature
```diff
@@ -2,6 +2,7 @@ Feature: Egress IP related features
 
   # @author bmeng@redhat.com
   # @case_id OCP-15465
+  @admin
```
akostadinov: Do we need `@admin` here? How did it work in the past?

akostadinov:
> If possible, we should avoid to run commands on the host in 4.x.

Do you see any place a command on the host is being executed?

features/networking/egress-ingress.feature
```diff
@@ -109,21 +109,18 @@ Feature: Egress-ingress related networking scenarios
...
     And I run the :create admin command with:
-      | f | dns-egresspolicy1.json |
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/egress-ingress/dns-egresspolicy1.json |
       | n | <%= cb.proj1 %> |
     Then the step should succeed
  
     # Check egress rule added in openflow
     Given I select a random node's host
-    When I run the ovs commands on the host:
-       | ovs-ofctl dump-flows br0 -O openflow13 \| grep <%= cb.yahoo_ip %> |
```
lihongan: Above command is executed on hosts and it need to ssh to the host. It can be replaced by `run command on the "<%= cb.node_name %>" node's sdn pod`.

akostadinov: It is removed (red lines). Do you see any command on the host that is not removed?

lihongan: No, all ovs commands on the host have been removed.
Sorry for my confusing comments. I means the fix in this PR looks good and we should follow it and update other cases like this, since this PR just updates some critical cases.
Except the `ovs` command there are some other commands on the host in networking cases, if I am not mistaken. I'd like to see the follow up PRs to update them.

features/networking/egress-ip.feature
lihongan: Interesting, seems the first step `Given I select a random node's host` need `@admin`, but I cannot tell how did it work in the past :(
```
Given /^I select a random node's host$/ do
  ensure_admin_tagged
```

lihongan: I found the initial PR: https://github.com/openshift/cucushift/pull/6792/files, maybe it is helpful but seems no logs after the review/update.

huiran0826: I just rerun the tests against V3 branch, it entered the same issue as master branch.  Here is the log errors:
Given I select a random node's host                                   # features/step_definitions/node.rb:9
      tag scenario @admin as you use admin access (RuntimeError)
      /home/jenkins/workspace/Runner-v3/lib/world.rb:66:in `ensure_admin_tagged'

It seems "Given I select a random node's host " really need @admin tag.
I am not sure how the original codes pass the tests,  but now after adding @admin tag, the case can pass.

PR #372, opened 2019-10-08
auto logging cases

QiaolingTang:
I'm automating some logging cases, @anpingli @pruan-rht @akostadinov  PTAL, thanks.

Cases: https://url.corp.redhat.com/c70d998

I added a new tag `@commonlogging`, when executing `@commonlogigng` cases, it will create clusterlogging instance with https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/example.yaml and will not delete the clusterlogging instance when the case finished.  I'm doing this to save some time.
The `@commonlogging` cases means no special settings in the logging cluster.

Run log:  https://url.corp.redhat.com/9f816f1


anpingli:
LGTM in function

anpingli:
@akostadinov  @pruan-rht  Could you review this PR?

features/step_definitions/logging.rb
```diff
@@ -309,3 +309,11 @@
     raise "Get metrics failed with error, #{@result[:response]}"
   end
 end
+
+Given /^default storageclass is stored in the#{OPT_SYM} clipboard$/ do | cb_name |
```
pruan-rht: @QiaolingTang perhaps this step belongs to `storage_class.rb` instead since it's `storageclass` related step

QiaolingTang: @pruan-rht OK, thank you.

features/step_definitions/storage_class.rb
```diff
@@ -228,3 +228,11 @@
     raise "failed to clone StorageClass from: #{src_sc}"
   end
 end
+
+Given /^default storageclass is stored in the#{OPT_SYM} clipboard$/ do | cb_name |
+  ensure_admin_tagged
+  cb_name = 'default_sc' unless cb_name
+  _sc = BushSlicer::StorageClass.get_matching(user: user) { |sc, sc_hash| sc.default? }.first
+  raise "Unable to get default storage class " unless _sc
+  cb[cb_name] = _sc
```
akostadinov: please add `cache_resources _sc`

features/logging/4.x/cluster_logging_operator.feature
```diff
@@ -0,0 +1,60 @@
...
+    Then the step should succeed
+    Given I get project clusterlogging named "instance" as YAML
+    Then the step should succeed
+    And the output should contain:
+      | nodeCount: 3 |
+    And I wait up to 600 seconds for the steps to pass:
+    """
+    And 3 pods become ready with labels:
+      | cluster-name=elasticsearch,component=elasticsearch,es-node-client=true,es-node-data=true,es-node-master=true |
+    """
```
akostadinov: why do you wrap into a wait step? `3 pods become ready with labels:` is already waiting

features/logging/4.x/cluster_logging_operator.feature
```diff
@@ -0,0 +1,60 @@
...
+      | nodeCount: 4 |
+    And I wait up to 600 seconds for the steps to pass:
+    """
+    And 4 pods become ready with labels:
+      | cluster-name=elasticsearch,component=elasticsearch,es-node-client=true,es-node-data=true |
+    And 3 pods become ready with labels:
+      | cluster-name=elasticsearch,component=elasticsearch,es-node-client=true,es-node-data=true,es-node-master=true |
+    And a pod becomes ready with labels:
+      | cluster-name=elasticsearch,component=elasticsearch,es-node-client=true,es-node-data=true,es-node-master=false |
+    """
```
akostadinov: Do we need wrapper wait step here? Same question applies to all other places.

features/logging/4.x/collector.feature
```diff
@@ -0,0 +1,40 @@
...
+  @admin
+  @destructive
+  @commonlogging
+  Scenario: All nodes logs had sent logs to Elasticsearch
+    Given I switch to cluster admin pseudo user
+    Given I use the "openshift-logging" project
+    # get node ip
+    #Given <%= daemon_set("fluentd").replica_counters[:desired] %> pods become ready with labels:
+    #  | component=fluentd |
+    And evaluation of `@pods.map {|n| n.node_ip}.uniq` is stored in the :node_ip clipboard
```
akostadinov: I suggest to name this `node_ips` as there would be multiple values.

features/logging/4.x/elasticsearch.feature
```diff
@@ -0,0 +1,65 @@
...
+    Then the step should succeed
+    And the output should contain:
+      | managementState: Unmanaged |
+    Given I get project elasticsearch named "elasticsearch" as YAML
+    Then the step should fail
+    When I run the :patch client command with:
+      | resource      | clusterlogging                                                                                        |
+      | resource_name | instance                                                                                              |
+      | p             | {"spec":{"logStore":{"elasticsearch":{"storage":{"storageClassName": "<%= cb.default_sc.name %>"}}}}} |
+      | type          | merge                                                                                                 |
```
akostadinov: do we need clean-up step as well here?

features/logging/4.x/elasticsearch.feature
```diff
@@ -0,0 +1,65 @@
...
+    Then the step should succeed
+    Given I get project clusterlogging named "instance" as YAML
```
akostadinov: looking at how many times we look at raw YAML, perhaps create a clusterlogging class? @pruan-rht ?

features/logging/4.x/elasticsearch.feature
```diff
@@ -0,0 +1,65 @@
...
+      | resource_name | instance                               |
+      | p             | {"spec":{"managementState":"Managed"}} |
+      | type          | merge                                  |
+    Then the step should succeed
+    Given I get project clusterlogging named "instance" as YAML
+    And the output should contain:
+      | storageClassName: <%= cb.default_sc.name %>   |
+      | managementState: Managed                      |
+    Given I wait for the "elasticsearch" elasticsearches to appear
+    Given I get project elasticsearch named "elasticsearch" as YAML
```
akostadinov: maybe create a class here too?

akostadinov: I will not block merging for these classes. But if @pruan-rht wants to help, that would be arwesome.

pruan-rht: @QiaolingTang @akostadinov +1 on writing a class interface for those two new resources.  I'll work on them Monday or perhaps tomorrow if I can find some time.

features/logging/4.x/elasticsearch.feature
```diff
@@ -0,0 +1,65 @@
...
+    Given a pod becomes ready with labels:
+      | component=elasticsearch |
+    And the expression should be true> pod.volume_claims.first.name.include? "elasticsearch-elasticsearch-cdm" and pod.volume_claims.first.name.include? cb.gen_uuid
+    Given I wait until the ES cluster is healthy
+    Given I delete the clusterlogging instance
+    Then the step should succeed
+    And I run the :delete client command with:
+      | object_type | pvc  |
+      | all         | true |
+    Then the step should succeed
```
akostadinov: is this a mandatory clean-up step? If it fails, will we fail to clean automatically?
If so, then we need this to be a clean-up step registered earlier. Otherwise we may remove these 2 steps.

features/logging/4.x/elasticsearch_operator.feature
```diff
@@ -0,0 +1,35 @@
...
+
+    Given I use the "openshift-logging" project
+    When I get project servicemonitor named "monitor-elasticsearch-cluster" as YAML
+    Then the step should succeed
+    And the output should contain:
+      | port: elasticsearch        |
+      | path: /_prometheus/metrics |
+    When I get project service named "elasticsearch-metrics" as YAML
+    Then the step should succeed
+    And evaluation of `@result[:parsed]['spec']['clusterIP']` is stored in the :service_ip clipboard
```
akostadinov: 3 steps can be folded into:
```
Given evaluation of `service("elasticsearch-metrics").ip` is stored in the :service_ip clipboard
```

features/logging/4.x/elasticsearch.feature
QiaolingTang: @pruan-rht @akostadinov I've add some classes, but I forget to update the case. https://github.com/openshift/verification-tests/pull/379/files#diff-b7b17bd879fb9cd97bf091211d2e34ea and https://github.com/openshift/verification-tests/pull/379/files#diff-bc2107c36e302368fe4b4afe755439de

features/logging/4.x/elasticsearch.feature
QiaolingTang: OK, I'll register a clean-up step in this case.

features/logging/4.x/elasticsearch.feature
```diff
@@ -0,0 +1,56 @@
...
+    And I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/example_unmanaged.yaml |
+    Then the step should succeed
+    And I register clean-up steps:
+    """
+    Given I delete the clusterlogging instance
+    Then the step should succeed
+    And I run the :delete client command with:
+      | object_type | pvc  |
+      | all         | true |
```
akostadinov: I think you better add `namespace` parameter here so that step properly works in case we end up with unexpected namespace at some point. Thank you.

features/logging/4.x/elasticsearch.feature
QiaolingTang: I've updated several classes in this PR. But for some fields in the elasticsearch CRD instance, I haven't figured out a good way to process them, so I keep using step `I get project elasticsearch named "elasticsearch" as YAML`.

features/logging/4.x/elasticsearch.feature
QiaolingTang: I added a clean-up step in the above, do I need to add one more clean-up step here? In my observation, one clean-up step in one case is enough as the clean-up step would be executed when the case failed or succeed.

features/logging/4.x/elasticsearch.feature
```diff
@@ -0,0 +1,69 @@
...
+    Then the step should succeed
+    And I register clean-up steps:
+    """
+    Given I delete the clusterlogging instance
+    Then the step should succeed
+    And I run the :delete client command with:
+      | object_type | pvc               |
+      | all         | true              |
+      | n           | openshift-logging |
+    Then the step should succeed
```
akostadinov: if no PVCs exist, will command fail?

features/logging/4.x/elasticsearch_operator.feature
```diff
@@ -0,0 +1,32 @@
...
+      |serviceaccount_name | prometheus-k8s |
+    Then the step should succeed
+    And evaluation of `@result[:response]` is stored in the :token clipboard
+
+    Given I use the "openshift-logging" project
+    When I get project servicemonitor named "monitor-elasticsearch-cluster" as YAML
+    Then the step should succeed
+    And the output should contain:
+      | port: elasticsearch        |
+      | path: /_prometheus/metrics |
```
akostadinov: We have a class `ServiceMonitor` already.  We better have here:
```
Given the expression should be true> service_monitor("monitor-elasticsearch-cluster").port == "elasticsearch"
And the expression should be true> service_monitor("monitor-elasticsearch-cluster").path == "/_prometheus/metrics"
```
`path` and `port` methods need to be written. I'll merge for now though as this is not critical here.

lib/openshift/cluster_logging.rb
```diff
@@ -165,5 +165,24 @@ def redundancy_policy(user: nil, quiet: false, cached: false)
...
+
+    def logstore_storageclassname(user: nil, cached: false, quiet: false)
+      return logstore_storage(user: user, cached: cached, quiet: quiet).dig('storageClassName')
+    end
+
+    def logstore_storagesize(user: nil, cached: false, quiet: false)
+      return logstore_storage(user: user, cached: cached, quiet: quiet).dig('size')
+    end
+
+    def logstore_nodecount(user: nil, cached: false, quiet: false)
```
akostadinov: could you split method names to words with underscores? `logstore_node_count`, `logstore_storage_size`, etc.?

lib/openshift/elasticsearch.rb
```diff
@@ -11,5 +11,11 @@ def redundancy_policy(user: nil, quiet: false, cached: false)
       rr = raw_resource(user: user, cached: cached, quiet: quiet)
       return rr.dig('spec', 'redundancyPolicy')
     end
+
+    def es_nodes(user: nil, quiet: false, cached: false)
```
akostadinov: how about name the method just `nodes`?

features/logging/4.x/elasticsearch.feature
QiaolingTang: I checked in my computer, when there is no pvc exist, the output is `No resources found`, and the command exit code is `0`. How does the cucushift determine whether a command is successfully executed?

lib/openshift/cluster_logging.rb
QiaolingTang: OK

lib/openshift/elasticsearch.rb
QiaolingTang: OK

features/logging/4.x/elasticsearch.feature
QiaolingTang: I checked this using cucushift, the command won't fail if there is no pvc.
```    And I run the :delete client command with:              # features/step_definitions/cli.rb:13
      [01:19:03] INFO> Shell Commands: oc delete pvc --all --config=/home/qitang/workdir/wlc-trust-142-qitang/ocp4_admin.kubeconfig -n openshift-logging
      No resources found
      
      [01:19:05] INFO> Exit Status: 0
      | object_type | pvc               |
      | all         | true              |
      | n           | openshift-logging |
    Then the step should succeed                            # features/step_definitions/common.rb:4
```

features/logging/4.x/elasticsearch_operator.feature
QiaolingTang: If using these non-exist methods, the case would fail. I think that's great to have `path` and `port` methods, I'll keep these steps in this scenario, and once we have the methods for `servicemonitor` object, I'll update the scenarios. What do you think?

QiaolingTang: @akostadinov  I wrote the `port` and `path` methods, PTAL, thanks.

features/logging/cluster_logging_operator.feature
```diff
@@ -0,0 +1,88 @@
...
+      | type          | merge                                                   |
+    Then the step should succeed
+    And the expression should be true> cluster_logging('instance').logstore_node_count == 3
+    And 3 pods become ready with labels:
+      | cluster-name=elasticsearch,component=elasticsearch,es-node-client=true,es-node-data=true,es-node-master=true |
+    When I run the :patch client command with:
+      | resource      | clusterlogging                                          |
+      | resource_name | instance                                                |
+      | p             | {"spec":{"logStore":{"elasticsearch":{"nodeCount":4}}}} |
+      | type          | merge                                                   |
```
akostadinov: Do we need clean-up for these patch commands? All scenarios need a separate clean-up. Or didyou add whole feature clean-up?

akostadinov: Other than that lgtm. Please do not mark as resolved my comments. This is how I keep track of which changes remain to be reviewed.

akostadinov:
/lgtm

please update clean-up later if needed per my comment above

features/logging/cluster_logging_operator.feature
QiaolingTang: I think I already have clean-up steps since I've set `| remove_logging_pods | true                                                                                                 |` when I create the clusterlogging CR instance, if the patch command failed, then the scenario will fail, and the clusterlogging instance will be deleted.  Please look at https://github.com/openshift/verification-tests/blob/master/features/step_definitions/logging.rb#L251-L280 and https://github.com/openshift/verification-tests/pull/372/files/6f40b7f3b9ca9cdd9a3ef333deb231531c784c6f#diff-62bc6201426dc6e375578edc55458329R59-R62

QiaolingTang: Sorry about marking your comments as resolved. I'll keep this in mind and won't change your comments' status later.

PR #376, opened 2019-10-10
utilty to summarize instance usages for various cloud vendors

pruan-rht:
usage is `./tools/cloud_cop.rb aws` to get summary of all regions or `./tools/cloud_cop.rb aws --region=us-east-2` to just get the summary for that region.  Currently only support AWS, will add other cloud providers later.  Here's an example output
http://pastebin.test.redhat.com/804638

### DIFF
--- a/lib/instance_summary.rb
+++ b/lib/instance_summary.rb
@@ -1,11 +1,8 @@
 require 'text-table'
+require 'thread'
+require 'thwait'
+
 module BushSlicer
-  # represents what we want to show in the summary output
-  # 1. platform   (aws, azure, gce, openstack, and etc)
-  # 2. instance_name
-  # 3. uptime
-  # 4. link to Flexy job ID that was used to spin it up (none if nont-Fles
-  #
   class InstanceSummary
     attr_accessor :jenkins
 
@@ -15,7 +12,6 @@ module BushSlicer
 
     ## print out summary in a text table format
     def print_summary(summary)
-
       table = Text::Table.new
       table.head = ['name', 'uptime', 'flexy_job_id', 'region']
       summary.each do | s |
@@ -35,15 +31,13 @@ module BushSlicer
         total += s[:inst_count]
         table.rows << [s[:platform], s[:region], s[:inst_count]]
       end
-      table.rows << ["Total instances:", "", total]
+      table.foot = ["Total instances", "", total]
       puts table
-
     end
-
   end
 
   class AwsSummary < InstanceSummary
-    attr_accessor :amz, :summary, :table
+    attr_accessor :amz
     def initialize(jenkins: nil)
       @amz = Amz_EC2.new
       @jenkins = jenkins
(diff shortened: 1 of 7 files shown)


liangxia:
Great. I like it.

pruan-rht:
@akostadinov this is ready for review now.  Thanks!

pruan-rht:
@akostadinov bump...

pruan-rht:
@akostadinov , bump, bump. :)

lib/launchers/openstack10.rb
```diff
@@ -822,6 +824,34 @@ def get_routers
...
+      raise res[:response] unless res[:success]
+      return res[:parsed]
+    end
+
+    def get_running_instances
+      threads = []
+      running_inst = {}
+      instances = get_objects("servers")
+      instances.each do |instance|
+        threads << Thread.new(instance) do | i|
```
akostadinov: better throttle this to 15 or so connections

akostadinov: use method `array#each_slice`

lib/instance_summary.rb
```diff
@@ -0,0 +1,287 @@
+require 'text-table'
+require 'thread'
+require 'thwait'
+
+module BushSlicer
+  class InstanceSummary
```
akostadinov: Could you add some description about what this class is doing? I looks like belonging to `tools/lib`.

lib/instance_summary.rb
```diff
@@ -0,0 +1,287 @@
...
+    attr_accessor :jenkins
+
+    def initialize(jenkins: nil)
```
akostadinov: you allow initialization without a jenkins instance. But later you call methods on `jenkins` without checking for `nil`. So perhaps `: nil` part needs to be removed. To make sure objects are always created with a `jenkins` object set.

lib/jenkins.rb
```diff
@@ -0,0 +1,74 @@
+require 'net/https'
+require 'thread'
+
+module BushSlicer
+  class Jenkins
+    attr_accessor :client, :build_map
+    def initialize
+      @client = JenkinsApi::Client.new(:server_url => 'https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com',
```
akostadinov: please do not hardcode this. it needs to be a config under `private/config/config.yaml`. Actually it is already there under ```
services:
  jenkins: ...
```

File: lib/jenkins.rb [hunk 5]
```diff
@@ -0,0 +1,74 @@
+require 'net/https'
+require 'thread'
+
+module BushSlicer
+  class Jenkins
```
akostadinov: This looks like more suitable for `tools/lib`.

lib/launchers/openstack10.rb
pruan-rht: @akostadinov what do you mean by using #each_slice?   We probably don't need to throttle as Jenkins only returns 100 entries max.  Also, if we do want to use threadpool, then that's another gem we need to install

akostadinov: 100 connections at the same time not very little but sounds not terrible.
wrt a thread pool, we can jst create that number of threads per each slice. Not needed to use a thread pool.

lib/jenkins.rb
pruan-rht: @akostadinov  how do I get the `conf`

PR #380, opened 2019-10-10
Centos 8 image release recently broke the Dockerfile, need to use cen…

pruan-rht:
…tos:7 instead


pruan-rht:
This is urgent since it's blocking other PRs from merging due to `ci/prow/images` test is failing.  Once it passes here, I'll merge it to unblock it.

pruan-rht:
merging this PR to unblock the merge queue.

akostadinov:
Thanks!

PR #385, opened 2019-10-14
auto case OCP-25599

QiaolingTang:
@anpingli PTAL, thanks.
Polarion case: https://url.corp.redhat.com/7335487

### DIFF
--- a/features/logging/4.x/kibana.feature
+++ b/features/logging/4.x/kibana.feature
@@ -1,0 +1,47 @@
+@clusterlogging
+@commonlogging
+Feature: Kibana related features
+
+  # @auther qitang@redhat.com
+  # @case_id OCP-25599
+  @admin
+  @destructive
+  Scenario: Show logs on Kibana web console according to different user role
+    Given I switch to the first user
+    Given I create a project with non-leading digit name
+    Given evaluation of `project` is stored in the :proj clipboard
+    When I run the :new_app client command with:
+      | file | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/loggen/container_json_unicode_log_template.json |
+    Then the step should succeed
+    Given a pod becomes ready with labels:
+      | run=centos-logtest,test=centos-logtest |
+    Given I switch to cluster admin pseudo user
+    Given I use the "openshift-logging" project
+    And I wait for the "kibana" route to appear
+    And I wait for the "project.<%= cb.proj.name %>" index to appear in the ES pod with labels "es-node-master=true"
+    Given I switch to the first user
+    And I login to kibana logging web console
+    And I perform the :kibana_find_index_pattern web action with:
+      | index_pattern_name | project.<%= cb.proj.name %>.<%= cb.proj.uid %>.* |
+    Then the step should succeed
+    Given cluster role "cluster-admin" is added to the "first" user
+    Then I login to kibana logging web console
+    Given evaluation of `[".operations.*", ".all", ".orphaned", "project.*"]` is stored in the :indices clipboard
+    And I run the :kibana_expand_index_patterns web action
+    Then the step should succeed
+    Given I repeat the following steps for each :index_name in cb.indices:
+    """
+    And I perform the :kibana_find_index_pattern web action with:
+      | index_pattern_name | #{cb.index_name} |
+    Then the step should succeed
+    """
+    And cluster role "cluster-admin" is removed from the "first" user
+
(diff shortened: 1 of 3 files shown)


QiaolingTang:
I'm not able to run the test in `Runner-v3` since https://github.com/openshift/verification-tests/pull/372 hasn't merged.

features/step_definitions/logging.rb
```diff
@@ -309,3 +309,26 @@
     raise "Get metrics failed with error, #{@result[:response]}"
   end
 end
+
+Given /^I open kibana console in a browser$/ do
```
akostadinov: What is the difference with `Given I login to kibana logging web console`? Can one step be used for kibana instead of two?

features/logging/4.x/kibana.feature
```diff
@@ -0,0 +1,52 @@
...
+    Given I switch to cluster admin pseudo user
+    Given I use the "openshift-logging" project
+    And I wait for the "kibana" route to appear
+    And I wait for the "project.<%= cb.proj.name %>" index to appear in the ES pod with labels "es-node-master=true"
+    Given I switch to the first user
+    And I open kibana console in a browser
+    And I perform the :kibana_find_index_pattern web action with:
+      | index_pattern_name | project.<%= cb.proj.name %>.<%= cb.proj.uid %>.* |
+    Then the step should succeed
+    Given cluster role "cluster-admin" is added to the "first" user
```
akostadinov: Could you use step `the first user is cluster-admin`?

features/logging/4.x/kibana.feature
```diff
@@ -0,0 +1,52 @@
...
+    #Then the step should succeed
+    Given evaluation of `[".operations.*", ".all", ".orphaned", "project.*"]` is stored in the :indices clipboard
+    Given I repeat the following steps for each :index_name in cb.indices:
+    """
+    And I run the :kibana_expand_index_patterns web action
+    And I perform the :kibana_find_index_pattern web action with:
+      | index_pattern_name | #{cb.index_name} |
+    Then the step should succeed
+    """
+    And cluster role "cluster-admin" is removed from the "first" user
```
akostadinov: can we instead use second user to check this?

features/step_definitions/logging.rb
QiaolingTang: The step `Given I login to kibana logging web console` could be used for logging 3.x, for logging 4.x, the login has changed. I'm not sure if I could modify the step or not, so I added a new step for logging 4.x .

features/logging/4.x/kibana.feature
QiaolingTang: no, the purpose of the case is to test whether the kibana could show indices correctly when the user role changes.

features/logging/4.x/kibana.feature
QiaolingTang: I have a question: what are the differences between step `cluster role "cluster-admin" is added to the "first" user` and step `the first user is cluster-admin` ?

features/step_definitions/logging.rb
akostadinov: the `master` branch is for 4.x and `v3` branch is for 3.x. So please change same step.

features/logging/4.x/kibana.feature
akostadinov: ack

features/logging/4.x/kibana.feature
akostadinov: No difference. Just slightly different semantics. But given later comment, I think keeping your original step makes more sense.

features/step_definitions/logging_metrics.rb
```diff
@@ -62,20 +62,22 @@
...
-    | username   | <%= user.name %>              |
-    | password   | <%= user.password %>          |
-    | kibana_url | <%= cb.logging_console_url %> |
+    | username   | <%= user.name %>                      |
+    | password   | <%= user.password %>                  |
+    | kibana_url | https://<%= cb.logging_console_url %> |
+    | idp        | <%= env.idp %>                        |
     })
+  step %Q/I run the :kibana_authorization_approve web action/
+  sleep 30
```
akostadinov: why do we hardcode sleep? Can't we wait for something to appear on page?

akostadinov: Other than that, things look nice!

QiaolingTang:
Run Log: https://url.corp.redhat.com/05f8130

PR #387, opened 2019-10-15
update step run command on node's sdn pod

huiran0826:
As there is one step in case OCP-9753 need run command fails, so update the file to remove the raise error if not success.

Any comments will be welcomed, thanks!
@zhaozhanqi @anuragthehatter @lihongan

The log is for PR https://github.com/openshift/cucushift/pull/7116 and this PR.
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/44963/console

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -671,7 +671,6 @@ Given /^I run command on the#{OPT_QUOTED} node's sdn pod:$/ do |node_name, table
   }.first
   cache_resources sdn_pod
   @result = sdn_pod.exec(network_cmd, as: admin)
-  raise "Failed to execute network command!" unless @result[:success]
 end
 
 Given /^the default interface on nodes is stored in the#{OPT_SYM} clipboard$/ do |cb_name|


akostadinov:
If this check is removed, then you need to make sure all other uses of this step are protected from failed commands as a separate step in the scenario. 

I did `rg -A 2 "node's sdn pod:"` and I see that there are a number of unprotected uses in `features/networking/egress-ingress.feature`. But please double check.

huiran0826:
> If this check is removed, then you need to make sure all other uses of this step are protected from failed commands as a separate step in the scenario.
> 
> I did `rg -A 2 "node's sdn pod:"` and I see that there are a number of unprotected uses in `features/networking/egress-ingress.feature`. But please double check.

Thanks Alex, your consideration is correct, I will update case 9753 in another way, this PR will be closed.

akostadinov:
On my end it is fine to make the change. Just there is one or two feature files that would need update. Not a huge problem. As you decide.

PR #388, opened 2019-10-16
add step for delete the specifid ovs pod

zhaozhanqi:
logs:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/44981/consoleFull

@huiran0826 @anuragthehatter @weliang1
Could you help review, thanks

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -683,3 +683,16 @@ Given /^the default interface on nodes is stored in the#{OPT_SYM} clipboard$/ do
   cb[cb_name] = @result[:response].chomp
   logger.info "The node's default interface is stored in the #{cb_name} clipboard."
 end
+
+Given /^I restart the ovs pod on the#{OPT_QUOTED} node$/ do | node_name |
+  ensure_admin_tagged
+  ensure_destructive_tagged
+
+  ovs_pod = BushSlicer::Pod.get_labeled("app=ovs", project: project("openshift-sdn", switch: false), user: admin) { |pod, hash|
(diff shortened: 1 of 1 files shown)


features/step_definitions/networking.rb
```diff
@@ -683,3 +683,15 @@
   cb[cb_name] = @result[:response].chomp
   logger.info "The node's default interface is stored in the #{cb_name} clipboard."
 end
+
+Given /^I restart the ovs pod on the#{OPT_QUOTED} node$/ do | node_name |
+  ensure_admin_tagged
```
akostadinov: if this restart can affect other users of the cluster running stuff at the same time, then this needs to be `ensure_destructive_tagged`.

akostadinov:
It is better, when you submit a new step, to also submit a scenario that uses it.

features/step_definitions/networking.rb
zhaozhanqi: updated, thanks

zhaozhanqi:
> It is better, when you submit a new step, to also submit a scenario that uses it.

yes, see https://github.com/openshift/cucushift/pull/7117

features/step_definitions/networking.rb
anuragthehatter: /lgtm. yep, it seems to be an ensure_destructive_tagged

akostadinov:
It is much preferable to submit at least one scenario in the same repo as an example. It doesn't matter tier of test.

PR #389, opened 2019-10-16
auto case OCP-22558 OCP-24292

QiaolingTang:
@anpingli @yapei  could you please help to review this? Thanks.

case: https://url.corp.redhat.com/42aeeb5
Log: https://url.corp.redhat.com/71579e0

### DIFF
--- a/features/logging/deploy_logging_operators_via_console.feature
+++ b/features/logging/deploy_logging_operators_via_console.feature
@@ -1,0 +1,65 @@
+Feature: Deploy logging operators via console
+
+  # @auther qitang@redhat.com
+  # @case_id OCP-22558
+  @admin
+  @destructive
+  Scenario: Deploy cluster-logging operator via web console.
+    # clear the logging related resources
+    Given logging service is removed successfully
+    Given the logging operators are redeployed after scenario
+    Given logging channel name is stored in the :logging_channel clipboard
+    Given I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/01_clo_ns.yaml |
+    Then the step should succeed
+    Given I register clean-up steps:
+    """
+    Given logging service is removed successfully
+    Then the step should succeed
+    """
+    Given "cluster-logging" packagemanifest's operator source name is stored in the :clo_opsrc clipboard
+    Given I switch to the first user
+    Given the first user is cluster-admin
+    Given I open admin console in a browser
+    # subscribe cluster-logging-operator
+    When I perform the :goto_operator_subscription_page web action with:
+      | package_name     | cluster-logging     |
+      | catalog_name     | <%= cb.clo_opsrc %> |
+      | target_namespace | openshift-logging   |
+    Then the step should succeed
+    And I perform the :set_custom_channel_and_subscribe web action with:
+      | update_channel    | <%= cb.logging_channel %> |
+      | install_mode      | OwnNamespace              |
+      | approval_strategy | Automatic                 |
+    Given cluster logging operator is ready
+
+  # @auther qitang@redhat.com
+  # @case_id OCP-24292
+  @admin
+  @destructive
+  Scenario: Deploy elasticsearch-operator via Web Console
+    # clear the logging related resources
+    Given logging service is removed successfully
+    Given the logging operators are redeployed after scenario
+    Given logging channel name is stored in the :logging_channel clipboard
+    Given "elasticsearch-operator" packagemanifest's operator source name is stored in the :eo_opsrc clipboard
+    Given I switch to the first user
+    Given the first user is cluster-admin
+    Given admin ensures "elasticsearch-operator" subscriptions is deleted from the "openshift-operators" project after scenario
+    Given I open admin console in a browser
+    When I perform the :goto_operator_subscription_page web action with:
+      | package_name     | elasticsearch-operator |
+      | catalog_name     | <%= cb.eo_opsrc %>     |
+      | target_namespace | openshift-operators    |
+    Then the step should succeed
+    When I perform the :set_custom_channel_and_subscribe web action with:
+      | update_channel    | <%= cb.logging_channel %> |
+      | install_mode      | AllNamespace              |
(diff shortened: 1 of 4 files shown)


QiaolingTang:
Please do not merge this before https://github.com/openshift/verification-tests/pull/395 being merged.

features/logging/4.x/deploy_logging_operators_via_console.feature
```diff
@@ -0,0 +1,89 @@
...
+  Scenario: Deploy cluster-logging operator via web console.
+    # clear the logging related resources
+    Given logging service is removed successfully
+    Given I switch to cluster admin pseudo user
+    Given logging channel name is stored in the :logging_channel clipboard
+    Given I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/clusterlogging/deploy_clo_via_olm/01_clo_ns.yaml |
+    Then the step should succeed
+    Given I switch to the first user
+    And cluster role "cluster-admin" is added to the "first" user
```
akostadinov: Preferably use `Given the first user is cluster-admin`.

lib/rules/web/admin_console/4.2/operators.xyaml
```diff
@@ -0,0 +1,54 @@
...
+  - selector:
+      xpath: //a[@class="pf-c-button pf-m-primary co-catalog-page__overlay-create" and contains(., "Install")]
+    op: click
+
+create_operator_subscription:
+  action:
+    if_element:
+      selector:
+        xpath: //button[@id="dropdown-selectbox" and @class="pf-c-dropdown__toggle"]
+    ref: select_target_namespace
```
yapei: I see `<project_name>` is passed in `goto_operatorhub_subscribe_page` so I think we don't need `select_target_namespace` any more

QiaolingTang: Yes, I'll update the rules. Thanks.

features/logging/4.x/deploy_logging_operators_via_console.feature
```diff
@@ -0,0 +1,76 @@
+Feature: Deploy logging operators via console
```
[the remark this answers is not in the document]
QiaolingTang: OK, I'll update all my feature files later. Thanks, @liangxia .

QiaolingTang:
@anpingli @yapei  I updated my PR, could you please help to review it again? Thanks.

features/logging/deploy_logging_operators_via_console.feature
```diff
@@ -0,0 +1,77 @@
+Feature: Deploy logging operators via console
+
+  # @auther qitang@redhat.com
+  # @case_id OCP-22558
+  @admin
+  @destructive
+  Scenario: Deploy cluster-logging operator via web console.
+    # clear the logging related resources
+    Given logging service is removed successfully
+    Given I switch to cluster admin pseudo user
```
akostadinov: switching to admin user not needed. Just do `I run the :create admin command with:` below. And in this way you will not have to use step `I switch to the first user` below.

features/logging/deploy_logging_operators_via_console.feature
```diff
@@ -0,0 +1,77 @@
...
+    When I perform the :goto_operator_subscription_page web action with:
+      | package_name     | cluster-logging     |
+      | catalog_name     | <%= cb.clo_opsrc %> |
+      | target_namespace | openshift-logging   |
+    Then the step should succeed
+    Given I register clean-up steps:
+    """
+    Given logging service is removed successfully
+    Then the step should succeed
+    """
```
akostadinov: Looks like clean-up step is not immediately before/after step creating the thing to be removed. This makes possible to miss clean-up if previous steps failed.

features/logging/deploy_logging_operators_via_console.feature
```diff
@@ -0,0 +1,77 @@
...
+    When I run the :delete client command with:
+      | object_type       | sub                    |
+      | object_name_or_id | elasticsearch-operator |
+      | namespace         | openshift-operators    |
+    Then the step should succeed
+    When I run the :delete client command with:
+      | object_type       | csv                 |
+      | object_name_or_id | <%= cb.eo_csv %>    |
+      | namespace         | openshift-operators |
+    """
```
akostadinov: if you want to delete something, better use step:
```
/^(I|admin) ensures? #{QUOTED} (\w+) is deleted(?: from the#{OPT_QUOTED} project)?( after scenario)?$/
```

Also clean-up steps need to be registered imediately before or after resource to be removed was created.

features/logging/deploy_logging_operators_via_console.feature
```diff
@@ -0,0 +1,65 @@
...
+    When I perform the :set_custom_channel_and_subscribe web action with:
+      | update_channel    | <%= cb.logging_channel %> |
+      | install_mode      | AllNamespace              |
+      | approval_strategy | Automatic                 |
+    Then the step should succeed
+
+    Given I use the "openshift-operators" project
+    Given I wait for the "elasticsearch-operator" subscriptions to appear
+    And evaluation of `subscription("elasticsearch-operator").current_csv` is stored in the :eo_csv clipboard
+    Given admin ensures "<%= cb.eo_csv %>" clusterserviceversions is deleted from the "openshift-operators" project after scenario
```
akostadinov: shouldn't this be `cluster_service_version`?

And should we register this clean-up step immediately before/after object was created?

QiaolingTang: it should be `clusterserviceversions`  https://github.com/openshift/verification-tests/blob/master/lib/openshift/cluster_service_version.rb#L5

I have to get the name of the csv object before I delete it. If the subscription is not created successfully, the csv object would not be created.

akostadinov: I believe `cluster_service_version` should work. These are registered as aliases - the readable name and the resource API name.

I'll update and merge. Let me know if you hit troubles with that. Thank you.

QiaolingTang: OK, thank you very much.

PR #392, opened 2019-10-17
rules for OCP-23044

SSshahan:
@yapei  please help to review, thanks


lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
```diff
@@ -29,6 +29,11 @@ click_button:
     selector:
       xpath: //button[contains(.,'<button_text>')]
     op: click
+click_button_text:
```
yapei: What's the purpose of adding `//button[text()='<button_text>']`?  What's the limitation of `//button[contains(.,'<button_text>')]`

lib/rules/web/admin_console/4.2/node.xyaml
```diff
@@ -0,0 +1,91 @@
...
+    selector:
+      <<: *dropdown
+    op: click
+  action: click_button_text
+add_taint:
+  elements:
+  - selector:
+      <<: *key
+    type: input
+    op: send_keys <affinity_key>
```
yapei: Why not combining these lines with 37 ~ 40 into one action? 
```
set_affinity_key:
  element: 
    selector: &key
      xpath: //div[text()='Key']/following-sibling::input
    type: input
    op: send_keys <affinity_key>
```

yapei: Then `add_taint` can be updated as
```
add_taint:
  action: set_affinity_key
  action: set_affinity_value
  action: set_affinity_effect

Comment (yapei) [reply lib/rules/web/admin_console/4.2/node.xyaml]: And `add_tolerations` can be 
```
add_tolerations:
  action: set_affinity_key
  action: set_affinity_operator
  action: set_affinity_value
  action: set_affinity_effect
```

File: lib/rules/web/admin_console/4.2/node.xyaml [hunk 3]
```diff
@@ -0,0 +1,91 @@
...
+  action: submit_changes
+node_affinity_modal_value:
+  element: 
+    selector: &value
+      xpath: //div[text()='Value']/following-sibling::input
+node_affinity_modal_key:
+  element: 
+    selector: &key
+      xpath: //div[text()='Key']/following-sibling::input
+node_affinity_modal_dropdown:
```
yapei: Is this action used?

yapei: OK I see, this action is only used to define an anchor. Then I have the same concern/comments as below, it can be updated as `cick_node_affinity_modal_dropdown` since it's always called with a `click` action

lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
SSshahan: The later would find more than one elements when trying to locate substring in one page, for instance, "NoSchedule" is a substring of `PreferNoSchedule`. However I would get two elements when I locate `NoSchedule`

yapei:
LGTM, thanks

SSshahan:
@akostadinov  @pruan-rht  please help to merge. Thanks

PR #393, opened 2019-10-17
auto case OCP-25365 OCP-18147

QiaolingTang:
@anpingli  PTAL, thanks.

Polarion cases: https://url.corp.redhat.com/5f6efcf

### DIFF
--- a/features/logging/collector.feature
+++ b/features/logging/collector.feature
@@ -2,6 +2,43 @@
 Feature: collector related tests
 
   # @auther qitang@redhat.com
+  # @case_id OCP-24837
+  @admin
+  @destructive
+  @commonlogging
+  Scenario: All nodes logs had sent logs to Elasticsearch
+    Given I switch to cluster admin pseudo user
+    Given I use the "openshift-logging" project
+    # get node ip
+    #Given <%= daemon_set("fluentd").replica_counters[:desired] %> pods become ready with labels:
+    #  | component=fluentd |
+    And evaluation of `@pods.map {|n| n.node_ip}.uniq` is stored in the :node_ips clipboard
+    #Given I store the schedulable nodes in the :nodes clipboard
+    #And evaluation of `cb.nodes.map {|n| n.name}` is stored in the :node_name clipboard
+
+    And I wait for the ".operations" index to appear in the ES pod with labels "es-node-master=true"
+    Given I get the ".operations" logging index information from a pod with labels "es-node-master=true"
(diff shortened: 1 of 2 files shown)


features/logging/4.x/collector.feature
```diff
@@ -0,0 +1,62 @@
...
+    Given evaluation of `pod` is stored in the :log_pod clipboard
+    Given I switch to cluster admin pseudo user
+    Given I use the "openshift-logging" project
+    Given evaluation of `cluster_logging('instance').collection_type` is stored in the :collection_type clipboard
+    And I wait up to 600 seconds for the steps to pass:
+    """
+    And I get the "project.<%= cb.proj.name %>" logging index information from a pod with labels "es-node-master=true"
+    Then the step should succeed
+    And the expression should be true> cb.index_data and cb.index_data.count > 0
+    """
```
akostadinov: @pruan-rht , I see this check two times already. As if it waits for logging data to be present. Do we have a special step?

@QiaolingTang , cold you confirm purpose of this waiting?

QiaolingTang: @akostadinov the ES may need some time to create the index and sometimes the collector needs time to collect the logs in projects, these condition always happen in a shared cluster due to limited resources, so I added the step to wait for the data to appear in the ES.

pruan-rht: @QiaolingTang to be consistent, let's assume we are always do this check so it's better to wrap it in a step as suggested by @akostadinov

akostadinov: @pruan-rht , so step like `Given log entries appear for the project`?

QiaolingTang: @pruan-rht @akostadinov seems we have a step `When /^I wait(?: (\d+) seconds)? for the #{QUOTED} index to appear in the ES pod(?: with labels #{QUOTED})?$/ do |seconds, index_name, pod_labels|`, I'll try this step.

akostadinov: Awesome, good catch!

akostadinov:
Conflict :/ , please resolve conflicts and will merge.

features/logging/4.x/collector.feature
pruan-rht: @QiaolingTang yeah, that was written a while back.

pruan-rht:
@QiaolingTang can you please resolve the conflicts?  Thx!

QiaolingTang:
Conflicts are resolved, thank you all.

PR #395, opened 2019-10-18
add new steps to get `opsrc` name

QiaolingTang:
@anpingli @akostadinov @pruan-rht  PTAL, thanks.

I'm adding a new environment variable `subscription_opsrc_name`, I add this because when running logging test, the opsrc we used to create subscription can be changed due to a lot of reasons.
And this variable could be used for all the testing related to creating subscriptions via OLM. I set the default value to `redhat-operators`.

Here is an example:
```
export BUSHSLICER_CONFIG='                                                        
environments:
  ocp4:
    version: "4.3"
    subscription_opsrc_name: "redhat-operators-art"
'
```
Log: https://url.corp.redhat.com/d7f0207

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -229,18 +229,6 @@ module BushSlicer
       return @system_docker_repo
     end
 
-    # the opsrc used to create subscription, default to "redhat-operators"
-    def subscription_opsrc_name
-      unless @subscription_opsrc_name
-        if opts[:subscription_opsrc_name]
-          @subscription_opsrc_name = opts[:subscription_opsrc_name]
-        else
-          @subscription_opsrc_name = "redhat-operators"
-        end
-      end
-      return @subscription_opsrc_name
-    end
-
     # helper parser
     def parse_version(ver_str)
       ver = ver_str.sub(/^v/,"")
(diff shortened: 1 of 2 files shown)


lib/environment.rb
```diff
@@ -229,6 +229,18 @@ def system_docker_repo
       return @system_docker_repo
     end
 
+    # the opsrc used to create subscription, default to "redhat-operators"
+    def subscription_opsrc_name
+      unless @subscription_opsrc_name
```
akostadinov: What does `opsrc` stand for?

akostadinov:
Can we auto-discover this `opsrc_name`? Some `oc get` command or something? Generally pr looks good. But adding new user options if highly undesirable and is worth spending some time to make automatic if possible.

lib/environment.rb
QiaolingTang: It is `Operator Source`.

QiaolingTang:
> Can we auto-discover this `opsrc_name`? Some `oc get` command or something? Generally pr looks good. But adding new user options if highly undesirable and is worth spending some time to make automatic if possible.

OK, let me try. Thanks a lot for the review.

QiaolingTang:
@akostadinov I updated my script, could you please review it again? Thanks a lot.

akostadinov:
Thank you!

PR #407, opened 2019-10-24
[Blocker] Replace obsoleted oadm_manage_node with oadm_cordon/uncordon_node

liangxia:
Tested with features/admin/pod.feature:58 on a 4.2 Azure cluster, and the scenario itself failed, but cordon node in the step and uncordon node in the clean-up are working fine.

@chengzhang1016 Help review the changes for pod.feature, @zhaozhanqi for node.rb

@akostadinov @pruan-rht

### DIFF
--- a/features/admin/pod.feature
+++ b/features/admin/pod.feature
@@ -69,14 +69,12 @@ Feature: pod related features
     And the pod named "pod-pull-by-tag" status becomes :running
     Given I register clean-up steps:
     """
-    I run the :oadm_manage_node admin command with:
-      | node_name   | <%= cb.pod_node %> |
-      | schedulable | true               |
+    I run the :oadm_uncordon_node admin command with:
+      | node_name | <%= cb.pod_node %> |
     the step should succeed
     """
-    When I run the :oadm_manage_node admin command with:
-      | node_name   | <%= cb.pod_node %> |
-      | schedulable | false              |
+    When I run the :oadm_cordon_node admin command with:
+      | node_name | <%= cb.pod_node %> |
     Then the step should succeed
     When I execute on the pod:
       | bash |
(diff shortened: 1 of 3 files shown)


liangxia:
Actually, the clean-up is using a command which does not exist causing auto execution to quit prematurely.

liangxia:
Adding @chao007 @duanwei33 as storage team is affected.

liangxia:
E.g, Runner-v3/46413
```
[08:02:06] INFO> === After Scenario: Drain a node that has cloud vendor volumes, Examples (#4) ===
[08:02:06] ERROR> Test Execution will finish prematurely.
unknown command oadm_manage_node (RuntimeError)
/home/jenkins/workspace/Runner-v3/lib/rules_command_executor.rb:232:in `build_command_line'
......
```

features/admin/pod.feature
```diff
@@ -69,14 +69,12 @@ Feature: pod related features
     And the pod named "pod-pull-by-tag" status becomes :running
     Given I register clean-up steps:
     """
-    I run the :oadm_manage_node admin command with:
-      | node_name   | <%= cb.pod_node %> |
-      | schedulable | true               |
+    I run the :oadm_uncordon_node admin command with:
+      | node_name | <%= cb.pod_node %> |
```
akostadinov: why not use `node schedulable status should be restored after scenario` step instead of this kind of clean-up?

akostadinov:
I have one suggestion but merging as it is not vital to get things working. Thank you.

/lgtm

features/admin/pod.feature
liangxia: @akostadinov Fixed in https://github.com/openshift/verification-tests/pull/413

PR #408, opened 2019-10-24
python-dnf doesn't exist in Fedora 30

rbbratta:


pruan-rht:
@rbbratta keep this would make it work for older feodra releases, no?

rbbratta:
It is python2-dnf, so it shouldn't be used because python2 is EOL.  This is part of a more general problem, Fedora 30 RPM ansible doesn't work with dnf because it doesn't use python3-dnf for some reason.

rbbratta:
well, you have to force it with `ansible_python_interpreter=/usr/bin/python3`
```
# ansible -i localhost, -c local -m dnf  -a 'name=* state=latest' all
[DEPRECATION WARNING]: Distribution fedora 30 on host localhost should use /usr/bin/python3, but is using /usr/bin/python for backward compatibility with prior Ansible releases. A future Ansible release will default to using the discovered platform python for this host. See https://docs.ansible.com/ansible/2.8/reference_appendices/interpreter_discovery.html for
more information. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
localhost | FAILED! => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python"
    },
    "changed": false,
    "cmd": "dnf install -y python2-dnf",
    "msg": "Could not import the dnf python module using /usr/bin/python (2.7.16 (default, Apr 30 2019, 15:54:43) [GCC 9.0.1 20190312 (Red Hat 9.0.1-0.10)]). Please install `python2-dnf` package or ensure you have specified the correct ansible_python_interpreter.",
    "rc": 1,
```

```
# ansible -i localhost, -c local -m dnf  -a 'name=* state=latest' all -e ansible_python_interpreter=/usr/bin/python3
localhost | SUCCESS => {
    "changed": false,
    "msg": "Nothing to do",
    "rc": 0,
    "results": []
}
```
```

akostadinov:
I think we can merge. I don't think we care about older Fedora versions. @pruan-rht , do you envision any problems?

pruan-rht:
@akostadinov if we don't care about Fedroa versions, then it's ok to merge this.

PR #412, opened 2019-10-25
use OCDebugAccessibleHost in OCP v4.x

akostadinov:


wjiangjay:
if we want to execute commands like before in a node, 
Take check kubelet service status as a example.
`oc debug nodes/<node_name> -- chroot /host systemctl status kubelet`

lib/host.rb
```diff
@@ -827,4 +828,73 @@ def clean_up
...
+        # note this will block until timeout if command does not exist remotely
+        # TODO: check debug pod status in the background
+        # note2: exit status is always 0, WRKLDS-98
+        node.env.admin.cli_exec(
+          :debug,
+          resource: "node/#{node.name}",
+          oc_opts_end: "",
+          exec_command_arg: commands,
+          _stdin: opts[:stdin],
+          _stdout: opts[:stdout]
```
wjiangjay: Here need to verify that no spam will come into stdout

wjiangjay: chekced with 
```
➜  ~ oc version 
Client Version: v4.3.0
Server Version: 4.2.1
Kubernetes Version: v1.14.6+868bc38
```

And the prompt is in stderr, so it's ok here.
```
➜  ~ oc debug nodes/ip-10-0-133-47.us-east-2.compute.internal -- chroot /host ls /
Starting pod/ip-10-0-133-47us-east-2computeinternal-debug ...
To use host binaries, run `chroot /host`
bin
boot
dev
etc
home
lib
lib64
media
mnt
opt
ostree
proc
root
run
sbin
srv
sys
sysroot
tmp
usr
var

Removing debug pod ...
➜  ~ oc debug nodes/ip-10-0-133-47.us-east-2.compute.internal -- chroot /host ls / 2>/dev/null
bin
boot
dev
etc
home
lib
lib64
media
mnt
opt
ostree
proc
root
run
sbin
srv
sys
sysroot
tmp
usr
var
```

akostadinov: So no concerns?

akostadinov:
wrt `chroot` question - I don't know. Up to team opinion.

lib/host.rb
wjiangjay: yes

wjiangjay:
others /LGTM

akostadinov:
I have changed execution to run debug pods in a unique project for the test executor. This should avoid inter-build interference. I have also tested steps `environment has at least 2 schedulable nodes` and `I select a random node's host`.

All works fine. 

Given lack of negative feedback and inability of old code to work properly I'm going to merge. As well I will remove detailed host specification using `oc` during install so that proxy support stops to be a problem.

wrt usnig `chroot` by default, let me know, I can add tomorrow or in the next few days.

PR #413, opened 2019-10-28
Replace manually registered clean-up with unified clean-up

liangxia:
@akostadinov @pruan-rht  Follow up for comments in #407  
https://github.com/openshift/verification-tests/pull/407/files#r339362339


akostadinov:
/lgtm

Thank you, great work!

PR #415, opened 2019-10-28
change default IDP to the last one

yapei:
@akostadinov  When defaulting IDP to the first one, it's always `kube:admin` and always failed (Found in CI).  So updated to use the last one, also the changes applies to our CI (which the only and last option is `flexy-htpasswd-provider`)


yapei:
Verified the login changes in https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/46606/console and it succeeded. Can you help merge if no objections then admin console cases are be un-blocked?

PR #416, opened 2019-10-28
Auto console notification case OCP-24316

xiaocwan:
@yapei Plz review. Thanks!

### DIFF
--- a/features/step_definitions/resources.rb
+++ b/features/step_definitions/resources.rb
@@ -125,7 +125,7 @@ end
 
 # tries to delete resource if it exists and make sure it disappears
 # example: I ensure "hello-openshift" pod is deleted
-Given /^(I|admin) ensures? #{QUOTED} (.+) is deleted(?: from the#{OPT_QUOTED} project)?( after scenario)?$/ do |by, name, type, project_name, after|
+Given /^(I|admin) ensures? #{QUOTED} (\w+) is deleted(?: from the#{OPT_QUOTED} project)?( after scenario)?$/ do |by, name, type, project_name, after|
   _user = by == "admin" ? admin : user
   _resource = resource(name, type, project_name: project_name)
   _seconds = 300
--- a/features/step_definitions/resources.rb
+++ b/features/step_definitions/resources.rb
@@ -125,7 +125,7 @@ end
 
 # tries to delete resource if it exists and make sure it disappears
 # example: I ensure "hello-openshift" pod is deleted
-Given /^(I|admin) ensures? #{QUOTED} (\w+) is deleted(?: from the#{OPT_QUOTED} project)?( after scenario)?$/ do |by, name, type, project_name, after|
+Given /^(I|admin) ensures? #{QUOTED} (.+) is deleted(?: from the#{OPT_QUOTED} project)?( after scenario)?$/ do |by, name, type, project_name, after|
   _user = by == "admin" ? admin : user
   _resource = resource(name, type, project_name: project_name)
   _seconds = 300
--- a/lib/openshift/console_notification.rb
+++ b/lib/openshift/console_notification.rb
@@ -1,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class ConsoleNotification < ClusterResource
+    RESOURCE = "consolenotifications"
+  end
+end
--- a/lib/openshift/console_notification.rb
+++ b/lib/openshift/console_notification.rb
@@ -1,7 +1,0 @@
-require 'openshift/cluster_resource'
-
-module BushSlicer
-  class ConsoleNotification < ClusterResource
-    RESOURCE = "consolenotification.console.openshift.io"
-  end
-end
--- a/lib/openshift/console_notification.rb
+++ b/lib/openshift/console_notification.rb
@@ -2,6 +2,6 @@ require 'openshift/cluster_resource'
 
 module BushSlicer
   class ConsoleNotification < ClusterResource
-    RESOURCE = "consolenotifications"
+    RESOURCE = "consolenotification.console.openshift.io"
   end
 end
--- a/lib/openshift/console_notification_console_openshift_io.rb
+++ b/lib/openshift/console_notification_console_openshift_io.rb
@@ -2,6 +2,6 @@ require 'openshift/cluster_resource'
 
 module BushSlicer
   class ConsoleNotificationConsoleOpenshiftIo < ClusterResource
-    RESOURCE = "console.notification.console.openshift.io"
+    RESOURCE = "consolenotification.console.openshift.io"
   end
 end
--- a/lib/openshift/console_notification_console_openshift_io.rb
+++ b/lib/openshift/console_notification_console_openshift_io.rb
@@ -1,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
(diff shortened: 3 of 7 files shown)


lib/openshift/console_notification.rb
```diff
@@ -0,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class ConsoleNotification < ClusterResource
+    RESOURCE = "consolenotifications"
```
akostadinov: we need to use full api name like `lib/openshift/config_imageregistry_operator_openshift_io.rb`

xiaocwan: @akostadinov If update to:
```
class ConsolenotificationConsoleOpenshiftIo < ClusterResource
    RESOURCE = "consolenotifications.console.openshift.io"
```
It will raise undefined step for the step "Admin ensure... is deleted after scenario"

akostadinov: You must have a typo somewhere. dots should not make a difference for that step.

lib/openshift/console_notification.rb
```diff
@@ -0,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class ConsoleNotification < ClusterResource
+    RESOURCE = "consolenotification.console.openshift.io"
```
akostadinov: my guess is this shall be `consolenotifications` (`s` at the end).

xiaocwan:
```
$ oc get consolenotification.console.openshift.io
No resources found.
```
This does work.

xiaocwan: @akostadinov 
```
      [08:40:04] INFO> Shell Commands: oc adm policy add-cluster-role-to-user cluster-admin testuser-45 --config=/home/xiaocwan/workdir/localhost-xiaocwan/ocp4_admin.kubeconfig
      clusterrole.rbac.authorization.k8s.io/cluster-admin added: "testuser-45"
      
      [08:40:43] INFO> Exit Status: 0
    Given admin ensures "notification3" consolenotifications.console.openshift.io is deleted after scenario # features/tierN/web/admin-console/customization.feature:241
      Undefined step: "admin ensures "notification3" consolenotifications.console.openshift.io is deleted after scenario" (Cucumber::Undefined)
      features/tierN/web/admin-console/customization.feature:241:in `Given admin ensures "notification3" consolenotifications.console.openshift.io is deleted after scenario'
```

xiaocwan: @akostadinov I guess there is something wrong with the step definition, `(\w+)` can not match the `consolenotifications.console.openshift.io` the for dot character. 
Suggest to use `(.+)`, it only not to match the white space. Need I update this in here or still use the previous spelling of the resource type? Otherwise it will block our PR or similar scenario.

lib/openshift/console_notification.rb
xiaocwan: @akostadinov Dot does matter for the string match inside step definition, updated, plz check, thanks!
Btw recent steps always failed on Job but could pass locally. (refer to https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/46634/console)

features/step_definitions/resources.rb
```diff
@@ -125,7 +125,7 @@
 
 # tries to delete resource if it exists and make sure it disappears
 # example: I ensure "hello-openshift" pod is deleted
-Given /^(I|admin) ensures? #{QUOTED} (\w+) is deleted(?: from the#{OPT_QUOTED} project)?( after scenario)?$/ do |by, name, type, project_name, after|
+Given /^(I|admin) ensures? #{QUOTED} (.+) is deleted(?: from the#{OPT_QUOTED} project)?( after scenario)?$/ do |by, name, type, project_name, after|
```
xiaocwan: @akostadinov Is that ok for you?

lib/rules/web/admin_console/4.2/customizations.xyaml
```diff
@@ -21,3 +21,15 @@ upload_file:
     visible: false
     type: filefield
     op: set <file_path>
+check_console_notification:
```
yapei: The changes looks good to me

features/step_definitions/resources.rb
akostadinov: Better use `[\w.]+`.

xiaocwan: I refered to clusterservicebroker in line 218 of svc-catalog.feature. But now I looked for other examples, there are cluster_service_broker. Will use underscore in this case.

akostadinov: Could you remove the change then? :)

lib/openshift/consolenotifications_console_openshift_io.rb
```diff
@@ -0,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class ConsolenotificationsConsoleOpenshiftIo < ClusterResource
```
akostadinov: `ConsoleNotificationsConsoleOpenShiftIo`

xiaocwan:
@akostadinov Updated and passed locally, must use `RESOURCE = "consolenotification.console.openshift.io"` cause `consolenotification` is a spelling of resource.

lib/world.rb
```diff
@@ -141,6 +141,7 @@ def #{snake_case}(*args, &block)
     alias scc security_context_constraints
     alias kubeapiserver kube_a_p_i_server
     alias openshiftapiserver open_shift_a_p_i_server
+    alias consolenotifications console_notification_console_openshift_io
```
akostadinov: `console_notifications`

lib/openshift/console_notification_console_openshift_io.rb
```diff
@@ -0,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class ConsoleNotificationConsoleOpenshiftIo < ClusterResource
+    RESOURCE = "consolenotification.console.openshift.io"
```
akostadinov: I can't find any such object in stage cluster but I think it must be `consolenotifications.console.openshift.io` with `s` at the end. This is how I see other APIs to be. If your cluster has such an object, then you can check for sure. What I see is that the `s` works:
```
🐚 oc --config=/tmp/kubeconfig get consolenotifications.console.openshift.io
No resources found.
```

That's why I suggested to put the `s` at the end.

To be sure, can you post here `oc get -o yaml` from such an object?

lib/world.rb
akostadinov: Actually `console_notification`. This alias here is used when we work with such objects in scenarios. e.g.
```
Then the expression should be true> console_notification.whatever == something
```
That's why I suggest to put it singular form here.

akostadinov: I see `openshiftapiserver` and `kubeapiserver` do not follow this convention. I need to go back and change them. Shouldn't be hard to grep them out.

lib/openshift/console_notification_console_openshift_io.rb
yapei: @xiaocwan  Please check new comments

xiaocwan:
@akostadinov Updated according to the last two comments. Please check.

xiaocwan:
@akostadinov @pruan-rht
Please review, thanks!

PR #418, opened 2019-10-29
More waiting time in case slow env as CI flaky failure hits

xingxingxia:
Why fix: https://url.corp.redhat.com/qe-ci-tracker-slow-env-flaky-failure (RH internal)
@yapei @akostadinov please review/merge, thanks!


xingxingxia:
grep -nR "I wait for the steps to pass:" features/ | wc -l
Gets 268 lines, thus fix in step def instead of case by case

akostadinov:
IMO this should be case by case. See #420 

At this point I don't think it is a good idea to increase the default to avoid excessive slowness. And reducing defaults also I don't find good to avoid too many failures.

xingxingxia:
OK, thanks, closing in favor of case by case https://github.com/openshift/cucushift/pull/7144
/close

PR #419, opened 2019-10-29
Add parameter for new-app cmd

xiuwang:
@openshift/devexp-qe Please help review, thx

### DIFF
--- a/features/step_definitions/registry.rb
+++ b/features/step_definitions/registry.rb
@@ -375,7 +375,7 @@ end
 # Generate default route
 Given /^I enable image-registry default route$/ do
   ensure_admin_tagged
-  ensure_destructive_tagged
+#  ensure_destructive_tagged
   step 'I run the :patch admin command with:', table(%{
       | resource      | configs.imageregistry.operator.openshift.io |
       | resource_name | cluster                                     |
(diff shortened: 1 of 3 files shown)


akostadinov:
Could you explain the need for these? I don't see a resource parameter for `new-app` command.

Do we need the `is` method? I find `image_stream` to look much more readable especially when used in code.

xiuwang:
@akostadinov the adding new-app parameter used in https://github.com/openshift/cucushift/pull/7142/files#diff-804ded6b9a7439c255b6dbcf9d8bdad9R434-R435
It's fine if `is` and `image_stream` both work https://github.com/openshift/cucushift/pull/7139/files#diff-878a6aeaa4a0ae5226030bbd4eacf809R115

akostadinov:
for `resource` I find it confusing, because we have many other commands accepting a resource and it is usually of the form `kind/name`. This is very different. In help I see
```
oc new-app (IMAGE | IMAGESTREAM | TEMPLATE | PATH | URL ...)
```
I would say that what you want is `path`. Or if you want to catch all possibilities with one parameter, then perhaps call it `source` or `source_spec` or something better if you have nice ideas.

wrt `is` I'm not strongly against it. Might be somehow confusing to see `is(..)` someehwere in the code where its meaning is not obvious. ALso would be hard to `grep` for it if needed. But maybe not a big deal and if you find it important, I can merge with it.

xiuwang:
@akostadinov I want to catch all possibilities with one parameter, and I'd like the `source_spec` placeholder. Indeed, `is` makes people a little confused, I will remove it.
Please help review it, thanks your advice.

PR #420, opened 2019-10-29
Wait 5 min (instead of 15 min) to speed up auto execution in CI

liangxia:
@akostadinov @pruan-rht @yapei


pruan-rht:
@liangxia I assume this PR is to address this [JIRA](https://projects.engineering.redhat.com/browse/OPENSHIFTQ-12802)?   We need to think carefully about changing the wait time for a particular resource to reach a ceratin desired state.  It's a trade of between stability and speed.

akostadinov:
Well, initially timeouts were shorter. Then gradually all were increased. Now we change back and my guess is we will start hitting a lot of transient failures.

We have always assumed that on a timeout failure, we just increased timeout instead of filing a bug. Thus I'm against such change.

PR #423, opened 2019-10-30
Automated and OCP-25652

jhou1:
Automated new test cases. Tested locally and passed.

### DIFF
--- a/features/machine/machine.feature
+++ b/features/machine/machine.feature
@@ -78,3 +78,21 @@ Feature: Machine features testing
     # node should be deleted
     Given I switch to cluster admin pseudo user
     Then I wait for the resource "node" named "<%= cb.new_node %>" to disappear within 600 seconds
+
+  # @author jhou@redhat.com
+  # @case_id OCP-25652
+  @admin
+  Scenario: MAO metrics is exposed on https
+    Given I switch to cluster admin pseudo user
+    And I use the "openshift-monitoring" project
+    And evaluation of `secret(service_account('prometheus-k8s').get_secret_names.first).token` is stored in the :token clipboard
+
+    When I run the :exec admin command with:
+      | n                | openshift-monitoring                                                                                                         |
+      | pod              | prometheus-k8s-0                                                                                                             |
+      | c                | prometheus                                                                                                                   |
+      | oc_opts_end      |                                                                                                                              |
+      | exec_command     | sh                                                                                                                           |
+      | exec_command_arg | -c                                                                                                                           |
+      | exec_command_arg | curl -v -s -k -H "Authorization: Bearer <%= cb.token %>" https://machine-api-operator.openshift-machine-api.svc:8443/metrics |
+    Then the step should succeed


features/machine/machine.feature
```diff
@@ -78,3 +78,111 @@ Feature: Machine features testing
...
+
+    Given I store the last provisioned machine in the :new_machine clipboard
+
+    # Phase "Provisioning"
+    When I run the :get admin command with:
+      | resource      | machine               |
+      | resource_name | <%= cb.new_machine %> |
+      | n             | openshift-machine-api |
+    Then the output should contain:
+      | Provisioning |
```
pruan-rht: @jhou1 I'm not sure how reliable the test can pass with this check...  From my experience that state comes and goes pretty quickly so it's hard to catch that state.  Perhaps skip this check?

features/machine/machine.feature
```diff
@@ -78,3 +78,111 @@ Feature: Machine features testing
...
+    # Phase "Provisioned"
+    Then I wait up to 300 seconds for the steps to pass:
+    """
+    When I run the :get admin command with:
+      | resource      | machine               |
+      | resource_name | <%= cb.new_machine %> |
+      | n             | openshift-machine-api |
+    Then the output should contain:
+      | Provisioned |
+    """
```
pruan-rht: perhaps add a step to check for the machine status

features/machine/machine.feature
akostadinov: Provisioning might take some time, not sure about `Provisioned`. My suggestion would be to just wait for some time until machine becomes Running and when removing - wait some time to see machine number becomes what we want.

Testing all different statuses a machine would go through IMO falls outside blackbox testing.

Maybe we need steps like:
```
When machine set ... is upscaled with 1 more machine
And I register clean-up steps
"""
When machine set ... is downscaled with 1 machine
Then machine set ... has the expected number of machines and they become running
"""
Then machine set ... has the expected number of machines and they become running
```

This will simplify such scenarios greatly. If we don't need a lot of scaling maybe not worth it. You decide. But I still support Peter, thant checking transient statuses is potentially problematic as we saw with pods and builds in the past. Please avoid if this is not an important feature that we need to test for out customers.

akostadinov: btw we can add a method `#ready?` to the MachineSet class and use this step to check machine is running:
```
/^(I|admin) waits? for the#{OPT_QUOTED} (\w+) to become ready(?: in the#{OPT_QUOTED} project)?(?: up to (\d+) seconds)?$/ 
```

pruan-rht: @akostadinov I can add that step(s) if @jhou1 is busy with other stuff

jhou1: @akostadinov @pruan-rht You are right, this test would probably be unreliable. The way how the machine phase is tested, can not ensure 100% pass ratio, especially when the cluster has pressure and api response is slow. I'll remove it, thanks

jhou1:
@pruan-rht thanks for the framework method. The test runs much much faster!

PR #426, opened 2019-10-31
add rules for operator subscription

yapei:
@QiaolingTang  @jianzhangbjz  @yanpzhan  Please help review the rules added for operator subscription. Let me know if there's anything need updates

### DIFF
--- a/lib/rules/web/admin_console/4.2/catalog.xyaml
+++ b/lib/rules/web/admin_console/4.2/catalog.xyaml
@@ -56,3 +56,35 @@ clear_filter_by_keyword:
   element:
     selector: *filter_by_keyword
     op: clear
+click_on_catalog_item:
+  element:
+    selector:
+      xpath: //a[@data-test='<operator_name>-<catalog_name>-openshift-marketplace']
+click_catalog_overlay_button:
+  element:
+    selector:
+      xpath: //a[contains(@class,'co-catalog-page__overlay-create')]
+    op: click
+choose_installmode:
+  params:
+    radio_text: <install_mode>
+  action: click_radio_input
+choose_update_channel:
+  params:
+    radio_text: <update_channel>
+  action: click_radio_input
+choose_approval_strategy:
+  params:
+    radio_text: <approval_strategy>
+  action: click_radio_input
+click_subscribe_button:
+  params:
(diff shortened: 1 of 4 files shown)


lib/rules/web/admin_console/4.2/catalog.xyaml
```diff
@@ -56,3 +56,40 @@ clear_filter_by_keyword:
   element:
     selector: *filter_by_keyword
     op: clear
+click_on_catalog_item:
+  element:
+    selector:
+      xpath: //a[@data-test='<operator_name>-redhat-operators-openshift-marketplace']
```
QiaolingTang: @yapei  I think it's better to use `<opsrc_name>` instead of `redhat-operators` because in our testing, we always use the operator source created by ourself not the default one.

lib/rules/web/admin_console/4.2/catalog.xyaml
```diff
@@ -56,3 +56,40 @@ clear_filter_by_keyword:
...
+  params:
+    radio_text: <approval_strategy>
+  action: click_radio_input
+click_subscribe_button:
+  params:
+    button_text: Subscribe
+  action: click_button_text
+set_default_params_for_operator_subscription:
+  params:
+    update_channel: '4.2'
```
QiaolingTang: The channel name should be a parameter because different operators have different channel name.

yapei: Yeah, this action is used for default params, the other action `set_customized_params_for_operator_subscription` is used to use customized params. I just wonder maybe someone need the default setting

lib/rules/web/admin_console/4.2/catalog.xyaml
yapei: good point, will update

lib/rules/web/admin_console/4.2/catalog.xyaml
QiaolingTang: Got it, but we still need a parameter to process the channel name.

QiaolingTang:
lgtm, thank you very much @yapei .

jianzhangbjz:
@yapei Thanks!

PR #427, opened 2019-10-31
multus automation

anuragthehatter:
Keeping the mutlus automation rolling while more on the way
Covered here are OCP-21949,OCP-21793 and OCP-21949

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -546,12 +546,14 @@ Feature: Multus-CNI related scenarios
     Then the step should fail
     And the output should contain:
       | admission webhook "multus-validating-config.k8s.io" denied the request|
+    And admin ensures "macvlan-bridge-21756" network_attachment_definition is deleted from the "default" project after scenario
     When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml" replacing paths:
       | ["metadata"]["name"] | macvlan-bridge@$ |
     Then the step should fail
     And the output should contain:
       |subdomain must consist of lower case alphanumeric characters|
-
+    And admin ensures "macvlan-bridge@$" network_attachment_definition is deleted from the "default" project after scenario
+    
   # @author anusaxen@redhat.com
   # @case_id OCP-21949
   @admin
@@ -563,14 +565,13 @@ Feature: Multus-CNI related scenarios
     When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-bridge.yaml" replacing paths:
       | ["metadata"]["name"] | macvlan-bridge-21456 |
     Then the step should succeed 
-    And admin ensures "macvlan-bridge-21456" network_attachment_definition is deleted from the "default" project after scenario
     # Create a pod consuming net-attach-def simulating wrong syntax in name
     When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-bridge.yaml" replacing paths:
       | ["metadata"]["generateName"] | macvlan-bridge-pod-$@ |
     Then the step should fail
     And the output should contain:
       | subdomain must consist of lower case alphanumeric characters |
-
+    
   # @author anusaxen@redhat.com
   # @case_id OCP-21793
   @admin
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -485,18 +485,12 @@ Feature: Multus-CNI related scenarios
     # Creating pod in the user's namespace which consumes the net-attach-def created in default namespace 
     Given I switch to the first user
     And I create a new project
-    
+    And evaluation of `project.name` is stored in the :project_name clipboard
+
     When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-bridge.yaml" replacing paths:
       | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | default/macvlan-bridge-pod |
     Then the step should succeed
-    And evaluation of `project.name` is stored in the :project_name clipboard
-    When I run the :get client command with:
-      | resource | pod                    |
-      | output   | json                   |
-      | n        | <%= cb.project_name %> |
-    Then the step should succeed
-    And  evaluation of `@result[:parsed]['items'][0]['metadata']['name']` is stored in the :pod_name clipboard
-
+    And evaluation of `@result[:response].match(/pod\/(.*) created/)[1]` is stored in the :pod_name clipboard
     And I wait up to 30 seconds for the steps to pass:
     """
     When I run the :describe client command with:
@@ -507,4 +501,3 @@ Feature: Multus-CNI related scenarios
     And the output should contain:
       | namespace isolation violation |
     """
-
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -470,14 +470,14 @@ Feature: Multus-CNI related scenarios
     # Create the net-attach-def via cluster admin
     Given I switch to cluster admin pseudo user
     And I use the "default" project
-    When I run the :create admin command with:
+    When I run the :create client command with:
 	    | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml |
     Then the step should succeed
     Given I register clean-up steps:
     """
     Given I switch to cluster admin pseudo user
     And I use the "default" project
-    When I run the :delete admin command with:
+    When I run the :delete client command with:
             | object_type       | net-attach-def |
             | object_name_or_id | macvlan-bridge |
     Then the step should succeed
@@ -490,7 +490,7 @@ Feature: Multus-CNI related scenarios
             | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | default/macvlan-bridge-pod |
     Then the step should succeed
     And evaluation of `project.name` is stored in the :project_name clipboard
-    When I run the :get admin command with:
+    When I run the :get client command with:
 	    | resource | pod                |
 	    | output   | json               |
 	    | n        | <%= cb.project_name %> |
@@ -499,7 +499,7 @@ Feature: Multus-CNI related scenarios
 
     And I wait up to 30 seconds for the steps to pass:
     """
-    When I run the :describe admin command with:
+    When I run the :describe client command with:
 	    | resource | pod                    |
 	    | name     | <%= cb.pod_name %>     |
        	    | n        | <%= cb.project_name %> |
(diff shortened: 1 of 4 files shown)


features/networking/multus.feature
```diff
@@ -470,14 +470,14 @@ Feature: Multus-CNI related scenarios
...
-    When I run the :create admin command with:
+    When I run the :create client command with:
 	    | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml |
     Then the step should succeed
     Given I register clean-up steps:
     """
     Given I switch to cluster admin pseudo user
     And I use the "default" project
-    When I run the :delete admin command with:
+    When I run the :delete client command with:
```
pruan-rht: @anuragthehatter, indentation should be 2, not 4.

anuragthehatter: Thanks @pruan-rht i knew that but somehow missed it. Also i need to work on my auto indent vim feature

pruan-rht: https://github.com/openshift/verification-tests/blob/master/doc/vim.adoc

anuragthehatter:
> @anuragthehatter technically, you can save one query and replace the `get` lines 493-498 by inserting the following after line 491 with `And evaluation of` `@result[:response].match(/pod\/(.*) created/)[1] is stored in the :pod_name clipboard`. The caveat is that if the output changes, your test will fail... not sure it's worth the saving for the sake of stability... @akostadinov WDYT?

@pruan-rht that's seems like a great suggestion. I expect pod to be created always. Your steps simply will check pod created or not else will fail which is acceptable

features/networking/multus.feature
```diff
@@ -416,4 +416,85 @@ Feature: Multus-CNI related scenarios
...
+    Then the step should succeed
+    Given I register clean-up steps:
+    """
+    Given I switch to cluster admin pseudo user
+    And I use the "default" project
+    When I run the :delete client command with:
+      | object_type       | net-attach-def |
+      | object_name_or_id | macvlan-bridge |
+    Then the step should succeed
+    """
```
akostadinov: you can replace all clean-up with step:
```
/^(I|admin) ensures? #{QUOTED} (\w+) is deleted(?: from the#{OPT_QUOTED} project)?( after scenario)?$/
```

When you set `after scenario` that will be registered as a clean-up step.

akostadinov: maybe you'll need a simple class for `net-attach-def` following existing convention in `lib/openshift`. I think this is `ClusterResource`. You only need to set `RESOURCE` in the class.

anuragthehatter: Thanks @akostadinov PTAL. Tested and works fine now.
"    And admin ensures "macvlan-bridge" networkattachmentdefinition is deleted from the "default" project after scenario"

anuragthehatter:
Hi @pruan-rht @akostadinov I would appreciate if we merge it if all comments looks good to leverage this in current release's tight scedule

lib/openshift/networkattachmentdefinition.rb
```diff
@@ -0,0 +1,9 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  # represnets an Openshift Namespace
+  class Networkattachmentdefinition < ClusterResource
+    RESOURCE = 'networkattachmentdefinition.k8s.cni.cncf.io'
```
akostadinov: could you paste `-o yaml` output of such object? I think it might be `networkattachmentdefinitions.k8s.cni.cncf.io`

lib/openshift/networkattachmentdefinition.rb
```diff
@@ -0,0 +1,9 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  # represnets an Openshift Namespace
+  class Networkattachmentdefinition < ClusterResource
```
akostadinov: Could you make Class name `NetworkAttachmentDefinition` and file name in snake case?

features/networking/multus.feature
```diff
@@ -416,4 +416,75 @@ Feature: Multus-CNI related scenarios
...
+  Scenario: User cannot consume the net-attach-def created in other project which is namespace isolated	
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+    # Create the net-attach-def via cluster admin
+    Given I switch to cluster admin pseudo user
+    And I use the "default" project
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml |
+    Then the step should succeed
+    And admin ensures "macvlan-bridge" networkattachmentdefinition is deleted from the "default" project after scenario
```
akostadinov: with changes described below this would be `network_attachment_definition`

lib/openshift/networkattachmentdefinition.rb
anuragthehatter: apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  creationTimestamp: "2019-11-12T00:20:43Z"
  generation: 1
  name: macvlan-bridge
  namespace: test
  resourceVersion: "190189"
  selfLink: /apis/k8s.cni.cncf.io/v1/namespaces/test/network-attachment-definitions/macvlan-bridge
  uid: fd2946d2-3608-4fa1-b174-52cd944ea057
spec:
  config: '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "eth0", "mode": "bridge",
    "ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100",
    "rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1"
    } }'

lib/openshift/networkattachmentdefinition.rb
anuragthehatter: For sure

anuragthehatter: Done. thanks

features/networking/multus.feature
anuragthehatter: Done. thanks

lib/openshift/network_attachment_definition.rb
```diff
@@ -0,0 +1,9 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  # represents an Object
+  class NetworkAttachmentDefinition < ClusterResource
+    RESOURCE = 'networkattachmentdefinition.k8s.cni.cncf.io'
```
akostadinov: I'd guess that there should be an `s` in `networkattachmentdefinitions`. Could you paste an `oc get -o yaml` example to be sure?

akostadinov: Actually I see now your  example output. It seems to be in fact `network-attachment-definitions`.

akostadinov: under `selfLink`

anuragthehatter: yep there is 's' but it works both ways 

apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
creationTimestamp: "2019-11-12T00:20:43Z"
generation: 1
name: macvlan-bridge
namespace: test
resourceVersion: "190189"
selfLink: /apis/k8s.cni.cncf.io/v1/namespaces/test/network-attachment-definitions/macvlan-bridge
uid: fd2946d2-3608-4fa1-b174-52cd944ea057
spec:
config: '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "eth0", "mode": "bridge",
"ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100",
"rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1"
} }'

lib/openshift/networkattachmentdefinition.rb
anuragthehatter: nope..no 's'
$ oc get networkattachmentdefinitions.k8s.cni.cncf.io
error: the server doesn't have a resource type "networkattachmentdefinitions"

features/networking/multus.feature
```diff
@@ -416,4 +416,75 @@ Feature: Multus-CNI related scenarios
...
+  Scenario: User cannot consume the net-attach-def created in other project which is namespace isolated	
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+    # Create the net-attach-def via cluster admin
+    Given I switch to cluster admin pseudo user
+    And I use the "default" project
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml |
+    Then the step should succeed
+    And admin ensures "macvlan-bridge" network_attachment_definition is deleted from the "default" project after scenario
```
akostadinov: my suggestion is to use a name like `macvlan-bridge-21793` or similar to avoid interference with other test scenarios using same name. Tests run in parallel. For creating you could also use step 
```
/^I run oc create( as admin)? (?:over|with) #{QUOTED} replacing paths:$/
```
It allows you to update some part of object before it is created. See existing usage of step for an example.

akostadinov:
Just two more simple things, thank you!

features/networking/multus.feature
anuragthehatter: Yep i am aware of  that step but but not sure if this would interfere with other as this is only case in default project and rest other scenario will create it in unique projects. Even if another default one lands in future, they would run sequentially not parallel and would cleanup before exit. Correct if i am wrong :)

lib/openshift/networkattachmentdefinition.rb
akostadinov: The thing we are looking at here is the `selfLink` where we actually see the surprising `network-attachment-definitions`. I see dashes for the first time.

features/networking/multus.feature
anuragthehatter: Taken care of it

anuragthehatter:
> Just two more simple things, thank you!

Took care of it. Thanks

anuragthehatter:
@pruan-rht Can you help merge it? Thanks. All comments have been taken care of. This will unblock  #506

features/networking/multus.feature
```diff
@@ -531,3 +531,78 @@ Feature: Multus-CNI related scenarios
...
+    
+  # @author anusaxen@redhat.com
+  # @case_id OCP-21793
+  @admin
+  Scenario: User cannot consume the net-attach-def created in other project which is namespace isolated	
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+    # Create the net-attach-def via cluster admin
+    Given I switch to cluster admin pseudo user
+    And I use the "default" project
```
pruan-rht: @anuragthehatter do you have to use the `default` system project?

anuragthehatter: @pruan-rht Yes this is multus feature. only a net-attach-def created under default project can be used by other projects if flag is enabled. So here we want to makes sure, with flag being disabled, the net-attach-def from default project can't be used in other project and when we try to use that e get violation error

pruan-rht: @anuragthehatter thanks for the explaination...LGTM.

PR #402, opened 2019-10-22
auto ocp21333

QiaolingTang:
@anpingli  PTAL, thanks.

Polarion cases: https://url.corp.redhat.com/bfc0843

Log: https://url.corp.redhat.com/3b95aa4

PR #432, opened 2019-11-03
More multus cases

anuragthehatter:
Some more multus cases automated 
OCP-24467,OCP-24489 and OCP-24488
@zhaozhanqi @weliang1 @huiran0826 @rbbratta

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -431,7 +431,7 @@ Feature: Multus-CNI related scenarios
       | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/bridge-host-local-novlan.yaml |
       | n | <%= cb.project_name %>                                                                                                                              |
     Then the step should succeed
-    #Creating no-vlan pod abosrbing above net-attach-def
+    #Creating no-vlan pod absorbing above net-attach-def
     When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
       | ["metadata"]["name"] | pod-novlan |
       | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | bridge3    |
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -416,4 +416,119 @@ Feature: Multus-CNI related scenarios
     Then the output should contain "net1"
     And the output should contain "net2"
     And the output should contain 2 times:
-      | macvlan mode bridge |
+	    | macvlan mode bridge |
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-24488
+  @admin
+  Scenario: Create pod with Multus bridge CNI plugin without vlan
+    # Make sure that the multus is enabled
+    Given the master version >= "4.0"
+    And the multus is enabled on the cluster
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    And evaluation of `project.name` is stored in the :project_name clipboard
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/bridge-host-local-novlan.yaml |
+      | n | <%= cb.project_name %>                                                                                               |
+    Then the step should succeed
+    #Creating no-vlan pod abosrbing above net-attach-def
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/pod-bridge-host-local-novlan.yaml |
+      | n | <%= cb.project_name %>                                                                                                   |
+    Then the step should succeed
+    And the pod named "pod-novlan" becomes ready
+    And evaluation of `pod` is stored in the :pod clipboard
+    When I execute on the pod:
+      | /usr/sbin/ip | -d | link |
+    Then the output should contain "net1"
+    #Entering into corresponding no eot make sure No VLAN ID information shown for secondary interface
+    When I run the :debug admin command with:
+      | resource         | node/<%= cb.pod.node_name %>  |
+      | t                |                               |
+      | oc_opts_end      |                               |
+      | exec_command     | chroot    		         |
+      | exec_command_arg | /host 		         |
+      | exec_command_arg | bridge 		         |
+      | exec_command_arg | vlan 		         |
+      | exec_command_arg | show 		         |
+    Then the step should succeed
+    And the output should contain 2 times:
+      | 1 PVID untagged |
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-24489
+  @admin
+  Scenario: Create pod with Multus bridge CNI plugin and vlan tag
+    # Make sure that the multus is enabled
+    Given the master version >= "4.0"
+    And the multus is enabled on the cluster
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    And evaluation of `project.name` is stored in the :project_name clipboard
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/bridge-host-local-vlan200.yaml |
+      | n | <%= cb.project_name %>                                                                                                |
+    Then the step should succeed
+    #Creating vlan pod abosrbing above net-attach-def
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/pod1-bridge-host-local-vlan200.yaml |
+      | n | <%= cb.project_name %>                                                                                                     |
+    Then the step should succeed
+    And the pod named "pod1-vlan200" becomes ready
+    And evaluation of `pod` is stored in the :pod clipboard
+    When I execute on the pod:
+      | /usr/sbin/ip | -d | link |
+    Then the output should contain: 
+      | net1 |
+    #Entering into corresponding no eot make sure VLAN ID information shown is for secondary interface
+    When I run the :debug admin command with:
+      | resource         | node/<%= cb.pod.node_name %>  |
+      | t                |                               |
+      | oc_opts_end      |                               |
+      | exec_command     | chroot                        |
+      | exec_command_arg | /host                         |
+      | exec_command_arg | bridge                        |
+      | exec_command_arg | vlan                          |
+      | exec_command_arg | show                          |
+    Then the step should succeed
+    And the output should contain 2 times:
+      | 200 |
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-24467
+  @admin
+  Scenario: CNO manager mavlan configured manually with static
+    # Make sure that the multus is Running
+    Given the master version >= "4.0"
+    And the multus is enabled on the cluster
+    Given the default interface on nodes is stored in the :default_interface clipboard 
+    #Patching simplemacvlan config in network operator config CRD
+    Given as admin I successfully merge patch resource "networks.config.openshift.io/cluster" with:
+      | {"spec":{"additionalNetworks":[{"name": "test-macvlan-case3","namespace": "openshift-multus","simpleMacvlanConfig": {"ipamConfig": {"staticIPAMConfig": {"addresses": [{"address": "10.128.2.100/23","gateway": "10.128.2.1"}]},"type": "static"},"master": "<%= cb.default_interface %>","mode": "bridge"},"type": "SimpleMacvlan"}]}} |
+    #Cleanup for bringing CRD to original
+    Given I register clean-up steps:
+    """
+    Given as admin I successfully merge patch resource "networks.config.openshift.io/cluster" with: 
+    | {"spec":{"additionalNetworks": null}} |
+    """
+ 
+    And I wait up to 30 seconds for the steps to pass:
+    """
+    When I run the :get client command with:
+      | object_type | net-attach-def   |
+      | n           | openshift-multus |
+    Then the step should succeed
+    And the output should contain:
+      | test-macvlan-case3 |
+    """
+    #Creating pod under openshift-multus project to absorb above net-attach-def
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-bridge.yaml" replacing paths:
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | test-macvlan-case3 |
+      | n | openshift-multus |
+    Then the step should succeed
+    And the pod named "macvlan-bridge-pod" becomes ready
+    When I execute on the pod:
+      | /usr/sbin/ip | -d | link |
+    Then the output should contain "net1"
+
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -439,13 +439,19 @@ Feature: Multus-CNI related scenarios
     Then the step should succeed
     And the pod named "pod-novlan" becomes ready
     And evaluation of `pod` is stored in the :pod clipboard
+    
+    #Clean-up required to erase bridge interfcaes created on node after this step
+    Given I register clean-up steps:
+    """
+    the bridge interface named "bridge3" is deleted from the "<%= cb.pod.node_name %>" node
+    """
     When I execute on the pod:
       | /usr/sbin/ip | -d | link |
     Then the output should contain "net1"
     #Entering into corresponding no eot make sure No VLAN ID information shown for secondary interface
     Given CNI vlan info is obtained on the "<%= cb.pod.node_name %>" node
     Then the step should succeed
-    And the output should contain:
+    And the output should contain 2 times:
       | 1 PVID untagged |
 
   # @author anusaxen@redhat.com
@@ -461,13 +467,20 @@ Feature: Multus-CNI related scenarios
       | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/bridge-host-local-vlan200.yaml |
       | n | <%= cb.project_name %>                                                                                                |
     Then the step should succeed
-    #Creating vlan pod abosrbing above net-attach-def
+    #Creating vlan pod absorbing above net-attach-def
     When I run the :create client command with:
       | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/pod1-bridge-host-local-vlan200.yaml |
       | n | <%= cb.project_name %>                                                                                                     |
     Then the step should succeed
     And the pod named "pod1-vlan200" becomes ready
     And evaluation of `pod` is stored in the :pod clipboard
+    
+    #Clean-up required to erase bridge interfcaes created on node after this step
+    Given I register clean-up steps:
+    """
+    the bridge interface named "mybridge" is deleted from the "<%= cb.pod.node_name %>" node
+    the bridge interface named "mybridge.200" is deleted from the "<%= cb.pod.node_name %>" node
+    """
     When I execute on the pod:
       | /usr/sbin/ip | -d | link |
     Then the output should contain: 
@@ -477,7 +490,7 @@ Feature: Multus-CNI related scenarios
     Then the step should succeed
     And the output should contain:
       | 200 |
-
+    
   # @author anusaxen@redhat.com
   # @case_id OCP-24467
   @admin
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -428,13 +428,14 @@ Feature: Multus-CNI related scenarios
     Given I have a project
     And evaluation of `project.name` is stored in the :project_name clipboard
     When I run the :create admin command with:
-      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/bridge-host-local-novlan.yaml |
-      | n | <%= cb.project_name %>                                                                                               |
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/bridge-host-local-novlan.yaml |
+      | n | <%= cb.project_name %>                                                                                                                              |
     Then the step should succeed
     #Creating no-vlan pod abosrbing above net-attach-def
-    When I run the :create client command with:
-      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/pod-bridge-host-local-novlan.yaml |
-      | n | <%= cb.project_name %>                                                                                                   |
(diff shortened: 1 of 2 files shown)


features/networking/multus.feature
```diff
@@ -416,4 +416,107 @@ Feature: Multus-CNI related scenarios
...
+    Then the step should succeed
+    And the output should contain 2 times:
+      | 1 PVID untagged |
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-24489
+  @admin
+  Scenario: Create pod with Multus bridge CNI plugin and vlan tag
+    # Make sure that the multus is enabled
+    Given the master version >= "4.0"
```
zhaozhanqi: This should be supported from 4.2

features/networking/multus.feature
```diff
@@ -416,4 +416,107 @@ Feature: Multus-CNI related scenarios
...
     And the output should contain "net2"
     And the output should contain 2 times:
-      | macvlan mode bridge |
+	    | macvlan mode bridge |
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-24488
+  @admin
+  Scenario: Create pod with Multus bridge CNI plugin without vlan
+    # Make sure that the multus is enabled
```
zhaozhanqi: This should be supported from 4.2

features/networking/multus.feature
```diff
@@ -416,4 +416,107 @@ Feature: Multus-CNI related scenarios
...
+    Given the master version >= "4.0"
+    And the multus is enabled on the cluster
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    And evaluation of `project.name` is stored in the :project_name clipboard
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/bridge-host-local-novlan.yaml |
+      | n | <%= cb.project_name %>                                                                                               |
```
zhaozhanqi: `\` is in an newline

features/networking/multus.feature
```diff
@@ -416,4 +416,107 @@ Feature: Multus-CNI related scenarios
...
+    Then the step should succeed
+    #Creating no-vlan pod abosrbing above net-attach-def
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/pod-bridge-host-local-novlan.yaml |
+      | n | <%= cb.project_name %>                                                                                                   |
```
zhaozhanqi: same `|`

zhaozhanqi:
others look good to me, Thanks @anuragthehatter

features/networking/multus.feature
anuragthehatter: I guess we shall remove such checks. Anyways when we run tests we select specific version in polarion and trigger on corresponding setup and polarion also has right version set. Not sure where such gates are useful.

Another view. oc version keeps breaking itself time to time like currently my client says "Client Version: openshift-clients-4.3.0-201909231341" which is wierd.

features/networking/multus.feature
anuragthehatter: Thanks @zhaozhanqi  Can you clarify more? Didn't understand

features/networking/multus.feature
anuragthehatter: The gating trend was there when 3.x and 4.x had same branch master. So may be its not useful now

features/networking/multus.feature
zhaozhanqi: oh, Nothing, should be the display issue in my side.

features/networking/multus.feature
anuragthehatter: removing version gates for now unless anybody has any reason to keep them

features/networking/multus.feature
anuragthehatter: thanks

features/networking/multus.feature
anuragthehatter: as discussed in other comment, removing version gates for now unless anybody has any reason to keep them

features/networking/multus.feature
anuragthehatter: ignoring it as per above discussion

anuragthehatter:
Hi @pruan-rht  @akostadinov  I would appreciate if we merge it if all comments looks good to leverage this in current release's tight schedule

features/step_definitions/networking.rb
```diff
@@ -678,9 +678,9 @@
   ensure_admin_tagged
   hosts = step "I select a random node's host"
   cb_name = "interface" unless cb_name
-  @result = host.exec_admin("/sbin/ip route show default | awk '/default/ {print $5}'")
+  @result = host.exec_admin("/sbin/ip route show default | awk 'NR==1{print $5}'")
```
rbbratta: was there bad input with multiple lines?  I'm the past I've seen `ip -o -4` to ensure single-line records and IPv4 only, but that might not be helpful in this case.

Also curiously `ip -j` outputs JSON now, which is exciting if we had `jq` installed everywhere.

```
ip -4  -j route show default | jq -r '.[0] | .dev'
```

features/networking/multus.feature
```diff
@@ -416,4 +416,104 @@ Feature: Multus-CNI related scenarios
...
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    And evaluation of `project.name` is stored in the :project_name clipboard
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/bridge-host-local-vlan200.yaml |
+      | n | <%= cb.project_name %>                                                                                                |
+    Then the step should succeed
+    #Creating vlan pod abosrbing above net-attach-def
```
rbbratta: s/abosrbing/absorbing/

features/step_definitions/networking.rb
anuragthehatter: Great suggestion @rbbratta but jq is not installed on nodes.
/sbin/ip route show default outputs like this:
default via 10.0.128.1 dev **ens3** 
default via 10.0.128.1 dev **ens3** proto dhcp metric 100 

There can be various way to grep ens3 but thought it should take only single line and fetch ens3 from that else it would fetch ens3 multiple times which needed a fix hence this one

features/networking/multus.feature
anuragthehatter: Done. Thanks for eagle eyes ;)

features/step_definitions/networking.rb
rbbratta: `host.exec_admin` is using ssh maybe?  

oddly, there seems to be `jq`  with `oc debug` and `chroot /host`

```
sh-4.4# cat /etc/redhat-release
Red Hat Enterprise Linux CoreOS release 4.3
sh-4.4# which jq
/usr/bin/jq
sh-4.4# jq --version
jq-1.5
```

anuragthehatter: host.exec_admin triggers oc debug. Hmm, with my oc debug on 4.3:

sh-4.4# cat /etc/redhat-release 
Red Hat Enterprise Linux CoreOS release 4.3
sh-4.4# which jq
which: no jq in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)

It is recommended to depend on basic linux commands/tools as across releases the support for tools keeps changing like for jq or git etc. Happened when we transitioned from 3.x to 4.x

rbbratta: exciting, must be different version of the debug pod or something.  

`    Image ID:      registry.redhat.io/rhel7/support-tools@sha256:6d0d66b55dbc1917f15b6c048158174b5bb5813bf10bb8d74c962a65dbb217dc`

akostadinov: Need to look deeper in this pull request. One thing discouraged since 4.x is running commands on nodes. People usually run commands on some pod when needed.

If possible do the same. Presently we are running commands on nodes using `oc debug node`. But it has a few nasty limitations:
https://bugzilla.redhat.com/show_bug.cgi?id=1771552
https://bugzilla.redhat.com/show_bug.cgi?id=1771549
https://bugzilla.redhat.com/show_bug.cgi?id=1771553

anuragthehatter: I am afraid that's not possible at the moment :( multus pods create their own interfaces different from node and we are interested in finding master interface on node. like multus pod create eth0 as primart interface but node create ensX. I tried to execute same and different command inside pod but couldn't get ensX in output

huiran0826: @anuragthehatter  can you try execute command on sdn pod instead of nodes?
     | bash | -c | /sbin/ip route show default | awk 'NR==1{print $5}  |

anuragthehatter: oops this is a general step for collecting master inf name. Sorry i misunderstood it as collecting master inf name from multus pod which is not possible. Yep, let me fix this to go inside pod for collecting master inf. It should be done via sdn pod or ovnkube pod if its OVN cluster.

anuragthehatter: @akostadinov committed. PTAL. Thanks

features/networking/multus.feature
akostadinov: It is useful when you want to test some framework change. Such checks let you avoid running impossible tests for your cluster.
e.g. grepping where some step is used and running such scenarios to see if your changes did not break something.
I agree it is not super important but is definitely helpful.

features/step_definitions/networking.rb
```diff
@@ -696,3 +685,35 @@
     raise "Fail to delete the ovs pod"
   end
 end
+
+Given /^the default interface on nodes is stored in the#{OPT_SYM} clipboard$/ do |cb_name|
+  ensure_admin_tagged
+  step "I select a random node's host"
+  node_name ||= node.name
+  cb_name = "interface" unless cb_name
```
akostadinov: `||=` operator means to set the value in case variable is `false` or `nil`. Thus above code will be nicer as:
```
node_name = node.name
cb_name ||= "interface"
```

features/step_definitions/networking.rb
```diff
@@ -696,3 +685,35 @@
...
+
+  sdn_pod = BushSlicer::Pod.get_labeled("app=sdn", project: project("openshift-sdn", switch: false), user: admin) { |pod, hash|
+    pod.node_name == node_name
+  }.first
+  cache_resources sdn_pod
+  @result = sdn_pod.exec("bash", "-c", "ip route show default | awk 'NR==1{print $5}'", as: admin)
```
akostadinov: I think the code above can be replaced with:
```
step "I run command on the \"node_name\" node's sdn pod:", table(
  "| bash | -c | ip route show default | awk 'NR==1{print $5}' |"
)
```
delegating to the step avoids duplicating logic to find SDN pod.

features/step_definitions/networking.rb
```diff
@@ -696,3 +685,35 @@
...
+  host = node.host
+  @result = host.exec_admin("/sbin/bridge vlan show")
+  raise "Failed to execute bridge vlan show command" unless @result[:success]
+end
+
+Given /^the bridge interface named "([^"]*)" is deleted from the "([^"]*)" node$/ do |bridge_name, node_name|
+  ensure_admin_tagged
+  node = node(node_name)
+  host = node.host
+  @result = host.exec_admin("/sbin/ip link delete #{bridge_name}")
```
akostadinov: whatever creates these bridges and vlans, isn't it running inside a pod? If so, can we deleted bridges and VLANs from pods instead of node hosts?

features/networking/multus.feature
```diff
@@ -416,4 +416,118 @@ Feature: Multus-CNI related scenarios
...
+    Given the default interface on nodes is stored in the :default_interface clipboard 
+    #Patching simplemacvlan config in network operator config CRD
+    Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with:
+      | {"spec":{"additionalNetworks":[{"name":"test-macvlan-case3","namespace":"openshift-multus","simpleMacvlanConfig":{"ipamConfig":{"staticIPAMConfig":{"addresses": [{"address":"10.128.2.100/23","gateway":"10.128.2.1"}]},"type":"static"},"master":"<%= cb.default_interface %>","mode":"bridge"},"type":"SimpleMacvlan"}]}} |
+    #Cleanup for bringing CRD to original
+    Given I register clean-up steps:
+    """
+    Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with: 
+    | {"spec":{"additionalNetworks": null}} |
+    """
```
akostadinov: btw will there be an issue if we remove and re-create the network operator instead of repatcing it? We have a ready step to re-create resoures. if you think that can work, then you could try it. See example in `features/step_definitions/oauth_client.rb`.
If this is not a good solution, I'm ok with existing one.

features/networking/multus.feature
```diff
@@ -416,4 +416,118 @@ Feature: Multus-CNI related scenarios
...
+  @admin
+  Scenario: Create pod with Multus bridge CNI plugin without vlan
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    And evaluation of `project.name` is stored in the :project_name clipboard
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/bridge-host-local-novlan.yaml |
+      | n | <%= cb.project_name %>                                                                                                                              |
```
akostadinov: you could skip `evaluation` step and put here:
```
| n | <%= project.name %>     |
```

features/networking/multus.feature
```diff
@@ -416,4 +416,118 @@ Feature: Multus-CNI related scenarios
...
+  @admin
+  Scenario: Create pod with Multus bridge CNI plugin and vlan tag
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    And evaluation of `project.name` is stored in the :project_name clipboard
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/bridge-host-local-vlan200.yaml |
+      | n | <%= cb.project_name %>                                                                                                |
```
akostadinov: same about `project.name`

features/networking/multus.feature
```diff
@@ -416,4 +416,118 @@ Feature: Multus-CNI related scenarios
...
+      | n | <%= cb.project_name %>                                                                                                     |
+    Then the step should succeed
+    And the pod named "pod1-vlan200" becomes ready
+    And evaluation of `pod` is stored in the :pod clipboard
+    
+    #Clean-up required to erase bridge interfcaes created on node after this step
+    Given I register clean-up steps:
+    """
+    the bridge interface named "mybridge" is deleted from the "<%= cb.pod.node_name %>" node
+    the bridge interface named "mybridge.200" is deleted from the "<%= cb.pod.node_name %>" node
```
akostadinov: Is it possible to randomize bridge names? Scenario is not marked as `destructive` so it shouldn't interfere with other scenarios running in parallel. If other scenarios use same bridge names `mybridge`, then they will likely interefere. And error will be hard to debug.

So wither we need to ensure scenarios do not interfere, or we need to tag as `@destructive`. As well put in polarion the `destructive` tag.

features/networking/multus.feature
```diff
@@ -416,4 +416,118 @@ Feature: Multus-CNI related scenarios
...
+    #Creating vlan pod absorbing above net-attach-def
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/pod1-bridge-host-local-vlan200.yaml |
+      | n | <%= cb.project_name %>                                                                                                     |
+    Then the step should succeed
+    And the pod named "pod1-vlan200" becomes ready
+    And evaluation of `pod` is stored in the :pod clipboard
+    
+    #Clean-up required to erase bridge interfcaes created on node after this step
+    Given I register clean-up steps:
```
akostadinov: btw instead of wrapping into this step, you can clean-up by using step
```
Given the bridge interface named "mybridge" is deleted from the "<%= cb.pod.node_name %>" node after scenario
```

features/networking/multus.feature
```diff
@@ -416,4 +416,118 @@ Feature: Multus-CNI related scenarios
...
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | test-macvlan-case3 |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | name=macvlan-bridge-pod |
+    And evaluation of `pod` is stored in the :pod clipboard
+
+    When I execute on the pod:
+      | /usr/sbin/ip | -d | link |
+    Then the output should contain "net1"
+    And admin ensures "<%= cb.pod.name %>" pod is deleted from the "openshift-multus" project after scenario
```
akostadinov: this needs to be registered in scenario before pod or immediately after pod is created. In case you need to obtain pod name after creation, then after creation you could use step `a pod is present with labels:`. Then register clean-up and then wait for it to become ready and whatever other tests.

The reason is to make sure failures do not lead to leaked resources.

features/networking/multus.feature
```diff
@@ -416,4 +416,118 @@ Feature: Multus-CNI related scenarios
...
+  # @author anusaxen@redhat.com
+  # @case_id OCP-24467
+  @admin
+  Scenario: CNO manager mavlan configured manually with static
+    # Make sure that the multus is Running
+    Given the multus is enabled on the cluster
+    Given the default interface on nodes is stored in the :default_interface clipboard 
+    #Patching simplemacvlan config in network operator config CRD
+    Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with:
+      | {"spec":{"additionalNetworks":[{"name":"test-macvlan-case3","namespace":"openshift-multus","simpleMacvlanConfig":{"ipamConfig":{"staticIPAMConfig":{"addresses": [{"address":"10.128.2.100/23","gateway":"10.128.2.1"}]},"type":"static"},"master":"<%= cb.default_interface %>","mode":"bridge"},"type":"SimpleMacvlan"}]}} |
```
akostadinov: Now thinking about this again, I think scenarios patching `"networks.operator.openshift.io/cluster"` need to be `@destructive` because they will surely interfere with each other if run simultaneously. See `doc/admin_features.adoc`.

features/networking/multus.feature
anuragthehatter: I guess existing one is fine. We have encountered bugs in past where oc patch randomly didn't work in CNO so we can catch that as well

features/step_definitions/networking.rb
zhaozhanqi: @anuragthehatter  I also added the comment in https://github.com/openshift/verification-tests/pull/502#discussion_r355254278 about this 
if we use sdn_pod to run this command `ip route`', then many cases using this step will cannot be run on OVN plugin.

features/step_definitions/networking.rb
```diff
@@ -696,3 +685,29 @@
     raise "Fail to delete the ovs pod"
   end
 end
+
+Given /^the default interface on nodes is stored in the#{OPT_SYM} clipboard$/ do |cb_name|
+  ensure_admin_tagged
+  step "I select a random node's host"
```
huiran0826: Can we move this step "I select a random node's host" to case level?

zhaozhanqi: @huiran0826 what do you mean for case level?  or any shortcoming for current steps?

huiran0826: @zhaozhanqi    Yes, it may cause some inconsistent under some conditions. For example: In  egressip, there are several steps need to execute on the node, such as get default interface ,get subnet for default interface, add valid ip on the node.  So here the node should be the same one, if using "I select a random node's host " in step definition, cannot assure the result from same node. So suggest  move this step to case : Given I select a random node's host

zhaozhanqi: mostly this step should be only executed once in one case, and after this step, you can store the node  to clipboard.  
you mean `get subnet for default interface, add valid ip on the node.`  need to use same step?

huiran0826: I mean, get default interface, get subnet for default interface, get valid ip on the node, basically all of these steps need to run command on the nodes.  Do you mean just use "I select a random node's host" in get default interface, but not in other steps?  If so ,  get default interface has to be the first step.  Actually why we just move this step "I select a random node's host" to the case?  If still not clear, let's discuss it offline.

zhaozhanqi: @huiran0826 Ok, I got you.  you want use this step after doing some changes in one node.  do we have this kind of scenario? if so. we can discuss how to deal with it. since if we move it out. we need to update some cases

huiran0826: @zhaozhanqi Yes, I have this kind of scenario, we can discuss it offline.

anuragthehatter: What you guys discuss offline and think, let me know :)

anuragthehatter:
@akostadinov  @zhaozhanqi  Can we merge this ASAP to support #402 and other automation blocked due to it? All comments have been taken care of and will continue looking as we progress OVN integeration. Please merge. Thanks!

features/step_definitions/networking.rb
anuragthehatter: This is fixed with same step integerating OVN check

features/networking/multus.feature
```diff
@@ -416,4 +416,118 @@ Feature: Multus-CNI related scenarios
...
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/bridge-host-local-vlan200.yaml |
+      | n | <%= project.name %>                                                                                                |
+    Then the step should succeed
+    #Creating vlan pod absorbing above net-attach-def
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/pod1-bridge-host-local-vlan200.yaml |
+      | n | <%= cb.project_name %>                                                                                                     |
```
pruan-rht: @anuragthehatter I don't see `cb.project_name` defined previously...I think you mean `projcect.name` instead?

features/networking/multus.feature
```diff
@@ -416,4 +416,118 @@ Feature: Multus-CNI related scenarios
...
+    #Cleanup for bringing CRD to original
+    Given I register clean-up steps:
+    """
+    Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with: 
+    | {"spec":{"additionalNetworks": null}} |
+    """
+ 
+    And I wait up to 30 seconds for the steps to pass:
+    """
+    When I run the :get admin command with:
```
pruan-rht: @anuragthehatter  TODO: we can perhaps use the framework's get properties instead of calling `:get` directly.

features/networking/multus.feature
anuragthehatter: Fixed. Apparently this remained unchanged when i was changing all to project.name. Thanks for catching :(

features/networking/multus.feature
pruan-rht: I will merge it now to unblock you guys…we can pick up the optimization later.

PR #435, opened 2019-11-05
Simplify machine steps

jhou1:
As a follow up of discussion at https://github.com/openshift/verification-tests/pull/423#discussion_r340892440, simplified the steps for scaling machinesets. The test is now more efficient by reducing the steps.

### DIFF
--- a/features/step_definitions/machine_set.rb
+++ b/features/step_definitions/machine_set.rb
@@ -4,3 +4,46 @@ Given /^I store all machinesets to the#{OPT_SYM} clipboard$/ do |cb_name|
   cb_name ||= :machinesets
   cb[cb_name] = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api"))
 end
+
+Given(/^I pick a random machineset to scale$/) do
+  ensure_admin_tagged
+  machine_sets = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api"))
+  cache_resources *machine_sets.shuffle
+end
+
+When(/^I scale the machineset to ([\+\-]?)#{NUMBER}$/) do | op, num |
+  ensure_destructive_tagged
+
+  case op
+  when "-"
+    replicas = machine_set.available_replicas - num.to_i
+  when "+"
+    replicas = machine_set.available_replicas + num.to_i
+  when ""
+    replicas = num.to_i
+  else
+    raise "wrong operation #{op.inspect} supplied"
+  end
+
+  step %Q/I run the :scale admin command with:/, table(%{
+    | n        | openshift-machine-api   |
+    | resource | machineset              |
+    | name     | <%= machine_set.name %> |
+    | replicas | #{replicas.to_s}        |
+  })
+end
+
+Then(/^the machineset should have expected number of running machines$/) do
+  machine_set.wait_till_ready(admin, 600)
+
+  machines_all = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api"))
+  machines = machines_all.select { |m| m.machine_set_name == machine_set.name }
+
+  # Each node should be running
+  machines.each do | machine |
+    machine.get
+    unless node(machine.node_name).ready?[:success]
+      raise "Node #{machine.node_name} has not become ready."
+    end
+  end
+end
--- a/lib/openshift/machine.rb
+++ b/lib/openshift/machine.rb
@@ -15,5 +15,10 @@ module BushSlicer
       instance_state = raw_resource(user: user, cached: cached, quiet: quiet).dig('status','providerStatus','instanceState')
       instance_state == 'running'
     end
+
+    def machine_set_name(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).
+        dig('metadata', 'labels', 'machine.openshift.io/cluster-api-machineset')
+    end
   end
 end
--- a/lib/openshift/machine_set.rb
+++ b/lib/openshift/machine_set.rb
@@ -7,16 +7,19 @@ module BushSlicer
 
     def desired_replicas(user: nil, cached: true, quiet: false)
       rr = raw_resource(user: user, cached: cached, quiet: quiet)
-      rr.dig('spec', 'replicas')
+      rr.dig('spec', 'replicas').to_i
     end
 
-    def current_replicas(user: nil, cached: true, quiet: false)
+    def available_replicas(user: nil, cached: true, quiet: false)
       rr = raw_resource(user: user, cached: cached, quiet: quiet)
-      rr.dig('status', 'replicas')
+      rr.dig('status', 'availableReplicas').to_i
     end
 
-    def healthy?
-      return desired_replicas == current_replicas
+    def ready?(user: nil, cached: false, quiet: false)
+      result = {}
+      result[:success] = (desired_replicas(user: user, cached: cached, quiet: quiet) ==
+        available_replicas(user: user, cached: cached, quiet: quiet))
+      return result
     end
   end
 end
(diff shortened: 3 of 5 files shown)


jhou1:

features/step_definitions/machine_set.rb
```diff
@@ -4,3 +4,38 @@
   cb_name ||= :machinesets
   cb[cb_name] = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api"))
 end
+
+Given(/^I pick a random machineset to scale$/) do
+  machine_sets = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api"))
+  cache_resources *machine_sets
+
+  @machine_set = machine_sets.shuffle.first
```
akostadinov: If you want to get a random machine set, then just cache like:
```
cache_resources *machine_sets.shuffle
```

Later just call method `machine_set` method to get the random one.

features/step_definitions/machine_set.rb
```diff
@@ -4,3 +4,38 @@
...
+end
+
+When(/^I scale the machineset by "([^"]*)"$/) do | by |
+  replicas = @machine_set.available_replicas + by.to_i
```
akostadinov: `replicas = machine_set.available_replicas + by.to_i`

akostadinov: Also we need a line `ensure_destructive_tagged`

lib/openshift/machine.rb
```diff
@@ -15,5 +15,10 @@ def ready?(user: nil, cached: true, quiet: false)
       instance_state = raw_resource(user: user, cached: cached, quiet: quiet).dig('status','providerStatus','instanceState')
       instance_state == 'running'
     end
+
+    def machine_set(user: nil, cached: true, quiet: false)
```
akostadinov: method returns a String thus better name it `machine_set_name`. To return the machine set you can do:
```
def machine_set(user: nil, cached: true, quiet: false)
  MachineSet.new(
    name: machine_set_name(user: user, cached: cached, quiet: quiet),
    project: project).tap { |m| m.default_user = default_user(user) }
  )
end
```

lib/openshift/machine_set.rb
```diff
@@ -7,16 +7,16 @@ class MachineSet < ProjectResource
...
-    def current_replicas(user: nil, cached: true, quiet: false)
+    def available_replicas(user: nil, cached: true, quiet: false)
       rr = raw_resource(user: user, cached: cached, quiet: quiet)
-      rr.dig('status', 'replicas')
+      rr.dig('status', 'availableReplicas').to_i
     end
 
     def healthy?
-      return desired_replicas == current_replicas
+      return desired_replicas == available_replicas
```
akostadinov: actually it makes more sense to add some params:
```
    def healthy?(user: nil, cached: true, quiet: false)
      return desired_replicas(user: user, cached: cached, quiet: quiet) == available_replicas(cached: true)
    end
```
Otherwise it might be ugly to call the method in a wait loop.

akostadinov: I would suggest also to rename this method to `ready?`. Then waiting for this will be as easy as calling
```
  machine_set.wait_till_ready(user, timeout)
```

features/step_definitions/machine_set.rb
```diff
@@ -4,3 +4,38 @@
   cb_name ||= :machinesets
   cb[cb_name] = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api"))
 end
+
+Given(/^I pick a random machineset to scale$/) do
```
akostadinov: we need a line `ensure_admin_tagged`

features/step_definitions/machine_set.rb
```diff
@@ -4,3 +4,38 @@
...
+  machine_sets = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api"))
+  cache_resources *machine_sets
+
+  @machine_set = machine_sets.shuffle.first
+end
+
+When(/^I scale the machineset by "([^"]*)"$/) do | by |
```
akostadinov: I would suggest this step to read:
```
When /^I scale the machineset to ([-+]?)NUMBER$/ do | op, num |
  case op
  when "-"
    replicas = @machine_set.available_replicas - num.to_i
  when "+"
    replicas = @machine_set.available_replicas + num.to_i
  when nil
    replicas = num.to_i
  else
    raise "wrong operation #{op.inspect} supplied"
  end
  ...
end
```

In this way the step can also be used to shorten clean-up too.

jhou1:
@akostadinov Thank you for the review, I've made updates accordingly.

lib/openshift/machine_set.rb
```diff
@@ -7,16 +7,19 @@ class MachineSet < ProjectResource
...
-      rr.dig('status', 'replicas')
+      rr.dig('status', 'availableReplicas').to_i
     end
 
-    def healthy?
-      return desired_replicas == current_replicas
+    def ready?(user: nil, cached: false, quiet: false)
+      result = {}
+      result[:success] = (desired_replicas(user: user, cached: cached, quiet: quiet) ==
+        available_replicas(user: user, cached: cached, quiet: quiet))
```
akostadinov: My suggestion was to use `cached: true` on the second call because you don't need to get full YAML twice. First call retrieves an updated YAML. Second call only needs to read it.

features/step_definitions/machine_set.rb
```diff
@@ -4,3 +4,46 @@
...
   cb[cb_name] = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api"))
 end
+
+Given(/^I pick a random machineset to scale$/) do
+  ensure_admin_tagged
+  machine_sets = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api"))
+  cache_resources *machine_sets.shuffle
+end
+
+When(/^I scale the machineset to ([\+\-]?)(\d)$/) do | op, num |
```
akostadinov: I have suggested to use `NUMBER` because it will also catch `<%= expression %>` format of specifying the number. But I had a mistake in my original suggestion. It should have been:
```
When /^I scale the machineset to ([\+\-]?)#{NUMBER}$/ do | op, num |
```
Sorry if this caused confusion.

akostadinov:
Thank you, just two simple suggestions.

lib/openshift/machine_set.rb
jhou1: Thanks @akostadinov. I was using `cached: true`, but as I had tested, it returns the machine_set with `desired_replicas=1` and `available_replicas=1` after I scale it to 3.  I set it to false to retrieve the data again so that it has new value.

jhou1: As explained above, then `machine_set.wait_till_ready(admin, 600)` will immediately returns  because it compares the cached `desired_replicas` with cached `available_replicas` and believes it's ready.  @akostadinov Will this method here make sense to you?

jhou1:
@akostadinov @pruan-rht Would you PTAL?

```
      [03:33:42] INFO> Exit Status: 0
      [03:33:42] INFO> === End After Scenario: Scale up and scale down a machineSet ===

1 scenario (1 passed)
10 steps (10 passed)
5m31.054s
```

lib/openshift/machine_set.rb
akostadinov: You need `cached: false` on first call to update the cache. The second call you can have `cached: true` because cache is already updated. If you keep `cached: false` in both places you perform 2 separate calls which is unnecessary.
Maybe I was not clear the first call should be `false` and only the second one `true`.

jhou1: Thanks. Now I get it, I'll address this in a subsequent PR.

PR #438, opened 2019-11-08
More n more multus auto

anuragthehatter:
OCP-24490, OCP24491 and OCP-24607
@weliang1 @zhaozhanqi @huiran0826 @rbbratta

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -537,14 +537,14 @@ Feature: Multus-CNI related scenarios
     # Create the net-attach-def with vlan 100 via cluster admin
     Given I have a project
     When I run the :create admin command with:
-      | f | https://raw.githubusercontent.com/anuragthehatter/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/bridge-host-local-vlan.yaml |
-      | n | <%= project.name %>                                                                                                                                  |
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/bridge-host-local-vlan.yaml |
+      | n | <%= project.name %>                                                                                                                               |
     Then the step should succeed 
     
     # Create the net-attach-def with vlan 200 via cluster admin
     When I run the :create admin command with:
-      | f | https://raw.githubusercontent.com/anuragthehatter/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/bridge-host-local-vlan-200.yaml |
-      | n | <%= project.name %>                                                                                                                                      |
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/bridge-host-local-vlan-200.yaml |
+      | n | <%= project.name %>                                                                                                                                   |
     Then the step should succeed 
     #Labeing a worker node to make sure all future pods to be scheduled on this node only
     When I run the :label admin command with:
@@ -635,8 +635,8 @@ Feature: Multus-CNI related scenarios
     # Create the net-attach-def without master pmtr via cluster admin
     Given I have a project
     When I run the :create admin command with:
-      | f | https://raw.githubusercontent.com/anuragthehatter/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-conf-without-master.yaml |
-      | n | <%= project.name %>                                                                                                                                       |
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-conf-without-master.yaml |
+      | n | <%= project.name %>                                                                                                                                    |
     Then the step should succeed
     
     #Creating a pod absorbing above net-attach-def
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -499,6 +499,15 @@ Feature: Multus-CNI related scenarios
       | ifconfig | net1 |
     Then the step should succeed
     And evaluation of `@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :pod3_net1_ip clipboard
+    
+    #Clean-up required to erase bridge interfcaes created on node
+    Given I register clean-up steps:
+    """
+    the bridge interface named "mybridge" is deleted from the "<%= cb.nodes[0].name %>" node
+    the bridge interface named "mybridge.100" is deleted from the "<%= cb.nodes[0].name %>" node
+    the bridge interface named "mybridge" is deleted from the "<%= cb.nodes[1].name %>" node
+    the bridge interface named "mybridge.100" is deleted from the "<%= cb.nodes[1].name %>" node
+    """  
     #making sure the pods on same node can ping while pods on diff nodes can't
 
     When I execute on the "<%= cb.pod1 %>" pod:
@@ -598,6 +607,13 @@ Feature: Multus-CNI related scenarios
     Then the step should succeed
     And evaluation of `@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :pod3_net1_ip clipboard
     
+    #Clean-up required to erase bridge interfcaes created on node
+    Given I register clean-up steps:
+    """
+    the bridge interface named "mybridge" is deleted from the "<%= cb.nodes[0].name %>" node
+    the bridge interface named "mybridge.100" is deleted from the "<%= cb.nodes[0].name %>" node
+    the bridge interface named "mybridge.200" is deleted from the "<%= cb.nodes[0].name %>" node
+    """  
     #making sure the pods in same vlan can communicate but in different vlans cannot
     When I execute on the "<%= cb.pod1 %>" pod:
       | ping | -c1 | -W2 | <%= cb.pod2_net1_ip %> |
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -610,6 +610,7 @@ Feature: Multus-CNI related scenarios
   # @author anusaxen@redhat.com
   # @case_id OCP-24490
   @admin
+  @destructive	
   Scenario: Pods can communicate each other with same vlan tag
     # Make sure that the multus is enabled
     Given the multus is enabled on the cluster
@@ -619,36 +620,13 @@ Feature: Multus-CNI related scenarios
       | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/bridge-host-local-vlan.yaml |
       | n | <%= project.name %>                                                                                                                               |
     Then the step should succeed
+    
     #Labeling a worker node to make sure couple of future pods to be scheduled on this node only
-    When I run the :label admin command with:
-      | resource | node                    |
-      | name     | <%= cb.nodes[0].name %> |
-      | key_val  | test=worker1            |
-    Then the step should succeed
+    Given  label "test=worker1" is added to the "<%= cb.nodes[0].name %>" node
     
-    Given I register clean-up steps:
-    """
-    When I run the :label admin command with:
-      | resource | node                    |
-      | name     | <%= cb.nodes[0].name %> |
-      | key_val  | test-                   |
-    Then the step should succeed
-    """ 
     #Labeing another worker node to make sure 3rd future pod to be scheduled on this node only
-    When I run the :label admin command with:
-      | resource | node                    |
-      | name     | <%= cb.nodes[1].name %> |
-      | key_val  | test=worker2            |
-    Then the step should succeed
+    Given  label "test=worker2" is added to the "<%= cb.nodes[1].name %>" node
     
-    Given I register clean-up steps:
-    """
-    When I run the :label admin command with:
-      | resource | node                    |
-      | name     | <%= cb.nodes[1].name %> |
-      | key_val  | test-            	   |
-    Then the step should succeed
-    """ 
     #Creating first pod in vlan 100
     When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/generic_multus_pod_nodeselector.yaml" replacing paths:
       | ["metadata"]["name"] | pod1-vlan100 |
@@ -729,6 +707,7 @@ Feature: Multus-CNI related scenarios
   # @author anusaxen@redhat.com
   # @case_id OCP-24491
   @admin
+  @destructive
   Scenario: Pods cannot communicate each other with different vlan tag
     # Make sure that the multus is enabled
     Given the multus is enabled on the cluster
@@ -747,20 +726,8 @@ Feature: Multus-CNI related scenarios
     Then the step should succeed 
     
     #Labeing a worker node to make sure all future pods to be scheduled on this node only
-    When I run the :label admin command with:
-      | resource | node                    |
-      | name     | <%= cb.nodes[0].name %> |
-      | key_val  | test=worker1            |
-    Then the step should succeed
+    Given  label "test=worker1" is added to the "<%= cb.nodes[0].name %>" node
     
-    Given I register clean-up steps:
-    """
-    When I run the :label admin command with:
-      | resource | node                    |
-      | name     | <%= cb.nodes[0].name %> |
-      | key_val  | test-                   |
-    Then the step should succeed
-    """ 
     #Creating first pod in vlan 100
     When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/generic_multus_pod_nodeselector.yaml" replacing paths:
       | ["metadata"]["name"] | pod1-vlan100 |
@@ -807,6 +774,7 @@ Feature: Multus-CNI related scenarios
     the bridge interface named "mybridge.100" is deleted from the "<%= cb.nodes[0].name %>" node
     the bridge interface named "mybridge.200" is deleted from the "<%= cb.nodes[0].name %>" node
     """  
+    
     #making sure the pods in same vlan can communicate but in different vlans cannot
     When I execute on the "<%= cb.pod1 %>" pod:
       | ping | -c1 | -W2 | <%= cb.pod2_net1_ip %> |
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -621,6 +621,11 @@ Feature: Multus-CNI related scenarios
       | n | <%= project.name %>                                                                                                                               |
     Then the step should succeed
     
+    #Clean-up required to erase bridge interfaces created due to above net-attach-def
+    Given I register clean-up steps:
+    """
+    the bridge interface named "mybridge" is deleted from the "<%= cb.nodes[0].name %>" node
+    """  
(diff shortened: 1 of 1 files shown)


features/networking/multus.feature
```diff
@@ -416,4 +416,232 @@ Feature: Multus-CNI related scenarios
...
+    Then the step should succeed
+    
+    Given I register clean-up steps:
+    """
+    When I run the :label admin command with:
+      | resource | node                    |
+      | name     | <%= cb.nodes[0].name %> |
+      | key_val  | test-                   |
+    Then the step should succeed
+    the step should succeed
```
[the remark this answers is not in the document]
anuragthehatter: yep thanks for catching that..its duplicate. fixed it

features/networking/multus.feature
```diff
@@ -416,4 +416,232 @@ Feature: Multus-CNI related scenarios
...
+  # @author anusaxen@redhat.com
+  # @case_id OCP-24491
+  @admin
+  Scenario: Pods cannot communicate each other with same vlan tag
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+    And I store all worker nodes to the :nodes clipboard
+    # Create the net-attach-def with vlan 100 via cluster admin
+    Given I have a project
+    When I run the :create admin command with:
```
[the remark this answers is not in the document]
anuragthehatter: oops, i don't understand what do you mean? My comment is kinda same
 "Create the net-attach-def with vlan 100 via cluster admin"

features/networking/multus.feature
```diff
@@ -416,4 +416,232 @@ Feature: Multus-CNI related scenarios
...
+    Then the step should fail
+    
+    When I execute on the "<%= cb.pod2 %>" pod:
+      | ping | -c1 | -W2 | <%= cb.pod3_net1_ip %> |
+    Then the step should fail
+    
+  # @author anusaxen@redhat.com
+  # @case_id OCP-24491
+  @admin
+  Scenario: Pods cannot communicate each other with same vlan tag
```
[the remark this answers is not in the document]
anuragthehatter: Yep, its in different vlans but like typo on title. Fixed it, Thanks

features/networking/multus.feature
```diff
@@ -416,4 +416,232 @@ Feature: Multus-CNI related scenarios
...
+      | resource | node                    |
+      | name     | <%= cb.nodes[0].name %> |
+      | key_val  | test-                   |
+    Then the step should succeed
+    the step should succeed
+    """ 
+    #Creating first pod in vlan 100
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
+      | ["metadata"]["name"] | pod1-vlan100 |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"]| bridgevlan100 |
```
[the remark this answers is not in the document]
anuragthehatter: Yep this was fixed immediately post generating this PR. Figured some overwrites. Ran tests again. works fine.

features/networking/multus.feature
```diff
@@ -416,4 +416,232 @@ Feature: Multus-CNI related scenarios
...
+    And evaluation of `pod.name` is stored in the :pod1 clipboard
+    And I execute on the pod:
+      | ifconfig | net1 |
+    Then the step should succeed
+    And evaluation of `@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :pod1_net1_ip clipboard
+    
+    #Creating 2nd pod on same node as first in vlan 100
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
+      | ["metadata"]["name"] | pod2-vlan100 |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"]| bridgevlan100 |
```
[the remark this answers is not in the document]
anuragthehatter: Yep this was fixed immediately post generating this PR. Figured some overwrites. Ran tests again. works fine.

features/networking/multus.feature
```diff
@@ -416,4 +416,232 @@ Feature: Multus-CNI related scenarios
...
+    And evaluation of `pod.name` is stored in the :pod2 clipboard
+    And I execute on the pod:
+      | ifconfig | net1 |
+    Then the step should succeed
+    And evaluation of `@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :pod2_net1_ip clipboard
+    
+    #Creating 3rd pod on same node but in vlan 200
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
+      | ["metadata"]["name"] | pod3-vlan200 |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"]| bridgevlan200 |
```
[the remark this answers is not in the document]
anuragthehatter: Yep this was fixed immediately post generating this PR. Figured some overwrites. Ran tests again. works fine.
Same as all above

features/networking/multus.feature
```diff
@@ -416,4 +416,232 @@ Feature: Multus-CNI related scenarios
...
+    Then the step should succeed
+    the step should succeed
+    """ 
+    #Creating first pod in vlan 100
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/generic_multus_pod_nodeselector.yaml" replacing paths:
+      | ["metadata"]["name"] | pod1-vlan100 |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"]| bridgevlan100 |
+      | ["spec"]["nodeSelector"]["test"] | worker1 |
+    Then the step should succeed
+    And the pod named "pod1-vlan100" becomes ready
```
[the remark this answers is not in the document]
anuragthehatter: This wont be specific time dependent. It will just poll until pod1-vlan100 become ready. It will time out with automation default timeout not sure how much is that.

features/networking/multus.feature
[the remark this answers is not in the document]
weliang1: Sorry, just found you have it in line 537

features/networking/multus.feature
```diff
@@ -416,4 +416,248 @@ Feature: Multus-CNI related scenarios
...
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+    And I store all worker nodes to the :nodes clipboard
+    Given I have a project
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/bridge-host-local-vlan.yaml |
+      | n | <%= project.name %>                                                                                                                               |
+    Then the step should succeed 
+    #Labeling a worker node to make sure couple of future pods to be scheduled on this node only
+    When I run the :label admin command with:
```
zhaozhanqi: @anuragthehatter  Thanks for the job. 
How about store the node name and using `nodeName` to make the pod to schedule the node you wanted. we better avoid updating the default label of worker.

anuragthehatter: @zhaozhanqi Yep, i made sure about this. This step is not overwriting default label but just appending this new label to default. During cleanup just the appended label will be cleaned up and default one's will stay same

features/networking/multus.feature
```diff
@@ -416,4 +416,248 @@ Feature: Multus-CNI related scenarios
...
+    # Create the net-attach-def without master pmtr via cluster admin
+    Given I have a project
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-conf-without-master.yaml |
+      | n | <%= project.name %>                                                                                                                                    |
+    Then the step should succeed
+    
+    #Creating a pod absorbing above net-attach-def
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | macvlan-conf |
```
akostadinov: pod perhaps needs namespace also set, otherwise it will go into `default` project

features/networking/multus.feature
```diff
@@ -416,4 +416,248 @@ Feature: Multus-CNI related scenarios
...
+    Then the step should succeed
+    
+    Given I register clean-up steps:
+    """
+    When I run the :label admin command with:
+      | resource | node                    |
+      | name     | <%= cb.nodes[0].name %> |
+      | key_val  | test-                   |
+    Then the step should succeed
+    """
```
akostadinov: we have a step ` /^label #{QUOTED} is added to the#{OPT_QUOTED} node$/`. It automatically registers clean-up so scenario will become a lot shorter.

features/networking/multus.feature
```diff
@@ -416,4 +416,248 @@ Feature: Multus-CNI related scenarios
...
+      | ["metadata"]["name"] | pod3-vlan200 |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"]| bridgevlan200 |
+      | ["spec"]["nodeSelector"]["test"] | worker1 |
+    Then the step should succeed
+    And the pod named "pod3-vlan200" becomes ready
+    And evaluation of `pod.name` is stored in the :pod3 clipboard
+    And I execute on the pod:
+      | ifconfig | net1 |
+    Then the step should succeed
+    And evaluation of `@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :pod3_net1_ip clipboard
```
akostadinov: Just asking whether it will be an option to get pod ip by calling `pod.ip` instead of running `ifconfgi`.
```
And evaluation of `pod.ip` is stored in the :pod3_net1_ip clipboard
```

features/networking/multus.feature
```diff
@@ -416,4 +416,248 @@ Feature: Multus-CNI related scenarios
...
+    
+    #Clean-up required to erase bridge interfcaes created on node
+    Given I register clean-up steps:
+    """
+    the bridge interface named "mybridge" is deleted from the "<%= cb.nodes[0].name %>" node
+    the bridge interface named "mybridge.100" is deleted from the "<%= cb.nodes[0].name %>" node
+    the bridge interface named "mybridge.200" is deleted from the "<%= cb.nodes[0].name %>" node
+    """
```
akostadinov: It appears to me that these bridges are created earlier, so could you move clean-up of bridges to the place immediately after they were created?

.. unless I am mistaken

akostadinov: btw is it not a bug that we deleted project but bridges remain?

features/networking/multus.feature
```diff
@@ -416,4 +416,248 @@ Feature: Multus-CNI related scenarios
     Then the output should contain "net1"
     And the output should contain "net2"
     And the output should contain 2 times:
-      | macvlan mode bridge |
+	    | macvlan mode bridge |
+
+ 
+  # @author anusaxen@redhat.com
+  # @case_id OCP-24490
```
akostadinov: same comments as for OCP-24491

features/networking/multus.feature
anuragthehatter: Thanks. pod.ip will fetch pod mgmt ip which will be eth0 of pod created. I am interested in net1 interface here created under same pod

features/networking/multus.feature
anuragthehatter: i had this discussion with dev last month. As pr CNF design, we are not supposed to cleanup bridge interfaces on node due to possibility of then for reuse. But in test we need to clear them up manually else they will interfere with output of next similar test

features/networking/multus.feature
anuragthehatter: hmm, wasn't aware of that, Thanks WTAL

features/networking/multus.feature
akostadinov: hmm, doesn't sound convincing to me but I'll pass this time

akostadinov: Still registering clean-up would make sense for your scenarios immediately before/after place where these are created.

anuragthehatter: Ya i was not convinced too :( https://bugzilla.redhat.com/show_bug.cgi?id=1768289
About cleanup steps, i will take a look. If cleanup is placed after pod creation then its fine because bridge actually gets created on a node when a pod gets created and landed up on that node

features/networking/multus.feature
anuragthehatter: afraid, this step "/^label #{QUOTED} is added to the#{OPT_QUOTED} node$/" is not getting cleanup post scenario as per testing

features/networking/multus.feature
akostadinov: There is no harm if you put the delete step even before creating the resource. The step logic is to ignore if object doesn't exist. Only fail if object exists and delete fails.
If there is any issue with pod and it doesn't start properly but bridge is created, registering later may never happen.
That's why I suggest moving earlier.

features/networking/multus.feature
akostadinov: Could you provide full run log where you observe label not being restored?

features/networking/multus.feature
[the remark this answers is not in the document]
akostadinov: @weliang1 , could you check whether your commends have been addressed and resolve conversations where you are satisfied?

features/networking/multus.feature
anuragthehatter: Line 1457 here http://pastebin.test.redhat.com/820129. I confirmed after the test by manaully login into cluster that label test=worker1 was still there on the node

features/networking/multus.feature
anuragthehatter: fixed. Adjusted in more sensible way

features/networking/multus.feature
anuragthehatter: @akostadinov @pruan-rht Any thoughts?

features/networking/multus.feature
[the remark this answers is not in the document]
weliang1: LGTM

features/networking/multus.feature
akostadinov: The issue has nothing to do with the label step. I see at the end
```
      [23:29:43] INFO> === After Scenario: Pods cannot communicate each other with different vlan tag ===
      [23:29:43] INFO> Step: the bridge interface named "mybridge.200" is deleted from the "ip-10-0-131-252.ap-northeast-2.compute.internal" node
      [23:29:43] ERROR> Test Execution will finish prematurely.
      Undefined dynamic step: "the bridge interface named "mybridge.200" is deleted from the "ip-10-0-131-252.ap-northeast-2.compute.internal" node" (Cucumber::UndefinedDynamicStep)
      /home/anusaxen/git/verification-tests/lib/world.rb:507:in `block (3 levels) in to_step_procs'
      /home/anusaxen/git/verification-tests/lib/world.rb:429:in `block in after_scenario'
      /home/anusaxen/git/verification-tests/lib/world.rb:429:in `reverse_each'
      /home/anusaxen/git/verification-tests/lib/world.rb:429:in `after_scenario'
      /home/anusaxen/git/verification-tests/lib/manager.rb:35:in `clean_up'
      /home/anusaxen/git/verification-tests/features/support/env.rb:68:in `block (2 levels) in <top (required)>'
      /home/anusaxen/git/verification-tests/lib/base_helper.rb:131:in `capture_error'
      /home/anusaxen/git/verification-tests/features/support/env.rb:66:in `After'
 ```

This means the clean-up step does not exist. So clean-up process ends before the label clean-up kicks in. Also the whole test run finishes pre-maturely.

File: features/networking/multus.feature (hunk 8 above)
Comment (akostadinov) [reply features/networking/multus.feature]: btw could you let original commenter to mark conversations as resolved? If you mark, then it is hard for commenter to figure out which changes are to be reviewed.

File: features/networking/multus.feature [hunk 14]
```diff
@@ -606,3 +606,252 @@ Feature: Multus-CNI related scenarios
...
+      | ["metadata"]["name"] | pod1-vlan100 |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"]| bridgevlan100 |
+      | ["spec"]["nodeSelector"]["test"] | worker1 |
+    Then the step should succeed
+    And the pod named "pod1-vlan100" becomes ready
+    And evaluation of `pod.name` is stored in the :pod1 clipboard
+    And I execute on the pod:
+      | ifconfig | net1 |
+    Then the step should succeed
+    And evaluation of `@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :pod1_net1_ip clipboard
```
akostadinov: btw will this work for IPv6-only clusters that we are trying to support?

features/networking/multus.feature
anuragthehatter: @akostadinov Yep, my bad. SOmething bad might have happened on that cluster as lot of tests were tried on it. I optimized the label steps across whole PR and ran tests 3 times. Test ends maturely and seems good to me now. 1650 and 4452 explains that - http://pastebin.test.redhat.com/822512.

features/networking/multus.feature
anuragthehatter: @akostadinov We are not sure about that yet. These are interfaces inside pods (multus) so its too early to say whether they will support ipv6 or not. AFAIK they might use dual stack approach..ipv4 routing via ipv6. We can see for optimization in future is its required.

features/networking/multus.feature
```diff
@@ -606,3 +606,252 @@ Feature: Multus-CNI related scenarios
...
+    And I execute on the pod:
+      | ifconfig | net1 |
+    Then the step should succeed
+    And evaluation of `@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :pod2_net1_ip clipboard
+    
+    #Clean-up required to erase bridge interfaces created due to above pods on same node
+    Given I register clean-up steps:
+    """
+    the bridge interface named "mybridge" is deleted from the "<%= cb.nodes[0].name %>" node
+    the bridge interface named "mybridge.100" is deleted from the "<%= cb.nodes[0].name %>" node
```
akostadinov: clean-up steps can still be moved up, immediately before or after pod is created. There should be no harm inusing these steps even if the resource have never been created. The parameters here are bridge name and node name, and these are known even before pod is created.

Something that I'm more concerned with, I think I commented in another PR about this, is that we create and delete cluster global resources like VLANs and bridges. We delete here the bridges. And they have a hardcoded name.
So my assumption is, in case multiple scenarios run simultaneously (what we do) then they will interfere with each other, likely resulting in the whole test run to be terminated breaking the whole pipeline.

If scenarios need to run one after the other, they need to be tagged `@destructive`. If dynamic names are used that make sure no interference happens, then `@destructive` is not needed and we run scenarios in parallel.

So please review scenarios here and in your other PRs that they will not interfere with each other if run simultaneously.

@pruan-rht and @liangxia can help you understand `destructive` and clean-up process if I am not available to review or answer in the following weeks.

anuragthehatter: @akostadinov Yes i did mark them as destructive already yesterday as i also evaluated them as destructive in case. And i have adjusted the cleanup properly. The problem with cleanup is bridges does get create when a pod gets created not a net-attach-def. That's why i have defined cleanup at the end of all pods creation. But for readability i have moved them under specific net-attach-def and pod for now

PR #446, opened 2019-11-14
some fixes for CI testing

QiaolingTang:
@akostadinov PTAL. Since @anpingli has added post actions to create the `qe-app-registry` to the clusters launched by jenkins, I modify the script to use `qe-app-registry` when there has the resources. 
Besides, there are some changes when log in to the kibana console, I add a new action to click the `Log in with OpenShift ` button if this button exists.
PTAL, thanks a lot.

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -268,6 +268,11 @@ Given /^I create clusterlogging instance with:$/ do | table |
 
   @result = admin.cli_exec(:create, f: crd_yaml, n: logging_ns)
   raise "Unable to create clusterlogging instance" unless @result[:success]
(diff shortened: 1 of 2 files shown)


QiaolingTang:
Run log: https://url.corp.redhat.com/00645e3

anpingli:
@QiaolingTang  I had created operators in post actions.    I think we needn't use the 'qe-app-registry' any more.

QiaolingTang:
> @QiaolingTang I had created operators in post actions. I think we needn't use the 'qe-app-registry' any more.

I just set a default value for the catalog source name, if there doesn't have `qe-app-registry`, we can subscribe other sources.

PR #448, opened 2019-11-14
Ocp 22606

geliu2016:
@akostadinov as I sent mail to you, pls help to review this upgrade testcase. thanks

### DIFF
--- a/features/test/upgrade.feature
+++ b/features/test/upgrade.feature
@@ -4,7 +4,6 @@ Feature: basic verification for upgrade testing
   @users=upuser1,upuser2
   # @case_id OCP-10017000
   Scenario: upgrade-prepre cakephp example works well after migrate
-
     When I run the :new_project client command with:
       | project_name | project-ocp10017 |
     Then the step should succeed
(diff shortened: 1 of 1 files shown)


features/test/upgrade.feature
```diff
@@ -1,22 +1,58 @@
...
+    And the output should contain:
+      | project-ocp10017 |
+  
+  #@author geliu@redhat.com
+  #@users=upuser1,upuser2
+  #@case_id OCP-10017
+  @admin
+  Scenario: cakephp example works well after migrate
+    Given I switch to cluster admin pseudo user
+    Given I use the "project-ocp10017" project
```
pruan-rht: @geliu2016 is having the project enough of a check

geliu2016: @pruan-rht  aha, this is an example testcase, you may find this case is very old and inactive testcase, I active it for my test, and it is very simple for other team to understand and follow this case format, and also i need 2 automation cases to verify the auto works well for multiple testcase(multiple means number >= 2), sorry for have not add 'example' sign for it, and it will be deleted soon, thx

PR #453, opened 2019-11-19
Fix OCP-11613

liangxia:
@chao007 @duanwei33


liangxia:
@akostadinov @pruan-rht Ready for review.

PR #455, opened 2019-11-19
Automate test case OCP-26102

jhou1:


jhou1:
I've got more similar tests, will switch to scenario outline. Closing

PR #456, opened 2019-11-19
Uncordon node so pods are schedulable

liangxia:
In 4.x cluster, generally there are 3 nodes in 3 zones. When we drain node (node in zone1 with persistent volume in zone1 ), those pods are not schedulable as there are no available node in zone1.
This PR uncordon the node so that pods are back after drain-node.

@chao007 @duanwei33


pruan-rht:
lgtm, will merge once other team members give the 'OK'

liangxia:
@chao007 @duanwei33 Could you check ?

PR #458, opened 2019-11-19
Fix 4.x oauth team test cases steps

barleyer:
There are some changes on output.So we need to modify the output logs checking.


barleyer:
@xiaojiey

PR #461, opened 2019-11-21
Update to use ruby-25-* images to do docker build

xiuwang:
@openshift/devexp-qe Please help review this change, thanks~


wewang58:
LGTM

PR #471, opened 2019-11-27
proper file location when loading ruby files

akostadinov:


akostadinov:
tested by installer team

PR #477, opened 2019-11-28
cli output changed

pruan-rht:


pruan-rht:
@akostadinov PTAL it's urgent.  The format change has been there in my 4.2.x client... I'm worried that our slaves are still using 4.1.x client for testing since automation run with Jenkins has not seen this.  Thanks!

akostadinov:
I believe this is product bug. We need first to check if devs are going to revert. You don't just change this output from 4.2.x to 4.2.y

akostadinov:
Also, in case we want to change on our end, we should change how we get version in the first place. This is `self.get_version_for`. It is IMO a regression to change YAML output of version between minor releases (even if you ask me between major releases).

pruan-rht:
I'll update this PR to change the correct place today

pruan-rht:
@akostadinov, I've fixed the `prase_version` method as well.  PTAL.  Thanks!

pruan-rht:
https://github.com/openshift/verification-tests/pull/509

PR #479, opened 2019-11-28
allow name suffix when matching scenarios to allow upgrade prepare test use case

akostadinov:
/hold until tested


akostadinov:
tested by team

PR #487, opened 2019-12-03
Upgrade ocp 22606

geliu2016:
@akostadinov @pruan-rht This should be a normal upgrade testcase script based on our final decision, pls help to review, thanks!
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/49815/console
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/49818/console

### DIFF
--- a/features/upgrade/etcd/upgrade.feature
+++ b/features/upgrade/etcd/upgrade.feature
@@ -1,0 +1,33 @@
+Feature: basic verification for upgrade testing
+  # @author geliu@redhat.com
+  # @case_id OCP-22606
+  @upgrade-prepare
+  @users=upuser1,upuser2
+  Scenario: etcd-operator and cluster works well after upgrade 
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/admin/subscription.yaml |
+    Then the step should succeed
+    When I use the "openshift-operators" project
+    Then status becomes :running of exactly 1 pods labeled:
+      | name=etcd-operator-alm-owned |
+    When I use the "default" project
+    Given I download a file from "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/admin/etcd-cluster.yaml"
+    When I run the :create client command with:
+      | f | etcd-cluster.yaml |
+    Then the step should succeed
+    Then status becomes :running of exactly 3 pods labeled:
+      | etcd_cluster=example |
+  
+  # @author geliu@redhat.com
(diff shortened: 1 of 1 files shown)


features/upgrade/etcd/upgrade.feature
```diff
@@ -0,0 +1,35 @@
...
+    When I use the "default" project
+    Given I download a file from "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/admin/etcd-cluster.yaml"
+    When I run the :create client command with:
+      | f | etcd-cluster.yaml |
+    Then the step should succeed
+    Then status becomes :running of exactly 3 pods labeled:
+      | etcd_cluster=example |
+  
+  # @author geliu@redhat.com
+  # @case_id OCP-22606
```
akostadinov: I think we want `@case_id` tag only before the check scenario. Otherwise our tools might be confused.

features/upgrade/etcd/upgrade.feature
```diff
@@ -0,0 +1,35 @@
+Feature: basic verification for upgrade testing
+  # @author geliu@redhat.com
+  # @case_id OCP-22606
```
akostadinov: I think we want @case_id tag only before the check scenario. Otherwise our tools might be confused.

features/upgrade/etcd/upgrade.feature
```diff
@@ -0,0 +1,35 @@
...
+  @upgrade-prepare
+  @users=upuser1,upuser2
+  @admin
+  Scenario: etcd-operator and cluster works well after upgrade
```
akostadinov: needs suffix ` - prepare` so that Runner-v3 will work with this. We chatted about it.

geliu2016: updated.

features/upgrade/etcd/upgrade.feature
geliu2016: ok

geliu2016: I deleted @case_id tag for upgrade-prepare, and kept @author tag, I think this will help to trace the case author, 
and I also curious about @case_id tag have been masked by '#', it will be ignored by code, right?

akostadinov: It is used by script to automatically update polarion, see DOC-1125811. Please use script instead of manually writing the YAML. Much easier and you make sure your tags are correct.

PR #489, opened 2019-12-04
automate eventrouter related case

QiaolingTang:
Add step to install eventrouter and automate eventrouter related case @anpingli  PTAL

Runner-v3 log: https://url.corp.redhat.com/1398aca

### DIFF
--- a/features/logging/eventrouter.feature
+++ b/features/logging/eventrouter.feature
@@ -1,0 +1,57 @@
+@clusterlogging
+@commonlogging
+Feature: eventrouter related test
+
+  # @author qitang@redhat.com
+  # @case_id OCP-21389
+  @admin
+  @destructive
+  Scenario: Deploy Openshift Event following Docs
+    Given I switch to the first user
+    Given I create a project with non-leading digit name
+    Then I run the :new_app client command with:
+      | app_repo | httpd-example |
+    Given I switch to cluster admin pseudo user
+    Given I use the "openshift-logging" project
+    Given logging eventrouter is installed in the cluster
+    Given a pod becomes ready with labels:
+      | component=eventrouter |
+    When I run the :logs admin command with:
+      | resource_name | <%= pod.name %>   |
+      | namespace     | openshift-logging |
+      | since         | 5m                |
+    Then the step should succeed
+    And the output should contain:
+      | "verb":  |
+      | "event": |
+      | "reason" |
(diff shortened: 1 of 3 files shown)


features/step_definitions/logging.rb
```diff
@@ -359,3 +359,22 @@
     step %Q/logging operators are installed successfully/
   }
 end
+
+Given /^logging eventrouter is installed in the cluster$/ do
+  step %Q/I switch to cluster admin pseudo user/
+  step %Q/I use the "openshift-logging" project/
```
liangxia: Duplicate steps
```
    Given I use the "openshift-logging" project
    Given logging eventrouter is installed in the cluster
```

features/logging/eventrouter.feature
```diff
@@ -0,0 +1,30 @@
+@clusterlogging @commonlogging
+Feature: eventrouter related test
+
+  # @author qitang@redhat.com
+  # @case_id OCP-21389
+  @admin @destructive
+  Scenario: Deploy Openshift Event following Docs
+    Given I switch to the first user
```
liangxia: Unnecessary step ?

features/step_definitions/logging.rb
```diff
@@ -359,3 +359,22 @@
...
+  unless deployment('eventrouter').exists?
+    @result = admin.cli_exec(:new_app, file: "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/eventrouter/internal_eventrouter.yaml")
+    unless @result[:success]
```
liangxia: Always do the clean-up not matter the above creation succeed or not, or we will left some resources there when met with partial success/failure.

features/step_definitions/logging.rb
QiaolingTang: I'll remove the steps  in the feature file
```Given I switch to cluster admin pseudo user
    Given I use the "openshift-logging" project```

features/logging/eventrouter.feature
QiaolingTang: I think no, because when I deploy logging resources, I have to use the cluster-admin user, so here if I don't switch user, the user would be a cluster-admin user.

QiaolingTang:
@anpingli @pruan-rht could you please help to review this PR? Thanks.

PR #491, opened 2019-12-04
rule fixes

yapei:
@xiaocwan  It looks like your auto cases OCP-24287, OCP-24316 are missing rules and wrong libs. I updated in this PR FYI
@yanpzhan  Can you help review? runner-v3 job is scheduled but not started yet #5​01​50


xiaocwan:
Cool, thanks for updating 4.3. @yapei

yanpzhan:
@yapei The update LGTM

Issue #492, opened 2019-12-04
result contain 'STDERR' information when host run command with 'oc debug'


zhaozhanqi:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/50163/console

see above logs:
I want to run command `'/sbin/ip\ route\ show\ default\ \|\ awk\ \'/default/\ \{print\ \$5\}\'` on host.  and get the result `ens5`, but the `Result` also included the following information:
```
      STDERR:
      Starting pod/ip-10-0-51-233us-east-2computeinternal-debug ...
      To use host binaries, run `chroot /host`
      
      Removing debug pod ...
```

@akostadinov

zhaozhanqi:
the logs should be removed
so I describe the issue again:
when execute `@result = host.exec_admin("/sbin/ip route show default | awk '/default/ {print $5}'")` , it will invoke 'oc debug node/xxxx' '/sbin/ip\ route\ show\ default\ \|\ awk\ \'/default/\ \{print\ \$5\}\' , but the output also contain the error:
```
      STDERR:
      Starting pod/ip-10-0-51-233us-east-2computeinternal-debug ...
      To use host binaries, run `chroot /host`
      
      Removing debug pod ...
```
So the return result include above error, it's not unexpected. Can we have a way to remove above error from the result.

akostadinov:
https://bugzilla.redhat.com/show_bug.cgi?id=1771549

akostadinov:
You will have to perhaps just call `ip route show default` and then parse in a ruby step.
```
evaluation of `...` is stored in the :ip clipboard
```

PR #502, opened 2019-12-06
update multus failed cases

zhaozhanqi:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/50818/console

@weliang1 @anuragthehatter @huiran0826 @rbbratta 

1. 
`| ["spec"]["config"]["eth0"]| <%= cb.default_interface %> |` cannot be parsed for now, and I did not find a way to replace one parameter (`master`) in `config`. so here I replace the whole `config` with specfied '<%= cb.default_interface %>'

2. 
```
cb[cb_name] = @result[:response].chomp.split[0] 
```
is for issue https://github.com/openshift/verification-tests/issues/492

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -8,18 +8,17 @@ Feature: Multus-CNI related scenarios
     Given the master version >= "4.0"
     And the multus is enabled on the cluster
     Given the default interface on nodes is stored in the :default_interface clipboard
-    And evaluation of `node.name` is stored in the :target_node clipboard    
+    And evaluation of `node.name` is stored in the :target_node clipboard
     # Create the net-attach-def via cluster admin
     Given I have a project
     When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml" replacing paths:
       | ["metadata"]["namespace"] | <%= project.name %> |    
-      | ["spec"]["config"]["eth0"]| <%= cb.default_interface %> |
-      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |      
+      | ["spec"]["config"]| '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "<%= cb.default_interface %>","mode": "bridge", "ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100", "rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1" } }' |
     Then the step should succeed
 
     # Create the first pod which consumes the macvlan custom resource
-    When I run the :create client command with:
-      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-bridge.yaml |
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-bridge.yaml" replacing paths:
+      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |
     Then the step should succeed
     And a pod becomes ready with labels:
       | name=macvlan-bridge-pod |
@@ -42,7 +41,7 @@ Feature: Multus-CNI related scenarios
 
     # Create the second pod which consumes the macvlan cr
     When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-bridge.yaml" replacing paths:
-      | ["spec"]["nodeName"] | "<%= cb.pod_node %>" |
+      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |
     Then the step should succeed
     And 2 pods become ready with labels:
       | name=macvlan-bridge-pod |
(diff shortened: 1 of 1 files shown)


features/step_definitions/networking.rb
```diff
@@ -680,7 +680,7 @@
   cb_name = "interface" unless cb_name
   @result = host.exec_admin("/sbin/ip route show default | awk '/default/ {print $5}'")
   raise "Failed to get the default interface of node" unless @result[:success]
-  cb[cb_name] = @result[:response].chomp
+  cb[cb_name] = @result[:response].chomp.split[0]
```
anuragthehatter: This is not advised due to oc debug randomly breaking connections. This is being taken care of in #432 where we replace this step to fetch default interface name from sdn pod instead of oc debug node. I will finish my comments on #432 today and that should solve this issue

zhaozhanqi: @anuragthehatter Thanks the advice. 
but if we use 'sdn' pod to run 'ip route' then this kind of case will cannot be run for OVN plugin. So we better to find a reason why oc debug randomly breaking connections.

features/networking/multus.feature
```diff
@@ -160,7 +155,7 @@ Feature: Multus-CNI related scenarios
 
     # Create the second pod which consumes the macvlan cr
     When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-vepa.yaml" replacing paths:
-      | ["spec"]["nodeName"] | "<%= cb.pod_node %>" |
+      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |
```
liangxia: Useless spaces added at the end of line.

features/step_definitions/networking.rb
anuragthehatter: @zhaozhanqi  Due to these limitations, oc debug is not advisable and guess we can keep our automation rolling while these bugs are still being fixed.
https://bugzilla.redhat.com/show_bug.cgi?id=1771552
https://bugzilla.redhat.com/show_bug.cgi?id=1771549
https://bugzilla.redhat.com/show_bug.cgi?id=1771553

On OVN cluster, ovnkube-node will give same ip route output like sdn-pod. So i can go ahead and adjust this def react accordingly as per network type. WDYT? like if its openshift-sdn go inside sdn pod and if its OVN, go inside ovnkube-node?

anuragthehatter: Adjustment will look like this 

Given /^the default interface on nodes is stored in the#{OPT_SYM} clipboard$/ do |cb_name|
  ensure_admin_tagged
  step "I select a random node's host"
  cb_name ||= "interface"

  @result = _admin.cli_exec(:get, resource: "network.operator", template: '{{.items[*].spec.defaultNetwork.type}}')
  if @result[:success] then
        networkType = @result[:response].strip
  case networkType
    when "OVNKubernetes"
      step %Q/I run command on the node's ovnkube pod:/, table("| bash | -c | ip route show default |")
    when "OpenShiftSDN"
      step %Q/I run command on the node's sdn pod:/, table("| bash | -c | ip route show default |")
    else
      raise "unknown networkType"
  end
  cb[cb_name] = @result[:response].split("\n").first.split(/\W+/)[7]
  logger.info "The node's default interface is stored in the #{cb_name} clipboard."
end

zhaozhanqi: @anuragthehatter seems this LGTM

huiran0826: @anuragthehatter  I think this is a good idea to make it work for both sdn and ovn.   But I thought if we make this adjustment in step like "the default interface on nodes is stored... ",  there are maybe some other steps also need this kind change, even in the case level. some cases just call " I run command on the node's sdn pod" directly.

Can we update this adjustment  in the step "I run command on the#{OPT_QUOTED} node's sdn pod" like:

Given /^I run command on the#{OPT_QUOTED} node's sdn pod:$/ do |node_name, table|
  ensure_admin_tagged
  network_cmd = table.raw
  node_name ||= node.name
  
  @result = _admin.cli_exec(:get, resource: "network.operator", template: '{{.items[*].spec.defaultNetwork.type}}')
 if @result[:success] then
 networkType = @result[:response].strip
 case networkType
   when "OVNKubernetes"
   sdn_pod = BushSlicer::Pod.get_labeled("app=ovnkube", project: project("openshift-OVNKubernetes", switch: false), user: admin) { |pod, hash|
     pod.node_name == node_name
    }.first
   when "OpenShiftSDN"
  sdn_pod = BushSlicer::Pod.get_labeled("app=sdn", project: project("openshift-sdn", switch: false), user: admin) { |pod, hash|
    pod.node_name == node_name
   }.first
   else
      raise "unknown networkType"
  end
   cache_resources sdn_pod
   @result = sdn_pod.exec(network_cmd, as: admin)
 end

anuragthehatter: @huiran0826 @zhaozhanqi yep, i am drilling down some steps which required such changes to support both networktypes. 
Anyways to your question i already have a step like
```
Given /^I run command on the#{OPT_QUOTED} node's ovnkube pod:$/ do |node_name, table|
```
Which will be more explanatory in terms of what type of cluster we have and will be reusable at lot of places in future. Lets keep sdn pod one as is in my view.

features/networking/multus.feature
zhaozhanqi: Thanks @liangxia

features/step_definitions/networking.rb
huiran0826: @anuragthehatter  Actually in the case, we have lots of steps like these: 
When I run command on the node's sdn pod:
If we cannot update step definition of "I run command on the#{OPT_QUOTED} node's sdn pod" to let it support both sdn and ovn, it will make it complicated to update this kind of case to support ovn.

zhaozhanqi: Thanks the advice. How about file another PR to update `I run command on the#{OPT_QUOTED} node's sdn pod` this step to make it can support OVN.
I'd like merge this ASAP, since these cases still in an abyss of misery for current auto run

huiran0826: It okay for me

anuragthehatter: requested to merge above code in #432 ASAP. Tested above code works fine with both networktypes. I will continue looking at OVN integeration keeping above steps in mind for now. Thanks

anuragthehatter: @zhaozhanqi  Okay so #432 is merged and that would take care of it now. You may want to pull back your changes from networking.rb? @huiran0826 I agree with your suggestion and that something i am evaluating. We definitely need above ovnkube pod step as well for standalone OVN tests( which won't be running for Openshift-sdn) to make code more readable in conjunction with OVN. Also a change in sdn pod step would be coming as well to support OVN. As a matter of fact, as per evaluation, more than 95% of tests(which are using this sdn pod step) wont be successful on OVN due to nature of difference in OVN outputs and change in ovs pmtrs.

zhaozhanqi: thanks, I rebased the latest code

zhaozhanqi:
@pruan-rht @chao007

PR #509, opened 2019-12-10
accomodate different cli prefix

pruan-rht:


pruan-rht:
For bug https://bugzilla.redhat.com/show_bug.cgi?id=1781909

PR #515, opened 2019-12-11
fix stderr output of shell commands in installer

akostadinov:
not sure how nobody noticed that STDERR was missing in shell commands


pruan-rht:
@akostadinov tested with this build https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Launch%20Environment%20Flexy/74371/console . I think it's ok to merge

PR #517, opened 2019-12-12
Some tweaks/addition to support OVN, more to come

anuragthehatter:

The changes have been tested on both OpenShiftSDN and OVNKubernetes clusters


anuragthehatter:
/hold
This might not work for 3.10-311. I will take a look at it tomorrow

features/step_definitions/networking.rb
```diff
@@ -745,4 +777,12 @@
...
     raise "Failed to execute network command!" unless @result[:success]
   end
-end
\ No newline at end of file
+end
+
+Given /^the env is using "([^"]*)" networkType$/ do |network_type|
+  ensure_admin_tagged
+  _admin = admin
+  step "I select a random node's host"
```
zhaozhanqi: is this step needed?

anuragthehatter:
> /hold
> This might not work for 3.10-311. I will take a look at it tomorrow

/hold cancel
Removed unnecessary 3.x steps as these changes will land in Master branch not v3. PTAL. Thanks

features/step_definitions/networking.rb
anuragthehatter: I believe this will be used as a gating in upcoming OVN specific test cases automation and will always be a check to make sure cluster we are using is indeed on OVN

zhaozhanqi: aha,misunderstood. I mean `step "I select a random node's host"` is necessary for this

features/step_definitions/networking.rb
```diff
@@ -664,15 +657,24 @@
...
+  _admin = admin
+   @result = _admin.cli_exec(:get, resource: "network.operator", output: "jsonpath={.items[*].spec.defaultNetwork.type}") 
+  if @result[:response] == "OpenShiftSDN"
+     sdn_pod = BushSlicer::Pod.get_labeled("app=sdn", project: project("openshift-sdn", switch: false), user: admin) { |pod, hash|
+       pod.node_name == node_name
+     }.first
+     cache_resources sdn_pod
+     @result = sdn_pod.exec(network_cmd, as: admin)
+  else
+     ovnkube_pod = BushSlicer::Pod.get_labeled("app=ovnkube-node", project: project("openshift-ovn-kubernetes", switch: false), user: admin) { |pod, hash|
```
zhaozhanqi: I do not think we have same result when run ovs command between sdn and OVN.  So do we need to change this ? we can add new step for OVN. since this will affect many version. if you intend to do this, please make sure 3.11 sdn and 4.x sdn and 4.3 OVN all works well.

anuragthehatter: > I do not think we have same result when run ovs command between sdn and OVN. So do we need to change this

I made this step inter-operable btw ovn/sdn clusters. The existing automated test cases such as ocp-18535 will continue run for ovn as well and this is host step so any ovn test want to execute ovs cmmand on host can use this step.


> please make sure 3.11 sdn and 4.x sdn and 4.3 OVN all works well

This wont impact 3.x as 3.x definition will be fetched from v3 branch. Above code is going into master which points to 4,1+ and hence will only applicable for 4.1+ tests.

anuragthehatter: @zhaozhanqi Can we merge this? This could will be leveraged in multus OVN testing.

features/step_definitions/networking.rb
```diff
@@ -1,29 +1,22 @@
...
-      pod.node_name == node.name
-    }.first
-    container_id = ovs_pod.containers.first.id
-    ovs_cmd = "runc exec #{container_id} " + ovs_cmd
+  @result = _admin.cli_exec(:get, resource: "network.operator", output: "jsonpath={.items[*].spec.defaultNetwork.type}")
+  if @result[:response] == "OpenShiftSDN" 
+     ovs_pod = BushSlicer::Pod.get_labeled("app=ovs", project: project("openshift-sdn", switch: false), user: admin) { |pod, hash|
+        pod.node_name == node.name
+     }.first
+     container_id = ovs_pod.containers.first.id
```
zhaozhanqi: thanks your updating this.  Could you help show the logs when using this step. I doubt the `container_id` is not invoking.

anuragthehatter: @zhaozhanqi Long story with my git stash pop. This step doesn't need modification at the moment. Works fine for sdn and ovn. However another steps modification like "node's sdn pod" is important at the moment. Pls let me know if you have any questions over there.

features/step_definitions/networking.rb
zhaozhanqi: LGTM, thanks

features/step_definitions/networking.rb
```diff
@@ -1,7 +1,7 @@
 Given /^I run the ovs commands on the host:$/ do | table |
   ensure_admin_tagged
   _host = node.host
-
+
```
liangxia: Intended ?

liangxia:
@anuragthehatter PR needs rebase.

features/step_definitions/networking.rb
anuragthehatter: Nope, fixed it, Thanks

xingxingxia:
This broken https://coreos.slack.com/archives/CPJQCSN49/p1577069066006800?thread_ts=1577064941.006300&cid=CPJQCSN49

PR #518, opened 2019-12-12
Fix the rhel8 images auto

xiuwang:

### DIFF
--- a/features/images/rhel8_images.feature
+++ b/features/images/rhel8_images.feature
@@ -1,0 +1,115 @@
+Feature: rhel8images.feature
+
+  # @author xiuwang@redhat.com
+  # @case_id OCP-22950
+  @admin
+  Scenario: Using new-app cmd to create app with ruby rhel8 image
+    Given I have a project
+    When I run the :tag admin command with:
+      | source           | registry.redhat.io/rhel8/ruby-25:latest |
+      | dest             | qe-ruby-25-rhel8:latest                 |
+      | reference_policy | local                                   |
+      | n                | openshift                               |
+    Then the step should succeed
+    When I run the :new_app client command with:
+      | image_stream | openshift/qe-ruby-25-rhel8                   |
+      | app_repo     | https://github.com/sclorg/s2i-ruby-container |
+      | context_dir  | 2.5/test/puma-test-app                       |
+      | name         | ruby25rhel8                                  |
+    Then the step should succeed
+    And the "ruby25rhel8-1" build completed
+    And a pod becomes ready with labels:
+      | deployment=ruby25rhel8-1 |
+    When I run the :logs client command with:
+      | resource_name | <%= pod.name %> |
+    Then the output should contain:
+      | Min threads: 0, max threads: 16 |
+    When I run the :set_env client command with:
+      | e        | PUMA_MIN_THREADS=1  | 
+      | e        | PUMA_MAX_THREADS=12 |
+      | e        | PUMA_WORKERS=5      |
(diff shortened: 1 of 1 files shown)


features/images/rhel8_images.feature
```diff
@@ -0,0 +1,115 @@
+Feature: rhel8images.feature
+
+  # @author xiuwang@redhat.com
+  # @case_id OCP-22950
+  @admin
+  Scenario: Using new-app cmd to create app with ruby rhel8 image
+    Given I have a project
+    When I run the :tag admin command with:
+      | source           | registry.redhat.io/rhel8/ruby-25:latest |
+      | dest             | ruby-25-rhel8:latest                    |
```
liangxia: What will happen if the image stream with the same name already exist ?

xiuwang: Just override the existing imagestream without error.

liangxia: Then it is not acceptable.
We may have that image stream integrate someday. And then this scenario will override default/existing imagestream.

Perhaps adding qe as prefix or suffix to distinguish that or you can find a better way if you do.

wzheng1: I don't think it is necessary to add any xxfix, since there is already a projected created, there is an order to recognize imagestream. the top priority is to find it under its own project; default one is under Openshift namespace.

xiuwang: @lxia Thanks your advice, update as comment, please help review again, thx

features/images/rhel8_images.feature
```diff
@@ -0,0 +1,115 @@
+Feature: rhel8images.feature
+
+  # @author xiuwang@redhat.com
+  # @case_id OCP-22950
+  @admin
+  Scenario: Using new-app cmd to create app with ruby rhel8 image
+    Given I have a project
+    When I run the :tag admin command with:
+      | source           | registry.redhat.io/rhel8/ruby-25:latest |
+      | dest             | qe-ruby-25-rhel8:latest                    |
```
liangxia: You have to fix the align now

xiuwang: updated

features/images/rhel8_images.feature
```diff
@@ -0,0 +1,115 @@
+Feature: rhel8images.feature
+
+  # @author xiuwang@redhat.com
+  # @case_id OCP-22950
+  @admin
+  Scenario: Using new-app cmd to create app with ruby rhel8 image
+    Given I have a project
+    When I run the :tag admin command with:
+      | source           | registry.redhat.io/rhel8/ruby-25:latest |
+     | dest             | qe-ruby-25-rhel8:latest                  |
```
liangxia: Sorry, bad luck again.

xiuwang: Updated, thanks~

lib/rules/cli/4.0.yaml
```diff
@@ -1012,6 +1012,7 @@
     :reference_policy: --reference-policy=<value>
     :scheduled: --scheduled=<value>
     :source_type: --source=<value>
+    :n: --namespace=<value>
```
pruan-rht: @xiuwang, I don't think you need this because `n` and `namespace` are both under the `global:` section.  https://github.com/openshift/verification-tests/blob/master/lib/rules/cli/4.0.yaml#L14-L15

xiuwang: Peter, thanks, I have removed this option.

PR #524, opened 2019-12-13
test will fail w/o @admin tag...

pruan-rht:


liangxia:
We are adding admin role to some users in the scenario, but didn't register clean-up. Failing this scenario will leave some users there with admin role, which will break following tests (Clean up projects with user which has admin role is too dangerous)

PR #526, opened 2019-12-13
Fix OCP-12400

xiuwang:


wzheng1:
Thanks for the implementation! LGTM

features/step_definitions/registry.rb
```diff
@@ -71,9 +71,9 @@
   end
   begin
     step %Q/I switch to cluster admin pseudo user/
-    step %Q/I use the "default" project/
+    step %Q/I use the "openshift-image-registry" project/
```
liangxia: The project name change works for all 4.x versions ?

xiuwang: For now, it's yes

liangxia: ack

PR #527, opened 2019-12-16
Add CLI 4.4

liangxia:
Changes made in this PR,
- Link 4.4 to 4.3, to support 4.4
- Move 4.0 to 4.1, as we do not have 4.0
- Remove 3.0/3.1/3.2, as it is EOF and no longer used


liangxia:
@akostadinov @pruan-rht  PTAL

PR #535, opened 2019-12-17
remove unnecessary steps

QiaolingTang:
Removing unnecessary steps.
@pruan-rht could you please help to merge it ? Thanks.

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -4,6 +4,7 @@
 ### none configurable, just use default parameters
 Given /^logging service has been installed successfully$/ do
   ensure_destructive_tagged
+  ensure_admin_tagged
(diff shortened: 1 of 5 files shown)


features/step_definitions/logging.rb
```diff
@@ -286,7 +286,6 @@
 
 Given /^I delete the clusterlogging instance$/ do
   ensure_admin_tagged
-  ensure_destructive_tagged
```
pruan-rht: @QiaolingTang why remove the `destructive` tags?

QiaolingTang: @pruan-rh I thought deleting the clusterlogging instance wouldn't affect the OCP cluster. But since deleting the CL instance could affect other logging scenarios, I'll add it back.  Thanks.

PR #537, opened 2019-12-17
OCP-24918

anuragthehatter:
@zhaozhanqi  @weliang1 @huiran0826 @rbbratta 
http://pastebin.test.redhat.com/822614

### DIFF
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -195,10 +195,9 @@ Feature: Operator related networking scenarios
     | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/list_for_pods.json |
   Then the step should succeed
   And a pod becomes ready with labels:
-      | name=test-pods |
+    | name=test-pods |
   Given I use the "test-service" service
   And evaluation of `service.ip(user: user)` is stored in the :service_ip clipboard
-  And I wait for the "test-service" service to become ready
   # Checking idling unidling manually to make sure it works fine before inducing flag feature
   When I run the :idle client command with:
     | svc_name | test-service |
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -210,7 +210,7 @@ Feature: Operator related networking scenarios
   Then the step should succeed
   And the output should contain:
     | Hello OpenShift |
-  #Inducing flag disablement here
+  #Inducing flag disablement here an wait 60 seconds for CNO to update it across the nodes
   Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with: 
     | {"spec":{"defaultNetwork":{"openshiftSDNConfig":{"enableUnidling" : false}}}} |
   # Cleanup required to move operator config back to normal
@@ -220,8 +220,8 @@ Feature: Operator related networking scenarios
     | {"spec":{"defaultNetwork":{"openshiftSDNConfig": null}}} |
   60 seconds have passed
   """
+  And 60 seconds have passed
   #We are idling service again and making sure it doesn't get unidle due to the above enableUnidling flag set to false
-  Given 60 seconds have passed
   When I run the :idle client command with:
     | svc_name | test-service |
   Then the step should succeed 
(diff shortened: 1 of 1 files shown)


features/networking/operator.feature
```diff
@@ -186,3 +186,57 @@ Feature: Operator related networking scenarios
...
+  Scenario: Service should not get unidle when config flag is disabled under CNO
+  Given I have a project
+  When I run the :create client command with:
+    | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/list_for_pods.json |
+  Then the step should succeed
+  And a pod becomes ready with labels:
+      | name=test-pods |
+  Given I use the "test-service" service
+  And evaluation of `service.ip(user: user)` is stored in the :service_ip clipboard
+  And I wait for the "test-service" service to become ready
```
weliang1: Should "And I wait for the "test-service" service to become ready", then
"And evaluation of `service.ip(user: user)` is stored in the :service_ip clipboard"?

features/networking/operator.feature
```diff
@@ -186,3 +186,57 @@ Feature: Operator related networking scenarios
...
+    | {"spec":{"defaultNetwork":{"openshiftSDNConfig":{"enableUnidling" : false}}}} |
+  # Cleanup required to move operator config back to normal
+  Given I register clean-up steps:
+  """
+  as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with: 
+    | {"spec":{"defaultNetwork":{"openshiftSDNConfig": null}}} |
+  60 seconds have passed
+  """
+  #We are idling service again and making sure it doesn't get unidle due to the above enableUnidling flag set to false
+  Given 60 seconds have passed
```
weliang1: Including above 60 seconds, does this mean 120 seconds passed?

anuragthehatter: above 60 seconds is a cleanup step which will execute after test finished up.

features/networking/operator.feature
anuragthehatter: We actually don't need to check service become ready step. Removed it for now. As if svc is not ready, it won't get IP assigned and it will eventually fail at evaluation step

features/networking/operator.feature
```diff
@@ -186,3 +186,56 @@ Feature: Operator related networking scenarios
...
+  Then the step should succeed 
+  And the output should contain:
+    | The service "<%= project.name %>/test-service" has been marked as idled |
+  When I execute on the "hello-pod" pod:
+    | /usr/bin/curl | --connect-timeout | 10 | <%= cb.service_ip %>:27017 |
+  Then the step should fail
+  #Moving CNO config back to normal and expect service to unidle then
+  Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with: 
+    | {"spec":{"defaultNetwork":{"openshiftSDNConfig": null}}} |
+  Given 60 seconds have passed
```
pruan-rht: @anuragthehatter instead of hard wait, can we check the `hello-pod` is ready instead?

anuragthehatter: @pruan-rht ```line 207``` is creating hello-pod actually and that should check the pod readiness. May be i can add a comment there saying that this will create a hello-pod?

pruan-rht: @anuragthehatter hard-code wait is always frowned upon...that's why I'm asking if there's an alternative.

anuragthehatter: @pruan-rht ah, there is confusion. This 60 seconds wait is for patch to be updated across the nodes by CNO. I clarified more providing comments and changed ```Given 60 seconds have passed``` to ```And 60 seconds have passed``` so it explains that its for patch step not hello-pod

pruan-rht: @anuragthehatter ok for now, but it would be good to into if it's possible to somehow verify the patch across the nodes by CNO.

anuragthehatter: Thanks Peter for suggestion but currently the only way to check whether network operator has taken patch successfully is to see message like "patch edited successfully" on CLI prompt and thats what we have been following as per dev. The operator logs is the other way which shows config has been reconciled across sdn pods but its not straight away easy as logs under operator has different parts and it wont be easy to collect different parts from logs and then verdict out that config has been reconciled across cluster. But agree that we can suggest something to dev to produce specifically in log or CLI

PR #533, opened 2019-12-17
new step to get available router ip

lihongan:
The step `I store default router IPs in the#{OPT_SYM} clipboard` cannot work for private cluster since no public DNS for route.
In the new step, we execute curl command on the pod inside the cluster so can solve above problem.
Actually, nslookup command can be used here but using curl can workaround the DNS misconfiguration issue for few clusters (vSphere/Baremetal and adding RHEL worker). When the DNS issue resolved then we can change back to use nslookup command.

PR #541, opened 2019-12-18
update all related cases to use the new step [follow up #533]

lihongan:
This one just a follow up PR of https://github.com/openshift/verification-tests/pull/533


lihongan:
@pruan-rht PTAL, THX

PR #543, opened 2019-12-18
fixed OCP-23543 to alleviate STDERR issues on debug pods

anuragthehatter:
@zhaozhanqi This will escape STDERR issues we are seeing for this test case on debug steps steps. We just grepping version data now etc to make things work irrespective of whether there is STDERR or not


anuragthehatter:
@zhaozhanqi http://pastebin.test.redhat.com/824682
PTAL. This will work fine now irrespective of whether we get STDERR or not

anuragthehatter:
duplicate of #544

PR #545, opened 2019-12-19
add methods and lib files for logging cases

QiaolingTang:
Adding methods and lib files for logging cases.
@pruan-rht could you please help to merge this PR? Thanks.

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -281,7 +281,6 @@ Given /^I create clusterlogging instance with:$/ do | table |
   # to wait for the status informations to show up in the clusterlogging instance
   sleep 10
   step %Q/I wait for clusterlogging with "#{log_collector}" log collector to be functional in the project/
-
 end
 
 Given /^I delete the clusterlogging instance$/ do
@@ -350,6 +349,8 @@ Given /^logging channel name is stored in the#{OPT_SYM} clipboard$/ do | cb_name
     cb[cb_name] = "4.2"
   when cluster_version('version').version.include?('4.3.')
     cb[cb_name] = "4.3"
+  when cluster_version('version').version.include?('4.4.')
+    cb[cb_name] = "4.4"
   end
 end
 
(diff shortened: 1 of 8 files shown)


QiaolingTang:
@pruan-rht could you please help to review this PR? Thanks.

features/step_definitions/logging.rb
```diff
@@ -386,3 +387,140 @@
...
   })
 end
+
+Given /^I generate certs for the#{OPT_QUOTED} receiver(?: in the#{OPT_QUOTED} project)?$/ do | receiver_name, project_name |
+  script_file = "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/logging/logforwarding/cert_generation.sh"
+  if project_name
+    proj = project_name
+  else
+    proj = "openshift-logging"
+  end
```
pruan-rht: @QiaolingTang, we can use the ruby shorthand `proejct_name ||= "openshift-logging"` and then use `project_name` instead of the temp variable `proj`

features/step_definitions/logging.rb
```diff
@@ -386,3 +387,140 @@
...
+  step %Q/I download a file from "#{script_file}"/
+  shell_cmd = "sh cert_generation.sh $(pwd) #{proj} #{receiver_name}"
+  system(shell_cmd)
+end
+
+Given /^fluentd receiver is deployed as (secure|insecure)(?: in the#{OPT_QUOTED} project)?$/ do | security, project_name |
+  if project_name
```
pruan-rht: @QiaolingTang same thing here.

QiaolingTang:
Steps testing pass log: https://url.corp.redhat.com/36da3d4

features/step_definitions/logging.rb
```diff
@@ -386,3 +387,128 @@
...
+
+Given /^fluentd receiver is deployed as (secure|insecure)(?: in the#{OPT_QUOTED} project)?$/ do | security, project_name |
+  project_name ||= "openshift-logging"
+
+  org_user = user
+  project(project_name)
+  step %Q/I ensures "fluentdserver" service_account is deleted from the "<%= project.name %>" project after scenario/
+  step %Q/I ensures "fluentdserver" config_map is deleted from the "<%= project.name %>" project after scenario/
+  step %Q/I ensures "fluentdserver" deployment is deleted from the "<%= project.name %>" project after scenario/
+  step %Q/I ensures "fluentdserver" service is deleted from the "<%= project.name %>" project after scenario/
```
pruan-rht: @QiaolingTang I see duplication of these 4 steps, perhaps we can merged them into a single step

QiaolingTang: Thanks @pruan-rht , I added a new single step to remove these resources, could you please review this PR again? If there is no other concerns, could you please help to merge it ?

QiaolingTang:
> @QiaolingTang I see these duplications, so I would recommend put that into a common step to save typing in the future.

Which duplication do you mean ?

QiaolingTang:
Hi @pruan-rht , I think I have resolved all the duplication, could you please help to review it again? 
Here is the pass log: https://url.corp.redhat.com/6246225

lib/openshift/cron_job.rb
```diff
@@ -10,5 +10,33 @@ def schedule(user: nil, cached: true, quiet: false)
...
+      specs = containers_spec(user: user, cached: cached, quiet: quiet)
+      target_spec = {}
+      specs.each do | spec |
+        target_spec = spec if spec.name == name
+      end
+      raise "No container spec found matching '#{name}'!" if target_spec.is_a? Hash
+      return target_spec
+    end
+
+    def node_selector(user: nil, cached: true, quiet: false)
```
pruan-rht: @QiaolingTang there are multiple `node_selector` methods defined in this PR, we should look into perhaps we can define that in the base-class so we don't have to keep defining the same thing over and over for new sub-classes.  WDYT @akostadinov ?

features/step_definitions/logging.rb
pruan-rht: @QiaolingTang I don't see this unified step you mentioned...can you pls point me to it?

QiaolingTang: @pruan-rht  https://github.com/openshift/verification-tests/pull/545/files#diff-0f0202f69f619cc471255d933a81861cR427-R455

lib/openshift/cron_job.rb
QiaolingTang: @pruan-rht which base-class do you mean? The json path of `node_selector` in different resources may not the same.

PR #547, opened 2019-12-19
fix  OCP-24316  OCP-21772

SSshahan:
@yapei please help to review, thanks

### DIFF
--- a/lib/rules/web/admin_console/4.0/projects.xyaml
+++ b/lib/rules/web/admin_console/4.0/projects.xyaml
@@ -8,7 +8,7 @@ create_project:
   action:
     if_param: 
     ref: set_project_description
-  action: click_create_button
+  action: submit_changes
 set_project_name:
   element:
     selector:
--- a/lib/rules/web/admin_console/4.2/customizations.xyaml
+++ b/lib/rules/web/admin_console/4.2/customizations.xyaml
@@ -24,7 +24,7 @@ upload_file:
 check_console_notification:
   elements:
   - selector:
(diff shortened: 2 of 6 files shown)


lib/rules/web/admin_console/4.0/projects.xyaml
```diff
@@ -8,7 +8,7 @@ create_project:
   action:
     if_param: 
     ref: set_project_description
-  action: click_create_button
+  action: submit_changes
```
SSshahan: The previous action we used will locate three elements

lib/rules/web/admin_console/4.2/customizations.xyaml
```diff
@@ -24,7 +24,7 @@ upload_file:
 check_console_notification:
   elements:
   - selector:
-      xpath: //div[@class='co-global-notification' and @data-test='<name>-<location>']//.[contains(text(), '<notification_text>')]
+      xpath: //div[@class='co-global-notification' and @data-test='<name>-<location>']//*[contains(text(), '<notification_text>')]
```
SSshahan: fix `cannot locate element ` error

lib/rules/web/admin_console/4.2/projects.xyaml
```diff
@@ -8,7 +8,11 @@ create_project:
   action:
     if_param: discription_name
     ref: set_project_description
-  action: click_create_button
+  action: submit_changes
+  element:
+    selector:
+      xpath: //span[contains(text(),'<project_name>')]
+    timeout: 30
```
SSshahan: Here have requests race condition, so we should promise the project created successfully after then perform next step.

PR #557, opened 2019-12-24
Add steps for OCP-26732

xiuwang:


xiuwang:
> @xiuwang do we really need steps for getting these parameters? It may look easier to just call the step
> 
> > evaluation of `daemon_set(ds_name).generation_number` is stored in the :xxx clipboard
@pruan-rht Thanks, your way is work and more easier, I have removed my steps. Could you help review again and https://github.com/openshift/cucushift/pull/7277/

PR #556, opened 2019-12-24
fix OCP-21772,OCP-19577

yapei:
@SSshahan  Can you help review?

PR #564, opened 2019-12-26
update rules create_project for OCP-19577

xiaocwan:
This rule should reuse for the negative scenario such as this case. No-matter project created successfully or not, need to check the Complete status for creating project, not only check for the Success status. (Success status has `//span` but the Error status need to check `//div`)
So update the rule in this PR.

@yanpzhan Please help to review, thanks!


xiaocwan:
Covered in https://github.com/openshift/verification-tests/pull/556/files , agree with the fix.
Closing it.

PR #566, opened 2019-12-26
Auto OCP-18994 and OCP-18998

xiuwang:

### DIFF
--- a/features/step_definitions/registry.rb
+++ b/features/step_definitions/registry.rb
@@ -399,7 +399,7 @@ Given /^current generation number of#{OPT_QUOTED} deployment is stored into#{OPT
 end
 
 Given /^docker config for default image registry is stored to the#{OPT_SYM} clipboard$/ do |cb_name|
-
+  cb_name ||= :dockercfg_file
   step %Q/I enable image-registry default route/
   step %Q/default image registry route is stored in the :integrated_reg_ip clipboard/
   step %Q/a 5 characters random string of type :dns is stored into the :dockercfg_name clipboard/
@@ -424,6 +424,7 @@ end
 
 Given /^certification for default image registry is stored to the#{OPT_SYM} clipboard$/ do |cb_name|
   ensure_admin_tagged
+  cb_name ||= :reg_crt_name
   
   step %Q/a 5 characters random string of type :dns is stored into the :regcrt clipboard/
   cb[cb_name] ="/tmp/#{cb.regcrt}.crt"
--- a/features/step_definitions/registry.rb
+++ b/features/step_definitions/registry.rb
@@ -397,3 +397,44 @@ Given /^current generation number of#{OPT_QUOTED} deployment is stored into#{OPT
   cb_name ||= :generation_number
   cb[cb_name] = deployment(name).generation_number(user: user, cached: false)
 end
+
+Given /^docker config for default image registry is saved$/ do
+  step %Q/I enable image-registry default route/
+  step %Q/default image registry route is stored in the :integrated_reg_ip clipboard/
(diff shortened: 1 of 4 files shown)


lib/rules/cli/4.1.yaml
```diff
@@ -397,6 +397,19 @@
     :l: -l <value>
     :resource-names-file: --resource-names-file=<value>
     :svc_name: <value>
+:image_mirror:
+  :cmd: oc image mirror
```
wewang58: maybe it's better    :cmd: oc image mirror <src> <des>
other LGTM

xiuwang: @wewang58 Updated, thanks

features/step_definitions/registry.rb
```diff
@@ -397,3 +397,44 @@
   cb_name ||= :generation_number
   cb[cb_name] = deployment(name).generation_number(user: user, cached: false)
 end
+
+Given /^docker config for default image registry is saved$/ do
```
pruan-rht: @xiuwang the step name is not very clear where `docker config` is saved to, please be more specific when the step name.

features/step_definitions/registry.rb
```diff
@@ -397,3 +397,44 @@
...
+  File.open("#{cb.dockercfg_file}", 'wb') { |f| 
+    f.write(
+       {   
+         "auths" => generated_cfg
+       }.to_json
+    )
+  }
+end
+
+Given /^certification for default image registry is saved$/ do
```
pruan-rht: @xiuwang same comment above

features/step_definitions/registry.rb
xiuwang: @pruan-rht Thanks your advice, I have addressed this comment, please help review again.

features/step_definitions/registry.rb
```diff
@@ -397,3 +397,45 @@
   cb_name ||= :generation_number
   cb[cb_name] = deployment(name).generation_number(user: user, cached: false)
 end
+
+Given /^docker config for default image registry is stored to the#{OPT_SYM} clipboard$/ do |cb_name|
+
+  step %Q/I enable image-registry default route/
+  step %Q/default image registry route is stored in the :integrated_reg_ip clipboard/
+  step %Q/a 5 characters random string of type :dns is stored into the :dockercfg_name clipboard/
+  cb[cb_name] ="/tmp/#{cb.dockercfg_name}"
```
pruan-rht: @xiuwang thanks for the update, but this step will fail if `cb_name` is not specified which will become `nil`.  You should add this a line like `cb_name ||= :dockercfg_name` Here's a good article if that explains what `||=` does http://www.rubyinside.com/what-rubys-double-pipe-or-equals-really-does-5488.html

features/step_definitions/registry.rb
```diff
@@ -397,3 +397,45 @@
...
+       }.to_json
+    )
+  }
+end
+
+Given /^certification for default image registry is stored to the#{OPT_SYM} clipboard$/ do |cb_name|
+  ensure_admin_tagged
+  
+  step %Q/a 5 characters random string of type :dns is stored into the :regcrt clipboard/
+  cb[cb_name] ="/tmp/#{cb.regcrt}.crt"
```
pruan-rht: @xiaocwan same comment as above

features/step_definitions/registry.rb
xiuwang: @pruan-rht Thanks, Peter~ Have updated all comments

xiuwang:
@pruan-rht @akostadinov Could you help review this again or merge it?
Thanks

PR #567, opened 2019-12-30
Fixes in various test cases

anuragthehatter:
Putting comments in each case explaining whats changed and why.
Thanks Zhanqi for tracking spreadsheet for failed cases

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -444,8 +444,9 @@ Feature: Multus-CNI related scenarios
     #Entering into corresponding no eot make sure No VLAN ID information shown for secondary interface
     Given CNI vlan info is obtained on the "<%= cb.pod.node_name %>" node
     Then the step should succeed
-    And the output should contain 2 times:
-      | 1 PVID untagged |
+    And the output should contain 1 times:
+      | 1 PVID   |
+      | untagged |
 
   # @author anusaxen@redhat.com
   # @case_id OCP-24489
@@ -677,29 +678,16 @@ Feature: Multus-CNI related scenarios
     the bridge interface named "mybridge.100" is deleted from the "<%= cb.nodes[1].name %>" node
     """  
     #making sure the pods on same node can ping while pods on diff nodes can't
-
     When I execute on the "<%= cb.pod1 %>" pod:
-      | ping | -c1 | -W2 | <%= cb.pod2_net1_ip %> |
+      | arping | -I | net1 |-c1 | -w2 | <%= cb.pod2_net1_ip %> |
     Then the step should succeed
-
-    When I execute on the "<%= cb.pod2 %>" pod:
-      | ping | -c1 | -W2 | <%= cb.pod1_net1_ip %> |
-    Then the step should succeed
-    
-    When I execute on the "<%= cb.pod1 %>" pod:
-      | ping | -c1 | -W2 | <%= cb.pod3_net1_ip %> |
-    Then the step should fail
 
     When I execute on the "<%= cb.pod3 %>" pod:
-      | ping | -c1 | -W2 | <%= cb.pod1_net1_ip %> |
+      | arping | -I | net1 |-c1 | -w2 | <%= cb.pod1_net1_ip %> |
     Then the step should fail
     
     When I execute on the "<%= cb.pod3 %>" pod:
-      | ping | -c1 | -W2 | <%= cb.pod2_net1_ip %> |
-    Then the step should fail
-    
-    When I execute on the "<%= cb.pod2 %>" pod:
-      | ping | -c1 | -W2 | <%= cb.pod3_net1_ip %> |
+      | arping | -I | net1 |-c1 | -w2 | <%= cb.pod2_net1_ip %> |
     Then the step should fail
     
   # @author anusaxen@redhat.com
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -444,7 +444,7 @@ Feature: Multus-CNI related scenarios
     #Entering into corresponding no eot make sure No VLAN ID information shown for secondary interface
     Given CNI vlan info is obtained on the "<%= cb.pod.node_name %>" node
     Then the step should succeed
-    And the output should contain 1 times:
+    And the output should contain 2 times:
       | 1 PVID   |
       | untagged |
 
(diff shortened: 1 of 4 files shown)


features/networking/multus.feature
```diff
@@ -677,29 +678,16 @@ Feature: Multus-CNI related scenarios
     the bridge interface named "mybridge.100" is deleted from the "<%= cb.nodes[1].name %>" node
     """  
     #making sure the pods on same node can ping while pods on diff nodes can't
-
     When I execute on the "<%= cb.pod1 %>" pod:
```
anuragthehatter: Too much ping combinations so i reduced to effective ones and the reason it failed sometime was because IPAM range in net-attach-def defined that subnet for each node. Let say range is 22.22.22.10 - 22.22.22.20 then its possible that one pod on a node gets 22.22.22.10 and other pod on new node might get same IP, so if that pod pings first pod, it will actually ping itself and throw a success. I used arping which eliminated the duplicacy issues. Tested again on ovn/sdn clusters

features/networking/operator.feature
```diff
@@ -228,14 +228,14 @@ Feature: Operator related networking scenarios
...
-    | /usr/bin/curl | --connect-timeout | 10 | <%= cb.service_ip %>:27017 |
+    | /usr/bin/curl | --connect-timeout | 30 | <%= cb.service_ip %>:27017 |
   Then the step should fail
   #Moving CNO config back to normal and expect service to unidle then
   Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with: 
     | {"spec":{"defaultNetwork":{"openshiftSDNConfig": null}}} |
   And 60 seconds have passed
   When I execute on the "hello-pod" pod:
-    | /usr/bin/curl | --connect-timeout | 10 | <%= cb.service_ip %>:27017 |
+    | /usr/bin/curl | --connect-timeout | 30 | <%= cb.service_ip %>:27017 |
```
anuragthehatter: Apparently it might take some time to idle the svc so 30 seconds seems okay to me. tested multiple times

features/step_definitions/networking.rb
```diff
@@ -790,3 +790,13 @@
   @result = _admin.cli_exec(:get, resource: "network.operator", output: "jsonpath={.items[*].spec.defaultNetwork.type}")
   raise "the networkType is not #{network_type}" unless @result[:response] == network_type
 end
+
```
anuragthehatter: This is useful definition and can be used in future as well. We can identify INTERNAL IP of any node using this.

features/networking/pod.feature
```diff
@@ -197,7 +197,8 @@ Feature: Pod related networking scenarios
   # @case_id OCP-23890
   @admin
   Scenario: A pod with or without hostnetwork cannot access the MCS port 22623 or 22624 on the master
-    Given evaluation of `env.master_hosts.first.local_ip` is stored in the :master_ip clipboard
+    Given I store the masters in the :masters clipboard
+    And the Internal IP of node "<%= cb.masters[0].name %>" is stored in the :master_ip clipboard
```
anuragthehatter: This is useful step and can be used in future as well. We can identify INTERNAL IP of any node using this. Designed rb for this and used in such cases

features/networking/pod.feature
```diff
@@ -197,7 +197,8 @@ Feature: Pod related networking scenarios
   # @case_id OCP-23890
   @admin
   Scenario: A pod with or without hostnetwork cannot access the MCS port 22623 or 22624 on the master
-    Given evaluation of `env.master_hosts.first.local_ip` is stored in the :master_ip clipboard
+    Given I store the masters in the :masters clipboard
```
anuragthehatter: Seems like ```env.master_hosts.first.local_ip``` no more works as env HOSTS now just take lb ip thats why i believe. Anyways this works great now.

features/networking/multus.feature
```diff
@@ -444,8 +444,9 @@ Feature: Multus-CNI related scenarios
     #Entering into corresponding no eot make sure No VLAN ID information shown for secondary interface
     Given CNI vlan info is obtained on the "<%= cb.pod.node_name %>" node
     Then the step should succeed
-    And the output should contain 2 times:
-      | 1 PVID untagged |
+    And the output should contain 1 times:
+      | 1 PVID   |
+      | untagged |
```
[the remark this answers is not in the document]
liangxia: I can understand splitting the output into two lines as it varies from different configurations.

What I'm not sure is that the output check changed from 2 times match to 1 time match. Any reason/change why is that happen ?

features/networking/pod.feature
zhaozhanqi: thanks the update. @anuragthehatter . they are LGTM. 
About the new added step for 'Given I store the masters in the :masters clipboard'. could you help also leave the run logs here?

features/networking/multus.feature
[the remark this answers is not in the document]
anuragthehatter: Yep, it should be 2 times. I was playing with 1 and 2 times results to check whether 2 times actually works fine or not and seems like i undoed my changes somehow and hence it reverted to 1.Thanks for catching that. Re-ran again to make sure http://pastebin.test.redhat.com/824270

features/networking/pod.feature
anuragthehatter: @zhaozhanqi ```Given I store the masters in the :masters clipboard``` is not i created. Its already there in node.rb. I created ```Internal IP of node``` step. Logs http://pastebin.test.redhat.com/824273
I ran it on OVN cluster and was supposed to fail due to Bug 1759338 so ignore the failures.

anuragthehatter: @zhaozhanqi ?

zhaozhanqi: thanks. LGTM

features/networking/multus.feature
weliang1: Agree, arping will resolve the issue of the duplicate IP address assigned to different pods in different nodes which cause the testing failed sometimes.

features/networking/multus.feature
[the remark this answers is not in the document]
weliang1: Not sure, just curious whether the testing will pass or failed if output of "/sbin/bridge vlan show" is like this:
 port      vlan ids
      bridge3
         100 PVID tagged
      veth7721622c
         1 PVID untagged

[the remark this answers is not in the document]
anuragthehatter: @weliang1 It will fail as we are not individually checking 1 but with "1 PVID" together. SO 100 PVID will fail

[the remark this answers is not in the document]
weliang1: @anuragthehatter got it, thanks!  Both changes in multus.feature including below one are /LGTM

PR #572, opened 2019-12-31
Remove comments from the step waiting up replication controller status

SSshahan:
@yapei  please help to review,thanks

### DIFF
--- a/features/step_definitions/rc.rb
+++ b/features/step_definitions/rc.rb
@@ -15,7 +15,6 @@ end
 #   rc(resource_name).wait_till_status(status.to_sym, user, ready_timeout)
 # end
 
-
 Given /^I wait until number of(?: "(.*?)")? replicas match "(\d+)" for replicationController "(.+)"$/ do |state, number, rc_name|
 ready_timeout = 300
   state = :running if state.nil?
--- a/features/step_definitions/rc.rb
+++ b/features/step_definitions/rc.rb
@@ -10,10 +10,10 @@ Given /^(I|admin) waits? until replicationController#{OPT_QUOTED} is ready$/ do 
   raise "replication controller #{rc.name} never became ready" unless @result[:success]
 end
 
-# Given /^I wait until the status of replication controller "(.+)" becomes :(.+)$/ do |resource_name, status|
-#   ready_timeout = 10 * 60
-#   rc(resource_name).wait_till_status(status.to_sym, user, ready_timeout)
-# end
+Given /^I wait until the status of replication controller "(.+)" becomes :(.+)$/ do |resource_name, status|
+  ready_timeout = 10 * 60
+  rc(resource_name).wait_till_status(status.to_sym, user, ready_timeout)
+end
 
 Given /^I wait until number of(?: "(.*?)")? replicas match "(\d+)" for replicationController "(.+)"$/ do |state, number, rc_name|
 ready_timeout = 300
--- a/features/step_definitions/rc.rb
+++ b/features/step_definitions/rc.rb
@@ -10,6 +10,11 @@ Given /^(I|admin) waits? until replicationController#{OPT_QUOTED} is ready$/ do 
   raise "replication controller #{rc.name} never became ready" unless @result[:success]
 end
 
+# Given /^I wait until the status of replication controller "(.+)" becomes :(.+)$/ do |resource_name, status|
+#   ready_timeout = 10 * 60
+#   rc(resource_name).wait_till_status(status.to_sym, user, ready_timeout)
+# end
+
 
 Given /^I wait until number of(?: "(.*?)")? replicas match "(\d+)" for replicationController "(.+)"$/ do |state, number, rc_name|
 ready_timeout = 300
--- a/features/step_definitions/rc.rb
+++ b/features/step_definitions/rc.rb
@@ -10,9 +10,17 @@ Given /^(I|admin) waits? until replicationController#{OPT_QUOTED} is ready$/ do 
   raise "replication controller #{rc.name} never became ready" unless @result[:success]
 end
 
-Given /^I wait until the status of replication controller "(.+)" becomes :(.+)$/ do |resource_name, status|
+#Given /^I wait until the status of replication controller "(.+)" becomes :(.+)$/ do |resource_name, status|
+#  ready_timeout = 10 * 60
+#  rc(resource_name).wait_till_status(status.to_sym, user, ready_timeout)
+#end
+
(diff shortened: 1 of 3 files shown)


features/step_definitions/rc.rb
```diff
@@ -10,10 +10,10 @@
...
 end
 
-# Given /^I wait until the status of replication controller "(.+)" becomes :(.+)$/ do |resource_name, status|
-#   ready_timeout = 10 * 60
-#   rc(resource_name).wait_till_status(status.to_sym, user, ready_timeout)
-# end
+Given /^I wait until the status of replication controller "(.+)" becomes :(.+)$/ do |resource_name, status|
+  ready_timeout = 10 * 60
+  rc(resource_name).wait_till_status(status.to_sym, user, ready_timeout)
+end
```
SSshahan: @pruan-rht  do you know why we commented the step about rc?

SSshahan:
@pruan-rht  could you help to merge it? Thanks!

features/step_definitions/rc.rb
pruan-rht: @SSshahan I don't know...git blame seems to indicate it's was done by @akostadinov.

liangxia:
The commit message does not reflect what's actually changed.

SSshahan:
@pruan-rht  Could you help to merge it firstly? The relevant feature file fix has been merged. Thank you

pruan-rht:
@SSshahan this PR's description doesn't match what is altered.

SSshahan:
@pruan-rht  updated,  please check

pruan-rht:
@SSshahan Let's hold off on this and check with @akostadinov after his comes back from vacation as to the rationale for commenting it out in the first place.

features/step_definitions/rc.rb
SSshahan: @akostadinov  do you know any detail about why we comment them out?

SSshahan: one web case need to use them

akostadinov: Maybe we better write a generic step `I wait until status of #{WORD} "#{QUOTED}" becomes #{SYM}`, where first parameter is resource type. In this case with one step we can cover all resource types instead of having a separate step for all resources that need.

You can see in `features/step_definitions/resources.rb` how to get resource object from word and name.

SSshahan:
@akostadinov  @pruan-rht  could you help to review the new step according to comment https://github.com/openshift/verification-tests/pull/572#discussion_r368404033

SSshahan:
@akostadinov @pruan-rht could you help to review ?

SSshahan:
@akostadinov @pruan-rht could you help to review ?

features/step_definitions/resources.rb
```diff
@@ -123,6 +123,13 @@
   end
 end
 
+Given /^I wait until the status of #{QUOTED} (\w+) becomes :(.+)$/ do |name, type, status|
+  _resource = resource(name, type)
+  ready_timeout = 10 * 60
+  @result = _resource.wait_till_status(status.to_sym, user, ready_timeout)
+end
```
akostadinov: nice

lib/openshift/replication_controller.rb
```diff
@@ -25,21 +25,21 @@ def status_reachable?(from_status, to_status)
...
-    #   }
-    #   res = describe(user, quiet: quiet)
-    #   if res[:success]
-    #     pods_status = res[:parsed][:pods_status]
-    #     res[:success] = (pods_status[status].to_i != 0)
-    #   end
-    #   return res
-    # end
+    def status?(user:, status:, quiet: false, cached: false)
+      statuses = {
```
akostadinov: this variable does not seem to be used anywhere

lib/openshift/replication_controller.rb
```diff
@@ -25,21 +25,21 @@ def status_reachable?(from_status, to_status)
...
+        running: "Running",
+        succeeded: "Succeeded",
+        failed: "Failed",
+        complete: "Complete",
+      }
+      res = describe(user, quiet: quiet)
+      if res[:success]
+        pods_status = res[:parsed][:pods_status]
+        res[:success] = (pods_status[status].to_i != 0)
+      end
```
akostadinov: What is the idea here? if we see any pod cound with desired status that we consider status reached? I'm not sure such logic will always apply.

I think this method has been commented out in the past because it was not well understood. Could you describe your use case and possibly scenario to use with so that we can figure out what is actually needed here. Whether we need such custom method or not. And with what implementation.

SSshahan: It seems we need the variable here, it will failed with `undefined method downcase for nil:NilClass (NoMethodError)
` if we commented out.
```
  Given I wait until the status of "hooks-2" replication_controller becomes :failed # features/step_definitions/resources.rb:126
      undefined method `downcase' for nil:NilClass (NoMethodError)
      /home/hasha/verification-tests/lib/openshift/resource.rb:251:in `phase'
      /home/hasha/verification-tests/lib/openshift/resource.rb:264:in `status?'
      /home/hasha/verification-tests/lib/openshift/resource.rb:360:in `block in wait_till_status'
      /home/hasha/verification-tests/lib/base_helper.rb:153:in `wait_for'
      /home/hasha/verification-tests/lib/openshift/resource.rb:359:in `wait_till_status'
      /home/hasha/verification-tests/features/step_definitions/resources.rb:129:in `/^I wait until the status of "(.+?)" (\w+) becomes :(.+)$/'
      features/tierN/web/admin-console/deployments.feature:169:in `Given I wait until the status of "hooks-2" replication_controller becomes :failed
```
One checkpoint of the case need to cancel dc rollout and waiting the latest rc becomes failed status, then check console page shows the expected states on page.

akostadinov: ok, it is fine to have the method. But it needs to make sense. Presently one variable is not used at all. Then one pod reaching proper status results in setting the status of the whole thing.

Or maybe we need to define the step better. See my other comment.

features/step_definitions/resources.rb
akostadinov: I like this step and I would like to keep it. The thing is that looking at replication controller it does not have a status of itself, we are looking at the statuses of its pod.
perhaps in your scenario you need to use step `a pod becomes ready with labels:` or `status becomes :([^\s]*?) of( exactly)? ([0-9]+) pods? labeled:`. WDYT?

SSshahan: @akostadinov  agree with you. I'll update the feature file if I hit scenario failure again. Then How about the pr, merge it or close it?

PR #573, opened 2019-12-31
add rules for oc 4.3

xiaojiey:
@xingxingxia could you pls help to review? Thanks.

### DIFF
--- a/lib/rules/cli/4.3.yaml
+++ b/lib/rules/cli/4.3.yaml
@@ -1,0 +1,1353 @@
+---
+:global_options:
+  # these are options as seen by `oc options`
+  :api_version: --api-version=<value>
+  :ca: --certificate-authority=<value>
+  :cluster: --cluster=<value>
+  :config: --config=<value>
+  :context: --context=<value>
+  :h: -h
+  :help: --help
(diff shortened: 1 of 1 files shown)


lib/rules/cli/4.3.yaml
```diff
@@ -0,0 +1,1353 @@
+---
```
liangxia: @xiaojiey  We already have 4.x rules.
```
[lxia@ocp verification-tests]$ ls lib/rules/cli/4.*
lib/rules/cli/4.1.yaml  lib/rules/cli/4.2.yaml  lib/rules/cli/4.3.yaml  lib/rules/cli/4.4.yaml
```

Any commands/options that you need does exist ? Add it into 4.1.yaml

xingxingxia: @xiaojiey yes, agree with @liangxia

xingxingxia: ls -l ls lib/rules/cli/4.*, you will see symlinks

xiaojiey:
Not needed. Replaced by another PR https://github.com/openshift/verification-tests/pull/651

PR #576, opened 2019-12-31
fix getting router subdomain

lihongan:
`Common::Net.dns_lookup(fqdn, multi: true)` doesn't work in private cluster since no public DNS for the routes.

We have implemented a new step in https://github.com/openshift/verification-tests/pull/533 to get default router IP, but the step `I store default router subdomain in the :subdomain clipboard` doesn't work in private cluster either. And we find that the method `get_routing_details` mixed those two options in one method and cause error even when you only want to get `router_subdomain`.

This PR fixes the issue by adding one argument to split them.

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -312,7 +312,7 @@ module BushSlicer
     # obtain router detals like default router subdomain and router IPs
     # @param user [BushSlicer::User]
     # @param project [BushSlicer::project]
-    def get_routing_details(user:, project:)
+    def get_routing_details(user:, project:, obj:)
       service_res = Service.create(by: user, project: project, spec: 'https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/service_with_selector.json')
       raise "cannot create service" unless service_res[:success]
       service = service_res[:resource]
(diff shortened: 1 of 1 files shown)


lihongan:
log as below (private cluster)
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/56838/consoleFull

```
    And I store default router subdomain in the :subdomain clipboard                                                                                            
# features/step_definitions/env.rb:4
<---snip--->
      [08:54:15] INFO> apps.zzhaodebug.qe.gcp.devcluster.openshift.com
```

lihongan:
@akostadinov @pruan-rht Ready for review. Thanks.

lib/environment.rb
```diff
@@ -321,24 +321,27 @@ def get_routing_details(user:, project:)
...
       fqdn = route.dns(by: user)
-      opts[:router_subdomain] = fqdn.split('.',2)[1]
-      opts[:router_ips] = Common::Net.dns_lookup(fqdn, multi: true)
-
+      if obj == 'subdomain'
+        opts[:router_subdomain] = fqdn.split('.',2)[1]
+      end
+      if obj == 'ips'
+        opts[:router_ips] = Common::Net.dns_lookup(fqdn, multi: true)
+      end
```
liangxia: Perhaps use if/else so we do not need to do if check twice every time.

liangxia:
/lgtm
we can do the optimisation in a separate PR.

lib/environment.rb
pruan-rht: @lihongan @liangxia's suggestion should be easy to implement.  Please correct amend this PR instead of filing another one.

lihongan: @liangxia @pruan-rht Thank you all, I've updated and use if/else.

pruan-rht: @lihongan do you foresee `get_routing_details` be called with object type other than `subdomain` and `ips`?

lihongan: @pruan-rht  No, I don't see any other caller. The two object type is added by this PR. 
The problem is the old `get_routing_details` gets both `router_ips` and `router_subdomain` at the same time even I just need one of them.
Another possible fix is removing `get_routing_details` and implementing each function in `router_ips(user:, project:)` and `router_default_subdomain(user:, project:)`, the two callers I can find.

pruan-rht: @lihongan Ok, I would suggest we separate them into two wrapper functions both of which call the base function but with the parameter pass in.  IMO, this will make it clearer to the user just by reading the method name instead of having to look up what parameters are accepted.  WDYT?

lihongan: No problem. Both solutions are accepted.
My requirement is very simple: just return what I want when I getting `router_ips` or `router_subdomain`. Don't get both even I just need one of them.
In our scenarios, most cases are just need one of them.

PR #582, opened 2020-01-03
rules for OCP-21111

SSshahan:
@yapei  please help to review, thanks

### DIFF
--- a/lib/rules/web/admin_console/4.3/catalog.xyaml
+++ b/lib/rules/web/admin_console/4.3/catalog.xyaml
@@ -128,3 +128,7 @@ set_custom_channel_and_subscribe:
   action: wait_box_loaded
   action: choose_update_channel
   action: click_subscribe_button
+check_image_tag_in_dropdown:
+  element:
+    selector:
+      xpath: //ul[contains(.,'<tag>') and not(contains(.,'<hidden_tag>'))]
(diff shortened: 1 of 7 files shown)


lib/rules/web/admin_console/4.3/catalog.xyaml
```diff
@@ -115,3 +115,7 @@ set_custom_channel_and_subscribe:
   action: wait_box_loaded
   action: choose_update_channel
   action: click_subscribe_button
+check_image_tag_in_dropdown:
```
yapei: This action seems not very re-usable. How about 
```
check_item_in_list:
  element:
    selector:
      xpath: //ul[contains(.,'<item>')]
```

yapei: Then in scenario, we can call it 
```
When I perform the :check_item_in_list web action with:
  | item | latest |
Then the step should succeed
When I perform the :check_item_in_list web action with:
  | item | <%=cb.hidden_version %> |
Then the step should fail
```

SSshahan:
[4.3 pass log](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/60219/console)
[4.2 pass log](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/60073/console)

PR #587, opened 2020-01-03
OCP-21846 and OCP-25787

anuragthehatter:
OVN, OCP-21846 and OCP-25787automation
http://pastebin.test.redhat.com/824700
http://pastebin.test.redhat.com/824718


features/networking/pod.feature
```diff
@@ -297,3 +297,45 @@ Feature: Pod related networking scenarios
...
+    #Tainting node to NoSchedule
+    When I run the :oadm_taint_nodes admin command with:
+      | node_name | <%= cb.nodes[0].name %> |
+      | key_val   | key2=value2:NoSchedule  |
+    Then the step should succeed
+    And I register clean-up steps:
+    """
+    I run the :oadm_taint_nodes admin command with:
+      | node_name | <%= cb.nodes[0].name %> |
+      | key_val   | key2-  |
```
weliang1: identation

features/networking/pod.feature
```diff
@@ -297,3 +297,45 @@ Feature: Pod related networking scenarios
...
+    Then the step should succeed
+    """
+    Given I have a project
+    #Makng sure test pods can' be sceduled on tha tainted node
+    And I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/pod-for-ping.json |
+    Then the step should succeed
+    And the pod named "hello-pod" status becomes :pending
```
weliang1: How to make sure this hello-pod will be created in the tainted node not the other running node?

features/networking/pod.feature
anuragthehatter: Thanks @weliang1. Will take care

features/networking/pod.feature
anuragthehatter: Good catch @weliang1 Thanks. It also needs [spec][nodename].

anuragthehatter: @weliang1 Made it more optimized. All nodes are tainted now in the beginning so no need to check which node hello-pod is being created on. Also added 60 sec timer for hello-pod in case it doesn't reach to pending state for 60 seconds,we should fail and know something is wrong. And then we check ovnkube pod should be deleted/recreated on a worker node.

anuragthehatter:
Added OCP-26035 as well for OVN (http://pastebin.test.redhat.com/824946) and it is expected to fail due to https://bugzilla.redhat.com/show_bug.cgi?id=1785360 -

features/networking/pod.feature
weliang1: OCP-21846 LGTM now

features/networking/sdn.feature
```diff
@@ -300,3 +300,34 @@ Feature: SDN related networking scenarios
...
+    #Checking controller iteration 1
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ovs-appctl -t ovn-controller connection-status |
+    Then the step should succeed
+    And the output should contain "connected"
+    #Checking controller iteration 2
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ls -l /var/run/openvswitch/ |
+    Then the step should succeed
+    And evaluation of `@result[:response].match(/ovn-controller.\d*\.ctl/)[0]` is stored in the :controller_pid_file clipboard
```
weliang1: What will the stored ":controller_pid_file clipboard" be used for? Does not see other step to use this clipboard

features/networking/sdn.feature
```diff
@@ -300,3 +300,34 @@ Feature: SDN related networking scenarios
...
+    And the output should contain "connected"
+    #Checking controller iteration 4
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ovs-ofctl dump-flows br-int \| wc -l |
+    Then the step should succeed
+    And the expression should be true> @result[:response].match(/\d*/)[0].to_i> 0
+    #Checking final iteration post all above iterations passed. In this iteration we expect CNi file to be created
+    And I run commands on the host:
+      | ls -l /var/run/multus/cni/net.d/ |
+    And the output should contain "ovn-kubernetes.conf"
```
weliang1: After reading test case Scenario, the testing steps should be like break all four controller iterations, and checking "ls -l /var/run/multus/cni/net.d/" should fail, then make one of four controller iterations back to work, then checking "ls -l /var/run/multus/cni/net.d/" should pass. How do you think? @anuragthehatter

features/networking/sdn.feature
anuragthehatter: @weliang1 Iteration 3 is using it. This will confirm ,ovn controller has programmed the openvswitch flows and has connection with it.

features/networking/sdn.feature
anuragthehatter: @weliang1 this case is making sure that if all 4 steps shows success, then there should be CNI file written. If all steps shows success and no CNI is found then we know something is wrong or if CNI is written and one of step is down. These iterations only run when OVN cluster comes first time. I also had same thinking but as per talk with Casey few days back, its not easy to break all iterations (lot of iptables destructiveness involved and no gurantees) and should be avoided for cluster turmoil. Also CNI written is not dynamic, once its written, it will be there. Passing or Failing of this case will give a clue about healthy/unhealthy OVN cluster and iterations issue if any.

features/networking/sdn.feature
weliang1: Sorry, I just saw it

features/networking/sdn.feature
```diff
@@ -300,3 +300,34 @@ Feature: SDN related networking scenarios
...
+    
+  # @author anusaxen@redhat.com
+  # @case_id OCP-25787
+  @admin
+  Scenario: Don't write CNI configuration file until ovn-controller has done at least one iteration
+    Given the env is using "OVNKubernetes" networkType
+    And I select a random node's host
+    #Checking controller iteration 1
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ovs-appctl -t ovn-controller connection-status |
```
zhaozhanqi: should be `ovn-appctl` instead of `ovs-appctl`

features/networking/sdn.feature
```diff
@@ -300,3 +300,34 @@ Feature: SDN related networking scenarios
...
+    Then the step should succeed
+    And the output should contain "connected"
+    #Checking controller iteration 2
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ls -l /var/run/openvswitch/ |
```
zhaozhanqi: the pid file has been changed to `/var/run/ovn/ovn-controller.pid`

features/networking/sdn.feature
anuragthehatter: Yea, the ovn 2.12. Thanks for catching that. Updated whole case and re-ran to make sure http://pastebin.test.redhat.com/843959

features/networking/sdn.feature
anuragthehatter: done

features/networking/pod.feature
```diff
@@ -297,3 +297,46 @@ Feature: Pod related networking scenarios
...
+  # @case_id OCP-21846
+  @admin
+  @destructive
+  Scenario: ovn pod can be scheduled even if the node taint to unschedule
+    Given the env is using "OVNKubernetes" networkType
+    And I store all worker nodes to the :nodes clipboard
+    #Tainting all worker nodes to NoSchedule
+    When I run the :oadm_taint_nodes admin command with:
+      | l         | node-role.kubernetes.io/worker |
+      | key_val   | key2=value2:NoSchedule         |
```
liangxia: Perhaps replace key2 with key21846 or something like that for easier debug ?

features/networking/pod.feature
```diff
@@ -297,3 +297,46 @@ Feature: Pod related networking scenarios
...
+    When I run the :get admin command with:
+      | resource      | pod                                   |
+      | fieldSelector | spec.nodeName=<%= cb.nodes[0].name %> |
+      | n             | openshift-ovn-kubernetes              |
+      | o             | jsonpath={.items[*].metadata.name}    |
+    Then the step should succeed
+    And evaluation of `@result[:response]` is stored in the :ovnkube_pod_name clipboard
+    And admin ensure "<%= cb.ovnkube_pod_name %>" pod is deleted from the "openshift-ovn-kubernetes" project
+    #Waiting 60 seconds for new ovnkube pod to get created and running on the same node it was deleted before
+    Given 60 seconds have passed
```
liangxia: Hard wait. Replace it with "I wait up to xxx seconds for the steps to pass"

features/networking/sdn.feature
liangxia: Perhaps use
```
      | ls /var/run/multus/cni/net.d/ovn-kubernetes.conf |
    Then the step should succeed
```

lib/rules/cli/4.1.yaml
```diff
@@ -1301,6 +1301,7 @@
     :node_name: <value>
     :output: --output <value>
     :overwrite: --overwrite=<value>
+    :l: -l <value>
```
liangxia: Sort it in alphanumeric order.
And you need to add it to 4.4.yaml as well.

anuragthehatter: Alphanumeric is done but 4.4,4.3 and 4.2 is all the way symlinked to 4.1 so the changes are being done in 4.1. Also the "-l" is supported on 4.1 and 4.4 as well

features/networking/pod.feature
anuragthehatter: Yep. I could i forget that. Thanks

features/networking/pod.feature
anuragthehatter: Yep this will be useful. Done. Thanks

features/networking/sdn.feature
anuragthehatter: Done and restested. Thanks for suggesting optimisation

features/networking/sdn.feature
```diff
@@ -340,4 +340,34 @@ Feature: SDN related networking scenarios
...
+    #Checking controller iteration 3
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ovn-appctl -t /var/run/ovn/<%= cb.controller_pid_file %> connection-status |
+    Then the step should succeed
+    And the output should contain "connected"
+    #Checking controller iteration 4
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ovs-ofctl dump-flows br-int \| wc -l |
+    Then the step should succeed
+    And the expression should be true> @result[:response].match(/\d*/)[0].to_i> 0
```
rbbratta: could we add a space before `>`

`@result[:response].match(/\d*/)[0].to_i > 0`

features/networking/sdn.feature
```diff
@@ -340,4 +340,34 @@ Feature: SDN related networking scenarios
...
+    Given the env is using "OVNKubernetes" networkType
+    And I select a random node's host
+    #Checking controller iteration 1
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ovn-appctl -t ovn-controller connection-status |
+    Then the step should succeed
+    And the output should contain "connected"
+    #Checking controller iteration 2
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ls -l /var/run/ovn/ |
```
rbbratta: do we always need the `bash -c`?

lib/rules/cli/4.1.yaml
```diff
@@ -1298,6 +1298,7 @@
   :options:
     :all: --all=<value>
     :key_val: <value>
+    :l: -l <value>
```
liangxia: 4.4.yaml also needs that (4.4 is no longer a soft link as there are incompatible change).
```
$ ls -l lib/rules/cli/4.*
-rw-rw-r--. 1 lxia lxia 38029 Mar 12 17:40 lib/rules/cli/4.1.yaml
lrwxrwxrwx. 1 lxia lxia     8 Dec  5 15:59 lib/rules/cli/4.2.yaml -> 4.1.yaml
lrwxrwxrwx. 1 lxia lxia     8 Dec  5 15:59 lib/rules/cli/4.3.yaml -> 4.2.yaml
-rw-rw-r--. 1 lxia lxia 38033 Mar 12 17:40 lib/rules/cli/4.4.yaml
lrwxrwxrwx. 1 lxia lxia     8 Mar 11 17:04 lib/rules/cli/4.5.yaml -> 4.4.yaml
```

features/networking/sdn.feature
anuragthehatter: Normally if we have a bash expression with redirects, pipes, etc. then we use bash as command and pass expression to bash with the -c param to run that whole command as one in subshell. We have seen issues in past if not use its specially when executing on pods. Here it may not be needed as i don't see any pipes,redirection,quotes etc. But guess most of us  are just accustomed to use bash -c now :)

features/networking/sdn.feature
anuragthehatter: Done, thanks

lib/rules/cli/4.1.yaml
anuragthehatter: Right, i just noticed that 4.3 is all the way symlinked to 4.1 but not 4.4. Thanks, updated!

features/networking/sdn.feature
```diff
@@ -340,4 +340,34 @@ Feature: SDN related networking scenarios
...
+    Then the step should succeed
+    And the output should contain "connected"
+    #Checking controller iteration 4
+    When I run command on the "<%= node.name %>" node's sdn pod:
+      | bash | -c | ovs-ofctl dump-flows br-int \| wc -l |
+    Then the step should succeed
+    And the expression should be true> @result[:response].match(/\d*/)[0].to_i > 0
+    #Checking final iteration post all above iterations passed. In this iteration we expect CNi file to be created
+    And I run commands on the host:
+      | ls /var/run/multus/cni/net.d/ovn-kubernetes.conf |
```
rbbratta: just failed for me.
```
ls: cannot access '/var/run/multus/cni/net.d/ovn-kubernetes.conf': No such file or directory
```

```
sh-4.4# find  /var/run/multus/cni/net.d -ls
    23600      0 drwxr-xr-x   2  root     root           60 Mar 13 14:41 /var/run/multus/cni/net.d
   205684      4 -rw-------   1  root     root           94 Mar 13 14:41 /var/run/multus/cni/net.d/10-ovn-kubernetes.conf
```
`4.4.0-0.nightly-2020-03-13-073111`

anuragthehatter: Thanks Ross for catching it. Yea it needs to be 10-ovn-kubernetes.conf which was right earlier before fix to this step. I am wondering why ```ls /var/run/multus/cni/net.d/ovn-kubernetes.conf``` is passing in automation despite it complains "ls: cannot access '/var/run/multus/cni/net.d/ovn-kubernetes.conf': No such file or directory". May be that's returning the exit code 0 based on . usuallu executing command in pod return exit status. Anyways i reverted back to my earlier commit which was specifically checking 10.ovn-kubernetest.conf for now it works okay now http://pastebin.test.redhat.com/844704

PR #591, opened 2020-01-06
add monitoring alerts/sliences url in console

juzhao:


juzhao:
There is not metrics UI in 4.1 admin console, so just added alerts/silences UI
also back port 4.3 change to 4.2

PR #592, opened 2020-01-06
OCP-11286 iptables rules test updated for 4.x

rbbratta:
For 4.4 if we partially remove rules it does not
trigger an iptablesSync (BZ-1810316)

Check if the partial removal works and then
do a full table flush and chain deletion.

This requires breaking the iptable verification step into two parts,
checking the rules and then deciding if the rule check should fail
depending on version

General fixes

`"openshift-ovs-networkpolicy"` is now the default in 4.x

Remove 3.x version specific iptables rules

When using oc debug commands the `:response` seems to always include
`STDERR` appended which doesn't work for matching the `MASQUERADE` rule so use
`:stdout.to_s` instead

### DIFF
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -38,7 +38,9 @@ Feature: SDN related networking scenarios
   @admin
   @destructive
   Scenario: iptables rules will be repaired automatically once it gets destroyed
-    Given the master version >= "4.1"
+    # we do not detect incomplete rule removal since ~4.3, BZ-1810316
+    # so only test on >= 4.3
+    Given the master version >= "4.3"
     Given I select a random node's host
     And the node iptables config is checked
     And the step succeeded
@@ -52,8 +54,10 @@ Feature: SDN related networking scenarios
     When the node standard iptables rules are removed
     And 35 seconds have passed
     When the node iptables config is checked
-    # we do not detect incomplete rule removal since ~4.3, BZ-1810316
-    Then the iptables partial deletion check should have succeeded depending on version
+    # Removing individual rules will not trigger automatic repair on < 4.3
+    # the check should fail
+    Then the step failed
+    # >= 4.3 we have to flush all the rules and tables to trigger a repair
     When the node standard iptables rules are completely flushed
     And 35 seconds have passed
     Then the node iptables config is checked
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -33,18 +33,32 @@ Feature: SDN related networking scenarios
     And the output should contain "Using userspace Proxier"
     """
 
-  # @author bmeng@redhat.com
+  # @author rbrattai@redhat.com
   # @case_id OCP-11286
   @admin
   @destructive
   Scenario: iptables rules will be repaired automatically once it gets destroyed
+    Given the master version >= "4.1"
     Given I select a random node's host
-    And the node iptables config is verified
+    And the node iptables config is checked
+    And the step succeeded
     And the node service is restarted on the host after scenario
+    And I register clean-up steps:
+    """
+    When the node iptables config is checked
+    Then the step succeeded
+    """
 
-    Given the node standard iptables rules are removed
-    Given 35 seconds have passed
-    Given the node iptables config is verified
+    When the node standard iptables rules are removed
+    And 35 seconds have passed
+    When the node iptables config is checked
+    # we do not detect incomplete rule removal since ~4.3, BZ-1810316
+    Then the iptables partial deletion check should have succeeded depending on version
+    When the node standard iptables rules are completely flushed
+    And 35 seconds have passed
+    Then the node iptables config is checked
+    Then the step succeeded
+
 
   # @author hongli@redhat.com
   # @case_id OCP-13847
@@ -188,7 +202,7 @@ Feature: SDN related networking scenarios
     Then the step should succeed
     And evaluation of `@result[:response].split("\n")[0]` is stored in the :sdn_pod_rules clipboard
     Then the expression should be true> cb.host_rules==cb.sdn_pod_rules
-    
+
   # @author huirwang@redhat.com
   # @case_id OCP-25707
   @admin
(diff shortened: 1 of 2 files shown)


features/step_definitions/networking.rb
```diff
@@ -122,17 +122,16 @@
   teardown_add net_verify
 end
 
+
+CLUSTER_NETWORK = "clusternetwork"
```
pruan-rht: @rbbratta, it's not recommended to define a variable like this outside of the step.  You can probably just use the specification in line 134 `clusternetwork`

rbbratta: is there way to define constants or symbols?

pruan-rht: @rbbratta why not just define that within your step?

rbbratta: I refactored the clusternetwork subnet lookup into a single step, but we still have lots of duplicate strings around, I'm afraid of random typos and such every time I duplicate a long string.

features/step_definitions/networking.rb
```diff
@@ -336,46 +203,27 @@
...
-    raise "failed to grep rule from the iptables nat table!" unless @result[:success]
-    nat_rule = @result[:response]
-
-    @resule = _host.exec("iptables -t nat -D #{nat_rule}")
-    raise "failed to delete iptables nat rule" unless @result[:success]
-  end
+  step "The subnet from the clusternetwork resource is stored in the clipboard"
+  subnet = cb.clusternetwork
+
+  @resule = _host.exec('iptables -D OPENSHIFT-FIREWALL-ALLOW -p udp -m udp --dport 4789 -m comment --comment "VXLAN incoming" -j ACCEPT')
```
anuragthehatter: @rbratta typo in instance variable name ```resule```  everywhere? You are using ```resule``` but checking status of ```result``` at most of the places

anuragthehatter: Also STDERR issue has 3 bugs open under openshift which i don't recall at the moment so we can design logic independent of whether STDERR is included in output or not.

rbbratta: > @rbratta typo in instance variable name `resule` everywhere? You are using `resule` but checking status of `result` at most of the places

Looks like the typo was in the old code and I just copy-pasted it.  I think I saw that and assumed it was on purpose.

Thanks for catching this.

rbbratta: > Also STDERR issue has 3 bugs open under openshift which i don't recall at the moment so we can design logic independent of whether STDERR is included in output or not.

I'm fine with using `@result[:stdout].to_s`. I'm not sure if the `.to_s` is necessary, haven't figured out what the underlying type of @result[:stdout]` is.

rbbratta:
this test case found a bug https://bugzilla.redhat.com/show_bug.cgi?id=1810316

rbbratta:
This needs update based on the new 4.4 iptables sync logic.

rbbratta:
4.3 logs http://pastebin.test.redhat.com/844377
4.4 logs http://pastebin.test.redhat.com/844379

rbbratta:
tested on 4.5  logs http://pastebin.test.redhat.com/859991

rbbratta:
ready for review

rbbratta:
fixed the logic and improved the error message.
http://file.rdu.redhat.com/rbrattai/ocp-11286.log

rbbratta:
It looks like they backported the iptables sync change to 4.3, so had to change the logic to check for 4.2

http://file.rdu.redhat.com/rbrattai/ocp-11286-4.4.log
http://file.rdu.redhat.com/rbrattai/ocp-11286-4.3.log
http://file.rdu.redhat.com/rbrattai/ocp-11286-4.2.log

rbbratta:
maybe we can just skip the test on version < 4.3

anuragthehatter:
> maybe we can just skip the test on version < 4.3

@rbbratta Thanks. Do you mean flush and delete doesn't work < 4.3? What we do on < 4.3 releases then ? Since you have investigated it so far and I have vague sense of it.

rbbratta:
The previous test was "When the node standard iptables rules are removed", we just individually removed a few standard rules and check that they were restored.  

This still happens on 4.2, if you remove individual rules one by one, then they are re-synced.

The new behavior you have to do the flush of each table before the rules are re-synced.

This PR removes all the old 3.x code, so we could just remove all the old  4.2 behavior and stuck with checking the new behavior.

zhaozhanqi:
Thanks @rbbratta 
/lgtm

anuragthehatter:
> The previous test was "When the node standard iptables rules are removed", we just individually removed a few standard rules and check that they were restored.
> 
> This still happens on 4.2, if you remove individual rules one by one, then they are re-synced.
> 
> The new behavior you have to do the flush all the tables and delete all the chains before the rules are re-synced.
> 
> This PR removes all the old 3.x code, so we could just remove all the old 4.2 behavior and stuck with checking the new behavior.

Thanks Ross for clarifications. Yep, keeping same behavior across releases sounds like a plan to me.

features/step_definitions/networking.rb
```diff
@@ -122,212 +122,104 @@
   teardown_add net_verify
 end
 
-Given /^the#{OPT_QUOTED} node iptables config is verified$/ do |node_name|
+
+Given /^The subnet from the clusternetwork resource is stored in the clipboard$/ do
```
liangxia, line 126: s/The/the/

rbbratta:
> Thanks Ross for clarifications. Yep, keeping same behavior across releases sounds like a plan to me.

I removed the 4.2 behavior and just marked the test >= 4.3.

rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/ocp-11286-4.5-updated.log

features/networking/sdn.feature
```diff
@@ -33,18 +33,36 @@ Feature: SDN related networking scenarios
...
+    And the step succeeded
     And the node service is restarted on the host after scenario
+    And I register clean-up steps:
+    """
+    When the node iptables config is checked
+    Then the step succeeded
+    """
+
+    When the node standard iptables rules are removed
+    And 35 seconds have passed
```
pruan-rht, line 55: @rbbratta is it possible to do this check and wait instead of a hard-code wait?

rbbratta: We have to wait for default `iptablesSyncPeriod` which is `30` seconds.  Err, well `iptablesSyncPeriod` can be changed post-install so we should query the current value, but it will be `30` seconds 99% of the time.

This step is a negative check, so we have to wait the full period to make sure the rules were not restored.  

Maybe I'll just add a comment explaining the 30 second wait.

```yaml
kubeProxyConfig:
     iptablesSyncPeriod: 30s
     proxyArguments:
          iptables-min-sync-period:
          - 30s
```

features/networking/sdn.feature
```diff
@@ -33,18 +33,36 @@ Feature: SDN related networking scenarios
...
+    When the node iptables config is checked
+    # Removing individual rules will not trigger automatic repair on < 4.3
+    # the check should fail
+    Then the step failed
+    # >= 4.3 we have to flush all the rules and tables to trigger a repair
+    When the node standard iptables rules are completely flushed
+    And 35 seconds have passed
```
rbbratta, line 62: I'll try replacing this with a `And I wait up to 35 seconds for the steps to pass:`

rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-07-20-21-27-26.log

rbbratta:
@anuragthehatter @pruan-rht  PTAL

rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-08-19-17-42-28.log

PR #595, opened 2020-01-07
add params for oc create for OCP-25763

xiaocwan:
@yapei @pruan-rht @liangxia 
Please review a hot fix for oc create with params, thanks!

### DIFF
--- a/lib/rules/web/admin_console/4.1/overview.xyaml
+++ b/lib/rules/web/admin_console/4.1/overview.xyaml
@@ -57,3 +57,45 @@ check_dc_available_action_menus:
       xpath: //li//a[text()='Edit Deployment Config']
   - selector:
       xpath: //li//a[text()='Delete Deployment Config']
+click_groupby_dropdown_button:
+  elements:
+  - selector:
+      xpath: //span[contains(.,'Group by')]/ancestor::button[contains(@class,'btn-dropdown')]
+    op: click
+check_groupby_dropdown_menu_item:
+  element:
+    selector: &group_by_dropdown_item
+      xpath: //li/a[text()='<dropdown_menu_item>']
+click_groupby_dropdown_menu_item:
+  element:
+    selector:
+      <<: *group_by_dropdown_item
+    op: click
+check_groupby_label_header_missing:
+  element:
+    selector:
+      xpath: //div[@class='dropdown-menu__header' and text()='Label']
+    missing: true
+check_overview_group_heading:
+  element:
+    selector:
+      xpath: //h2[contains(@class,'project-overview__group-heading') and text()='<group_heading_name>']
+check_overview_group_items:
+  elements:
+  - selector:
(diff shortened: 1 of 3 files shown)


xiaocwan:
Rebased latest fix by @liangxia and make up with the optional params

lib/rules/cli/4.1.yaml
```diff
@@ -147,6 +147,8 @@
     :f: -f <value>
     :filename: --filename=<value>
     :record: --record=<value>
+    :reource_type: <reource_type>
```
yapei: it should be `resource_type` and `resource_name`

xiaocwan: @yapei thx, updated.

wewang58: There is no resouce_type ,when "oc create --help"

lib/rules/cli/4.1.yaml
```diff
@@ -147,6 +147,8 @@
     :f: -f <value>
     :filename: --filename=<value>
     :record: --record=<value>
+    :resource_type: <resource_type>
+    :resouce_name: <resouce_name>
```
liangxia: 1. Still typos
2. Where are the parameters used ?

xiaocwan: @liangxia 
1. Updated.
2. In my Open PR: https://github.com/openshift/cucushift/pull/7303/files

lib/rules/cli/4.1.yaml
```diff
@@ -147,6 +147,8 @@
     :f: -f <value>
     :filename: --filename=<value>
     :record: --record=<value>
+    :resource_type: <resource_type>
+    :resource_name: <resource_name>
```
liangxia: Sort the options in alpha order, so we can easily find/add/remove options.

liangxia:
/hold
We need to squash the commit as it contains only 2 line addition, which is quite simple. 5 commits for that does not look good.
Also the commit message and description of the PR needs update, it is no longer a hot fix.

xiaocwan:
@liangxia Is this good now?

liangxia:
We have command "oc create user", which is added in https://github.com/openshift/verification-tests/pull/628

xiaocwan:
> We have command "oc create user", which is added in #628

It's better to decouple the Create command, why not? Please tell me HDYT, thanks!

liangxia:
Your command is actually using "oc create user",
$ oc create user --help
This command can be used to create a user object.

 Typically, users are created automatically during login. If automatic creation is disabled (by using the "lookup"
mapping method), users must be created manually.

 Corresponding identity and useridentitymapping objects must also be created to allow logging in as the created user.

Usage:
  oc create user USERNAME [flags]

Examples:
  # Create a user with the username "ajones" and the display name "Adam Jones"
  oc create user ajones --full-name="Adam Jones"

xiaocwan:
Will Update OCP-25763 from feature step after the PR https://github.com/openshift/verification-tests/pull/628 merge. This PR could be closed after that rule merge.

PR #602, opened 2020-01-08
add a goto rule to 3.11 for OCP-19577

xiaocwan:
@yapei Plz review, thanks!


pruan-rht:
@xiaocwan if this only applies to 3.x, perhaps submit it to the `v3` branch instead of going into `master` which is for OCP >= 4.0

PR #609, opened 2020-01-09
Added missing CMD policy in add/remove role/cluster-role

liangxia:
@pruan-rht @akostadinov

### DIFF
--- a/features/cli/policy.feature
+++ b/features/cli/policy.feature
@@ -10,7 +10,7 @@ Feature: change the policy of user/service account
       | o             | wide        |
     Then the output should match:
       | admin.*(<%= @user.name %>)? |
-    When I run the :oadm_add_role_to_user client command with:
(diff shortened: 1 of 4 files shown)


liangxia:
> @liangxia do you know how these tests were able to pass before?

@pruan-rht 
The alias is missing `policy`, but it points to the correct oc commands with `policy`.
It works without this PR, but this PR just make the commands consistent.

E.g, prreviously, it is,
```
:oadm_remove_cluster_role_from_group:
  :cmd: oc adm policy remove-cluster-role-from-group <role_name> <group_name>
```
Now, it is
```
:oadm_policy_remove_cluster_role_from_group:
  :cmd: oc adm policy remove-cluster-role-from-group <role_name> <group_name>
```

The old one will be removed in a separate PR to not break existing automation before all the scenarios are updated.

PR #611, opened 2020-01-09
turn off more logs

yapei:


yapei:
Hi @pruan-rht  Can you help merge this PR? I turned off debug logs since we already have collected some useful logs when running automation in CI. Keeping them as open bring too much logs

PR #615, opened 2020-01-10
Alert testing for machine-api operators

jhou1:
Verify that alert is fired when its relevant operator is down.
I've verified my tests locally.

```
    [07:41:26] INFO> Exit Status: 0
    [07:41:26] INFO> === End After Scenario: Alert should be fired when operator is down, Examples (#2) ===

2 scenarios (2 passed)
26 steps (26 passed)
6m38.954s
```

### DIFF
--- a/features/machine/alerting.feature
+++ b/features/machine/alerting.feature
@@ -1,0 +1,64 @@
+Feature: Alerting for machine-api
+
+  # @author jhou@redhat.com
+  @admin
+  @destructive
+  Scenario Outline: Alert should be fired when operator is down
+    Given I switch to cluster admin pseudo user
+
+    # scale down cvo and operators
+    When I run the :scale admin command with:
+      | resource | deployment                |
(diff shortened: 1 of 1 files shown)


jhou1:
@liangxia @pruan-rht Would you please review?

features/machine/alerting.feature
```diff
@@ -0,0 +1,64 @@
...
+      | replicas | 1                     |
+      | n        | openshift-machine-api |
+    Then the step should succeed
+    """
+
+    And I use the "openshift-monitoring" project
+    And evaluation of `secret(service_account('prometheus-k8s').get_secret_names.find {|s| s.match('token')}).token` is stored in the :token clipboard
+
+    # allow time for the alert to fire
+    Given 180 seconds have passed
```
liangxia: How about use wait up to instead of hard waiting ?

jhou1: Good suggestion, I updated it.

PR #619, opened 2020-01-13
Replace obsoleted "oc env" with "oc set env"

liangxia:
\# The cmd :env can be removed after 3.1 EOL (nov 2018).
\# We have both :env and :set_env since 3.2.
\# Removed :env in 3.11.

@pruan-rht @akostadinov @zhouying7780


liangxia:
/hold
until https://github.com/openshift/cucushift/pull/7328 is merged.

liangxia:
/unhold
as openshift/cucushift#7328 was merged.

PR #623, opened 2020-01-13
Add multus default route overwrite

weliang1:
Auto testing log: http://pastebin.test.redhat.com/826842

@zhaozhanqi @huiran0826 @anuragthehatter @rbbratta

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -1027,3 +1027,33 @@ Feature: Multus-CNI related scenarios
       | /usr/sbin/ip | a  |
     Then the output should contain:
       | c2:b0:57:49:47:f1 |
+
+  # @author weliang@redhat.com
+  # @case_id OCP-25915
+  @admin
+  @destructive
+  Scenario: Multus default route overwrite
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/ipam-static.yaml" replacing paths:
+      | ["metadata"]["namespace"]  | <%= project.name %> | 
+      | ["metadata"]["name"]       | bridge-static       |
+      | ["spec"]["config"]| '{ "cniVersion": "0.3.0", "type": "bridge", "master": "ens3", "ipam": {"type":"static","addresses": [{"address": "22.2.2.22/24","gateway": "22.2.2.1"}]}}' |
+    Then the step should succeed
+      
+    # Create a pod absorbing above net-attach-def
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/multus-default-route-pod.yaml   |
+      | n | <%= project.name %>                                                                                                    |
+    Then the step should succeed
+    And the pod named "multus-default-route-pod" becomes ready
+
+    # Check created pod has correct default route
+    When I execute on the pod:
+      | /usr/sbin/ip | route             |
+    Then the output should contain:
+      | default via 22.2.2.254 dev net1  |
+
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -1031,7 +1031,6 @@ Feature: Multus-CNI related scenarios
   # @author weliang@redhat.com
   # @case_id OCP-25915
   @admin
-  @destructive
   Scenario: Multus default route overwrite
     # Make sure that the multus is enabled
     Given the multus is enabled on the cluster
@@ -1046,8 +1045,8 @@ Feature: Multus-CNI related scenarios
       
     # Create a pod absorbing above net-attach-def
     When I run the :create client command with:
-      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/multus-default-route-pod.yaml   |
-      | n | <%= project.name %>                                                                                                    |
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/multus-default-route-pod.yaml |
+      | n | <%= project.name %>                                                                                                         |
     Then the step should succeed
     And the pod named "multus-default-route-pod" becomes ready
 


features/networking/multus.feature
```diff
@@ -1008,3 +1008,30 @@ Feature: Multus-CNI related scenarios
...
+    Given I have a project
+    When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/ipam-static.yaml" replacing paths:
+      | ["metadata"]["namespace"]  | <%= project.name %> | 
+      | ["metadata"]["name"]       | bridge-static       |
+      | ["spec"]["config"]| '{ "cniVersion": "0.3.0", "type": "bridge", "master": "ens3", "ipam": {"type":"static","addresses": [{"address": "22.2.2.22/24","gateway": "22.2.2.1"}]}}' |
+    Then the step should succeed
+      
+    # Create a pod absorbing above net-attach-def
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/multus-default-route-pod.yaml   |
```
pruan-rht: @weliang1 pls try to avoid using personal git-repo.  Test templates should be pushed to `openshift-qe/v3-testfils` instead.

weliang1: Update to use openshift-qe/v3-testfiles. Thanks!

features/networking/multus.feature
```diff
@@ -1008,3 +1008,32 @@ Feature: Multus-CNI related scenarios
...
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/multus-default-route-pod.yaml |
+      | n | <%= project.name %>                                                                                                         |
+    Then the step should succeed
+    And the pod named "multus-default-route-pod" becomes ready
+
+    # Check created pod has correct default route
+    When I execute on the pod:
+      | /usr/sbin/ip | route            |
+    Then the output should contain
```
anuragthehatter: This test will fail unless you have colon at the end of this line like ```Then the output should contain:```
Refer https://github.com/anuragthehatter/verification-tests/blob/master/features/step_definitions/common.rb#L33

features/networking/multus.feature
```diff
@@ -1008,3 +1008,32 @@ Feature: Multus-CNI related scenarios
...
+    When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/ipam-static.yaml" replacing paths:
+      | ["metadata"]["namespace"]  | <%= project.name %> | 
+      | ["metadata"]["name"]       | bridge-static       |
+      | ["spec"]["config"]| '{ "cniVersion": "0.3.0", "type": "bridge", "master": "<%= cb.default_interface %>", "ipam": {"type":"static","addresses": [{"address": "22.2.2.22/24","gateway": "22.2.2.1"}]}}' |
+    Then the step should succeed
+    
+    # Create a pod absorbing above net-attach-def
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/multus-default-route-pod.yaml |
+      | n | <%= project.name %>                                                                                                         |
```
anuragthehatter: You dont need this. 1020 will be creating course to use that project only

weliang1: Fixed. Thanks!

features/networking/multus.feature
weliang1: Fixed. Thanks!

pruan-rht: @weliang1 please run do a sanity check of your PR before submitting it by running it locally and/or against jenkins (preferable the latter) to make sure it work.

weliang1: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/520/consoleFull

features/networking/multus.feature
```diff
@@ -1008,3 +1008,31 @@ Feature: Multus-CNI related scenarios
...
+    # Make sure that the multus is enabled
+    Given the multus is enabled on the cluster
+    Given the default interface on nodes is stored in the :default_interface clipboard
+
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/ipam-static.yaml" replacing paths:
+      | ["metadata"]["namespace"]  | <%= project.name %> | 
+      | ["metadata"]["name"]       | bridge-static       |
+      | ["spec"]["config"]| '{ "cniVersion": "0.3.0", "type": "bridge", "master": "<%= cb.default_interface %>", "ipam": {"type":"static","addresses": [{"address": "22.2.2.22/24","gateway": "22.2.2.1"}]}}' |
```
zhaozhanqi: hi, @weliang1 I saw this case is working after 4.2 version.  So in 4.2  we do not have to add  `master` parameter in the net-attach-def. Actually. it's a bug IMO in 4.1. 
So I suggest here we can delete the `"master"` , thus above line 1017 can also be removed.

weliang1: @zhaozhanqi , still get error when define net-attach-def under normal users:
STDERR:
      Error from server (Forbidden): error when creating "STDIN": network-attachment-definitions.k8s.cni.cncf.io is forbidden: User "testuser-0" cannot create resource "network-attachment-definitions" in API group "k8s.cni.cncf.io" in the namespace "qr26w"

anuragthehatter: @weliang1 I believe what Zhanqi mentioned is you don't need "master": "<%= cb.default_interface %>" in net-attach-def and hence line 1017 can also be removed. Test should pass that way

weliang1: @anuragthehatter @zhaozhanqi  Got it now, removed "master": "<%= cb.default_interface %>" and testing passed. Thanks!

pruan-rht:
@weliang1 this test looks destructive, if it's then we need to add the `destructive` tag.  Also, should there be a clean-up of the interface you modified/created?

pruan-rht:
@weliang1 looking at more, it looks like the interface is created under the user project, so it should be removed automatically when the test finishes and the framework does its cleanup steps.

weliang1:
@pruan-rht  add @destructive and do the rebase, re run the test and pass, here is running log: http://pastebin.test.redhat.com/829315

features/networking/multus.feature
```diff
@@ -1027,3 +1027,33 @@ Feature: Multus-CNI related scenarios
...
+    Given I have a project
+    When I run oc create as admin over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/NetworkAttachmentDefinitions/ipam-static.yaml" replacing paths:
+      | ["metadata"]["namespace"]  | <%= project.name %> | 
+      | ["metadata"]["name"]       | bridge-static       |
+      | ["spec"]["config"]| '{ "cniVersion": "0.3.0", "type": "bridge", "master": "ens3", "ipam": {"type":"static","addresses": [{"address": "22.2.2.22/24","gateway": "22.2.2.1"}]}}' |
+    Then the step should succeed
+      
+    # Create a pod absorbing above net-attach-def
+    When I run the :create client command with:
+      | f | https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/multus/multus-default-route-pod.yaml   |
```
pruan-rht: @weliang1 your rebase changed it back to using your personal repo again.  Please update to use the `v3-testfiles` repo.  Also, if this test can be run concurrently w/o the possibility  of interference, then we don't need the `destructive` tag.

PR #626, opened 2020-01-13
OCP-14354: add steps to restore deleted node from clipboard

rbbratta:
For OCP-14354 we need to delete a node, but the test will fail
unless we re-create it at the end.

Add two steps, one to save the node yaml to a clipboard and
the other to re-create the node from the clipboard yaml

### DIFF
--- a/features/step_definitions/node.rb
+++ b/features/step_definitions/node.rb
@@ -24,7 +24,7 @@ Given /^I store the( schedulable| ready and schedulable)? (node|master|worker)s 
   else
     cb[cbname] = env.nodes.select { |n| n.ready? && n.schedulable? }
   end
-  
+
   if role == "worker"
     cb[cbname] = cb[cbname].select { |n| n.is_worker? }
   elsif role == "master"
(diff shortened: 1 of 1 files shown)


qinpingli:
LGTM

features/step_definitions/node.rb
```diff
@@ -591,3 +591,33 @@
   nodes = BushSlicer::Node.list(user: admin)
   cb[cb_name] = nodes.select { |n| n.is_worker? }
 end
+
+Given /^I store the node #{QUOTED} to the#{OPT_SYM} clipboard$/ do |node, cb_name|
```
anuragthehatter: @rbbratta Can we change it to ```Given /^I store the node #{QUOTED} **YAML** to the#{OPT_SYM} clipboard$```. As right now it would seen to the reader as like we are saving node's name into the clipboard.

rbbratta: make sense

anuragthehatter: Thanks @rbbratta ! You need to change scenario using this step as well.

rbbratta: ah, the `**` mess up the regex.  maybe just all caps `YAML` works?

anuragthehatter: Sorry for confusion i meant ```Given I store the node #{QUOTED} yaml to the#{OPT_SYM} clipboard```

PR #628, opened 2020-01-14
Add command "oc create user"

liangxia:


xiaocwan:
IMO It's better to decouple the  command, since each command has `options` which means step could and could not use the params, it all depends on the command
Why add the specific sub-command for each `:create`? Then we will need more sub-commands as below. Please tell me how do you think, thanks!

Available Commands: (for `oc create`)

-   clusterresourcequota Create cluster resource quota resource.
-   clusterrole          Create a ClusterRole.
-   clusterrolebinding   Create a ClusterRoleBinding for a particular ClusterRole
-   configmap            Create a configmap from a local file, directory or literal value
-   cronjob              Create a cronjob with the specified name.
-   deployment           Create a deployment with the specified name.
-   deploymentconfig     Create deployment config with default options that uses a given image.
-   identity             Manually create an identity (only needed if automatic creation is disabled).
-   imagestream          Create a new empty image stream.
-   imagestreamtag       Create a new image stream tag.
-   job                  Create a job with the specified name.
-   namespace            Create a namespace with the specified name
-   poddisruptionbudget  Create a pod disruption budget with the specified name.
-   priorityclass        Create a priorityclass with the specified name.
-   quota                Create a quota with the specified name.
-   role                 Create a role with single rule.
-   rolebinding          Create a RoleBinding for a particular Role or ClusterRole
-   route                Expose containers externally via secured routes
-   secret               Create a secret using specified subcommand
-   service              Create a service using specified subcommand.
-   serviceaccount       Create a service account with the specified name
-   user                 Manually create a user (only needed if automatic creation is disabled).
-   useridentitymapping  Manually map an identity to a user.

liangxia:
Yes, that's what we are doing and are the commands we need to test.
We have part of them which means we have test coverage.
```
$ grep -r ":create_" lib/rules/cli/4.4.yaml 
:create_clusterresourcequota:
:create_clusterrolebinding:
:create_configmap:
:create_namespace:
:create_quota:
:create_rolebinding:
:create_route_edge:
:create_route_passthrough:
:create_route_reencrypt:
:create_secret:
:create_service:
:create_service_loadbalancer:
:create_serviceaccount:
```

xiaocwan:
LGTM
@pruan-rht Is this good to merge?

PR #629, opened 2020-01-14
Rewrite scenarios to be non-destructive

liangxia:
They are marked as destructive because we have created PV in those scenarios, and PV can interfere with PVCs created in other scenarios.
Now we limit PV/PVC in specific storage classes, so they no longer interfere with other PVs/PVCs.


duanwei33:
/LGTM
Good to remove the destructive tag, thanks for the update.

liangxia:
@pruan-rht  Could you check ?

PR #634, opened 2020-01-15
Automate test case OCP-25628

jhou1:
Automate test case OCP-25628.

The test covers that when a machine is start at Deleting state, an alert is fired. Tested locally

```
      [10:08:38] INFO> === End After Scenario: Alert should be fired when machine is in 'Deleting' state for too long ===

1 scenario (1 passed)
12 steps (12 passed)
1m18.511s
```

@liangxia @pruan-rht can you please review?

### DIFF
--- a/features/machine/alerting.feature
+++ b/features/machine/alerting.feature
@@ -62,3 +62,57 @@ Feature: Alerting for machine-api
       | cluster-autoscaler-operator | openshift-machine-api              | ClusterAutoscalerOperatorDown | # @case_id OCP-26250
       | machine-api-operator        | openshift-machine-api              | MachineAPIOperatorDown        | # @case_id OCP-26248
       | machine-approver            | openshift-cluster-machine-approver | ClusterMachineApproverDown    | # @case_id OCP-26110
+
+  # @author jhou@redhat.com
+  # case_id OCP-25628
+  @admin
+  @destructive
+  Scenario: Alert should be fired when machine is in 'Deleting' state for too long
+    Given I have an IPI deployment
+    And I switch to cluster admin pseudo user
+    And I pick a random machineset to scale
+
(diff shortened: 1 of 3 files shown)


features/machine/alerting.feature
```diff
@@ -62,3 +62,57 @@ Feature: Alerting for machine-api
...
+    When I run the :annotate client command with:
+      | n            | openshift-machine-api                       |
+      | resource     | machine                                     |
+      | resourcename | <%= machine.name %>                         |
+      | overwrite    | true                                        |
+      | keyval       | machine.openshift.io/exclude-node-draining= |
+    Then the step should succeed
+    """
+
+    Given I run the :delete admin command with:
```
pruan-rht: @jhou1 perhaps we can use `admin ensure` step instead here.

features/step_definitions/machine_healthcheck.rb
```diff
@@ -22,6 +22,7 @@
     | ["spec"]["nodeName"] | #{machine.node_name}  |
   })
   step %Q{the step should succeed}
+  step %Q{admin ensures "kubelet-killer" pod is deleted from the "openshift-machine-api" project after scenario}
```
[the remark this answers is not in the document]
jhou1: It's defined in this step https://github.com/openshift/verification-tests/blob/942bab5dc9b868305b1e57560c1b5d2539218a2d/features/step_definitions/machine_healthcheck.rb#L20

features/machine/alerting.feature
jhou1: @pruan-rht this step is to create the `machine is in 'Deleting' state for too long` scenario. The deletion won't be successful until we add the annotation in line#85 in tear down. Then the `admin ensure` is in the last line of this test could delete the machine to ensure a consistent test env.

PR #635, opened 2020-01-15
fix warnings introuduced with ruby 2.7, which warns about backward co…

pruan-rht:
…mpatiblity against ruby 3.0  for example:

> DEPRECATION WARNING: Ruby 2.5+ (required by Rails 6) provides Hash#transform_values natively, so requiring active_support/core_ext/hash/transform_values is no longer necessary. Requiring it will raise LoadError in Rails 6.1. (called from require at /home/pruan/workspace/bushslicer/features/step_definitions/dc.rb:4)
> /home/pruan/workspace/bushslicer/lib/host.rb:532: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

### DIFF
--- a/features/step_definitions/dc.rb
+++ b/features/step_definitions/dc.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/hash/slice.rb'
-require 'active_support/core_ext/hash/transform_values.rb'
+#require 'active_support/core_ext/hash/transform_values.rb'
 
 Given /^I wait until the status of deployment "(.+)" becomes :(.+)$/ do |resource_name, status|
   ready_timeout = 10 * 60
--- a/features/step_definitions/dc.rb
+++ b/features/step_definitions/dc.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/hash/slice.rb'
-#require 'active_support/core_ext/hash/transform_values.rb'
 
 Given /^I wait until the status of deployment "(.+)" becomes :(.+)$/ do |resource_name, status|
   ready_timeout = 10 * 60
@@ -150,6 +149,15 @@ Given /^(I|admin) redeploys? #{QUOTED} dc( after scenario)?$/ do |who, dc_name, 
 end
 
 Given /^master CA is added to the#{OPT_QUOTED} dc$/ do |name|
+  ensure_admin_tagged
+
+  org_user = @user
+  org_proj_name = project.name
+  @user = admin
+  @result = secret("router-certs-default", project("openshift-ingress", switch: false)).value_of("tls.crt")
+  File.open("/tmp/openshift.crt", 'wb') { |f|
+    f.write(@result)
+  }
 
   step %Q/certification for default image registry is stored to the :reg_crt_name clipboard/
   step %Q/I run the :create_configmap client command with:/, table(%{
--- a/features/step_definitions/deployment.rb
+++ b/features/step_definitions/deployment.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/hash/slice.rb'
-require 'active_support/core_ext/hash/transform_values.rb'
+#require 'active_support/core_ext/hash/transform_values.rb'
 
 ### Deployment related steps
 
--- a/features/step_definitions/deployment.rb
+++ b/features/step_definitions/deployment.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/hash/slice.rb'
-#require 'active_support/core_ext/hash/transform_values.rb'
 
 ### Deployment related steps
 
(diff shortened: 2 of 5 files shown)


pruan-rht:
for reference https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/ . Tested against ruby 2.6.3  it worked

features/step_definitions/dc.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/hash/slice.rb'
-require 'active_support/core_ext/hash/transform_values.rb'
+#require 'active_support/core_ext/hash/transform_values.rb'
```
rbbratta: could we just remove the line instead of commenting it out?

pruan-rht: @rbbratta sure, done.

pruan-rht:
@akostadinov PTAL.  Thanks!

features/step_definitions/deployment.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/hash/slice.rb'
-require 'active_support/core_ext/hash/transform_values.rb'
```
akostadinov: I know this is causing deprecation warning. But we still support running on 2.4. Perhaps require it conditionally if RUBY version is above 2.5?

lib/environment_manager.rb
```diff
@@ -37,7 +37,7 @@ def conf
       raise "no '#{key}' environment configuration found" if ! env_class || env_class.empty?
 
       env_opts[:key] = key # lets have it for a reference
-      return BushSlicer.const_get(env_class).new(env_opts)
+      return BushSlicer.const_get(env_class).new(**env_opts)
```
akostadinov: why do we need this change?

lib/host.rb
```diff
@@ -529,9 +529,9 @@ def delete(file, opts={})
       else
         exec_method = :exec
       end
-      public_send(exec_method, "rm #{r} -f -- #{file}", opts)
+      public_send(exec_method, "rm #{r} -f -- #{file}", **opts)
       opts[:quiet] = true
-      res = public_send(exec_method, "ls -d -- #{file}", opts)
+      res = public_send(exec_method, "ls -d -- #{file}", **opts)
```
akostadinov: You sure we need these changes?

features/step_definitions/deployment.rb
akostadinov: or shall we stop supporting ruby 2.4? Maybe ask @jianlinliu

lib/environment_manager.rb
pruan-rht: @akostadinov because w/o the change we will see this warning  `warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`

lib/host.rb
pruan-rht: @akostadinov yes.  https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

features/step_definitions/deployment.rb
pruan-rht: @jianlinliu WDYT about the stop support for ruby 2.4?

pruan-rht: @jianlinliu can you please comment on this?

jianlinliu: I am not sure what is meaning of support for ruby 2.4. I have no any cucushift test cases.

features/step_definitions/dc.rb
```diff
@@ -150,6 +149,15 @@
 end
 
 Given /^master CA is added to the#{OPT_QUOTED} dc$/ do |name|
+  ensure_admin_tagged
```
cben: this chunk seems unrelated to this PR?

features/step_definitions/deployment.rb
pruan-rht: @akostadinov @jianlinliu with F32 out now, the default ruby is 2.7.1, can we merge this PR?  Otherwise, the cucumber output will have lots of warnings.  @jianlinliu what it means is that we would update our jenkins slave pods to have ruby > 2.5

tools/find_image.rb
```diff
@@ -32,14 +32,14 @@ def run
...
-        doc = Nokogiri::HTML(open(options.url).read)
+        doc = Nokogiri::HTML(URI.open(options.url).read)
         links = doc.xpath("//a[contains(text(), '#{options.build_type}')]")
 
         target_links = links.map { |l|  options.url + l.attributes['href'].value }
         threads = []
         target_links.each do |link|
           threads << Thread.new(link) do |i|
-            doc = Nokogiri::HTML(open(link).read)
+            doc = Nokogiri::HTML(URI.open(link).read)
```
akostadinov, line 42: @pruan-rht , Ithink the other fixes are already merged, do you want to submit only changes to this file?

PR #638, opened 2020-01-15
Few fixes and optimizations

anuragthehatter:
Found during nightly CI runs
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/534/console
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/535/console
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/536/console
@pruan-rht Can you help merging it?

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -446,7 +446,7 @@ Feature: Multus-CNI related scenarios
     Then the step should succeed
     And the output should contain 2 times:
       | 1 PVID   |
-      | untagged |
+      | ntagged |
 
   # @author anusaxen@redhat.com
   # @case_id OCP-24489
@@ -490,12 +490,13 @@ Feature: Multus-CNI related scenarios
   @admin
   @destructive
   Scenario: CNO manager mavlan configured manually with static
-    # Make sure that the multus is Running
     Given the multus is enabled on the cluster
+    And I store all worker nodes to the :nodes clipboard
     Given the default interface on nodes is stored in the :default_interface clipboard 
     #Patching simplemacvlan config in network operator config CRD
+    Given I have a project
     Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with:
-      | {"spec":{"additionalNetworks":[{"name":"test-macvlan-case3","namespace":"openshift-multus","simpleMacvlanConfig":{"ipamConfig":{"staticIPAMConfig":{"addresses": [{"address":"10.128.2.100/23","gateway":"10.128.2.1"}]},"type":"static"},"master":"<%= cb.default_interface %>","mode":"bridge"},"type":"SimpleMacvlan"}]}} |
+	    | {"spec":{"additionalNetworks":[{"name":"test-macvlan-case3","namespace":"<%= project.name %>","simpleMacvlanConfig":{"ipamConfig":{"staticIPAMConfig":{"addresses": [{"address":"10.128.2.100/23","gateway":"10.128.2.1"}]},"type":"static"},"master":"<%= cb.default_interface %>","mode":"bridge"},"type":"SimpleMacvlan"}]}} |
     #Cleanup for bringing CRD to original
     Given I register clean-up steps:
     """
@@ -506,23 +507,21 @@ Feature: Multus-CNI related scenarios
     And I wait up to 30 seconds for the steps to pass:
     """
     When I run the :get admin command with:
-      | resource | net-attach-def   |
-      | n        | openshift-multus |
+      | resource | net-attach-def      |
+      | n        | <%= project.name %> |
     Then the step should succeed
     And the output should contain:
       | test-macvlan-case3 |
     """
     #Creating pod under openshift-multus project to absorb above net-attach-def
-    Given I switch to cluster admin pseudo user
-    Given I use the "openshift-multus" project
     When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/Pods/1interface-macvlan-bridge.yaml" replacing paths:
-      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | test-macvlan-case3 |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | test-macvlan-case3      |
+      | ["metadata"]["namespace"]                                  | <%= project.name %>     |
+      | ["spec"]["nodeName"]                                       | <%= cb.nodes[0].name %> |
     Then the step should succeed
     And a pod becomes ready with labels:
       | name=macvlan-bridge-pod |
     And evaluation of `pod` is stored in the :pod clipboard
-    And admin ensures "<%= cb.pod.name %>" pod is deleted from the "openshift-multus" project after scenario
-    
     When I execute on the pod:
       | /usr/sbin/ip | -d | link |
     Then the output should contain "net1"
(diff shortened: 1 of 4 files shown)


features/networking/multus.feature
```diff
@@ -446,7 +446,7 @@ Feature: Multus-CNI related scenarios
     Then the step should succeed
     And the output should contain 2 times:
       | 1 PVID   |
-      | untagged |
+      | ntagged |
```
anuragthehatter: I kept it like this. At this point this will work. Its not good to play with changing this definition as 1000's of cases are using it.
https://github.com/anuragthehatter/verification-tests/blob/master/features/step_definitions/common.rb#L72

features/networking/multus.feature
```diff
@@ -490,12 +490,13 @@ Feature: Multus-CNI related scenarios
   @admin
   @destructive
   Scenario: CNO manager mavlan configured manually with static
-    # Make sure that the multus is Running
     Given the multus is enabled on the cluster
+    And I store all worker nodes to the :nodes clipboard
```
anuragthehatter: In this case i am simply forcing test to execute on a worker node. Earlier it was on master

features/networking/service.feature
```diff
@@ -109,7 +109,8 @@ Feature: Service related networking scenarios
   # @case_id OCP-23895
   @admin
   Scenario: User cannot access the MCS by creating a LoadBalancer service that points to the MCS
-    Given evaluation of `env.master_hosts.first.local_ip` is stored in the :master_ip clipboard
+    Given I store the masters in the :masters clipboard
```
anuragthehatter: ```env.master_hosts.first.local_ip``` won't work anymore guess due to now ens is just taking lb. Fixed it with workaround

features/networking/operator.feature
```diff
@@ -186,16 +186,21 @@ Feature: Operator related networking scenarios
     And the output should not contain "NotReady"
     Given the status of condition "Progressing" for network operator is :False
     """
+
   # @author anusaxen@redhat.com
   # @case_id OCP-24918
```
anuragthehatter: http://pastebin.test.redhat.com/828146
Changed scenario a more smarter way included polling to check a partcular keyword in sdn logs to make sure flag is updated across the sdn and

features/step_definitions/networking.rb
```diff
@@ -886,4 +886,23 @@
   @result = host.exec_admin("ifconfig #{cb.interface}")
   cb[cb_ipaddr]=@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]
   logger.info "The Internal IP of node is stored in the #{cb_ipaddr} clipboard."
-end
\ No newline at end of file
+end
+
+Given /^I store "([^"]*)" node's corresponding sdn or ovnkube pod name in the#{OPT_SYM} clipboard$/ do |node_name,cb_pod_name|
```
anuragthehatter: This step def will be savior in future for all sdn and ovn tests

pruan-rht: @anuragthehatter IMO, this step name should be something like `I store ...node's default network type pod name...`

anuragthehatter: Commited! Smart suggestion @pruan-rht , thanks

features/networking/operator.feature
```diff
@@ -204,23 +209,34 @@ Feature: Operator related networking scenarios
...
   as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with: 
     | {"spec":{"defaultNetwork":{"openshiftSDNConfig": null}}} |
-  120 seconds have passed
   """
-  And 120 seconds have passed
+  And I wait up to 300 seconds for the steps to pass:
+    """
+    When I run the :logs admin command with:
+      | resource_name | <%= cb.sdn_pod %> |
+      | namespace     | openshift-sdn     |
```
zhaozhanqi: @anuragthehatter Thanks for fixing issue.  I saw you add one step to get the sdn or ovn pod in line 202.  but here the namespace is 'openshift-sdn'. So if ovn the ns should be 'openshift-ovn-kubernetes'

anuragthehatter: @zhaozhanqi As we can see, this step is reusable in future irrespective of networktype. Also this particular case is specific to openshift-sdn. This feature won't be supported on OVN as per talk with Casey during this story. So we can keep as is

pruan-rht:
@anuragthehatter what I mean is these steps are duplicated within the `if/else` block.  They look the same except for a couple of parameters.  	 So you still need the `if/else`  block, but to just set the parameters like `app=xxx` and `project` parameter.

> cb[cb_pod_name] = BushSlicer::Pod.get_labeled("app=sdn", project: project("openshift-sdn", switch: false), user: admin) { |pod, hash|
>        pod.node_name == node_name
>      }.first.name
>      logger.info "node's corresponding sdn pod name is stored in the #{cb_pod_name} clipboard."

anuragthehatter:
> @anuragthehatter what I mean is these steps are duplicated within the `if/else` block. They look the same except for a couple of parameters. So you still need the `if/else` block, but to just set the parameters like `app=xxx` and `project` parameter.
> 
> > cb[cb_pod_name] = BushSlicer::Pod.get_labeled("app=sdn", project: project("openshift-sdn", switch: false), user: admin) { |pod, hash|
> > pod.node_name == node_name
> > }.first.name
> > logger.info "node's corresponding sdn pod name is stored in the #{cb_pod_name} clipboard."

@pruan-rht understand, now, That optimization make sense. Thanks for bringing this up.

anuragthehatter:
> @anuragthehatter what I mean is these steps are duplicated within the `if/else` block. They look the same except for a couple of parameters. So you still need the `if/else` block, but to just set the parameters like `app=xxx` and `project` parameter.
> 
> > cb[cb_pod_name] = BushSlicer::Pod.get_labeled("app=sdn", project: project("openshift-sdn", switch: false), user: admin) { |pod, hash|
> > pod.node_name == node_name
> > }.first.name
> > logger.info "node's corresponding sdn pod name is stored in the #{cb_pod_name} clipboard."

Fixed and tested on both networks. Thanks!

features/step_definitions/networking.rb
```diff
@@ -886,4 +886,23 @@
...
+  _admin = admin
+  @result = _admin.cli_exec(:get, resource: "network.operator", output: "jsonpath={.items[*].spec.defaultNetwork.type}")
+  if @result[:response] == "OpenShiftSDN"
+     app="app=sdn"
+     project_name="openshift-sdn"
+  else
+     app="app=ovnkube-node"
+     project_name="openshift-ovn-kubernetes"
+  end   
+  raise "Unable to find corresponding networkType pod name" unless @result[:success]
```
pruan-rht: @anuragthehatter how about move this up before the `if/else` block?

features/step_definitions/networking.rb
```diff
@@ -886,4 +886,23 @@
...
+  cb[cb_pod_name] = BushSlicer::Pod.get_labeled(app, project: project(project_name, switch: false), user: admin) { |pod, hash|
```
pruan-rht: @anuragthehatter, we need to assign a default value for `cb_pod_name`, otherwise it will be `nil` and go to the default which might have some values there already.

features/step_definitions/networking.rb
anuragthehatter: Yep agree, that should be before if/else. Done!

features/step_definitions/networking.rb
anuragthehatter: Done, thanks! I usually do that but missed here

features/networking/operator.feature
zhaozhanqi: @anuragthehatter thanks. then it's ok

PR #643, opened 2020-01-17
Fix clone from storage class without parameters

liangxia:
When we use below step to clone from storage class which does not have parameters(E.g, on OpenStack), it will fail.
```
And admin clones storage class "storageclass-<%= project.name %>" from ":default" with:
  | ["parameters"]["fstype"] | <fstype> |
```

Pass [log](https://privatebin-it-iso.int.open.paas.redhat.com/?9c771e5fc0409aa7#mIAssdpGUnPIA8ALPWQSyPplA5UTYpJnHvtqzyZA1eM=) on OpenStack.

@chao007 @duanwei33 @pruan-rht @akostadinov


chao007:
LGTM. 
Thanks

PR #644, opened 2020-01-19
Update OCP-9639

chao007:
Step `When I run commands on the host` is interrupted sometimes, so add the tag `@destructive`
@liangxia @duanwei33 Please help reiview.


liangxia:
> Step `When I run commands on the host` is interrupted sometimes, so add the tag `@destructive`

I'm wondering how is the issue resolved with the tag added ?

chao007:
This case can be executed alone. And I test it can be passed

qinpingli:
@chao007 Do we hit this issue in the automation tests recently? If no, we can close this pr.

PR #650, opened 2020-01-19
Remove unused parameters

liangxia:
Those parameters are no-longer used in the scenarios.


liangxia:
closing as we do not want to break v3.

PR #655, opened 2020-01-20
Add a step for rsh command in pod

xiaocwan:
@xingxingxia @xiaojiey Please check is that what you need? Any param needs to add?
Tested with a simple test as below and PASS locally: http://pastebin.test.redhat.com/828646
```
Feature: test rsh
  # @author xiaocwan@redhat.com
  # @case_id OCP-test
  @admin
  Scenario: test pod rsh
    Given I have a project
      | template | django-psql-example     | 
    Given the "django-psql-example-1" build completes
    Given 1 pods become ready with labels:
      | deployment=postgresql-1 |
    When I rsh on the pod:
      | ls | -al |
    Then the step should succeed
```


xiaocwan:
Seems not using it now, will close this PR.

PR #660, opened 2020-01-23
added method to return images on a node

pruan-rht:


pruan-rht:
@akostadinov PTAL, thanks!

akostadinov:
could you show example output? Can `images` method return an array of  `Image` objects instead of some structure? Similar to `Node#taints`.

lib/openshift/node.rb
```diff
@@ -255,5 +255,14 @@ def max_pod_count_schedulable(user: nil, cached: true, quiet: false, **pod_reque
       return remaining_pods_count if pod_requests.empty?
       return [ max_pod_count_capacity(user: user, cached: false, quiet: quiet, **pod_requests), remaining_pods_count].min
     end
+
+    def images(user: nil, cached: true, quiet: false)
```
akostadinov, line 259: you can rename this to `private def raw_images(...)` and add a method `def images(...)` like in 
```
    def images(user: nil, cached: true, quiet: false)
      unless cached && props[:images]
        props[:images] = raw_images(user: user, cached: cached, quiet: quiet).map do |rimage|
          Image.new(name: rimage['names'].first, env: @owner.env)
        end
      end
      return props[:images]
    end
```
My concern though is whether there images are OpenShift resources of just some list of images that do not exist in the environment in any form. Do you know this? I assume they should exist.

akostadinov (commented):
This fell through the cracks. Lets finish this PR.

PR #661, opened 2020-01-24
Changing definition to support mixed nodes

anuragthehatter:
The existing definition is breaking on a cluster with mixed nodes as noted in CI runs. The reason is cluster with rhel7+rhcos nodes might have different master interface names on them respectively. Existing definition calls another step in it which changes course from existing node to a different one. We need to make sure the definition should execute on a single node itself from starting to end.

We could have shortened this definition by executing on host but i prefer to get inf names via network pods due to sometime we get STDERR on host commands.

Confirmed via logs that this whole def is executing on a single node now http://pastebin.test.redhat.com/830405

@pruan-rht Can you help merging it?

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -881,9 +881,24 @@ end
 Given /^the Internal IP of node "([^"]*)" is stored in the#{OPT_SYM} clipboard$/ do |node_name,cb_ipaddr|
   ensure_admin_tagged
   node = node(node_name)
+  _admin = admin
   host = node.host
-  step "the default interface on nodes is stored in the clipboard"
-  @result = host.exec_admin("ifconfig #{cb.interface}")
+  cb_ipaddr ||= "ip_address"
+  @result = _admin.cli_exec(:get, resource: "network.operator", output: "jsonpath={.items[*].spec.defaultNetwork.type}")
+  if @result[:success] then
+     networkType = @result[:response].strip
+  end
+  case networkType
+  when "OVNKubernetes"
+    step %Q/I run command on the node's ovnkube pod:/, table("| bash | -c | ip route show default |")
+  when "OpenShiftSDN"
+    step %Q/I run command on the node's sdn pod:/, table("| bash | -c | ip route show default |")
+  else
+    raise "unknown networkType"
+  end
+  def_inf = @result[:response].split("\n").first.split(/\W+/)[7]
+  logger.info "The node's default interface is #{def_inf}"
+  @result = host.exec_admin("ifconfig #{def_inf}")
   cb[cb_ipaddr]=@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]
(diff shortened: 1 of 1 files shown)


features/step_definitions/networking.rb
```diff
@@ -881,9 +881,24 @@
 Given /^the Internal IP of node "([^"]*)" is stored in the#{OPT_SYM} clipboard$/ do |node_name,cb_ipaddr|
   ensure_admin_tagged
   node = node(node_name)
+  _admin = admin
```
akostadinov: what is the point to create this local variable?

akostadinov: Other than that lgtm.

anuragthehatter: Indeed we don't need it, Thanks for catching that redundancy. Fixed.

features/step_definitions/networking.rb
```diff
@@ -881,9 +881,24 @@
...
+  when "OVNKubernetes"
+    step %Q/I run command on the node's ovnkube pod:/, table("| bash | -c | ip route show default |")
+  when "OpenShiftSDN"
+    step %Q/I run command on the node's sdn pod:/, table("| bash | -c | ip route show default |")
+  else
+    raise "unknown networkType"
+  end
+  def_inf = @result[:response].split("\n").first.split(/\W+/)[7]
+  logger.info "The node's default interface is #{def_inf}"
+  @result = host.exec_admin("ifconfig #{def_inf}")
```
pruan-rht: @anuragthehatter should we use `ip addr` instead of `ifconfig`

features/step_definitions/networking.rb
```diff
@@ -882,8 +882,22 @@
...
   cb[cb_ipaddr]=@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]
```
pruan-rht: @anuragthehatter I've seen this before...perhaps we should think about putting it into a method call #ip_addr or something.  @akostadinov, suggestion as to where we might to put this?  base_helper.rb?

pruan-rht:
@anuragthehatter FYI, added https://github.com/openshift/verification-tests/pull/662

features/step_definitions/networking.rb
akostadinov: If internal IP of node is different from IP reported by `oc get node -o yaml`, then we can add method `Node#internal_ip`. Good suggestion.

I'm just curious why node ip reported by API should be different from what we get in this way.

anuragthehatter: @akostadinov @pruan-rht Internal IP of node is "not" different from IP reported by oc get node -o yaml. I don't think so we support node.ip etc hence this was was designed for use specifically for networking use cases.

pruan-rht: @anuragthehatter, ok, I see...  Let's merge this for now and we can put a TODO for enhancement later.

PR #662, opened 2020-01-24
added network object support

pruan-rht:
helper for #661

### DIFF
--- a/lib/openshift/network.rb
+++ b/lib/openshift/network.rb
@@ -1,0 +1,14 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  # @note represents an OpenShift environment Storage Class
+  class Network < ClusterResource
+    RESOURCE = "networks"
+
+    def network_type(user: nil, cached: true, quiet: false)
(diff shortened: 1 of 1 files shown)


pruan-rht:
@akostadinov PTAL.  thanks

lib/openshift/network.rb
```diff
@@ -0,0 +1,14 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  # @note represents an OpenShift environment Storage Class
+  class Network < ClusterResource
+    RESOURCE = "networks"
```
akostadinov: usually RESOURCE is `io.openshift...` nowadays. You will see this in YAML output of objects.

akostadinov: other than that lg

lib/openshift/network.rb
```diff
@@ -0,0 +1,14 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  # @note represents an OpenShift environment Storage Class
```
liangxia: Comment/Code mismatch

PR #664, opened 2020-01-31
Cloud cop 2

pruan-rht:


pruan-rht:
@akostadinov this PR addresses your comments from #376 .  PTAL.  Thanks!

akostadinov:
If three files are a bundle, perhaps you can put them under some sub-directory of `tools`. e.g. `cloud-cop` or something. If they are to be used by other commands under tools, then perhaps leave them. But I'm more inclined to put into a sub-dir.

 You can merge as you decide.

pruan-rht:
@akostadinov I'll merge this as is...I forsee the `jenkins.rb` file will be used elsewhere.

PR #670, opened 2020-02-04
Simplify steps to query prometheus

jhou1:
Per the [discussion](https://github.com/openshift/cucushift/pull/7356#issuecomment-577745707) with @pruan-rht and @liangxia , we'd like to simplify the prometheus queriy steps. Instead of running the `oc exec` on the `prometheus-k8s-0` or `prometheus-k8s-1` pod, we query the prometheus route, this makes sure the data is always consistent.

Verified my tests locally:
```
    [07:43:19] INFO> === End After Scenario: Alert should be fired when operator is down, Examples (#3) ===

3 scenarios (3 passed)
24 steps (24 passed)
4m25.926s
```

### DIFF
--- a/features/machine/alerting.feature
+++ b/features/machine/alerting.feature
@@ -38,23 +38,13 @@ Feature: Alerting for machine-api
     Then the step should succeed
     """
 
-    And I use the "openshift-monitoring" project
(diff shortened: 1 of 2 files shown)


jhou1:

akostadinov:
Looks nice! 
@pruan-rht ?

P.P. we still need step `I run curl command on the (CLO|ES|fluentd) pod to get metrics with:`?

PR #681, opened 2020-02-06
Fixing dumb-init path

jhutar:


pruan-rht:
@jhutar LG, but would you mind adding some logs to show how it is broken and the fix is correct?  Thanks!

jhutar:
Hello @pruan-rht. I have used this Dockerfile as a base for my own and when I have ran resulting container, it was complaining `/usr/bin/dumb-init: No such file or directory` or something like that. I just had to fix the path.

pruan-rht:
@jhutar I'm not familiar with that executable, but after reading this https://github.com/Yelp/dumb-init it seems that you have to install it?

jhutar:
Hello. It is being downloaded and chmod-ed here:

https://github.com/openshift/verification-tests/pull/681/files#diff-52f522915d70485a0cbd72c22bb05829L24-L25

Hope I'm not missing your point.

PR #684, opened 2020-02-10
Ensure machines and nodes are restored after scenario

jhou1:
The machine number is restored after scenario, but we did not wait for the node number to be restored. This PR address the enhancement.

@qinpingli Will you PTAL?


jhou1:
Thank you @qinpingli  @liangxia for your review. I will merge this because this PR is safe to merge(does not touch any library or step definition). Once this is merged it will not cause disruption to the environment.

PR #688, opened 2020-02-10
Run OCP-9565 in network-policy by default

weliang1:
Delete "Given the env is using multitenant network" in the script to let it run in network-policy by default

Testing log:
http://pastebin.test.redhat.com/834589

@zhaozhanqi @huiran0826 @rbbratta


akostadinov:

-- saved reply

PR #695, opened 2020-02-12
OCP-24465: select bridge IP from range to avoid bind error

rbbratta:
We try to create a bridge with an fixed IP, but somehow we keep
getting collision and thus bind errors.

Originally on 88.8.8.200 and now on 88.8.8.191

```
dnsmasq[27878]: dnsmasq: failed to create listening socket for 88.8.8.200: Address already in use
dnsmasq[187398]: dnsmasq: failed to create listening socket for 88.8.8.191: Address already in use
```

Insetad we can use a for loop to try to pick an address that doesn't
already exist on the system.

Slightly complicated, but we run `ip -4 -o addr show` and use sed to
pick out the IP address, then compare it to the number we selected
and hopefully it works.

Add a debug `ip -4 -o addr show` before the test to see the exsiting
interfaces

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -436,7 +436,7 @@ end
 Given /^the cluster network plugin type and version and stored in the clipboard$/ do
   ensure_admin_tagged
   _host = node.host
-  
+
   step %Q/I run command on the node's sdn pod:/, table([["ovs-ofctl"],["dump-flows"],["br0"],["-O"],["openflow13"]])
   unless @result[:success]
     raise "Unable to execute ovs command successfully. Check your command."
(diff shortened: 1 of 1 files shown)


zhaozhanqi:
thanks the update @rbbratta 
/lgtm

anuragthehatter:
@rbbratta Thanks. yep range will definitely solve this. This might have caused when such DHCP tests ran back to back and earlier test might not have released the IP on time for the next test.

/LGTM

rbbratta:
I think #705 might be the actual fix

liangxia:
Shall we close this one as #705 is merged ?

rbbratta:
yep, closing

PR #696, opened 2020-02-12
add one step to make sure all nodes are changed to ready

zhaozhanqi:
add one step to make sure all nodes changed to 'ready' otherwise it will affect the next case execution with error eg. error cannot get the token for user


zhaozhanqi:
@huiran0826 @weliang1 @anuragthehatter @rbbratta

PR #697, opened 2020-02-12
fix click_specified_idp

yapei:
@SSshahan  Can you help review? Thanks


SSshahan:
@pruan-rht  could you help to merge it? Thanks!

PR #700, opened 2020-02-12
add a rule to make sure kebab actions list open before next click

xiaocwan:
Update for both 4.3 and 4.4 for stable:
1) expect the kebab actions list open then choose the item
2) submit_changes actually existed then click it

Console 4.x framework is using ::after and ::before a lot, which makes our xpath unstable.
Refer to https://css-tricks.com/almanac/selectors/a/after-and-before/

Suggest solutions:
1) use expect before the click option (or before the next move, such as goto actions)
2) use CSS instead of xpath for more stable
Either one is okay.

This affects **OCP-25763** and **OCP-25800**, used to be failed: http://pastebin.test.redhat.com/835297
So I update it by 1st solution.
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/68535/console

@yapei @SSshahan @yanpzhan I suggest to have a discussion about this.
@yapei For this case, please review the change, thanks！

### DIFF
--- a/lib/rules/web/admin_console/4.3/actions.xyaml
+++ b/lib/rules/web/admin_console/4.3/actions.xyaml
@@ -8,9 +8,10 @@ click_resource_action_button:
     op: click
     timeout: 30
 choose_action_item_from_list:
-  action: check_actions_items_open
-  element:
-    selector:
+  elements:
+  - selector:
+      xpath: //*[@data-test-id="action-items"] 
+  - selector:
       xpath: //button[@data-test-action='<item>' and not(@disabled)]
     op: click
 choose_item_from_list:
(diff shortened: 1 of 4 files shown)


xiaocwan:
One more related case[ fixed by this, PASS on Jenkins](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/68536/console)

[Used to be failed, refer to the Failed log](http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/02/11/04:30:25/check_MachineHealthCheck_page_on_console/console.html)

PR #680, opened 2020-02-06
Update to run cucushift on oc 4.4 using --kubeconfig instead of --config

xiaocwan:
Hi @akostadinov and @pruan-rht 
Could we use `--kubeconfig` instead of `--config` since `--config` has been deprecated? I suggest to switch to the right flag for the oc 4.4.

$ oc version
Client Version: openshift-clients-4.4-23-g37623bc8
Server Version: 4.4.0-0.nightly-2020-02-05-181112

Mac using the latest oc version failed to run cucushift unless change `--config` to `--kubeconfig` as this PR did.

PR #702, opened 2020-02-13
Revert part of #680 to unblock 4.1/4.2/4.3 CI

liangxia:
https://github.com/openshift/verification-tests/pull/680

### DIFF
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -6,7 +6,7 @@
   :api_version: --api-version=<value>
   :ca: --certificate-authority=<value>
   :cluster: --cluster=<value>
-  :config: --kubeconfig=<value>
+  :config: --config=<value>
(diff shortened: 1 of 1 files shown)


liangxia:
Once the 'oc config' bug is fix, we can switch to --kubeconfig

liangxia:
/hold
Choose one of #702 and #703

liangxia:
/hold
To have more time/people to think about those issues.

lib/rules/cli/4.1.yaml
```diff
@@ -1110,7 +1110,7 @@
   :cmd: oc adm diagnostics
   :options:
     :cluster-context: --cluster-context=<value>
-    :config: --kubeconfig=<value>
+    :config: --config=<value>
```
akostadinov: this can be removed given it is already in global options

PR #694, opened 2020-02-12
Remove error/warning info from "oc version"

liangxia:
```
      [06:18:46] INFO> cleaning-up user testuser-0 projects
      [06:18:46] INFO> Shell Commands: rm  -f -- /home/jenkins/workspace/Runner-v3/workdir/ocp4_testuser-0.kubeconfig
      
      [06:18:46] INFO> Exit Status: 0
      [06:18:46] INFO> Shell Commands: oc version -o yaml --config=/tmp/kubeconfig20200211-364-1nbppq6
      Flag --config has been deprecated, use --kubeconfig instead
      clientVersion:
        buildDate: "2020-02-10T17:21:47Z"
        compiler: gc
        gitCommit: ee05f83cb8f0bee84c0a1516591437cd71befdc8
        gitTreeState: clean
        gitVersion: v4.4.0
        goVersion: go1.13.4
        major: ""
        minor: ""
        platform: linux/amd64
      
      
      [06:18:52] INFO> Exit Status: 0
      (<unknown>): mapping values are not allowed in this context at line 2 column 14 (Psych::SyntaxError)
      /home/jenkins/workspace/Runner-v3/lib/cli_executor.rb:68:in `get_version_for'
      /home/jenkins/workspace/Runner-v3/lib/cli_executor.rb:261:in `version'
      /home/jenkins/workspace/Runner-v3/lib/cli_executor.rb:251:in `executor'
      /home/jenkins/workspace/Runner-v3/lib/cli_executor.rb:293:in `user_opts'
      /home/jenkins/workspace/Runner-v3/lib/cli_executor.rb:303:in `exec'
      /home/jenkins/workspace/Runner-v3/lib/api_accessor.rb:71:in `cli_exec'
      /home/jenkins/workspace/Runner-v3/lib/api_accessor_owner.rb:28:in `public_send'
      /home/jenkins/workspace/Runner-v3/lib/api_accessor_owner.rb:28:in `block (2 levels) in <module:APIAccessorOwner>'
      /home/jenkins/workspace/Runner-v3/lib/openshift/cluster_resource.rb:162:in `get_matching'
      /home/jenkins/workspace/Runner-v3/lib/openshift/user.rb:143:in `projects'
      /home/jenkins/workspace/Runner-v3/lib/openshift/user.rb:113:in `block in clean_projects'
      /home/jenkins/workspace/Runner-v3/lib/base_helper.rb:153:in `wait_for'
[...]

PR #703, opened 2020-02-13
[Blocker] Switch to --kubeconfig only in 4.4 as some cmd in 4.3 does not support it.

liangxia:
For 4.1/4.2/4.3, `oc config --kubeconfig` does not generate kubeconfig , then automation will fail with error,
error: stat /home/jenkins/workspace/Runner-v3/workdir/ocp4_admin.kubeconfig: no such file or directory

For 4.4, when --config is used, there is a waring message which will cause an issue describe in https://github.com/openshift/verification-tests/pull/694

So we copy 4.1 to 4.4, and switch to --kubeconfig in 4.4

### DIFF
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -6,7 +6,7 @@
   :api_version: --api-version=<value>
   :ca: --certificate-authority=<value>
   :cluster: --cluster=<value>
-  :config: --kubeconfig=<value>
+  :config: --config=<value>
   :context: --context=<value>
   :h: -h
   :help: --help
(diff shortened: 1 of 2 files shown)


xiaocwan:
Yes, I agree to split rules for different oc versions.

liangxia:
/hold
Choose one of #702 and #703

jhou1:
Thank you @liangxia .

I'm not a fan of #702, reverting it back means we still have to work around 4.4. I prefer this approach.

lib/rules/cli/4.1.yaml
```diff
@@ -1110,7 +1110,7 @@
   :cmd: oc adm diagnostics
   :options:
     :cluster-context: --cluster-context=<value>
-    :config: --kubeconfig=<value>
+    :config: --config=<value>
```
akostadinov: this can be removed given it is already in global options

akostadinov:
btw I wonder if we shall make newer versions override things from older versions instead of full copy of all rules each. But for the time being we do this

/lgtm

lib/rules/cli/4.1.yaml
liangxia: @akostadinov Tracked in https://github.com/openshift/verification-tests/pull/708

PR #705, opened 2020-02-13
update dnsmasq config file

zhaozhanqi:
we often met error : 192.168.1.1 is already used .
try to debug, found sometimes the dnsmasq.conf will be append two times.
 So update '>>' to ">'. and file the dnsmasq config with 
https://github.com/openshift-qe/v3-testfiles/pull/1234

@anuragthehatter @huiran0826 @weliang1 @rbbratta

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -805,12 +805,13 @@ Given /^a DHCP service is configured for interface "([^"]*)" on "([^"]*)" node w
   host = node.host
   dhcp_status_timeout = 30
   #Following will take dnsmasq backup and append curl contents to the dnsmasq config after
-  @result = host.exec_admin("cp /etc/dnsmasq.conf /etc/dnsmasq.conf.bak;curl https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/dnsmasq_for_testbridge.conf | sed s/testbr1/#{br_inf}/g | sed s/88.8.8.100,88.8.8.110,24h/#{add_lease}/g >> /etc/dnsmasq.conf;systemctl restart dnsmasq --now")
+  @result = host.exec_admin("cp /etc/dnsmasq.conf /etc/dnsmasq.conf.bak;curl https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/dnsmasq_for_testbridge.conf | sed s/testbr1/#{br_inf}/g | sed s/88.8.8.100,88.8.8.110,24h/#{add_lease}/g > /etc/dnsmasq.conf;systemctl restart dnsmasq --now")
   raise "Failed to configure dnsmasq service" unless @result[:success]
   wait_for(dhcp_status_timeout) {
     if host.exec_admin("systemctl status dnsmasq")[:response].include? "running"
       logger.info("dnsmasq service is running fine")
     else
+      host.exec_admin("cp /etc/dnsmasq.conf.bak /etc/dnsmasq.conf && systemctl restart dnsmasq --now")
       raise "Failed to start dnsmasq service. Check you cluster health manually"
     end
   }


rbbratta:
ah this probably fixes OCP-24465 as well and shows that #695 is not the fix.

anuragthehatter:
@zhaozhanqi I am skeptical about this change as > will overwrite the dnsmasq file while >> would just append so we won't be interfering the existing config of dnsmasq. However this still needs to be evaluated whether its okay to overwrite whole dnsmasq file for test duration to make sure its not impacting other cluster requirements.

anuragthehatter:
Actually, by looking at the logs carefully here http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/02/03/23:49:51/Multus_CNI_type_bridge_with_DHCP/console.html

1) It seems like cleanup step of "a DHCP service is deconfigured on the "<%= cb.nodes[0].name %>" node" won't execute if corresponding step is failed which is happening above in logs. So the cluster will be left with provided IP configured under dnsmasq. And henceteh re-run of test will complain IP exists.

2) We can do one thing - irrespective of whether cleanup happens or not, we can copy the original dnsmasq file in this step https://github.com/anuragthehatter/verification-tests/blob/master/features/step_definitions/networking.rb#L814

So that even cleanup doesn't occur, we will have original dnsmasq at place

What say?

zhaozhanqi:
> Actually, by looking at the logs carefully here http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/02/03/23:49:51/Multus_CNI_type_bridge_with_DHCP/console.html
> 
> 1. It seems like cleanup step of "a DHCP service is deconfigured on the "<%= cb.nodes[0].name %>" node" won't execute if corresponding step is failed which is happening above in logs. So the cluster will be left with provided IP configured under dnsmasq. And henceteh re-run of test will complain IP exists.
> 2. We can do one thing - irrespective of whether cleanup happens or not, we can copy the original dnsmasq file in this step https://github.com/anuragthehatter/verification-tests/blob/master/features/step_definitions/networking.rb#L814
> 
> So that even cleanup doesn't occur, we will have original dnsmasq at place
> 
> What say?

yes, it's also a good idea, I added line
`      host.exec_admin("cp /etc/dnsmasq.conf.bak /etc/dnsmasq.conf && systemctl restart dnsmasq --now")`

features/step_definitions/networking.rb
```diff
@@ -805,12 +805,13 @@
...
   dhcp_status_timeout = 30
   #Following will take dnsmasq backup and append curl contents to the dnsmasq config after
-  @result = host.exec_admin("cp /etc/dnsmasq.conf /etc/dnsmasq.conf.bak;curl https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/dnsmasq_for_testbridge.conf | sed s/testbr1/#{br_inf}/g | sed s/88.8.8.100,88.8.8.110,24h/#{add_lease}/g >> /etc/dnsmasq.conf;systemctl restart dnsmasq --now")
+  @result = host.exec_admin("cp /etc/dnsmasq.conf /etc/dnsmasq.conf.bak;curl https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/dnsmasq_for_testbridge.conf | sed s/testbr1/#{br_inf}/g | sed s/88.8.8.100,88.8.8.110,24h/#{add_lease}/g > /etc/dnsmasq.conf;systemctl restart dnsmasq --now")
   raise "Failed to configure dnsmasq service" unless @result[:success]
   wait_for(dhcp_status_timeout) {
     if host.exec_admin("systemctl status dnsmasq")[:response].include? "running"
       logger.info("dnsmasq service is running fine")
     else
+      host.exec_admin("cp /etc/dnsmasq.conf.bak /etc/dnsmasq.conf && systemctl restart dnsmasq --now")
```
liangxia: We can use `mv` instead of `cp` here. Faster, and we no longer left a unused .bak file on the system.

anuragthehatter: @liangxia Great observation. mv could have make more sense but such tests are complicated and having a bak file on cluster is safe during this step. This might be mostly required in debugging in case of something exits pre-maturely/crash etc hence needed to be looked into cluster manually in such cases. Having bak file in had would be safe. Anyways such steps will always be followed by ```DHCP service is deconfigured on the "([^"]*) node``` which will eventually remove the bak file so we are good there. We are seeing hiccups in such tests from time to time and such bak file would def be useful sometime.

liangxia: ack

PR #708, opened 2020-02-14
Remove duplicate global options from specific command

liangxia:
From https://github.com/openshift/verification-tests/pull/703/files#r379050913


liangxia:
Fixed by https://github.com/openshift/verification-tests/pull/777

PR #716, opened 2020-02-18
Fix OCP-10944

wewang58:
@openshift/devexp-qe  please review it, thanks

### DIFF
--- a/features/cli/build.feature
+++ b/features/cli/build.feature
@@ -644,26 +644,26 @@ Feature: build 'apps' with CLI
   # @case_id OCP-10944
   Scenario: Simple error message return when no value followed with oc build-logs
     Given I have a project
-    When I run the :build_logs client command with:
-      | build_name | |
+    When I run the :logs client command with:
+      | resource_name | |
     Then the step should fail
-    And the output should contain:
-      | resource name may not be empty |
+    And the output should match:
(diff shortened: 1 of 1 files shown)


features/cli/build.feature
```diff
@@ -644,26 +644,23 @@ Feature: build 'apps' with CLI
...
   Scenario: Simple error message return when no value followed with oc build-logs
     Given I have a project
-    When I run the :build_logs client command with:
-      | build_name | |
+    When I run the :logs client command with:
+      | resource_name | |
     Then the step should fail
-    And the output should contain:
-      | resource name may not be empty |
+    And the output should match "resource name may not be empty||must provide one or more resources by argument or filename"
```
chengzhang1016: In fact, I do not suggest to use strict matching, we need frequent code change to against developer's any little change.

wewang58: When exeute comand build-logs, it will prompt "resource name may not be empty", now using logs command, just display "must provide one or more resources by argument or filename", so delete "resource name may not be empty", please check it again, thanks

wewang58:
@chengzhang1016  Could you help to review my script again, already updated it

chengzhang1016:
LGTM. Thanks update.

PR #720, opened 2020-02-19
update OCP-18482 and OCP-9650

lihongan:
OCP-18482: Since the max concurrent connections is related the number of router pods as well as the clould LB, we cannot tested it with `curl` anymore. The new solution is more straightforward and just check haproxy.conf in router pod. 

OCP-9650: remove the redundant matching string, if the output contains `Hello-OpenShift` then that means the HTTP status code should be `200`.

### DIFF
--- a/features/routing/rate-limit.feature
+++ b/features/routing/rate-limit.feature
@@ -2,13 +2,20 @@ Feature: Testing haproxy rate limit related features
 
   # @author hongli@redhat.com
   # @case_id OCP-18482
+  @admin
   Scenario Outline: limits backend pod max concurrent connections for unsecure, edge, reen route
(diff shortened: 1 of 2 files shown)


lihongan:
4.3 log:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/70925/consoleFull
4.4 log:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/70940/consoleFull

lihongan:
@zhaozhanqi PTAL, thanks

akostadinov:
auto lg, team should review

PR #725, opened 2020-02-20
helper method to get loadBalancer ingress

pruan-rht:


pruan-rht:
@akostadinov PTAL.  thx!

akostadinov:
I prefer to see example senario where things are used going forward. Not only some isolated changes.

/lgtm

PR #726, opened 2020-02-21
Ensure goto action complete loading current page

xiaocwan:
I've been wanting to suggest the change since 3.x. Since 4.x has some changes on console rendering, automation sometimes fail when page is directing and automation is doing some operation(like click) simultaneously.

The reason I update goto rules:
Test step itself should ensure the current action succeed, not the following step wait for the following checkpoint, it’s not clear for debug.
The wait_box_loaded (_css: .loading-box__loaded_) will be succeed even when the page has no item (but not a total empty crashed page), it makes sure the current page finished loading, that's the complete operation for goto action. IMO this can not rely on following steps, also the timeout for next step may not be enough.

I tested each page updated for the goto action.

So far I only update the goto_xxx_list page, will also update the goto_one_resource_xxx_page, this css selector does also work on this page. [Here comes an example, error should raise on goto_one_dc_page, not check_popover_info. Also the error page may raised by click action is VERY closed to goto action when page is still loading](http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/02/17/01:27:39/Check_popover_help_on_resource_detail_page/console.html) (refer to the [doc for pseudo-elements](https://www.w3.org/TR/CSS2/generate.html#before-after-content) )
@akostadinov @pruan-rht  Please tell me your thought and concern, thanks!

### DIFF
--- a/lib/rules/web/admin_console/4.4/goto.xyaml
+++ b/lib/rules/web/admin_console/4.4/goto.xyaml
@@ -150,8 +150,10 @@ goto_catalog_page:
   url: catalog/ns/<project_name>
 goto_create_app_from_imagestream_page:
   url: catalog/source-to-image?imagestream=<is_name>&imagestream-ns=openshift&preselected-ns=<project_name>
+  action: wait_form_loaded
 goto_create_app_from_template_page:
   url: catalog/instantiate-template?template==<template_name>&template-ns=openshift&preselected-ns=<project_name>
+  action: wait_form_loaded
 
 # monitoring
 goto_monitoring_metrics_page:
@@ -180,6 +182,7 @@ goto_one_namespace_page:
 # operator hub
 goto_operator_hub_page:
   url: operatorhub/ns/default
+  action: wait_box_loaded
 goto_installed_operators_page:
   url: k8s/ns/<project_name>/operators.coreos.com~v1alpha1~ClusterServiceVersion
 goto_operator_subscription_page:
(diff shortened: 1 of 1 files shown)


xiaocwan:
@pruan-rht @akostadinov @yapei 
Could anyone take a look at this? 
Thanks!

yapei:
@xiaocwan   Since `goto_xxx` is the smallest unit we can reuse for most actions, adding `wait_xxx_loaded` may reduce the flexibility. `wait_xxx_load` should be called on demand

xiaocwan:
`.loading-box__loaded` is basic check page loading success after goto. Only to check if page turns out to be empty, not to check anything actually. It's clear for debug, not to depend on following checkpoint fails. 

It's not the same thing as usual check rules.

liangxia:
There are conflicts.

akostadinov:
Nice idea. I am wondering if this can be used after each click too? Or if we can insert it automatically after each operation.

xiaocwan:
> Nice idea. I am wondering if this can be used after each click too? Or if we can insert it automatically after each operation.

Yes, suggest to add checkpoint after each Click action in smallest unit theoretically, such as the button is disappear or disabled. But sometimes Click will browse to another page even external page, so far I can't find a general unit check after operation automatically.

xiaocwan:
Conflict resolved. Thanks @liangxia

xiaocwan:
Please /hold .
Still discuss it with team members

liangxia:
@xiaocwan  Commands needs to be in one line by itself. Like this,
/hold

xiaocwan:
/hode cancel
Only keep `action: wait_box_loaded` for goto_projects_list_page and ran all related cases for regression test. There is no issue.
@yapei @liangxia  Please review.

akostadinov:
/hold cancel

there was a typo above :)

PR #727, opened 2020-02-21
add lib for scheduler cluster CR

zhouying7780:
@xingxingxia @kasturinarra  could you please help review ? thanks

### DIFF
--- a/features/step_definitions/scheduler.rb
+++ b/features/step_definitions/scheduler.rb
@@ -6,3 +6,23 @@ Given /^the #{QUOTED} scheduler priorityclasses is restored after scenario$/ do 
     raise "Cannot delete priorityclass: #{name}" unless @result[:success]
   }
 end
+
+Given /^the #{QUOTED} scheduler CR is restored after scenario$/ do |name|
+  ensure_admin_tagged
+  ensure_destructive_tagged
+  org_scheduler = {}
+  @result = admin.cli_exec(:get, resource: 'scheduler', resource_name: name, o: 'yaml')
+  if @result[:success]
+    org_scheduler['spec'] = @result[:parsed]['spec']
+    logger.info "scheduler restore tear_down registered:\n#{org_scheduler}"
+  else
+    raise "Could not get scheduler: #{name}"
+  end
+  patch_json = org_scheduler.to_json
+  _admin = admin
+  teardown_add {
+    opts = {resource: 'scheduler', resource_name: name, p: patch_json, type: 'merge' }
+    @result = _admin.cli_exec(:patch, **opts)
+    raise "Cannot restore scheduler: #{name}" unless @result[:success]
+  }
+end


zhouying7780:
@pruan-rht  Could you please help review ? thanks.

features/step_definitions/scheduler.rb
```diff
@@ -0,0 +1,19 @@
...
+  else
+    raise "Could not get scheduler: #{name}"
+  end
+  patch_json = org_scheduler.to_json
+  _admin = admin
+  teardown_add {
+    opts = {resource: 'scheduler', resource_name: name, p: patch_json, type: 'merge' }
+    @result = _admin.cli_exec(:patch, **opts)
+    raise "Cannot restore scheduler: #{name}" unless @result[:success]
+    sleep 60
```
pruan-rht: @zhouying7780, can you please elaborate the reason for the 60 seconds hard-coded sleep?

xingxingxia: Agree, for any magic number, always encouraged to comment the reason

features/step_definitions/scheduler.rb
```diff
@@ -0,0 +1,19 @@
...
+  }
+end
```
xingxingxia: Seems what this step does is common for many other CRs, could check if others defined. If not, seems better to make it common, not only to scheduler CR

features/step_definitions/scheduler.rb
```diff
@@ -0,0 +1,19 @@
+Given /^the #{QUOTED} scheduler CRD is restored after scenario$/ do |name|
```
xingxingxia: CR looks more accurate :), CR is Resource, CRD is Definition :)

features/step_definitions/scheduler.rb
zhouying7780: @xingxingxia features/step_definitions/oauth.rb has defined the oauth CRD.

features/step_definitions/scheduler.rb
zhouying7780: @pruan-rht @xingxingxia  I just copy the method from features/step_definitions/oauth.rb, it         has defined the 60 seconds sleep.

features/step_definitions/scheduler.rb
zhouying7780: Have updated.

features/step_definitions/scheduler.rb
xingxingxia: Just a suggestion to avoid same code multiple times. Fine to keep PR as is.

xingxingxia:
lgtm with no more comment
@pruan-rht help checke / merge, thanks

pruan-rht:
@zhouying7780 please rebase.  thanks!

zhouying7780:
@pruan-rht  have rebased, please help review  and if no more need update please help merge,thanks.

zhouying7780:
@liangxia   Could you please help review and merge , thanks.

features/step_definitions/scheduler.rb
```diff
@@ -4,3 +4,23 @@
   @result = admin.cli_exec(:delete, object_type: 'priorityclasses', object_name_or_id: name)
   raise "priority classes were not deleted successfully!" unless @result[:success]
 end
+
+Given /^the #{QUOTED} scheduler CRD is restored after scenario$/ do |name|
+  ensure_admin_tagged
```
liangxia: Also need ensure_destructive_tagged as scheduler change can cause failures for other scenarios when running in parallel

liangxia:
There are several unresolved comments need action.

zhouying7780:
@liangxia   all comments have addressed ,  please help review again. thanks.

features/step_definitions/scheduler.rb
zhouying7780: deleted , and retest , no issue.

PR #731, opened 2020-02-24
 OCP27627

miyadav:
This is to implement the test case for checking - https://bugzilla.redhat.com/show_bug.cgi?id=1796599  
Successful job using the test : https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/72505/console

### DIFF
--- a/features/machine/machine.feature
+++ b/features/machine/machine.feature
@@ -108,3 +108,11 @@ Feature: Machine features testing
     Then the step should succeed
     And the output should match "Provider ID:\s+<%= cb.providerID %>"
 
+  # @author miyadav@redhat.com
+  # @case_id OCP-27627
+  @admin
+  Scenario: Verify all machine instance-state should be consistent with their providerStats.instanceState
+    Given I have an IPI deployment
+    And evaluation of `BushSlicer::Machine.list(user: admin, project: project('openshift-machine-api'))` is stored in the :machines clipboard
+    Then the expression should be true> cb.machines.select{|m|m.instance_state == m.annotation_instance_state}.count == cb.machines.count
+ 


pruan-rht:
@miyadav there are conflicts...please fix.

features/machine/machine.feature
```diff
@@ -108,3 +108,24 @@ Feature: Machine features testing
...
+      | o         | jsonpath='{range .items[*]}{.metadata.annotations}{.status.providerStatus.instanceState}{"/\\n"}{end}'|
+
+    And evaluation of `@result[:response].split(":")` is stored in the :console_output_array clipboard
+
+    #Evaluating Result for each line
+    When I repeat the following steps for each :console_output in cb.console_output_array:
+    """
+    And evaluation of `cb.console_output_array.pop.split("]")[0]` is stored in the :provider_status clipboard
+    And evaluation of `cb.console_output_array.pop.split(":").pop.split("]")[1].split("/")[0]` is stored in the :instance_state clipboard
+    Then the expression should be true> cb.provider_status == cb.instance_state
```
pruan-rht: @miyadav do you know if status.phase == status.providerStatus.instanceState  they look the same to me.  If that's the case we can use existing method and not have to do the string splitting.

pruan-rht: @miyadav please see #746 for explanation on how to use the new methods.  It should shorten it to only 2 lines of code.  Please let me know if it's unclear

miyadav: thanks @pruan-rht  , so as I understand  your suggestion is to replace line numbers 116 to 130 with " And evaluation of `BushSlicer::Machine.list(user: admin, project: project('openshift-machine-api'))` is stored in the :machines clipboard
   Then the expression should be true> cb.machines.select {|m| m.instance_state == m.annotation_instance_state }.count == cb.machines.count"  once helper PR is merged?

jhou1: I think so,  #746  gets merged, your steps can be shorten to the two lines.

miyadav: thanks @pruan-rht  , going over it will ask when have issues relating to them.

miyadav: Made changes  and tested working fine locally

PR #732, opened 2020-02-25
fix kebab list item issue

yapei:
During debugging for OCP-24232, I found that the action `click_one_operation_in_kebab|Cancel Build` always fail at clicking at `Cancel Build` button at the first time, it only succeeded at the 2nd time

```
    When I perform the :click_one_operation_in_kebab web action with:
      | resource_name  | ruby-sample-3 |
      | kebab_item     | Cancel Build  |
    Then the step should succeed
    When I perform the :confirm_cancel_action web action with:
      | cancel | true |
    Then the step should succeed
```
Then I found when we open kebab button, the list items in kebab are disabled at first then enabled if user has enough permission. So I added `not(contains(@class,'pf-m-disabled')` to make sure kebab list item is enabled then it can be clicked at anytime

```
//button[contains(@class,'pf-c-dropdown__menu-item') and not(contains(@class,'pf-m-disabled')) and text()='<kebab_item>']
```
Also only kebab list item has issues, Actions list items don't

### DIFF
--- a/lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
+++ b/lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
@@ -277,15 +277,19 @@ click_link_with_text:
     selector:
       xpath: //a[contains(.,'<text>') and contains(@href,'<link_url>')]
     op: click
+click_kebab_item:
+  element:
+    selector:
+      xpath: //button[contains(@class,'pf-c-dropdown__menu-item') and not(contains(@class,'pf-m-disabled')) and text()='<kebab_item>']
+    op: click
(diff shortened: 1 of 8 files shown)


yapei:
@yanpzhan  @SSshahan  @xiaocwan  Please help check these changes.

yanpzhan:
LGTM

xiaocwan:
If ` not(contains(@class,'pf-m-disabled'))` will be disappeared in the given time on console, it LGTM to replace the `wait for` step.

yapei:
@pruan-rht  Can you help merge if it looks good to you?

PR #735, opened 2020-02-25
update AB routing cases

lihongan:
The A/B routing test cases are not stable when just depending on `curl` and counting result. 
In the OCP 4.x test env, the number of router pod is not a fixed value and always more than one, and the config in each router is reloaded asynchronously, so the cases are random failed and need to debug manually. 
In order to make them more stable I'd like to update the cases to check the config directly (weight 0-256). 

logs:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/72779/consoleFull

### DIFF
--- a/features/routing/abrouting.feature
+++ b/features/routing/abrouting.feature
@@ -2,15 +2,25 @@ Feature: Testing abrouting
 
   # @author yadu@redhat.com
   # @case_id OCP-12076
+  @admin
   Scenario: Set backends weight for unsecure route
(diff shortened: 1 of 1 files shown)


lihongan:
@zhaozhanqi @quarterpin PTAL, thanks.

lihongan:
@akostadinov @pruan-rht Ready for review. Thanks.

PR #739, opened 2020-02-27
add rules for OCP-26734

yapei:
@xiaocwan  @SSshahan  Can you help review?


lib/rules/web/admin_console/4.4/user_management.xyaml
```diff
@@ -33,3 +33,26 @@ click_stop_impersonation:
   element:
     selector: *stop_impersonation
     op: click
+check_data_in_users_table:
```
yapei: Move rules in `users.xyaml` here so we can reduce one file

lib/rules/web/admin_console/4.4/goto.xyaml
```diff
@@ -222,4 +222,6 @@ goto_all_machines_page:
 goto_all_machine_sets_page: 
   url: k8s/all-namespaces/machine.openshift.io~v1beta1~MachineSet
 
-
+# global config
+goto_alertmanagerconfig_page:
+  url: monitoring/alertmanagerconfig
```
xiaocwan: Highly suggest  to make sure each `goto` works by the rule itself. Needs a `action wait_box_loaded` here. Sometimes it needs a long time out to loading page and next action may fail during loading. 
Generally I suggest the logical order : `first do some operation -> then check` in one rule, and all same (`first do some operation -> then check`) in other rule. 
I mean each single rule should follow operate -> check order. So that `Then the step should succeed` make sense.
Refer to [this PR](https://github.com/openshift/verification-tests/pull/726) opened several days ago.
HDYT?

lib/rules/web/admin_console/4.4/user_management.xyaml
SSshahan: If change the rules location, seems we also need update them for 4.3 rule?

yapei: We can update for 4.3, also it seems harmless if don't change to 4.3, right?

yapei:
@pruan-rht  Can you help lgtm label if no objections?

PR #740, opened 2020-02-27
Auto OCP-20984

xiaocwan:
@yapei Please review, thanks!

### DIFF
--- a/lib/rules/web/admin_console/4.4/actions.xyaml
+++ b/lib/rules/web/admin_console/4.4/actions.xyaml
@@ -7,6 +7,12 @@ click_resource_action_button:
       xpath: //button[contains(.,'Actions')]
     op: click
     timeout: 30
+click_resource_action_button_on_sidebar:
+  elements:
+  - selector:
+      xpath: //div[contains(@class, 'overview__sidebar-pane')]//button[contains(.,'Actions')]
+    op: click
+    timeout: 30
 choose_action_item_from_list:
   elements:
   - selector:
--- a/lib/rules/web/admin_console/4.4/overview.xyaml
+++ b/lib/rules/web/admin_console/4.4/overview.xyaml
@@ -6,12 +6,12 @@ wait_overview_items_loaded:
 check_list_heading_shown:
   element:
     selector:
-      xpath: //button[contains(@class,'project-overview__item-heading--name') and text()='<heading>' and not(contains(@class,'pf-m-disabled'))]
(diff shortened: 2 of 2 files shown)


lib/rules/web/admin_console/4.4/actions.xyaml
```diff
@@ -7,6 +7,12 @@ click_resource_action_button:
       xpath: //button[contains(.,'Actions')]
     op: click
     timeout: 30
+click_resource_action_button_on_sidebar:
```
yapei: Any differences with `click_resource_action_button` ?

lib/rules/web/admin_console/4.4/overview.xyaml
```diff
@@ -6,11 +6,12 @@ wait_overview_items_loaded:
 check_list_heading_shown:
   element:
     selector:
-      xpath: //button[contains(@class,'project-overview__item-heading--name') and text()='<heading>']
+      xpath: //button[contains(@class,'project-overview__item-heading--name') and text()='<heading>' and not(contains(@class,'pf-m-disabled'))]
```
yapei: Why adding `pf-m-disabled` ?

lib/rules/web/admin_console/4.4/overview.xyaml
```diff
@@ -6,11 +6,12 @@ wait_overview_items_loaded:
...
 check_list_item:
   elements: 
   - selector: &overview_item
       xpath: //button[contains(@class,'project-overview__item-heading--name') and text()='<resource_name>']/ancestor::div[contains(@class,'project-overview__item--<resource_kind>')]
+    timeout: 90
```
yapei: Do we need so long time here?

lib/rules/web/admin_console/4.4/actions.xyaml
yapei: Ok, I see the differences

lib/rules/web/admin_console/4.4/overview.xyaml
xiaocwan: Just for incase the button is disabled for tmp, will remove if it's not needed.

lib/rules/web/admin_console/4.4/overview.xyaml
xiaocwan: Yes, it's loading for a long time on my local side. Parallel running jobs may be also slow.

xiaocwan:
Updated

PR #741, opened 2020-02-27
Automated OCP-27654

SSshahan:
@yapei  please help to review. Thx!

### DIFF
--- a/lib/rules/web/admin_console/4.4/notification.xyaml
+++ b/lib/rules/web/admin_console/4.4/notification.xyaml
@@ -1,0 +1,52 @@
+click_notification_drawer:
+  elements:
+  - selector:
+      xpath: //button[contains(@aria-label,'Notification Drawer')]
+    op: click
+    timeout: 30
+open_notification_drawer:
+  action: click_notification_drawer
+  element:
+    selector: &header_title
+      xpath: //h1[contains(.,'Notifications')]
+close_notification_drawer:
+  action: click_notification_drawer
+  element:
+    selector: *header_title
+    missing: true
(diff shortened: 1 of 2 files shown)


lib/rules/web/admin_console/4.4/notification.xyaml
```diff
@@ -0,0 +1,52 @@
+click_notification_drawer:
```
yapei: This can be used with `click_masthead_button`

yapei: I would recommend create a new file `masthead.xyaml` with all masthead related rules into it

lib/rules/web/admin_console/4.4/notification.xyaml
```diff
@@ -0,0 +1,52 @@
...
+  elements:
+  - selector:
+      xpath: //button[contains(@aria-label,'Notification Drawer')]
+    op: click
+    timeout: 30
+open_notification_drawer:
```
yapei: Normally we'd better not mix click & check operation in one action. Let me try to explain
Given these scenarios in feature 
```
    When I run the :open_notification_drawer web action
    Then the step should succeed
    When I run the :close_notification_drawer web action
    Then the step should succeed
```
At the first glance, we just notice that we're opening & close notification drawer, didn't know we also checked that `Notification` headings appear & disappear

yapei: This time we can just leave them as they are

SSshahan:
@pruan-rht please help to merge, thanks

PR #742, opened 2020-02-27
Automate clusterautoscaler test case

jhou1:
Automated clusterautoscaler test case, @sunzhaohua2 @miyadav please review.

Tested locally.
```
      [09:39:09] INFO> === End After Scenario: Cluster should automatically scale up and scale down ===

1 scenario (1 passed)
19 steps (19 passed)
15m47.689s
```

### DIFF
--- a/features/machine/autoscaler.feature
+++ b/features/machine/autoscaler.feature
@@ -1,0 +1,54 @@
+Feature: Cluster Autoscaler Tests
+
+  # @author jhou@redhat.com
+  # @case_id OCP-28108
+  @admin
+  @destructive
+  Scenario: Cluster should automatically scale up and scale down with clusterautoscaler deployed
+    Given I have an IPI deployment
+    And I switch to cluster admin pseudo user
+
+    Given I store the number of machines in the :num_to_restore clipboard
+    And admin ensures node number is restored to "<%= cb.num_to_restore %>" after scenario
+
+    Given I clone a machineset named "machineset-clone"
+    And admin ensures "machineset-clone" machineset is deleted after scenario
+
+    # Create clusterautoscaler
+    Given I use the "openshift-machine-api" project
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/cloud/cluster-autoscaler.yml |
+    Then the step should succeed
+    And admin ensures "default" clusterautoscaler is deleted after scenario
+    And 1 pods become ready with labels:
+      | cluster-autoscaler=default,k8s-app=cluster-autoscaler |
+
+    # Create machineautoscaler
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/cloud/machine-autoscaler.yml" replacing paths:
+      | ["metadata"]["name"]               | maotest                 |
+      | ["spec"]["minReplicas"]            | 1                       |
+      | ["spec"]["maxReplicas"]            | 3                       |
+      | ["spec"]["scaleTargetRef"]["name"] | <%= machine_set.name %> |
+    Then the step should succeed
+    And admin ensures "maotest" machineautoscaler is deleted after scenario
+
+    # Create workload
+    When I run the :create admin command with:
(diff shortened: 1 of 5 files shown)


jhou1:

features/machine/autoscaler.feature
```diff
@@ -0,0 +1,67 @@
...
+    And admin ensures "machineset-clone" machineset is deleted after scenario
+
+    # Create clusterautoscaler
+    Given I use the "openshift-machine-api" project
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/cloud/cluster-autoscaler.yml |
+    Then the step should succeed
+    And 1 pods become ready with labels:
+      | cluster-autoscaler=default,k8s-app=cluster-autoscaler |
+    And I register clean-up steps:
```
liangxia: Perhaps register the clean-up earlier. E.g, if the above pod failed to become ready, then the clean-up is not registered successfully.

lib/openshift/machine_set.rb
```diff
@@ -18,7 +18,7 @@ def available_replicas(user: nil, cached: true, quiet: false)
     def ready?(user: nil, cached: false, quiet: false)
       result = {}
       result[:success] = (desired_replicas(user: user, cached: cached, quiet: quiet) ==
-        available_replicas(user: user, cached: cached, quiet: quiet))
+        available_replicas(user: user, cached: true, quiet: quiet))
```
liangxia: Update in step implemention ? I though it should be on step definition, that's line 18

jhou1: This is intended. Both methods call `raw_resource` method, the second call could use the cached resource to save time. This is @akostadinov's suggestion when I first implemented it.

features/machine/autoscaler.feature
jhou1: Thanks, addressed that.

lib/openshift/machine_set.rb
liangxia: Ack

features/machine/autoscaler.feature
```diff
@@ -0,0 +1,67 @@
...
+
+    # Create clusterautoscaler
+    Given I use the "openshift-machine-api" project
+    When I run the :create admin command with:
+      | f | https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/cloud/cluster-autoscaler.yml |
+    Then the step should succeed
+    And I register clean-up steps:
+    """
+    When I run the :delete admin command with:
+      | object_type       | clusterautoscaler |
```
pruan-rht: @jhou1 please use the step `I ensure...` instead.

features/machine/autoscaler.feature
```diff
@@ -0,0 +1,67 @@
...
+    # Create machineautoscaler
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/cloud/machine-autoscaler.yml" replacing paths:
+      | ["metadata"]["name"]               | maotest                 |
+      | ["spec"]["minReplicas"]            | 1                       |
+      | ["spec"]["maxReplicas"]            | 3                       |
+      | ["spec"]["scaleTargetRef"]["name"] | <%= machine_set.name %> |
+    Then the step should succeed
+    And I register clean-up steps:
+    """
+    When I run the :delete admin command with:
```
pruan-rht: @jhou1 please use the step I ensure... instead.

features/machine/autoscaler.feature
jhou1: Oh, I can remember why I didn't use `I ensure ...` in the first place. the `clusterautoscaler` and `machineautoscaler` are custom resources, we don't have a library defined around it. I'll try implement that.

pruan-rht: @jhou1 [machine_autoscaler.rb](https://github.com/openshift/verification-tests/blob/master/lib/openshift/machine_autoscaler.rb) is there.  You can use that as a template for `clusterautoscaler`

jhou1: It's updated now.

PR #746, opened 2020-02-28
helper for pr731

pruan-rht:
@miyadav once this PR is merged, then you can reduce your scenario to these two lines
```
   And evaluation of `BushSlicer::Machine.list(user: admin, project: project('openshift-machine-api'))` is stored in the :machines clipboard
   Then the expression should be true> cb.machines.select {|m| m.instance_state == m.annotation_instance_state }.count == cb.machines.count
```

### DIFF
--- a/lib/openshift/machine.rb
+++ b/lib/openshift/machine.rb
@@ -26,6 +26,16 @@ module BushSlicer
         dig('status','phase')
     end
 
+    def instance_state(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).
+          dig('status', 'providerStatus', 'instanceState')
+    end
+
+    def annotation_instance_state(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).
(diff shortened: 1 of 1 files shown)


pruan-rht:
helper for PR #731

jhou1:
Looks good! Thanks!

pruan-rht:
@akostadinov PTAL...should be a simple one

lib/openshift/machine.rb
```diff
@@ -26,6 +26,16 @@ def phase(user: nil, cached: true, quiet: false)
...
 
+    def instance_state(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).
+          dig('status', 'providerStatus', 'instanceState')
+    end
+
+    def annotation_instance_state(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).
+          dig('metadata', 'annotations', 'machine.openshift.io/instance-state')
+    end
```
akostadinov: @pruan-rht , we already have a method toread annotations in `resource.rb`. Could you remove this redundant method?

PR #753, opened 2020-03-03
Add upgrade case

wewang58:


features/upgrade/build/build-upgrade.feature
```diff
@@ -0,0 +1,34 @@
...
+      | deploymentconfig=ruby-hello-world |
+
+  # @author wewang@redhat.com
+  # @case_id OCP-13025
+  @upgrade-check
+  @users=upuser1,upuser2
+  Scenario: Check docker and sti build works well after upgrade
+    Given I switch to the first user
+    When I use the "devexp-upgrade" project
+    And status becomes :running of 1 pods labeled:
```
xiuwang: It’s better to trigger a new build for two type builds

wzheng1: agree

wewang58: Updated, please check it again, thanks

features/upgrade/build/build-upgrade.feature
```diff
@@ -0,0 +1,41 @@
...
+      | project_name | build-upgrade |
+    Then the step should succeed
+    When I run the :new_app client command with:
+      | app_repo | openshift/ruby~https://github.com/openshift/ruby-ex |
+    Then the step should succeed
+    When I run the :new_app client command with:
+      | app_repo | openshift/ruby~https://github.com/openshift/ruby-hello-world |
+      | strategy | docker                                                       |
+    Then the step should succeed
+    Given I use the "build-upgrade" project
```
liangxia: Looks like we do not need this step.

wewang58: It will report "no projects in cache" when there's no the step

liangxia: Ack

PR #757, opened 2020-03-04
OCP-26089: add step to find OVN Raft leader pod

rbbratta:
Run `ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound`
on any ovnkube-master node and parse the output to find the leader pod


anuragthehatter:
Thanks Ross. 
/LGTM

anuragthehatter:
OVN 2.12 is also backported to 4.3 so /var/run/ovn/ovnnb_db.ctl path will be present on 4.3 and 4.4..Nice

PR #758, opened 2020-03-04
Enhance machine healthcheck test for 4.3

jhou1:
In 4.3, to make a node NotReady, kill the `hyperkube` process; in >= 4.4, kill the `kubelet` process. Add a condition so that the case runs well for 4.3.

At the same time, increase the timeout because on Azure and OSP, the instance creation/deletion is slower. 

@sunzhaohua2 @miyadav @liangxia PTAL


jhou1:
This is a simple one and it's verified by zhsun and me. There is no risk merging this myself.

PR #774, opened 2020-03-09
auto OCP-23745

sunzhaohua2:
Automated autoscaler test case OCP-23745.
```
[02:57:36] INFO> === End After Scenario: Delete machineautoscaler related machineset doesn't exist ===
1 scenario (1 passed)
10 steps (10 passed)
0m27.352s
```

@jhou1 @miyadav please help to review, thanks.

### DIFF
--- a/features/machine/autoscaler.feature
+++ b/features/machine/autoscaler.feature
@@ -176,3 +176,23 @@ Feature: Cluster Autoscaler Tests
     Then the step should succeed
     And the output should match "Annotations:\s+<none>"
 
+  # @author zhsun@redhat.com
+  # @case_id OCP-23745
+  @admin
+  @destructive
+  Scenario: Machineautoscaler can be deleted when its referenced machineset does not exist
+    Given I have an IPI deployment
+    And I switch to cluster admin pseudo user
+
+    Given I store the number of machines in the :num_to_restore clipboard
+    And admin ensures node number is restored to "<%= cb.num_to_restore %>" after scenario
+
+    Given I use the "openshift-machine-api" project
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/cloud/machine-autoscaler.yml" replacing paths:
+      | ["metadata"]["name"]               | maotest |
+      | ["spec"]["minReplicas"]            | 1       |
+      | ["spec"]["maxReplicas"]            | 3       |
+      | ["spec"]["scaleTargetRef"]["name"] | invalid |
+    Then the step should succeed
+    And admin ensures "maotest" machineautoscaler is deleted
+


features/machine/autoscaler.feature
```diff
@@ -179,3 +179,129 @@ Feature: Cluster Autoscaler Tests
     Then the step should succeed
     And the output should match "Annotations:\s+<none>"
 
+# @author zhsun@redhat.com
```
jhou1: Not aligned

features/machine/autoscaler.feature
```diff
@@ -179,3 +179,129 @@ Feature: Cluster Autoscaler Tests
...
+  # @case_id OCP-23745
+  @admin
+  @destructive
+  Scenario: Delete machineautoscaler related machineset doesn't exist
```
jhou1: Prefer a more understandable scenario description like **Machineautoscaler can be deleted when its referenced machineset does not exist**

features/machine/autoscaler.feature
```diff
@@ -179,3 +179,129 @@ Feature: Cluster Autoscaler Tests
...
+    the expression should be true> machine_set.desired_replicas(cached: false) == 3
+    """
+    Then the machineset should have expected number of running machines
+
+    Given I use the "machineset-clone1" machineset
+    And I wait up to 60 seconds for the steps to pass:
+    """
+    the expression should be true> machine_set.desired_replicas(cached: false) == 3
+    """
+    Then the machineset should have expected number of running machines
```
jhou1: Maybe save these lines simply by using  `the expression should be true> machine_set("machineset-clone1").desired_replicas(cached: false) == 3`

features/machine/autoscaler.feature
```diff
@@ -179,3 +179,129 @@ Feature: Cluster Autoscaler Tests
...
+    And the output should contain:
+      | Splitting scale-up between 2 similar node groups |
+    """
+
+    Given I use the "machineset-clone" machineset
+    And I wait up to 120 seconds for the steps to pass:
+    """
+    the expression should be true> machine_set.desired_replicas(cached: false) == 3
+    """
+    Then the machineset should have expected number of running machines
```
jhou1: Maybe save these lines simply by using `the expression should be true> machine_set("machineset-clone").desired_replicas(cached: false) == 3`

features/machine/autoscaler.feature
```diff
@@ -179,3 +179,129 @@ Feature: Cluster Autoscaler Tests
...
+    Then the step should succeed
+    And the output should contain:
+      | Splitting scale-up between 2 similar node groups |
+    """
+
+    Given I use the "machineset-clone" machineset
+    And I wait up to 120 seconds for the steps to pass:
+    """
+    the expression should be true> machine_set.desired_replicas(cached: false) == 3
+    """
```
jhou1: Maybe save these lines simply by using `the expression should be true> machine_set("machineset-clone").desired_replicas(cached: false) == 3`

features/machine/autoscaler.feature
```diff
@@ -179,3 +179,129 @@ Feature: Cluster Autoscaler Tests
...
+    Then the machineset should have expected number of running machines
+
+    Given I use the "machineset-clone1" machineset
+    And I wait up to 60 seconds for the steps to pass:
+    """
+    the expression should be true> machine_set.desired_replicas(cached: false) == 3
+    """
```
jhou1: Maybe save these lines simply by using `the expression should be true> machine_set("machineset-clone1").desired_replicas(cached: false) == 3`

features/machine/autoscaler.feature
```diff
@@ -179,3 +179,129 @@ Feature: Cluster Autoscaler Tests
...
+
+    # Delete workload
+    Given admin ensures "workload" job is deleted from the "openshift-machine-api" project
+ 
+    # Check cluster auto scales down
+    Given I use the "machineset-clone" machineset
+    And I wait up to 180 seconds for the steps to pass:
+    """
+    the expression should be true> machine_set.desired_replicas(cached: false) == 1
+    """
```
jhou1: Same with previous comment

features/machine/autoscaler.feature
```diff
@@ -179,3 +179,129 @@ Feature: Cluster Autoscaler Tests
...
+    Then the machineset should have expected number of running machines
+
+    Given I use the "machineset-clone1" machineset
+    And I wait up to 180 seconds for the steps to pass:
+    """
+    the expression should be true> machine_set.desired_replicas(cached: false) == 1
+    """
```
jhou1: Same with previous comment

jhou1:
@sunzhaohua2 Thanks, the test covers the autoscaler well. A few comments added. 
The first test look good to me.
The second test is more appropriate to put in cucushift repo as tier 2, can you please do that?

Thanks

sunzhaohua2:
@jhou1 Thanks, have updated. OCP-20108 is here: https://github.com/openshift/cucushift/pull/7441

features/machine/autoscaler.feature
```diff
@@ -175,3 +175,27 @@ Feature: Cluster Autoscaler Tests
...
+    And admin ensures node number is restored to "<%= cb.num_to_restore %>" after scenario
+
+    Given I use the "openshift-machine-api" project
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/cloud/machine-autoscaler.yml" replacing paths:
+      | ["metadata"]["name"]               | maotest |
+      | ["spec"]["minReplicas"]            | 1       |
+      | ["spec"]["maxReplicas"]            | 3       |
+      | ["spec"]["scaleTargetRef"]["name"] | invalid |
+    Then the step should succeed
+    And admin ensures "maotest" machineautoscaler is deleted after scenario
```
jhou1: I think we can remove this line. Because the falling step also deletes the MAO and verifies the deletion.

jhou1: Other than that looks all good!

features/machine/autoscaler.feature
```diff
@@ -175,3 +175,27 @@ Feature: Cluster Autoscaler Tests
...
+
+    Given I use the "openshift-machine-api" project
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/cloud/machine-autoscaler.yml" replacing paths:
+      | ["metadata"]["name"]               | maotest |
+      | ["spec"]["minReplicas"]            | 1       |
+      | ["spec"]["maxReplicas"]            | 3       |
+      | ["spec"]["scaleTargetRef"]["name"] | invalid |
+    Then the step should succeed
+
+    When I run the :delete admin command with:
```
pruan-rht: @sunzhaohua2 I think you can use `admin ensure...` step instead.

sunzhaohua2: @pruan-rht Thanks, I have updated.

jhou1:
Please squash the commits @sunzhaohua2

features/machine/autoscaler.feature
jhou1: This is the actual step being tested, so I'd prefert it be deleted in this scenario(not after). Remove the `after scenario` looks better.

sunzhaohua2: So if it is testing delete we use delete command, if it is clean up environment we use `after scenario`.

jhou1: @pruan-rht Can you please advise? We want to test the deletion is successful here. Should we be using `I run the delete ..` or `admin ensures ... `?

jhou1: @liangxia Can you please advise?

liangxia: Generally ,we do prepare/configuration in Before Scenario. And clean-up in After Scenario.
`ensure xxx is deleted` actually do a `oc delete` to trigger the deletion, then do `oc get` to verify the deletion is done.
Adding `after scenario` do the same operation, the difference is that it happens in the clean-up, while without it the operations happens during the scenario execution.

For this special case, the deletion happens just after the creation, and the deletion happens to be the last step of the scenario. Both way are OK for me.

liangxia:
By the way, the PR summary reference two test cases, but the PR contains only one.

features/machine/autoscaler.feature
jhou1: Thanks, then it seems removing the 'after scenario' would do

PR #775, opened 2020-03-09
add step to deploy syslog server

QiaolingTang:
Adding steps to deploy rsyslog server.
Currently, we didn't figure out how to deploy rsyslog server as secure, so only insecure rsyslog server will be deployed.  I will add steps for the secure rsyslog server later.


QiaolingTang:
@anpingli @pruan-rht could you please take a look?

PR #777, opened 2020-03-09
Remove adm_diagnostics from 4.x as it is an 3.x command

liangxia:
Hawkular is replaced by cluster monitoring.
`oc adm diagnostics` is replaced by operator-based diagnostics.


akostadinov:
/lgtm

Thanks!

PR #781, opened 2020-03-10
rules added for OCP-27901

yapei:


lib/rules/web/admin_console/4.1/overview.xyaml
```diff
@@ -62,11 +62,11 @@ click_groupby_dropdown_button:
   - selector:
       xpath: //span[contains(.,'Group by')]/ancestor::button[contains(@class,'btn-dropdown')]
     op: click
-check_groupby_dropdown_menu_item:
```
yapei: rule can be reused, so changed to a common name

lib/rules/web/admin_console/4.1/overview.xyaml
```diff
@@ -62,11 +62,11 @@ click_groupby_dropdown_button:
...
+check_dropdown_menu_item:
   element:
     selector: &group_by_dropdown_item
       xpath: //li/a[text()='<dropdown_menu_item>']
-click_groupby_dropdown_menu_item:
```
yapei: same as above, changed the name since it can be reused in different places

PR #790, opened 2020-03-11
helper step to close the current browser

pruan-rht:
@yapei, @xueli181114, PTAL if it addresses your need.  Thx!


xueli181114:
LGTM

PR #792, opened 2020-03-12
Add clusterrolebinding command option

pdhamdhe:
@xiaojiey Could you please help to review ? Thanks!

### DIFF
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -166,6 +166,7 @@
   :options:
     :clusterrole: --clusterrole=<value>
     :group: --group=<value>
+    :serviceaccount: --serviceaccount=<value>
 :create_configmap:
   :cmd: oc create configmap
   :options:
--- a/lib/rules/cli/4.4.yaml
+++ b/lib/rules/cli/4.4.yaml
@@ -166,6 +166,7 @@
   :options:
     :clusterrole: --clusterrole=<value>
     :group: --group=<value>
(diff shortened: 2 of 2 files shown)


lib/rules/cli/4.1.yaml
```diff
@@ -166,6 +166,7 @@
   :options:
     :clusterrole: --clusterrole=<value>
     :group: --group=<value>
+    :serviceaccount: --serviceaccount=<value>
```
liangxia: We also need to add it to 4.4.yaml
```
$ ls -l lib/rules/cli/4.*
-rw-rw-r--. 1 lxia lxia 37983 Mar 12 10:20 lib/rules/cli/4.1.yaml
lrwxrwxrwx. 1 lxia lxia     8 Dec  5 15:59 lib/rules/cli/4.2.yaml -> 4.1.yaml
lrwxrwxrwx. 1 lxia lxia     8 Dec  5 15:59 lib/rules/cli/4.3.yaml -> 4.2.yaml
-rw-rw-r--. 1 lxia lxia 37987 Mar 12 10:20 lib/rules/cli/4.4.yaml
lrwxrwxrwx. 1 lxia lxia     8 Mar 11 17:04 lib/rules/cli/4.5.yaml -> 4.4.yaml
```

pdhamdhe: @liangxia I have added the same option in 4.4.yaml, please check. Thanks!

PR #800, opened 2020-03-13
Bug fix for volume attached failed when volume and pod are in different zones

qinpingli:
For 4.x, AWS, GCE, and Azure will be deployed in multiple zones by default, so for cases OCP-9700 and OCP-9699 will hit the volume attach failed error, if the volume and pod are not in the same zone.

This pr bases on all the default sc's binding mode is `WaitForFirstConsumer` in OCP4.x, this will avoid the attached failed issue in a multiple zones cluster.

So, in this pr, use the pvc instead of pv.

### DIFF
--- a/features/storage/security.feature
+++ b/features/storage/security.feature
@@ -1,20 +1,27 @@
 Feature: storage security check
 
   # @author lxia@redhat.com
+  # @author piqin@redhat.com
   @admin
   Scenario Outline: [origin_infra_20] volume security testing
     Given I have a project
-    And I have a 1 GB volume and save volume id in the :vid clipboard
+    When I run oc create over "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/storage/misc/pvc.json" replacing paths:
+      | ["metadata"]["name"] | mypvc1 |
+    Then the step should succeed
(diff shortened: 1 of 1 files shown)


features/storage/security.feature
```diff
@@ -93,10 +101,10 @@ Feature: storage security check
...
 
     Examples:
-      | storage_type         | volume_name | type   |
-      | gcePersistentDisk    | pdName      | gce    | # @case_id OCP-9700
-      | awsElasticBlockStore | volumeID    | ebs    | # @case_id OCP-9699
-      | cinder               | volumeID    | cinder | # @case_id OCP-9721
+      | storage_type         |
+      | gcePersistentDisk    | # @case_id OCP-9700
+      | awsElasticBlockStore | # @case_id OCP-9699
+      | cinder               | # @case_id OCP-9721
```
liangxia: Updating parameters will require update in Polarion, thus break 3.11 automation for those test cases.

qinpingli: Yes, once the PR is merged, I'll update polarion.
This branch is for OCP 4.x only, right?

qinpingli: updated the PR to keep the parameters.

liangxia:
LGTM.  Still need feature team to review.

PR #803, opened 2020-03-16
some rule fixes

yapei:
[4.4 pass log](https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/75884/console)


yapei:
@pruan-rht  @liangxia  can you help lgtm label if no objections?

PR #808, opened 2020-03-16
Add support to get pod nominatedNodeName

kasturinarra:
This function is to get pod.nominatedNodeName to test priority/preemption scheduling


kasturinarra:
@zhouying7780 can you please help review ? Thanks

PR #809, opened 2020-03-16
Add support for resource type priority classes

kasturinarra:
Add support for resource type priority classes and also adds library which helps to delete a priority class after scenario.

### DIFF
--- a/features/step_definitions/scheduler.rb
+++ b/features/step_definitions/scheduler.rb
@@ -1,0 +1,6 @@
+Given /^the #{QUOTED} scheduler priorityclasses is restored after scenario$/ do |name|
+  ensure_admin_tagged
+  ensure_destructive_tagged
+  @result = admin.cli_exec(:delete, object_type: 'priorityclasses', object_name_or_id: name)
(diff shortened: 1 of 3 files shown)


kasturinarra:
@zhouying7780 can you please help review ? Thanks

lib/openshift/priority_class.rb
```diff
@@ -0,0 +1,8 @@
+require 'openshift/cluster_resource'
```
pruan-rht: @kasturinarra FYI, you don't really to add this for deletion if you are calling `admin.cli_exec...`

zhouying7780:
LGTM.

features/step_definitions/scheduler.rb
```diff
@@ -0,0 +1,5 @@
+Given /^the #{QUOTED} scheduler priorityclasses is restored after scenario$/ do |name|
+  ensure_admin_tagged
```
liangxia: Also `ensure_destructive_tagged` so it does not interfere with other scenarios.

PR #812, opened 2020-03-17
Add prepare for machines upgrade

liangxia:


jhou1:
@liangxia That is a good solution! Just one question, does the prepare steps need case mapping in polarsion?

liangxia:
> Just one question, does the prepare steps need case mapping in polarion?

@jhou1  I'm not quite sure.  So need @geliu2016 to review/answer.

features/upgrade/machines/machines.feature
```diff
@@ -1,6 +1,11 @@
 Feature: Machine-api components upgrade tests
+  @upgrade-prepare
+  Scenario Outline: Cluster operator should be available after upgrade - prepare
+    Given the expression should be true> "True" == "True"
```
pruan-rht: @liangxia can you please explain what this check is for?

liangxia: @pruan-rht  Act as a place-holder.

pruan-rht: @liangxia placeholder for what?

liangxia: @pruan-rht  According to our upgrade workflow, we need an upgrade-prepare and upgrade-check for each scenario. But some of them do not need any prepare steps, which lead to errors "can not find scenarios" in the log, so we just add a simple/useless step here to get rid of the errors in the log.

pruan-rht: ok, I see, perhaps we should add this note into the PR

liangxia: @pruan-rht  Added the note into the PR.

PR #815, opened 2020-03-17
Fixing autoscaler test flakes

jhou1:
This should fix the issue in https://issues.redhat.com/browse/OCPQE-34. It includes fixes to:
-  `the expression should be true` was incorrectly used which always returns true
- Update the `#ready?` to firmly check machineset has consistent `availableReplicas`, `readyReplicas`, `replicas` and `fullyLabeledReplicas`. Description of these fields are in https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_03_machineset.crd.yaml#L223

I've tested all autoscaler test on GCP which we fail often, now they are all passing.

@sunzhaohua2 @miyadav PTAL

### DIFF
--- a/lib/openshift/machine_set.rb
+++ b/lib/openshift/machine_set.rb
@@ -15,10 +15,10 @@ module BushSlicer
       rr.dig('status', 'availableReplicas').to_i
     end
 
-    def ready?(user: nil, cached: false, quiet: false)
+    def ready?(user: nil, quiet: false)
       result = {}
-      result[:success] = (desired_replicas(user: user, cached: cached, quiet: quiet) ==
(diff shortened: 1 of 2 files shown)


lib/openshift/machine_set.rb
```diff
@@ -15,10 +15,10 @@ def available_replicas(user: nil, cached: true, quiet: false)
       rr.dig('status', 'availableReplicas').to_i
```
sunzhaohua2: \
[receipts.pdf](https://github.com/openshift/verification-tests/files/4357637/receipts.pdf)
\

PR #820, opened 2020-03-18
add automation scenario for logging upgrading

QiaolingTang:
@anpingli @geliu2016 could you please help to review this PR? Thanks.

### DIFF
--- a/features/upgrade/logging/upgrade.feature
+++ b/features/upgrade/logging/upgrade.feature
@@ -1,0 +1,96 @@
+Feature: Logging upgrading related features
+
+  # @author qitang@redhat.com
+  @admin
+  @destructive
+  @upgrade-prepare
+  @users=upuser1,upuser2
+  Scenario: Cluster logging checking during cluster upgrade - prepare 
+    # prepare data
+    Given I switch to the first user
+    When I run the :new_project client command with:
+      | project_name | logging-upgrade-data-check |
+    Then the step should succeed
+    Given I obtain test data file "logging/loggen/container_json_log_template.json"
+    When I run the :new_app client command with:
+      | file | container_json_log_template.json |
+    Then the step should succeed
+
+    # deploy clusterlogging, enable pvc for ES
+    Given logging operators are installed successfully
+    And default storageclass is stored in the :default_sc clipboard
+    Given I obtain test data file "logging/clusterlogging/clusterlogging-storage-template.yaml"
+    When I process and create:
+      | f | clusterlogging-storage-template.yaml    |
+      | p | STORAGE_CLASS=<%= cb.default_sc.name %> |
+      | p | PVC_SIZE=20Gi                           |
+      | p | ES_NODE_COUNT=3                         |
+      | p | REDUNDANCY_POLICY=SingleRedundancy      |
+    Then the step should succeed
+    Given I wait for the "fluentd" daemonset to appear up to 300 seconds
+    And I wait until ES cluster is ready
+    And I wait until "fluentd" log collector is ready
+    And I wait until kibana is ready
+    Given I wait for the project "logging-upgrade-data-check" logs to appear in the ES pod
+    # check cron jobs
+    When I check the cronjob status
+    Then the step should succeed
+
+    Given I switch to the first user
+    When I login to kibana logging web console
+    Then the step should succeed
+    Given the "logging-upgrade-data-check" project is deleted
+
+  # @case_id OCP-22911
+  # @author qitang@redhat.com
+  @admin
+  @upgrade-check
+  @users=upuser1,upuser2
+  Scenario: Cluster logging checking during cluster upgrade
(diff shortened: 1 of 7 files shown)


features/upgrade/logging/upgrade.feature
```diff
@@ -0,0 +1,130 @@
...
+    Then the step should succeed
+    When I run the :new_app client command with:
+      | file | <%= BushSlicer::HOME %>/testdata/logging/loggen/container_json_log_template.json |
+    Then the step should succeed
+
+    # deploy clusterlogging, enable pvc for ES
+    Given logging operators are installed successfully
+    And default storageclass is stored in the :default_sc clipboard
+    When I process and create:
+      | f | <%= BushSlicer::HOME %>/testdata/logging/clusterlogging/clusterlogging-storage-template.yaml |
```
anpingli, line 22: As we only run once cases during upgrade. we should prepare a ES clusters. Let's use nodecount=3.

QiaolingTang: Updated the template and the table, added a parameter `ES_NODE_COUNT`, set default value to `1`.

QiaolingTang:
Runner-v3 logs:
upgrade-prepare: https://url.corp.redhat.com/02100ea
upgrade-check: https://url.corp.redhat.com/10a223a

QiaolingTang:
Runner-v3 logs:
upgrade-prepare: https://url.corp.redhat.com/54baad2
upgrade-check: https://url.corp.redhat.com/4512750

@anpingli could you please help to review the PR?

features/upgrade/logging/upgrade.feature
```diff
@@ -0,0 +1,111 @@
...
+    # ensure there are no new PVCs after upgrading
+    And the expression should be true> BushSlicer::PersistentVolumeClaim.list(user: user, project: project).count == cluster_logging('instance').logstore_node_count
+    # check if kibana console is accessible
+    Given I switch to the first user
+    And the first user is cluster-admin
+    Given I login to kibana logging web console
+    Then the step should succeed
+
+    # upgrade EO if needed
+    Given I make sure the "elasticsearch-operator" operator version in the "openshift-operators-redhat" namespace matches the cluster version
```
anpingli, line 75: @QiaolingTang Could we have common check steps?
1.  All fluent pods are running.
2.  All ES pods are running. the ES is healthy.  All index in green status.
3.  Kibana can be accessible.
4. Curator is executed and in completed status.  
5. Re-index and index deleted pods are triggered and in completed status.
And the above common check steps after all pods are restarted.

QiaolingTang: @anpingli for check point 1~3, you can find them in [Line 55~72](https://github.com/openshift/verification-tests/pull/820/files#diff-b62bdfbba92de6867c219315222e8d66f85b158b28a374ce9a9fbeb7242eb6a1R55-R72) [Line 85](https://github.com/openshift/verification-tests/pull/820/files#diff-b62bdfbba92de6867c219315222e8d66f85b158b28a374ce9a9fbeb7242eb6a1R85), [Line 90~96](https://github.com/openshift/verification-tests/pull/820/files#diff-b62bdfbba92de6867c219315222e8d66f85b158b28a374ce9a9fbeb7242eb6a1R90-R96) and [Line 110](https://github.com/openshift/verification-tests/pull/820/files#diff-b62bdfbba92de6867c219315222e8d66f85b158b28a374ce9a9fbeb7242eb6a1R110)

QiaolingTang: For check step 4 and 5, I'm not sure if we could do this.  For curator cronjob, although we can set the schedule time ourselves, but it may be scheduled when the ES pod are under upgrading, in this condition, the job will fail, it's hard for us to judge the condition.
For rollover/delete cronjobs, the default schedule time is every 15 minutes, we can't change it unless we set the management state to `Unmanaged`, and they have the same problem as the curator cronjobs.

anpingli: @QiaolingTang  If no way to change the time. wait 15 minutes after the ES was restarted are acceptable.

QiaolingTang: OK, let me try

QiaolingTang: I added a step to check the cronjob status, @anpingli could you please take a look again? Thanks.

QiaolingTang:
Runner-v3 logs:
upgrade from 4.4 to 4.5:
upgrade-prepare: https://url.corp.redhat.com/b1e2460
upgrade-check: https://url.corp.redhat.com/9b88058

upgrade from 4.5 to 4.6
upgrade-prepare: https://url.corp.redhat.com/ac0f1f5
upgrade-check: https://url.corp.redhat.com/478142a

anpingli (commented):
/lgtm

PR #822, opened 2020-03-18
Modified the func for teardown

kasturinarra:
I actually wanted this function to be called at teardown so modified the function accordingly


kasturinarra:
@zhouying7780 can you please help review ? Thanks !!

PR #823, opened 2020-03-19
Integrate test data and test scripts

liangxia:
Address the issue of OCPQE-52

Only touch verification-tests part in this PR, will follow up with cucushift when we get closer to the final decision.


xltian:
Do you want to add a separate folder to group all the test files? for eg: testdata?

liangxia:
> Do you want to add a separate folder to group all the test files? for eg: testdata?

@xltian Good catch. They are grouped into the folder testdata, but forgot/lost that subpath in the substitution part.  Correct the substitution now.

liangxia:
Merging so team can test it and report issues.

PR #827, opened 2020-03-19
OCP-25294, OCP-25787 and OCP-26373

anuragthehatter:
This test case was created during 4.2 but eluded testing on 4.3 and 4.4 but its working as expected..phewww. We don't know why the future release versions were never updated in Polarion for this case. I have manually fixed it now and get this automated.
Logs: http://pastebin.test.redhat.com/846486
@weliang1 @zhaozhanqi @huiran0826 @rbbratta 

(Thanks @pruan-rht for helping catching an issue here as i was checking pod running status instead should have checked Ready. Learnt that in pod state table, running comes earlier than Ready :) . So if you check a pod ip on running status it may may not get assigned but definitely get assigned in Ready state)

### DIFF
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -426,15 +426,25 @@ Feature: Pod related networking scenarios
   # @case_id OCP-25294
   @admin
   Scenario: Pod should be accesible via node ip and host port
-    Given I select a random node's host
+    Given I store the workers in the :workers clipboard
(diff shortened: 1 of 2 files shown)


anuragthehatter:
Added OCP-25787 as well 
Logs: http://pastebin.test.redhat.com/846586

anuragthehatter:
OCP-26373 as well http://pastebin.test.redhat.com/847028

PR #828, opened 2020-03-19
fix service_project.exists? race condition

rbbratta:
Problem:
We re-use the service project each time we do step
"run commands on the host"
the clean_up method deletes the project.

When running lots of commands on the host
the after cleanup and before the next access
the service_project can be in the `Terminating` phase

```
  [15:43:58] INFO> Shell Commands: oc get projects tests-cucushift-oc44-standard-slave-gtknl-0 --output=yaml --kubeconfig=/home/jenkins/workspace/Runner-v3/workdir/ocp4_admin.kubeconfig
  apiVersion: project.openshift.io/v1
  kind: Project
....
  status:
    phase: Terminating

  [15:43:59] INFO> Exit Status: 0
  [15:43:59] INFO> Shell Commands: oc debug  --kubeconfig=/home/jenkins/workspace/Runner-v3/workdir/ocp4_admin.kubeconfig -n tests-cucushift-oc44-standard-slave-gtknl-0 node/qeci-3419-8s8xx-control-plane-0 -- chroot /host/ bash -c mkdir\ -v\ -p\ \'/tmp/workdir/cucushift-oc44-standard-slave-gtknl-0\'

  STDERR:
  Error from server (Forbidden): pods "qeci-3419-8s8xx-control-plane-0-debug" is forbidden: unable to create new content in namespace tests-cucushift-oc44-standard-slave-gtknl-0 because it is being terminated

  [15:43:59] INFO> Exit Status: 1
```

The `project.exists?` check doesn't check the phase, so it thinks
a `Terminating` project exists when it doesn't.

Check `project.active?(cached: true)` and if not active wait for it to
terminate.

Also change the cleanup to use `ensure_deleted` so we wait for it to
delete properly

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -415,11 +415,14 @@ module BushSlicer
     def service_project
       unless @service_project
         project = Project.new(name: "tests-" + EXECUTOR_NAME.downcase, env: self)
-        unless project.exists?
+        unless project.active?
+          project.wait_to_disappear(admin)
           res = project.create(by: admin, clean_up_registered: true)
           unless res[:success]
             raise "failed to create service project #{project.name}, see log"
           end
+          # we must update the cache, since we just waited for the previously active project to disappear
+          project.reload
         end
         @service_project = project
       end


rbbratta:
phase `Terminating` failure log 

http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/03/17/15:13:58/User_cannot_access_the_MCS_by_creating_a_service_that_maps_to_non-MCS_port_to_port_22623__3647356555/console.html

lib/environment.rb
```diff
@@ -415,7 +415,7 @@ def nodes(user: admin, refresh: false)
     def service_project
       unless @service_project
         project = Project.new(name: "tests-" + EXECUTOR_NAME.downcase, env: self)
-        unless project.exists?
+        unless project.exists? && project.active?
```
akostadinov: For better performance I'd suggest using `project.active?(cached: true)`.

akostadinov:
Good job debugging this! Just I think we can save an API call. See my line comment.

lib/environment.rb
rbbratta: okay I'll update and try to figure out a good was to test triggering the `Terminating` phase.

rbbratta:
ah, this won't work, the create fails

```
[01:02:16] INFO> Exit Status: 0

[01:02:16] INFO> Shell Commands: oc get projects tests-rbrattai-rbrattai --output=yaml --kubeconfig=/home/rbrattai/Documents/deployments/85655/workdir/ocp4_admin.kubeconfig

apiVersion: project.openshift.io/v1
kind: Project
spec:
  finalizers:
  - kubernetes
status:
  phase: Terminating


[01:02:16] INFO> Exit Status: 0

[01:02:16] INFO> Shell Commands: oc new-project tests-rbrattai-rbrattai --kubeconfig=/home/rbrattai/Documents/deployments/85655/workdir/ocp4_admin.kubeconfig


STDERR:
Error from server (AlreadyExists): project.project.openshift.io "tests-rbrattai-rbrattai" already exists

```

rbbratta:
stressed about 30 times, seems to work.

http://pastebin.test.redhat.com/846649

rbbratta:
retesting features/networking

https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/751/console

lib/environment.rb
```diff
@@ -416,6 +416,9 @@ def service_project
       unless @service_project
         project = Project.new(name: "tests-" + EXECUTOR_NAME.downcase, env: self)
         unless project.exists?
+          unless project.active?(cached: true)
+            project.wait_to_disappear(admin)
```
rbbratta: we might not need this, or maybe it should be `project.ensure_deleted(user: nil)` as well.

rbbratta: I guess `phase:` is either `Active` or `Terminating` for namespaces, so we might not need `ensure_deleted` because `!active?` means it is `Terminating`.

anuragthehatter:
> stressed about 30 times, seems to work.
> 
> http://pastebin.test.redhat.com/846649

Awesome debug, Ross. I believe with your changes this should work fine now. I wonder if we can create a "terminating phase" situation and see what happens. Not sure though at the moment how that can be done. Can look at it a bit today. Thanks

rbbratta:
yep, still not quite fixed, need to tweak the `active?` check

rbbratta:
cycle this about 20 times, replace project with `EXECUTOR_NAME.downcase`
```
  @admin
  Scenario: Repeat delete and creation
    Given I store the schedulable nodes in the :nodes clipboard
    Given I use the "<%= cb.nodes[0].name %>" node
    When admin creates a project with:
      | project_name | tests-rbrattai-rbrattai |
    When I run the :delete admin command with:
      | object_type       | project                 |
      | object_name_or_id | tests-rbrattai-rbrattai |
      | date |
    Then the step should succeed

  @admin
  Scenario: Repeat delete and creation
    Given I store the schedulable nodes in the :nodes clipboard
    Given I use the "<%= cb.nodes[0].name %>" node
      | date |
    Then the step should succeed
```

lib/environment.rb
```diff
@@ -450,7 +451,7 @@ def clean_up
       @user_manager.clean_up if @user_manager
       @hosts.each {|h| h.clean_up } if @hosts
       if @service_project
-        @service_project.delete_graceful(by: nil)
+        @service_project.ensure_deleted(user: nil)
```
akostadinov: THe original idea behind `delete_graceful` was not to wait for project to disappear because next scenario may nor even need service project at all. So it will have time to disappear in the background. Is this change needed to avoid problems? Or is it something you did just for the peace of mind?

rbbratta: the current behaviour is that the the project stays in `phase: Terminating` for so long that the next test already starts to use it and fails to create debug pod because the project is terminating.

```
  STDERR:
  Error from server (Forbidden): pods "qeci-3419-8s8xx-control-plane-0-debug" is forbidden: unable to create new content in namespace tests-cucushift-oc44-standard-slave-gtknl-0 because it is being terminated
```

We could keep this lazy delete and then just use the `wait_for_disapppear` before the `project.create`.

rbbratta: well I guess the other problem is that `delete_graceful` doesn't clear the project cache.

rbbratta: retesting without `ensure_deleted`

rbbratta: latest test without `ensure_deleted`

http://pastebin.test.redhat.com/847052

lib/environment.rb
```diff
@@ -415,11 +415,14 @@ def nodes(user: admin, refresh: false)
...
         project = Project.new(name: "tests-" + EXECUTOR_NAME.downcase, env: self)
-        unless project.exists?
+        unless project.active?
+          project.wait_to_disappear(admin)
           res = project.create(by: admin, clean_up_registered: true)
           unless res[:success]
             raise "failed to create service project #{project.name}, see log"
           end
+          # we must update the cache, since we just waited for the previously active project to disappear
+          project.reload
```
akostadinov: Thanks a lot @rbbratta . Thanks a lot for investigating and fixing this really tricky issue.

anuragthehatter:
Thanks Ross for investing time in this critical debug. cc'ing @zhaozhanqi @huiran0826 @weliang1 This should solve the project related hiccups we have been seeing in last some weeks in CI runs.

zhaozhanqi:
Awesome @rbbratta

PR #832, opened 2020-03-23
Add lib to patch clusterrole system:kube-scheduler

kasturinarra:
Add lib to patch clusterrole system:kube-scheduler

### DIFF
--- a/features/step_definitions/scheduler.rb
+++ b/features/step_definitions/scheduler.rb
@@ -38,3 +38,13 @@ Given /^the #{QUOTED} clusterrole is restored after scenario$/ do |name|
     raise "Cannot restore clusterrole: #{name}" unless @result[:success]
   }
 end
+
+Given /^I delete all custom scheduler resources after scenario$/ do
+  ensure_admin_tagged
+  ensure_destructive_tagged
+  teardown_add {
+    step %Q/admin ensures "my-scheduler" deployment is deleted from the "kube-system" project/
+    step %Q/admin ensures "my-scheduler" service_account is deleted from the "kube-system" project/
+    step 'admin ensures "my-scheduler-as-kube-scheduler" clusterrolebinding is deleted'
+  }
+end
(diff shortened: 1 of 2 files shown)


kasturinarra:
@zhouying7780 could you please help review ? Thanks

features/step_definitions/scheduler.rb
```diff
@@ -26,3 +26,14 @@
     raise "Cannot restore scheduler: #{name}" unless @result[:success]
   }
 end
+
+Given /^the #{QUOTED} clusterrole is restored after scenario$/ do |name|
+  ensure_admin_tagged
```
zhouying7780: Since this is destructive, add `ensure_destructive_tagged` too

features/step_definitions/scheduler.rb
```diff
@@ -26,3 +26,25 @@
...
+  ensure_destructive_tagged
+  patch_json = '[{"op": "remove", "path": "/rules/2/resourceNames/1"}]'
```
akostadinov: This appears to be a very scenario specific way to restore this object.
How about using `hidden recreate cluster resource( after scenario)?` step? Pls check how is it already used.

features/step_definitions/scheduler.rb
```diff
@@ -26,3 +26,25 @@
...
+  }
+end
+
+Given /^I delete all custom scheduler resources after scenario$/ do
+  ensure_admin_tagged
+  ensure_destructive_tagged
+  teardown_add {
+    step %Q/admin ensures "my-scheduler" deployment is deleted from the "kube-system" project/
+    step %Q/admin ensures "my-scheduler" service_account is deleted from the "kube-system" project/
+    step 'admin ensures "my-scheduler-as-kube-scheduler" clusterrolebinding is deleted'
```
akostadinov: Can we have these 3 steps in scenario directly? Or do we keep a convention that custom schedulers in all scenarios will keep that naming?

kasturinarra:
Can i close this PR as both of these are not needed now ?

zhouying7780:
@kasturinarra could close.

kasturinarra:
closing the PR based on above comments, thanks !!

PR #833, opened 2020-03-23
Automate OCP-25691

jhou1:
Scenario: prevent remediation with `maxUnhealthy`. Also combines the test scenario with OCP-25687 for testing the together in automation is easier.

Tested and passed

```
      [10:06:53] INFO> === End After Scenario: Use "maxUnhealthy" to prevent automated remediation ===

1 scenario (1 passed)
15 steps (15 passed)
11m49.742s
```

@sunzhaohua2 @miyadav PTAL

### DIFF
--- a/features/machine/machine_healthcheck.feature
+++ b/features/machine/machine_healthcheck.feature
@@ -72,3 +72,47 @@ Feature: MachineHealthCheck Test Scenarios
     When I create the 'Ready' unhealthyCondition
 
     Then the machine should be remediated
+
+  # @author jhou@redhat.com
+  # @case_id OCP-25691
+  @admin
+  @destructive
+  Scenario: Use "maxUnhealthy" to prevent automated remediation
+    Given I have an IPI deployment
+    And I switch to cluster admin pseudo user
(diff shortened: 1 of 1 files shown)


features/machine/machine_healthcheck.feature
```diff
@@ -72,3 +72,51 @@ Feature: MachineHealthCheck Test Scenarios
...
+  # @author jhou@redhat.com
+  # @case_id OCP-25691
+  @admin
+  @destructive
+  Scenario: Use "maxUnhealthy" to prevent automated remediation
+    Given I have an IPI deployment
+    And I switch to cluster admin pseudo user
+
+    Given I store the number of machines in the :num_to_restore clipboard
+    And admin ensures node number is restored to "<%= cb.num_to_restore %>" after scenario
```
miyadav: @jhou  , Here do we have to take care of number of nodes , as after the cloned  machineset is deleted , will that also be taken care by itself ?

jhou1: Yes, we must take care of the number of nodes, wait until it's deleted to exit. This is to ensure it  does not break the next test following it.

PR #834, opened 2020-03-23
Add arp command to help debug

anuragthehatter:
@pruan-rht Can you help merging this?
This step sometimes fails on misc clusters complaining say 192.168.1.1 is already in use. arp -a output will always help us regardless of fail or pass. Its not reproducible on personal setups.Thanks!


anuragthehatter:
> /lgtm. @anuragthehatter hopefully this doesn't dump too many debug information out...TMI...

Nope it just print less than 10 entries
http://pastebin.test.redhat.com/847540

PR #836, opened 2020-03-24
add some files for logging

QiaolingTang:
Hi @liangxia , I noticed you have changed the test files and scripts' url, some of my test files ([1], [2]) are missed, I'm adding these files, PTAL, thanks.

[1]https://github.com/openshift/verification-tests/blob/master/features/step_definitions/logging.rb#L406
[2]https://github.com/openshift/verification-tests/blob/master/features/step_definitions/logging.rb#L477-L478


liangxia:
/unhold
The script only see these three files missing and a parameter change in another file.

PR #843, opened 2020-03-25
Fix admin/scc.feature after integrate test data

liangxia:
Tested with a 4.4.0-rc.4 cluster locally, it passed.
/cc @zhouying7780


liangxia:
Remove scc_with_confilict_capabilities.yaml as it is not used by any scenarios.
```
[lxia@ocp verification-tests]$ grep -r scc_with_confilict ../verification-tests/
[lxia@ocp verification-tests]$ echo $?
1
```

PR #844, opened 2020-03-25
DRAFT: allow changing oc debug image

rbbratta:

so we can use can existing image, e.g. ovs image or any other

Add the `using image ""` option to the step and pass
the :image option all the way down to `exec_raw`

This might not work for all callers of `exec_raw`

### DIFF
--- a/features/step_definitions/node.rb
+++ b/features/step_definitions/node.rb
@@ -86,11 +86,15 @@ Given /^environment has( at least| at most) (\d+)( schedulable)? nodes?$/ do |cm
 end
 
 # @host from World will be used.
-Given /^I run( background)? commands on the host:$/ do |bg, table|
+Given /^I run( background)? commands on the host(?: using image "(.+?)")?:$/ do |bg, image, table|
   ensure_admin_tagged
 
   raise "You must set a host prior to running this step" unless host
-  @result = host.exec(*table.raw.flatten, background: !!bg)
+  if image
+    @result = host.exec(*table.raw.flatten, background: !!bg, image: image)
+  else
+    @result = host.exec(*table.raw.flatten, background: !!bg)
+  end
 end
 
 Given /^I run commands on the host after scenario:$/ do |table|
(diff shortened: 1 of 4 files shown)


features/step_definitions/node.rb
```diff
@@ -86,11 +86,15 @@
 end
 
 # @host from World will be used.
-Given /^I run( background)? commands on the host:$/ do |bg, table|
+Given /^I run( background)? commands on the host(?: using image "(.+?)")?:$/ do |bg, image, table|
   ensure_admin_tagged
 
   raise "You must set a host prior to running this step" unless host
-  @result = host.exec(*table.raw.flatten, background: !!bg)
+  if image
```
zhaozhanqi: how about we get the image name in here and use it directly. thus we do not need to update the steps `Given /^I run( background)? commands on the host`
eg. using openshift-multus pod since both openshift-sdn and OVN have this kind of pod.
oc get -n openshift-multus pods -l app=multus -o yaml | awk '/image:/ {print $2 ; exit}'

anuragthehatter: Thanks Ross for working on this. This step is being used in all modules not just sdn and If it breaks, it will break most of the scenarios so def need to work on this carefully.  Yea i think the better way is to fall over on multus image ```if``` support tools is out of reach.

rbbratta: I'm thinking about adding a `BUSHSLICER_HOST_DEBUG_IMAGE=multus` env variable and using that to override globally, then we could set it CI for disconnected clusters.

Unless there is some other tag or way of detecting disconnect clusters.

anuragthehatter: Hmm not sure about this @pruan-rht @akostadinov any comments?

rbbratta: alternate PR using `OPENSHIFT_ENV_OCP4_HOST_DEBUG_IMAGE=multus`  env var  https://github.com/openshift/verification-tests/pull/855

zhaozhanqi: hmm, I think we can use multus image as the default image whatever the cluster is disconnect or not.

rbbratta: replaced with https://github.com/openshift/verification-tests/pull/983

PR #848, opened 2020-03-25
add hpa_rubyex testfile

SSshahan:


liangxia:
Where is this file used ?

SSshahan:
my new automation case which I am working on.

liangxia:
Then the test file should be add in the same PR with the scenario itself

SSshahan:
does it matter? if we merge the testfile along with the scenario itself,
then we cannot give the jenkins pass log in this kind merge process.
Although we can change it to local directory when running on local host but
not sure how to do it on jenkins.

liangxia:
>  if we merge the testfile along with the scenario itself, then we cannot give the jenkins pass log in this kind merge process.

Why not ?  The test file can be referenced by <%= ENV['BUSHSLICER_HOME'] %>/testdata/xxxxxxx

Sending test files alone make no sense, as there are no context to understand how the file is used.

SSshahan:
OK, let me update.

liangxia:
Adding @yapei, seems Han does not aware of OCPQE-52

SSshahan:
I just heard of changes and i believe more of team members who did not
attend the weekly automation meeting also did not aware of the changes.

SSshahan:
https://github.com/openshift/verification-tests/pull/852 revert pr, please help to merge. Thanks

liangxia:
> I just heard of changes and i believe more of team members who did not attend the weekly automation meeting also did not aware of the changes.

Thank you for the remind. I just drop an email to aos-qe for awareness.

SSshahan:
@Liang Xia <lxia@redhat.com>  seems we don't support this on 3.11, right?

PR #850, opened 2020-03-25
update image tag to digest

zhaozhanqi:
@huiran0826 @anuragthehatter @weliang1 @rbbratta 
since for disconnect cluster need to use 'digest' for image

### DIFF
--- a/testdata/networking/aosqe-pod-for-ping.json
+++ b/testdata/networking/aosqe-pod-for-ping.json
@@ -10,7 +10,7 @@
   "spec": {
       "containers": [{
         "name": "hello-pod",
-        "image": "aosqe/pod-for-ping",
+        "image": "docker.io/aosqe/pod-for-ping@sha256:f61ca8fd1dd7dfb8d403b66df549ed5a94fc4626d3b19e2530af48aefa75e5ea",
         "resources":{
           "limits":{
             "memory":"340Mi"
--- a/testdata/networking/net_admin_cap_pod.yaml
+++ b/testdata/networking/net_admin_cap_pod.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   containers:
   - name: network-pod
-    image: docker.io/bmeng/centos-network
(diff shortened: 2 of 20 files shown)


testdata/networking/aosqe-pod-for-ping.json
```diff
@@ -10,7 +10,7 @@
   "spec": {
       "containers": [{
         "name": "hello-pod",
-        "image": "aosqe/pod-for-ping",
+        "image": "docker.io/aosqe/pod-for-ping@sha256:f61ca8fd1dd7dfb8d403b66df549ed5a94fc4626d3b19e2530af48aefa75e5ea",
```
liangxia: Prefix it with docker.io will not work in disconnected cluster even if you have mirror the image.

zhaozhanqi: are you sure? I think the `docker.io` is required. @wzheng1  please correct me if I'm wrong

wzheng1: yes, docker.io is required when using imagecontentsourcepolicy in disconencted cluster. We need to add it to image mirror list and create ICSP for it. @xiuwang

liangxia: Ack

testdata/networking/net_admin_cap_pod.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
 spec:
   containers:
   - name: network-pod
-    image: docker.io/bmeng/centos-network
+    image: docker.io/bmeng/centos-network@sha256:48da37205f9b43424e0983d4c5e7e07f77b7ba1504bbe35e2f264c75dcb4cd15
```
liangxia: The personal account should be replaced.

zhaozhanqi: aha. missed this one. thanks for catching this.  updated

PR #851, opened 2020-03-25
[OSP] add proxy support for private OSP

wjiangjay:
```
services:
  ......
  openstack_osp16_titan25:
    tenant_name: "shiftstack"
    user: shiftstack_user
    password: xxx
    domain: { name: "Default" }
    url: http://xx.xx.xx.xx:5000/v3/auth/tokens
    key: openshift-qe
    host_opts:
      user: root
      ssh_private_key: xxx.pem
      class: SSHAccessibleHost
    cloud_type: openstack
    image: rhcos-xxx
    proxy: http://xx.xx.xx.xx:8888
```

```
[1] pry(#<BushSlicer::DefaultWorld>)> iaas = iaas_by_service("openstack_osp16_titan25")                   
[2] pry(#<BushSlicer::DefaultWorld>)> iaas.get_floating_ips                                              
=> [{"id"=>"079ba68e-8f88-493b-bccb-780b1dbea2a5",                                                       
  "tenant_id"=>"af00b96789454b3d8b86b4140e836427",                                                       
  "floating_ip_address"=>"10.0.0.231",                                                                   
  "floating_network_id"=>"30a276c7-8bab-4a99-af75-f856bfe98d6f",                                       
  "router_id"=>nil,                                                                                      
  "port_id"=>nil,                                                                                                                                                                                                  
  "fixed_ip_address"=>nil,                                                                               
  "status"=>"DOWN",                                                                                      
  "description"=>"",                                                                                     
  "dns_domain"=>"",                                                                                      
  "dns_name"=>"",                                                                                                                                                                                                  
  "tags"=>[],                   
  "created_at"=>"2020-03-24T07:17:56Z",                                                                  
  "updated_at"=>"2020-03-24T07:17:56Z",             
  "revision_number"=>0,                                                                                  
  "project_id"=>"af00b96789454b3d8b86b4140e836427"}]
```

### DIFF
--- a/lib/launchers/openstack.rb
+++ b/lib/launchers/openstack.rb
@@ -28,6 +28,11 @@ module BushSlicer
                      'openstack_qeos10'
       opts = default_opts(service_name).merge options
 
+      # for private OSP behind proxy
+      if opts[:proxy]
+        ENV['http_proxy'] = opts[:proxy]
+      end
+
       ## poor's man OSP version detection
       url = opts[:url].match(%r{(^https?://[-:.a-zA-Z0-9]+)(/.*)?$})&.to_a&.fetch(1)
       raise "cannot parse auth URL #{opts[:url].inspect}" unless url


wjiangjay:
@akostadinov help review this, thx

lib/launchers/openstack.rb
```diff
@@ -28,6 +28,11 @@ def self.instance(**options)
                      'openstack_qeos10'
       opts = default_opts(service_name).merge options
 
+      # for private OSP behind proxy
+      if opts[:proxy]
+        ENV['http_proxy'] = opts[:proxy]
```
wjiangjay: Due to https://github.com/openshift/verification-tests/blob/master/lib/http.rb#L44

wjiangjay:
since it modifies the ENV, so will make all HTTP requests goes with the proxy once we initial the iaas by iaas_by_name.

lib/launchers/openstack.rb
akostadinov: I had to revert because it got merged before I finished my comment. I'm wondering why do we need an OSP behind proxy support. It is fine, just curious why we use such setup.

I think we should pass the `proxy` option to `Http` instead of altering the env variable.

PR #855, opened 2020-03-25
DRAFT: allow global override of oc debug image

rbbratta:

so we can use can existing image, e.g. ovs image or any other

Add the `OPENSHIFT_ENV_OCP_HOST_DEBUG_IMAGE` option change
the global oc debug image

quick hack to enable using the multus image


rbbratta:
replaced with https://github.com/openshift/verification-tests/pull/983

PR #856, opened 2020-03-26
[OSP][IaaS scope] add proxy support for private OSP

wjiangjay:
follow-up https://github.com/openshift/verification-tests/pull/851 

```
[5] pry(#<BushSlicer::DefaultWorld>)> iaas = iaas_by_service("openstack_osp16_titan25")
[6] pry(#<BushSlicer::DefaultWorld>)> iaas.proxy_set?
=> true
[7] pry(#<BushSlicer::DefaultWorld>)> iaas.get_networks.length
=> 2
[8] pry(#<BushSlicer::DefaultWorld>)> iaas = iaas_by_service("openstack_osp16_aosqe_rdu2")
[9] pry(#<BushSlicer::DefaultWorld>)> iaas.proxy_set?
=> false
[10] pry(#<BushSlicer::DefaultWorld>)> iaas.get_networks.length
=> 5

```

### DIFF
--- a/lib/launchers/openstack10.rb
+++ b/lib/launchers/openstack10.rb
@@ -34,6 +34,8 @@ module BushSlicer
                      'openstack_upshift'
       @opts = default_opts(service_name).merge options
 
+      @proxy = opts[:proxy] || ENV["http_proxy"]
+
       @os_user = ENV['OPENSTACK_USER'] || opts[:user]
       unless @os_user
         Timeout::timeout(120) do
@@ -78,6 +80,11 @@ module BushSlicer
       self.get_token()
     end
 
+
+    def proxy_set?()
+      return @proxy ? true : false
+    end
+
     # @return [ResultHash]
     # @yield [req_result] if block is given, it is yielded with the result as
     #   param
@@ -93,6 +100,10 @@ module BushSlicer
         :read_timeout => read_timeout,
         :open_timeout => open_timeout
       }
+
+      if proxy_set?()
+        req_opts.merge!({:proxy => @proxy})
+      end
 
       case method
       when "GET", "DELETE"
--- a/lib/launchers/openstack10.rb
+++ b/lib/launchers/openstack10.rb
@@ -101,7 +101,7 @@ module BushSlicer
         :open_timeout => open_timeout
       }
 
-      if proxy_set?()
+      if proxy_set?
         req_opts.merge!({:proxy => @proxy})
       end
 
--- a/lib/launchers/openstack4.rb
+++ b/lib/launchers/openstack4.rb
@@ -94,7 +94,7 @@ module BushSlicer
         :open_timeout => open_timeout
       }
 
-      if proxy_set?()
+      if proxy_set?
         req_opts.merge!({ :proxy => @proxy })
       end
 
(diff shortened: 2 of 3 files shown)


wjiangjay:
@akostadinov help review this, thx

lib/launchers/openstack4.rb
```diff
@@ -66,6 +68,10 @@ def initialize(**options)
       self.get_token()
     end
 
+    def proxy_set?()
+        return @proxy ? true : false
+    end
+
```
liangxia: Is openstack4.rb still used ?

wjiangjay: no idea, but since it's OpenSourced, so better with it.

akostadinov: we need to remove openstack4

lib/launchers/openstack10.rb
```diff
@@ -78,6 +80,11 @@ def initialize(**options)
       self.get_token()
     end
 
+
+    def proxy_set?()
+      return @proxy ? true : false
+    end
```
akostadinov: Please rename method to `proxy?`.

lib/launchers/openstack10.rb
```diff
@@ -94,6 +101,10 @@ def rest_run(url, method, params, token = nil, read_timeout = 60, open_timeout =
         :open_timeout => open_timeout
       }
 
+      if proxy_set?()
```
akostadinov: no need to use parentheses `()`

lib/launchers/openstack4.rb
akostadinov: but it is fine to keep it for the time being. Just saying it is not really needed

akostadinov:
Thank you, just two style comments, otherwise looks good

lib/launchers/openstack10.rb
wjiangjay: I choose the same name with RestClient https://www.rubydoc.info/gems/rest-client/RestClient#proxy_set%3F-class_method

lib/launchers/openstack10.rb
wjiangjay: ok, on it

lib/launchers/openstack10.rb
wjiangjay: any strong reason to use `proxy?` ?

akostadinov: Strong reason is avoiding length when it gives no additional information to the reader. `set` in this case tells nothing to help a code reader. And for code writer it is just some suffix to remember.

If you want to clarify the method like `os_api_access_proxy` (I'm not sure this is a very good name but giving a quick example) then fine.

At the end it is not only about correctness but also style the project maintainer tries to establish. I'm not saying what you did is wrong. I'm just asking you to change to a shorter version unless you have an even better idea.

Actually having a method `proxy?` or `proxy_set?` is not needed at all because it adds code length that is not really necessary to achieve end goal. See my other comment.

akostadinov:
Superseded by #879

akostadinov:
Decided to get rid of some old code.

lib/launchers/openstack10.rb
wjiangjay: Thanks

PR #862, opened 2020-03-26
Add files required for custom scheduler

kasturinarra:
Add files required to test custom scheduler


kasturinarra:
@zhouying7780 can you please help review ?

liangxia:
Add your scenarios with test files in the same PR.  Not sure how/what to review when there are test files only but no context to understand how it is used.

PR #863, opened 2020-03-26
add create poddisruptionbudget lib

zhouying7780:
@liangxia  Could you please help review , thanks.

### DIFF
--- a/lib/rules/cli/4.4.yaml
+++ b/lib/rules/cli/4.4.yaml
@@ -186,6 +186,19 @@
     :validate: --validate=<value>
 :create_namespace:
   :cmd: oc create namespace <name>
+:create_poddisruptionbudget:
+  :cmd: oc create poddisruptionbudget <name>
+  :options:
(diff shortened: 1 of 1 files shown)


lib/rules/cli/4.4.yaml
```diff
@@ -186,6 +186,19 @@
     :validate: --validate=<value>
 :create_namespace:
   :cmd: oc create namespace <name>
+:create_poddisruptionbudget:
```
liangxia: We also need to add it into 4.1.yaml if the command exist in 4.1/4.2/4.3

liangxia:
@zhouying7780  The command already exist in 4.4.yaml
```
[lxia@ocp verification-tests]$ grep -r create_poddisruptionbudget *
features/tierN/scheduler/alert.feature:    When I run the :create_poddisruptionbudget client command with:
lib/rules/cli/4.1.yaml::create_poddisruptionbudget:
lib/rules/cli/4.4.yaml::create_poddisruptionbudget:
```
Perhaps we are safe to close this PR ?

PR #866, opened 2020-03-26
Fixing a step definition impacted by BUSHSLICER_HOME changes

anuragthehatter:
Fixing a step definition impacted by BUSHSLICER_HOME changes. Its causing couple of dependent cases to fail. Also making quick debug steps more clear. 
@pruan-rht PTAL and help merging it
Re-ran Logs: http://pastebin.test.redhat.com/848845

cc @zhaozhanqi

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -806,13 +806,17 @@ Given /^a DHCP service is configured for interface "([^"]*)" on "([^"]*)" node w
   host = node.host
   dhcp_status_timeout = 30
   #Following will take dnsmasq backup and append curl contents to the dnsmasq config after
-  @result = host.exec_admin("cp /etc/dnsmasq.conf /etc/dnsmasq.conf.bak;curl #{ENV['BUSHSLICER_HOME']}/testdata/networking/multus-cni/dnsmasq_for_testbridge.conf | sed s/testbr1/#{br_inf}/g | sed s/88.8.8.100,88.8.8.110,24h/#{add_lease}/g > /etc/dnsmasq.conf;systemctl restart dnsmasq --now")
+  logger.info("Logging ARP entries on dnsmasq.conf node which might help in debugging later if required")
+  host.exec_admin("arp -a")
+  logger.info("Logging last 10 lines of dnsmasq.conf to check if required config is not appended already which might help in debugging later")
+  host.exec_admin("tail -10 /etc/dnsmasq.conf")
+  @result = host.exec_admin("cp /etc/dnsmasq.conf /etc/dnsmasq.conf.bak;curl https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/dnsmasq_for_testbridge.conf | sed s/testbr1/#{br_inf}/g | sed s/88.8.8.100,88.8.8.110,24h/#{add_lease}/g > /etc/dnsmasq.conf;systemctl restart dnsmasq --now")
   raise "Failed to configure dnsmasq service" unless @result[:success]
   wait_for(dhcp_status_timeout) {
     if host.exec_admin("systemctl status dnsmasq")[:response].include? "running"
       logger.info("dnsmasq service is running fine")
     else
-      host.exec_admin("arp -a;cp /etc/dnsmasq.conf.bak /etc/dnsmasq.conf && systemctl restart dnsmasq --now")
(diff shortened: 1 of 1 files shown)


features/step_definitions/networking.rb
```diff
@@ -806,13 +806,17 @@
   host = node.host
   dhcp_status_timeout = 30
   #Following will take dnsmasq backup and append curl contents to the dnsmasq config after
-  @result = host.exec_admin("cp /etc/dnsmasq.conf /etc/dnsmasq.conf.bak;curl #{ENV['BUSHSLICER_HOME']}/testdata/networking/multus-cni/dnsmasq_for_testbridge.conf | sed s/testbr1/#{br_inf}/g | sed s/88.8.8.100,88.8.8.110,24h/#{add_lease}/g > /etc/dnsmasq.conf;systemctl restart dnsmasq --now")
+  logger.info("Logging ARP entries on dnsmasq.conf node which might help in debugging later if required")
+  host.exec_admin("arp -a")
+  logger.info("Logging last 10 lines of dnsmasq.conf to check if required config is not appended already which might help in debugging later")
+  host.exec_admin("tail -10 /etc/dnsmasq.conf")
+  @result = host.exec_admin("cp /etc/dnsmasq.conf /etc/dnsmasq.conf.bak;curl https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/networking/multus-cni/dnsmasq_for_testbridge.conf | sed s/testbr1/#{br_inf}/g | sed s/88.8.8.100,88.8.8.110,24h/#{add_lease}/g > /etc/dnsmasq.conf;systemctl restart dnsmasq --now")
```
anuragthehatter: This curl will execute inside the debug node so i don't think BUSHSLICER_HOME will wotk here as per my quick experiment. SO reverting this corner case back to its originality

pruan-rht: @anuragthehatter, reverting to the `v3-testfiles` would mean we need to keep track of two repos, which is not ideal.  How about using the URL to the  `testdata` instead?  https://raw.githubusercontent.com/openshift/verification-tests/master/testdata/networking/multicast-rc.json.  One area of concerns is that the v3-testfiles is public

anuragthehatter: Thats right. oops i noticed i pointed it to public repo. ya thats what i wanted..guess too much parallel tabs. Fixed it . Thanks!

rbbratta: can we use `host.copy_to(local_path, remote_path)` ?

rbbratta: two steps, copy, then sed I guess.   The sed can be combined into a single sed command with `;`

`sed 's/testbr1/#{br_inf}/g;s/88.8.8.100,88.8.8.110,24h/#{add_lease}/g'`

anuragthehatter: @rbbratta Just changing the curl path in this definition due to ENV changes. Nothing more. I suggest we merge this asap to avoid further failures in automation. ```host.copyto``` might need additional validation and wait. Can play with that tweak later i guess. WDYT

rbbratta: something like this might work

```ruby
  dns_masq_conf = %Q(
conf-dir=/etc/dnsmasq.d,.rpmnew,.rpmsave,.rpmorig
no-resolv
expand-hosts
bogus-priv
domain=mydomain.net
local=/mydomain.net/
interface=#{br_inf}
dhcp-range=#{add_lease}
bind-interfaces
)
  @result = host.exec_admin("cat > /etc/dnsmasq.conf.new", stdin: dns_masq_conf)
  # cp -b uses default ~ suffix
  @result = host.exec_admin("cp -b /etc/dnsmasq.conf.new /etc/dnsmasq.conf;systemctl restart dnsmasq --now")
  raise "Failed to configure dnsmasq service" unless @result[:success]
  wait_for(dhcp_status_timeout) {
    if host.exec_admin("systemctl status dnsmasq")[:response].include? "running"
      logger.info("dnsmasq service is running fine")
    else
      host.exec_admin("arp -a;cp /etc/dnsmasq.conf~ /etc/dnsmasq.conf && systemctl restart dnsmasq --now"
```

anuragthehatter: Thanks @rbbratta for the idea. As we discussed i am not re-engineering this already working step but just changing the referenced curl to now point to private repo. I would definitely keep this in mind in future. Guess we are heavily dependent on fetching files via repos being an ```opensource org``` ..hehe

PR #867, opened 2020-03-26
modifed OCP-25787 to support 4.3 and 4.4 releases

anuragthehatter:
This CI case found errorneous during CI run on 4.3 however works fine on 4.4. Made it compatible on both 
@zhaozhanqi @pruan-rht 
4.3 logs: http://pastebin.test.redhat.com/848931
4.4 logs: http://pastebin.test.redhat.com/848927

### DIFF
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -349,32 +349,50 @@ Feature: SDN related networking scenarios
   @admin
   Scenario: Don't write CNI configuration file until ovn-controller has done at least one iteration
     Given the env is using "OVNKubernetes" networkType
-    And I select a random node's host
-    #Checking controller iteration 1
-    When I run command on the "<%= node.name %>" node's sdn pod:
-      | bash | -c | ovn-appctl -t ovn-controller connection-status |
+    And I store the masters in the :master clipboard
+    And I store "<%= cb.master[0].name %>" node's corresponding default networkType pod name in the :ovnkube_pod clipboard
+    Given I switch to cluster admin pseudo user
+    And I use the "openshift-ovn-kubernetes" project
+    #Fetching ovn master pod name to be used later 
+    When I run the :get client command with:
+      | resource      | pods                                   |
+      | fieldSelector | spec.nodeName=<%= cb.master[0].name %> |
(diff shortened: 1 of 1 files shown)


features/networking/sdn.feature
```diff
@@ -349,32 +349,50 @@ Feature: SDN related networking scenarios
...
-    And I select a random node's host
-    #Checking controller iteration 1
-    When I run command on the "<%= node.name %>" node's sdn pod:
-      | bash | -c | ovn-appctl -t ovn-controller connection-status |
+    And I store the masters in the :master clipboard
+    And I store "<%= cb.master[0].name %>" node's corresponding default networkType pod name in the :ovnkube_pod clipboard
+    Given I switch to cluster admin pseudo user
+    And I use the "openshift-ovn-kubernetes" project
+    #Fetching ovn master pod name to be used later 
+    When I run the :get client command with:
```
pruan-rht: @anuragthehatter, here's a nugget for you.  You can replace line 357-363 with, after which the `pod` object will have the pod information you need.  This step is useful to find the pod name by label.  
```
    And a pod is present with labels:
      | l | app=ovnkube-master |
```

anuragthehatter: @pruan-rht Thanks , let me try but i guess this would randomly get any pod with that label under openshift-ovn-kubernetes project? This whole case needs to be on ovn-master and ovnkube-node for same node to make sure iterations are working well.

anuragthehatter: @pruan-rht gave it a try, apparently its saving random master pod name. Currently 357-363 is fetching master pod name from a master node and 353 is fetching ovnkube pod name corresponding to same master node. As i mentioned we need same node for all iterations to make sense. Let me know if you have other suggestions to improve this. I guess i would implement in future a step like ```Give I fetch a pod name from node < node_name> with label " blah" ```

pruan-rht: ok, thanks for the explaination...we can merge this for now and think about enhancement later...

anuragthehatter:
Seems like this was approved but didn't merge. No sure what happended. Reopening this one

PR #870, opened 2020-03-27
Shorten the parameters

liangxia:
Using short parameters can adapt to the URL change, e.g, from v3-testfiles to integrate test files (or any other changes like this in the future), without any change in Polarion in future.
Also, a shorter parameter is more readable.

/cc @xiuwang @zhaozhanqi @lihongan


lihongan:
LGTM. The `Examples` part looks cleaner. Thank you.

PR #876, opened 2020-03-27
add lib for inspect command

zhouying7780:
@kasturinarra Could you please help review ? thanks.

### DIFF
--- a/lib/rules/cli/4.4.yaml
+++ b/lib/rules/cli/4.4.yaml
@@ -1136,6 +1136,13 @@
     :sync_config: --sync-config=<value>
     :type: --type=<value>
     :whitelist: --whitelist=<value>
+:oadm_inspect:
+  :cmd: oc adm inspect
+  :options:
+    :all_namespaces: --all-namespaces=<value>
+    :dest_dir: --dest-dir=<value>
+    :resource: <value>
+    :resource_name: <value>
 :oadm_new_project:
   :cmd: oc adm new-project <project_name>
   #:expected:


pruan-rht:
@zhouying7780 does this apply for OCP < 4.4 as well? If that's the case we need to add that to the 4.1 yaml as well.

lib/rules/cli/4.4.yaml
```diff
@@ -1136,6 +1136,13 @@
     :sync_config: --sync-config=<value>
     :type: --type=<value>
     :whitelist: --whitelist=<value>
+:oadm_inspect:
+  :cmd: oc adm inspect
+  :options:
+    :all_namespaces: --all-namespaces=<value>
```
akostadinov: is this not a global option already?

akostadinov:
@pruan-rht is correct. WIll let this merge in the mean time though

/lgtm

lib/rules/cli/4.4.yaml
pruan-rht: @zhouying7780  `all_namespaces` is a already defined, so I think we can take it out.

pruan-rht: @zhouying7780 please ignore comment...all_namespaces is apparently not defined as a global option...perhaps it should.

akostadinov: That was my initial reaction. But `oc options` does not return it. So it should not be in globals.

liangxia:
> does this apply for OCP < 4.4 as well? If that's the case we need to add that to the 4.1 yaml as well.

@pruan-rht  We have the command in 4.1.yaml, just some option names are not consistent in 4.1 and 4.4. Fixed the inconsistent in https://github.com/openshift/verification-tests/pull/881
```
$ grep -r 'adm_inspect' *
features/registry/registry.feature:    When I run the :oadm_inspect admin command with:
lib/rules/cli/4.1.yaml::oadm_inspect:
lib/rules/cli/4.4.yaml::oadm_inspect:
```

PR #877, opened 2020-03-27
Making test more stable

jhou1:
Increase wait time to make the tests more stable. Currently scale down takes longer than scale up, and sometimes test exit before nodes are scaled down. @sunzhaohua2 @miyadav PTAL

### DIFF
--- a/features/step_definitions/machine.rb
+++ b/features/step_definitions/machine.rb
@@ -63,7 +63,7 @@ Then(/^admin ensures node number is restored to #{QUOTED} after scenario$/) do |
   teardown_add {
     num_actual = 0
 
-    success = wait_for(600, interval: 30) {
+    success = wait_for(900, interval: 20) {
       num_actual = BushSlicer::Node.list(user: admin).length
       num_expected == num_actual.to_s
     }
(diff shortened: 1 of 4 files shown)


features/step_definitions/machine.rb
```diff
@@ -63,7 +63,7 @@
   teardown_add {
     num_actual = 0
 
-    success = wait_for(600, interval: 30) {
+    success = wait_for(1200, interval: 20) {
```
pruan-rht: @jhou1 IMO, 20 minutes is big long...it might be a OCP issue if we have to wait for that long

jhou1: Yes, normally in my test, the test could properly clean up and finish. It is in the ci, that the cluster didn't scale down successfully in test tear down and I can not reproduce it locally.

```
Failed to restore cluster, expected number of nodes 6, got 7 (RuntimeError)
```

Scale down takes longer time because there is a node drain logic during scale down and the drain must be successful. If there are more nodes to scale down, it probably increases the wait time. My wait time is too aggressive, I'll try to increase to 900 and test on more clusters to see if this fixes the test flake.

sunzhaohua2:
/lgtm
Tested autoscaler cases on aws, gcp, azure all passed. Couldn’t pass on osp. There maybe bug with osp or performance issue. But these changes looks good to me. Thanks.

PR #882, opened 2020-03-30
Clean-up 3.x CLI rules from master/4.x branch

liangxia:
We are not using those 3.x CLI rules in master/4.x branch, and leaving them here makes new comers confusing. Also potentially slow down the clone/load of the automation.

From automation perspective,
/cc @akostadinov @pruan-rht  
For cli/kubectl,
/cc @xingxingxia @zhouying7780 
For apb/svcat,
/cc @chengzhang1016 @jianzhangbjz


akostadinov:
LGTM

PR #887, opened 2020-03-31
rules added for fixing OCP-19532

yapei:
@yanpzhan  @SSshahan  Please help review thanks

### DIFF
--- a/lib/webauto/web4cucumber.rb
+++ b/lib/webauto/web4cucumber.rb
@@ -428,6 +428,16 @@ require_relative 'chrome_extension'
           end
         end
 
+        if action_body[:if_not_param]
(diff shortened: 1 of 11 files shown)


lib/webauto/web4cucumber.rb
```diff
@@ -428,6 +428,16 @@ def handle_action(action_body, **user_opts)
           end
         end
 
+        if action_body[:if_not_param]
+          res_param[:success] = true
+          if action_body[:if_not_param].kind_of? String
+            if user_opts.has_key? action_body[:if_not_param].to_sym
```
yapei: Didn't add too much extra logic currently, we can add later if more is needed

PR #889, opened 2020-03-31
fix occacional failure of ci job

SSshahan:
@akostadinov  @liangxia please help to review. Thanks.
  https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_verification-tests/887/pull-ci-openshift-verification-tests-master-images/2872/build-log.txt .  Parse the output of `curl -L -sS https://api.github.com/repos/mozilla/geckodriver/releases/latest` using` jq `

### DIFF
--- a/tools/openshift-ci/Dockerfile.centos
+++ b/tools/openshift-ci/Dockerfile.centos
@@ -10,9 +10,11 @@ ADD . /verification-tests/
 RUN set -x && \
     SCL_BASE_PKGS="centos-release-scl scl-utils-build" && \
     INSTALL_PKGS="rh-ruby26 rh-ruby26-ruby-devel rh-git218 bsdtar" && \
+    JQ_PKG="jq" &&\
     yum -y update && \
     yum install -y --enablerepo=centosplus $SCL_BASE_PKGS && \
     yum install -y --enablerepo=centosplus $INSTALL_PKGS && \
+    yum install -y --enablerepo=centosplus $JQ_PKG && \
     GECKODRIVER_DOWNLOAD_URL="$(curl -sSL https://api.github.com/repos/mozilla/geckodriver/releases/latest | jq . | grep -E 'browser_download_url.*linux64' | sed -E 's/.*(https[^"]*).*/\1/')" && \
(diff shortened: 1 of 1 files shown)


tools/openshift-ci/Dockerfile.centos
```diff
@@ -10,15 +10,18 @@ ADD . /verification-tests/
 RUN set -x && \
     SCL_BASE_PKGS="centos-release-scl scl-utils-build" && \
     INSTALL_PKGS="rh-ruby26 rh-ruby26-ruby-devel rh-git218 bsdtar" && \
+    JQ_PKG="jq" && \
+    curl -L https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm -o /tmp/epel-release-latest-7.noarch.rpm && \
+    yum install -y /tmp/epel-release-latest-7.noarch.rpm && \
+    yum install -y $JQ_PKG && \
```
akostadinov: could you integrate this in same command we download `SCL_BASE_PKGS`? Running `yum` twice is taking a lot of time. I would say, download all `SCL_BASE_PKGS`, `JQ_PKG` and `INSTALL_PKGS` with one `yum` command. Also please use option `--setopt=skip_missing_names_on_install=False`, otherwise we need a separate `rpm -V` line which is another needless slowdown. Better fail early.

akostadinov:
btw build failed `No package rh-ruby26 available.`

SSshahan:
`yum install -y --setopt=skip_missing_names_on_install=False --enablerepo=centosplus centos-release-scl scl-utils-build jq rh-ruby26 rh-ruby26-ruby-devel rh-git218 bsdtar` does the failure relavant to the command? it just passed in last time

akostadinov:
@SSshahan , command failed and this option is what allowed that. I'm saying in other `yum` commands the option should be present as well. It makes sure that all specified packages have been installed. Otherwise we need to catch that with `rpm -V` but htis is slower and one can miss to add the check.

In summary:

1. integrate `yum` commands into single call where possible
2. add the option if missing
3. remove `rpm -V` command

liangxia:
I have another try with https://github.com/openshift/verification-tests/pull/898 without adding new package.

PR #897, opened 2020-03-31
OCP-26373 and OVN funtionality check def

anuragthehatter:
http://pastebin.test.redhat.com/850360
@rbbratta @weliang1 @zhaozhanqi @huiran0826

### DIFF
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -494,10 +494,10 @@ Feature: Pod related networking scenarios
     | l | app=ovnkube-node |
   #Removing CNI config file from container to check readiness probe functionality
   When I run the :exec client command with:
-    | pod              | <%= pod.name %>    		       |
+    | pod              | <%= pod.name %>                       |
     | c                | ovnkube-node                          |
     | oc_opts_end      |                                       |
-    | exec_command     | rm      	                       |
+    | exec_command     | rm                                    |
     | exec_command_arg | /etc/cni/net.d/10-ovn-kubernetes.conf |
   Then the step should succeed
   #Deleting ovnkube-pod will force CNO to rewrite the conf file and bring cluster back to normal after scenario
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -492,7 +492,7 @@ Feature: Pod related networking scenarios
   And I use the "openshift-ovn-kubernetes" project
   And a pod is present with labels:
     | l | app=ovnkube-node |
-  #Removing CNI config file from container to check readiness probe funtionality
+  #Removing CNI config file from container to check readiness probe functionality
   When I run the :exec client command with:
     | pod              | <%= pod.name %>    		       |
     | c                | ovnkube-node                          |
@@ -506,7 +506,7 @@ Feature: Pod related networking scenarios
   admin ensure "<%= pod.name %>" pod is deleted from the "openshift-ovn-kubernetes" project
   OVN is functional on the cluster
   """
-  #Now make sure readinessprobe checking above file will cause one of the two ovnkube-node containers to go down
+  #Now make sure readiness probe checking above file will cause one of the two ovnkube-node containers to go down
   Given 30 seconds have passed
   When I run the :get admin command with:
     | resource      | pod             |
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -484,8 +484,8 @@ Feature: Pod related networking scenarios
   # @author anusaxen@redhat.com
   # @case_id OCP-26373
   @admin
-  @destructive  
-    Scenario: Pod readiness check for OVN		
+  @destructive
+  Scenario: Pod readiness check for OVN
     Given the env is using "OVNKubernetes" networkType
     And OVN is functional on the cluster
     Given I switch to cluster admin pseudo user
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -506,13 +506,13 @@ Feature: Pod related networking scenarios
   admin ensure "<%= pod.name %>" pod is deleted from the "openshift-ovn-kubernetes" project
   And OVN is functional on the cluster
   """
-  #Now make sure readiness probe checking above file will cause one of the two ovnkube-node containers to go down
+  #Now make sure readiness probe checking above file will cause one of the two ovnkube-node containers to go down and container ready status change to false
   Given I wait up to 30 seconds for the steps to pass:
   """
   When I run the :get admin command with:
-    | resource      | pod             |
-    | resource_name | <%= pod.name %> |
+    | resource      | pod                                             |
+    | resource_name | <%= pod.name %>                                 |
+    | o             | jsonpath='{.status.containerStatuses[1].ready}' |
   Then the step should succeed
-  And the output should contain: 
-    | 1/2 |
+  And the output should contain "false"
   """
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -485,35 +485,35 @@ Feature: Pod related networking scenarios
   # @case_id OCP-26373
   @admin
   @destructive  
-  Scenario: Pod readiness check for OVN		
-  Given the env is using "OVNKubernetes" networkType
-  And OVN is functional on the cluster
-  #Need this cleanup to restore cluster to healthy state due to future destructive steps
-  And I register clean-up steps:
-  """
-  And OVN is functional on the cluster
-  """
-  Given I switch to cluster admin pseudo user
-  And I use the "openshift-ovn-kubernetes" project
-  And a pod is present with labels:
-    | app=ovnkube-node |
-  #Removing CNI config file from container to check readiness probe functionality
-  When I run the :exec client command with:
-    | pod              | <%= pod.name %>                       |
-    | c                | ovnkube-node                          |
-    | oc_opts_end      |                                       |
-    | exec_command     | rm                                    |
-    | exec_command_arg | /etc/cni/net.d/10-ovn-kubernetes.conf |
-  Then the step should succeed
-  #Deleting ovnkube-pod will force CNO to rewrite the conf file and bring cluster back to normal after scenario
-  And admin ensure "<%= pod.name %>" pod is deleted from the "openshift-ovn-kubernetes" project after scenario
-  #Now make sure readiness probe checking above file will cause one of the two ovnkube-node containers to go down and container ready status change to false
-  Given I wait up to 30 seconds for the steps to pass:
-  """
-  When I run the :get admin command with:
-    | resource      | pod                                                                     |
-    | resource_name | <%= pod.name %>                                                         |
-    | o             | jsonpath='{.status.containerStatuses[?(@.name=="ovnkube-node")].ready}' |
-  Then the step should succeed
-  And the output should contain "false"
-  """
+    Scenario: Pod readiness check for OVN		
+    Given the env is using "OVNKubernetes" networkType
+    And OVN is functional on the cluster
+    Given I switch to cluster admin pseudo user
+    And I use the "openshift-ovn-kubernetes" project
+    And a pod is present with labels:
+      | app=ovnkube-node |
+    #Removing CNI config file from container to check readiness probe functionality
+    When I run the :exec client command with:
+      | pod              | <%= pod.name %>                       |
+      | c                | ovnkube-node                          |
+      | oc_opts_end      |                                       |
+      | exec_command     | rm                                    |
+      | exec_command_arg | /etc/cni/net.d/10-ovn-kubernetes.conf |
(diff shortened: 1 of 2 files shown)


features/networking/pod.feature
```diff
@@ -480,3 +480,37 @@ Feature: Pod related networking scenarios
...
+  OVN is functional on the cluster
+  """
+  #Now make sure readiness probe checking above file will cause one of the two ovnkube-node containers to go down
+  Given 30 seconds have passed
+  When I run the :get admin command with:
+    | resource      | pod             |
+    | resource_name | <%= pod.name %> |
+  Then the step should succeed
+  And the output should contain: 
+    | 1/2 |
```
pruan-rht: What is the `1/2` meaning?

features/networking/pod.feature
```diff
@@ -480,3 +480,37 @@ Feature: Pod related networking scenarios
...
+    | exec_command_arg | /etc/cni/net.d/10-ovn-kubernetes.conf |
+  Then the step should succeed
+  #Deleting ovnkube-pod will force CNO to rewrite the conf file and bring cluster back to normal after scenario
+  And I register clean-up steps:
+  """
+  admin ensure "<%= pod.name %>" pod is deleted from the "openshift-ovn-kubernetes" project
+  OVN is functional on the cluster
+  """
+  #Now make sure readiness probe checking above file will cause one of the two ovnkube-node containers to go down
+  Given 30 seconds have passed
```
pruan-rht: @anuragthehatter it's probably more reliable to use a loop than having a hard-coded sleep time.

features/step_definitions/networking.rb
```diff
@@ -975,3 +975,12 @@
   raise "Failed to kill the #{ovndb} database daemon" unless @result[:success]
 end
 
+Given /^OVN is functional on the cluster$/ do
+  ensure_admin_tagged
+  desired_ovnkube_node_replicas = daemon_set('ovnkube-node', project('openshift-ovn-kubernetes')).replica_counters(user: admin)[:desired]
+  available_ovnkube_node_replicas = daemon_set('ovnkube-node', project('openshift-ovn-kubernetes')).replica_counters(user: admin)[:available]
+  desired_ovnkube_master_replicas = daemon_set('ovnkube-master', project('openshift-ovn-kubernetes')).replica_counters(user: admin)[:desired]
+  available_ovnkube_master_replicas = daemon_set('ovnkube-master', project('openshift-ovn-kubernetes')).replica_counters(user: admin)[:available]
```
rbbratta: does `daemon_set` cache results?  or can we use temp variables to avoid calling the function multiple times?

or maybe we can use `.values_at` on the hash.

```ruby
  ovnkube_node_ds = daemon_set('ovnkube-node', project('openshift-ovn-kubernetes')).replica_counters(user: admin)
  desired_ovnkube_node_replicas, available_ovnkube_node_replicas = ovnkube_node_ds.values_at(:desired, :available)

```

features/networking/pod.feature
anuragthehatter: Peter, couldn't guess other way to check READY value. Like in example below. Any suggestions are welcome though
# oc get pod ovnkube-node-w9f9h
NAME                 READY   STATUS    RESTARTS   AGE
ovnkube-node-w9f9h   1/2     Running   0          3h10m

features/networking/pod.feature
anuragthehatter: Yep i agree. How could i forget here?

features/step_definitions/networking.rb
anuragthehatter: hmm, i have to check on that Ross. Thanks

features/networking/pod.feature
anuragthehatter: I can parse json though for READY bu though above would be easy way

anuragthehatter: json parsing for container status would make more sense..Fixed it.

features/step_definitions/networking.rb
anuragthehatter: Shortened it via temp variables and values_at array makes sense, thanks

zhaozhanqi:
BTW, if CNO can monitor the ovn pod is not ready and give the correct message?

anuragthehatter:
> @anuragthehatter the steps LGTM, just wondering if CNO can monitor the ovn pod is not ready and give the correct message?

Yes this will trigger problem in CNO reconcillation as well. Network operator would change progressing to true due to ovnkube-node dameonset availability reduced by 1. So in short, CNO can also monitor and i guess will message out like xxx damenonset is not available and changing progressing state to blah or true.

anuragthehatter:
Hey guys, any comments? Else it can be merged to hangout with other daily CI runs.

features/networking/pod.feature
```diff
@@ -480,3 +480,39 @@ Feature: Pod related networking scenarios
...
+  admin ensure "<%= pod.name %>" pod is deleted from the "openshift-ovn-kubernetes" project
+  And OVN is functional on the cluster
+  """
+  #Now make sure readiness probe checking above file will cause one of the two ovnkube-node containers to go down and container ready status change to false
+  Given I wait up to 30 seconds for the steps to pass:
+  """
+  When I run the :get admin command with:
+    | resource      | pod                                             |
+    | resource_name | <%= pod.name %>                                 |
+    | o             | jsonpath='{.status.containerStatuses[1].ready}' |
```
rbbratta, line 515: we might be able to use a file in jsonpath and not rely on ordering of containers.

`'{.status.containerStatuses[?(@.name=="ovnkube-node")].ready}'`

features/networking/pod.feature
```diff
@@ -480,3 +480,39 @@ Feature: Pod related networking scenarios
...
+  # @case_id OCP-26373
+  @admin
+  @destructive  
+  Scenario: Pod readiness check for OVN		
+  Given the env is using "OVNKubernetes" networkType
+  And OVN is functional on the cluster
+  Given I switch to cluster admin pseudo user
+  And I use the "openshift-ovn-kubernetes" project
+  And a pod is present with labels:
+    | l | app=ovnkube-node |
```
liangxia, line 494: It should be
```
  And a pod is present with labels:
    | app=ovnkube-node |
```

features/networking/pod.feature
```diff
@@ -480,3 +480,39 @@ Feature: Pod related networking scenarios
...
+    | pod              | <%= pod.name %>                       |
+    | c                | ovnkube-node                          |
+    | oc_opts_end      |                                       |
+    | exec_command     | rm                                    |
+    | exec_command_arg | /etc/cni/net.d/10-ovn-kubernetes.conf |
+  Then the step should succeed
+  #Deleting ovnkube-pod will force CNO to rewrite the conf file and bring cluster back to normal after scenario
+  And I register clean-up steps:
+  """
+  admin ensure "<%= pod.name %>" pod is deleted from the "openshift-ovn-kubernetes" project
```
liangxia, line 506: If you use `admin ensure "<%= pod.name %>" pod is deleted from the "openshift-ovn-kubernetes" project after scenario`, then we do not need the step `I register clean-up steps`

features/networking/pod.feature
anuragthehatter: Thats a great suggestion, Ross. Was never able to crack this json code..heh. Fixed!

features/networking/pod.feature
anuragthehatter: Anyways it works both ways. Thanks Moved it to more optimised version as you suggested, Liang.

features/networking/pod.feature
anuragthehatter: Yep, aware of that but i actually want ```And OVN is functional on the cluster``` step at very last immediately post after this pod deletion step to make sure i leave cluster in healthy state so i put them under register back to back. WDYT

anuragthehatter: Nevermind, adjusted it in more sensible way now http://pastebin.test.redhat.com/853969.

features/networking/pod.feature
```diff
@@ -480,3 +480,40 @@ Feature: Pod related networking scenarios
...
+  @admin
+  @destructive  
+  Scenario: Pod readiness check for OVN		
+  Given the env is using "OVNKubernetes" networkType
+  And OVN is functional on the cluster
+  #Need this cleanup to restore cluster to healthy state due to future destructive steps
+  And I register clean-up steps:
+  """
+  And OVN is functional on the cluster
+  """
```
akostadinov, line 495: You can use a system verification step here. They run at the moment as well during clean-up.
```
Given system verification steps are used:
```

features/networking/pod.feature
akostadinov: I saw this. Original version also was good. In the new way, better use verification steps as I commented.

features/step_definitions/networking.rb
```diff
@@ -975,3 +975,13 @@
   raise "Failed to kill the #{ovndb} database daemon" unless @result[:success]
 end
 
+Given /^OVN is functional on the cluster$/ do
+  ensure_admin_tagged
+  ovnkube_node_ds = daemon_set('ovnkube-node', project('openshift-ovn-kubernetes')).replica_counters(user: admin)
```
akostadinov, line 980: Most object methods use cached values to avoid multiple API calls. It makes sense here to call `replica_counters(user: admin, cached: false)` to make sure you are getting values at the moment. The next step can still use cached value because we already updated the object cache. No need for a second API call.

features/step_definitions/networking.rb
```diff
@@ -975,3 +975,13 @@
...
+  ovnkube_master_ds = daemon_set('ovnkube-master', project('openshift-ovn-kubernetes')).replica_counters(user: admin)
```
akostadinov, line 981: Most object methods use cached values to avoid multiple API calls. It makes sense here to call `replica_counters(user: admin, cached: false)` to make sure you are getting values at the moment.

features/networking/pod.feature
```diff
@@ -480,3 +480,40 @@ Feature: Pod related networking scenarios
     | ip route |
   Then the output should contain:
     | <%= cb.pod_cidr %> dev <%= cb.tunnel_inf_name %> |
+    
+  # @author anusaxen@redhat.com
+  # @case_id OCP-26373
+  @admin
+  @destructive  
+  Scenario: Pod readiness check for OVN		
+  Given the env is using "OVNKubernetes" networkType
```
liangxia, line 489: The steps should be one more level indention than the Scenario name.

features/step_definitions/networking.rb
anuragthehatter: Make sense, we don't wan't cached results. Thanks

features/networking/pod.feature
```diff
@@ -480,3 +480,40 @@ Feature: Pod related networking scenarios
     | ip route |
   Then the output should contain:
     | <%= cb.pod_cidr %> dev <%= cb.tunnel_inf_name %> |
+    
+  # @author anusaxen@redhat.com
+  # @case_id OCP-26373
+  @admin
+  @destructive  
+    Scenario: Pod readiness check for OVN
```
rbbratta, line 488: trailing whitespace on this line

features/step_definitions/networking.rb
```diff
@@ -975,3 +975,13 @@
...
 
+Given /^OVN is functional on the cluster$/ do
+  ensure_admin_tagged
+  ovnkube_node_ds = daemon_set('ovnkube-node', project('openshift-ovn-kubernetes')).replica_counters(user: admin,cached: false)
+  ovnkube_master_ds = daemon_set('ovnkube-master', project('openshift-ovn-kubernetes')).replica_counters(user: admin,cached: false)
+  desired_ovnkube_node_replicas, available_ovnkube_node_replicas = ovnkube_node_ds.values_at(:desired, :available)
+  desired_ovnkube_master_replicas, available_ovnkube_master_replicas = ovnkube_master_ds.values_at(:desired, :available)
+  
+  raise "OVN is not running correctly! Check one of your ovnkube-node pod" unless desired_ovnkube_node_replicas == available_ovnkube_node_replicas && available_ovnkube_node_replicas != 0
+  raise "OVN is not running correctly! Check one of your ovnkube-master pod" unless desired_ovnkube_master_replicas == available_ovnkube_master_replicas && available_ovnkube_master_replicas != 0
```
rbbratta, line 986: I dunno, I guess technically this might be better as `> 0` instead of `!= 0`.   I don't think we can have negative replicates, but it is still an `int`.

anuragthehatter: hmm, doesn't matter if we are keeping either ```>``` or ```!=```  as we know negative replicas won't be possible or we are not checking how counters work in this case but our intention is to make sure available replicas are always positive integer so making ```> 0``` would also fill the requirements here.

PR #900, opened 2020-04-01
Better HOME

liangxia:
/cc @pruan-rht @akostadinov

### DIFF
--- a/features/admin/pod.feature
+++ b/features/admin/pod.feature
@@ -5,7 +5,7 @@ Feature: pod related features
   Scenario: Endpoints should update in time and no delay
     Given I have a project
     When I run the :create client command with:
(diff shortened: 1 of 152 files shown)


akostadinov:
/lgtm

while on it, I'm sorry for gradually adding more to this. But this comes to me in an evolutionary way. I think that ideally we need to write a new method so that, in places where we need such resources:
```
<%= test_file "testdata/networking/list_for_pods.json" %>
```
In this way we can always change location of test files, even move them back remote at some point and keep tests the same. If you think you are up to another round of this renaming, I can help with the method implementation. Perhaps implement in `DefaultWorld` class.

liangxia:
> In this way we can always change location of test files, even move them back remote at some point and keep tests the same. If you think you are up to another round of this renaming, I can help with the method implementation. Perhaps implement in `DefaultWorld` class.

@akostadinov  Let's do it. We may have to move again in the near future.

PR #903, opened 2020-04-01
fix OCP-26910

yapei:
@SSshahan  please help review,thanks


lib/rules/web/admin_console/4.4/projects.xyaml
```diff
@@ -1,6 +1,6 @@
 #---check project related action
 create_project:
-  action: click_create_button
+  action: click_create_project_button
```
yapei, line 3: if we use `click_create_button` then another Create button will be clicked, see
Screen Shot 2020-04-01 at 6 39 00 PM

PR #908, opened 2020-04-02
Automated OCP-21926

wzheng1:
Here is test result: https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/81811/console

### DIFF
--- a/features/cli/oc_registry.feature
+++ b/features/cli/oc_registry.feature
@@ -1,0 +1,40 @@
+Feature: oc registry command scenarios
+
+  # @author wzheng@redhat.com
+  # @case_id OCP-21926
+  Scenario: Check function of oc registry command
+    Given I have a project
+    When I run the :registry_info client command with:
+      | internal | true |
+    Then the step should succeed
+    And the output should contain:
+      | image-registry.openshift-image-registry.svc:5000 |
+    When I run the :registry_info client command with:
+      | internal | true |
+      | quiet    | true |
+    Then the step should fail
+    And the output should contain:
+      | error |
+    When I run the :registry_info client command with:
+      | internal | 123 |
+    Then the step should fail
+    And the output should contain:
+      | invalid |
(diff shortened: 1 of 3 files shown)


wzheng1:
@openshift/devexp-qe Please review, thanks!

features/cli/oc_registry.feature
```diff
@@ -0,0 +1,40 @@
...
+
+  # @author wzheng@redhat.com
+  # @case_id OCP-21926
+  Scenario: Check function of oc registry command
+    Given I have a project
+    When I run the :registry_info client command with:
+      | internal | true |
+    Then the step should succeed
+    And the output should contain:
+      | image-registry.openshift-image-registry.svc:5000 |
```
pruan-rht, line 11: @wzheng1 does this output changes depending on the cluster configuration?  Like in-cluster registry?

wzheng1: No, all 4.x cluster has the same name like this.

lib/rules/cli/4.1.yaml
```diff
@@ -1327,3 +1327,13 @@
...
+:registry_info:
+  :cmd: oc registry info
+  :options:
+    :internal: --internal=<value>
+    :quiet: --quiet=<value>
+:registry_login:
+  :cmd: oc registry login
+  :options:
+    :z: -z default
+    :skip-check: --skip-check=<value>
```
liangxia, line 1339: Sort them with alphanumeric order.

wzheng1: Done

features/cli/oc_registry.feature
xiuwang: @wzheng1 If this will report error when the route for default regtistry is enabled?

wzheng1: No, that's why I use internal=true option

PR #913, opened 2020-04-03
add lib for oc extension plugin

zhouying7780:
@kasturinarra Please help review , thanks.

### DIFF
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -391,6 +391,11 @@
     :template: --template=<value>
     :type: --type=<value>
     :wildcardpolicy: --wildcard-policy=<value>
+:exec_extension_oc_cmd:
+  :cmd: oc
+  :options:
+    :arg: <value>
+    :cmd_name: <value>
 :extract:
   :cmd: oc extract
   :options:
--- a/lib/rules/cli/4.4.yaml
+++ b/lib/rules/cli/4.4.yaml
@@ -391,6 +391,11 @@
     :template: --template=<value>
     :type: --type=<value>
     :wildcardpolicy: --wildcard-policy=<value>
+:exec_extension_oc_cmd:
+  :cmd: oc
+  :options:
+    :arg: <value>
+    :cmd_name: <value>
 :extract:
   :cmd: oc extract
   :options:


lib/rules/cli/4.1.yaml
```diff
@@ -391,6 +391,10 @@
     :template: --template=<value>
     :type: --type=<value>
     :wildcardpolicy: --wildcard-policy=<value>
+:exec_extension_oc_cmd:
+  :cmd: oc
+  :options:
+    :arg: <value>
```
kasturinarra, line 397: Can we use the lib which is already existing which looks similar to this ? Below is what i see can be used instead of creating this new one. what do you think  ?

exec_raw_oc_cmd_for_neg_tests

zhouying7780: Then we should use `When I run the :exec_raw_oc_cmd_for_neg_tests client command with:`   
that's make no sense.

kasturinarra: May i know why it does not make any sense ? I see it is very similar to the one you have written here but just the name is not suitable.

:exec_raw_oc_cmd_for_neg_tests:   # used for negative testing in which we want to deviate from yaml
  :cmd: oc
  :options:
    :arg: <value>
    :test_do_not_use: <value>

zhouying7780: @kasturinarra Have updated, make it more readable.

liangxia: Perhaps a scenario with the command can help us understand the command.

zhouying7780:
@liangxia @kasturinarra  Could you please help review ? thanks.

zhouying7780:
@kasturinarra I've updated the lib, please help review again, thanks.

kasturinarra:
@zhouying7780 when you said you made it more readable , may i know what has been added here to make it more readable ? Looks like i am missing something.

zhouying7780:
@kasturinarra  options add the 'cmd_name' to distinguish 'arg'. then we use plugin command like 'oc cmd1  arg1' will more readable.

zhouying7780:
@kasturinarra @liangxia could you please help review ? thanks.

PR #914, opened 2020-04-07
remove cases that are only applicable to the 3.x branch

pruan-rht:


pruan-rht:
these cases are only applicable to 3.x branch.  Removing them from master branch.  FYI, @mgahagan73

PR #916, opened 2020-04-07
Node proxy config verification

jhou1:
This covers the proxy config validation for node, verified it works. @sunilcio @weinliu @lyman9966 PTAL.

```
    [09:10:07] INFO> Exit Status: 0
    [09:10:07] INFO> === End After Scenario: Proxy config should be applied to kubelet and crio, Examples (#2) ===

2 scenarios (2 passed)
16 steps (16 passed)
0m43.251s
```

### DIFF
--- a/features/node/proxy.feature
+++ b/features/node/proxy.feature
@@ -1,0 +1,31 @@
+Feature: Node proxy configuration tests
+
+  # @author jhou@redhat.com
+  @admin
+  Scenario Outline: Proxy config should be applied to kubelet and crio
+    Given I switch to cluster admin pseudo user
+    When I run the :get client command with:
+      | resource      | proxy   |
+      | resource_name | cluster |
+      | o             | yaml    |
+    Then the step should succeed
+    And evaluation of `YAML.load @result[:response]` is stored in the :proxy clipboard
+
+    Given I store the ready and schedulable workers in the :workers clipboard
+    When I run the :debug client command with:
+      | resource     | node/<%= cb.workers.first.name %> |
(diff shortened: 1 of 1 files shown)


features/node/proxy.feature
```diff
@@ -0,0 +1,31 @@
+Feature: Node proxy configuration tests
+
+  # @author jhou@redhat.com
+  @admin
+  Scenario Outline: Proxy config should be applied to kubelet and crio
+    Given I switch to cluster admin pseudo user
+    When I run the :get client command with:
```
pruan-rht, line 7: @jhou1 try to avoid calling `:get` directly if possible.  It should be easy to add the class...I did it for you here https://github.com/openshift/verification-tests/pull/920.  After that's merged, then you can replace lines `7-12` with ```And the evalution of `proxy('cluster').raw_resource['spec']` is stored in the :proxy clipboard```

jhou1: Thanks @pruan-rht , I will replace it after this PR merges. Meanwhile, would you please enlighten me why we should avoid calling `:get` if possible?

pruan-rht: @jhou1 sure thing... The main reason we want to avoid using the raw `:get` is to adheres to the OOP paradigm.  By using encapsulation, we can address possible changes in the future easier.  In fact, that's one of the top comments that I've seen from reviewing PRs.  I can give a quick demo to the team or write a mojo about it...perhaps I should...

jhou1: Thank you @pruan-rht. I will follow the best practice.

PR #920, opened 2020-04-07
added skeleton for 'proxy' class

pruan-rht:


lib/openshift/proxy.rb
```diff
@@ -0,0 +1,6 @@
+module BushSlicer
+  class Proxy < ClusterResource
+    RESOURCE = "proxies"
```
akostadinov, line 3: @pruan-rht , please use full type as you can see in `oc get proxies -o yaml`. Now we see something like `org.k8s....` or `com.openshift...` making it up but hope you get the idea.

pruan-rht: @akostadinov hmm...I am not sure what you mean full type.

rbbratta: :thinking:  trying to figure this out as well, but it looks like `RESOURCES` is supposed to be the `api-resource`?   It looks like `proxies` to me. 

```
$ oc api-resources | grep -i prox
proxies                                                config.openshift.io                   false        Proxy
```

Also the code in `resource.rb` says it is also what is used in the REST API call

```
    # this needs to be set per sub class
    # represents the string we use with `oc get ...`
    # also represents string we use in REST call URL path
    # e.g. RESOURCE = "pods"
```

` GET https://example.com:6443/apis/config.openshift.io/v1/proxies?limit=500 `

akostadinov: @rbbratta , @pruan-rht , according to above output, it seems to be `proxies.config.openshift.io`

pruan-rht: ok, PR updated.... PTAL

pruan-rht:
@akostadinov looks like the github-bot doesn't want anything else in the `/lgtm` comment, otherwise the PR won't be merged automagically.  Mind doing a  `/lgtm` comment only?

PR #923, opened 2020-04-08
Get version info from ClusterVersion

liangxia:
We have to specify version info when trigger automation. We would like to get the version info automatically (from ClusterVersion).

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -278,11 +278,14 @@ module BushSlicer
     # @return [Array<String>] raw version, major and minor number
     def get_version(user:)
       if opts[:version]
-        @major_version, @minor_version = parse_version(opts[:version])
-        return opts[:version], @major_version, @minor_version
+        _version = opts[:version]
+      else
+        _version = admin.cli_exec(:get, resource: "clusterversion", resource_name: "version", o: "jsonpath={.status.desired.version}")
       end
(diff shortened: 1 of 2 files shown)


pruan-rht:
@akostadinov PTAL

lib/environment.rb
```diff
@@ -278,11 +278,14 @@ def parse_version(ver_str)
...
-        return opts[:version], @major_version, @minor_version
+        _version = opts[:version]
+      else
+        _version = admin.cli_exec(:get, resource: "clusterversion", resource_name: "version", o: "jsonpath={.status.desired.version}")
       end
+      @major_version, @minor_version = parse_version(_version)
+      return _version, @major_version, @minor_version
 
-      raise "getting version of cluster as user not possible presently"
+      #raise "getting version of cluster as user not possible presently"
```
akostadinov, line 288: We already set default version in `config/config.yaml` so we need to comment out that value if we want this to work OOB.

Also I think we better do like this:
```
if opst[:version]
  ...
elsif admin?
  ...
else
  raise "cluster version not set and getting without admin access not possible presently"
end
```

liangxia: @akostadinov  Updated.

akostadinov:
Thank you!
/lgtm
/unhold

PR #928, opened 2020-04-09
Add imagepruners resource

wzheng1:
Add imagepruners resource.

### DIFF
--- a/lib/openshift/imagepruners_imageregistry_operator_openshift_io.rb
+++ b/lib/openshift/imagepruners_imageregistry_operator_openshift_io.rb
@@ -1,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class ImageprunersImageregistryOperatorOpenshiftIo < ClusterResource
+    RESOURCE = 'imagepruners.imageregistry.operator.openshift.io'
+  end
+end


lib/openshift/imagepruners_imageregistry_operator_openshift_io.rb
```diff
@@ -0,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class ImageprunersImageregistryOperatorOpenshiftIo < ClusterResource
```
pruan-rht, line 4: Do we have to use such a long class name?

wzheng1: I just copy from other resources, is it fine to define shorter?

wzheng1: In order to be consistent with other resources, I suggest to keep it like this.

pruan-rht: @wzheng1 that's a lot of typing... I think it's fine for the `RESOURCE=...` line but for the class name, I think `ImagePruners` is sufficient.  WDYT @akostadinov ?

akostadinov: I would even call it `ImagePrunersImageregistryOperatorOpenshiftIo`, then the short name will be `image_prunner` which reads much better.

akostadinov: Using the full name for the class avoids conflicts like between `config_imageregistry_operator_openshift_io.rb` and `config_samples_operator_openshift_io.rb`. If we think that `ImagePrunners` will not conflict with anything, then class name can be shorter. Still keeping current `RESOURCE`.

pruan-rht:
https://github.com/openshift/verification-tests/pull/1106

PR #930, opened 2020-04-09
Workaround for download file with relative path

liangxia:
We still have lots of scenarios contain the step 'download a file from' with relative path.
Before those scenarios are fixed, we need a workaround to not failing those tests.

The idea is that we only do real download if the url contains '://', or else we copy the file from the relative path.
We can revert the workaround after those scenarios are fixed.

/cc @akostadinov @pruan-rht @xltian

### DIFF
--- a/features/step_definitions/net.rb
+++ b/features/step_definitions/net.rb
@@ -63,27 +63,36 @@ When /^I open web server via the(?: "(.+?)")? url$/ do |url|
 end
 
 Given /^I download a file from "(.+?)"(?: into the "(.+?)" dir)?$/ do |url, dl_path|
-  retries = 3
-  while true
-    @result = BushSlicer::Http.get(url: url, cookies: cb.http_cookies)
-    if @result[:success]
-      if dl_path
-        file_name = File.join(dl_path, File.basename(URI.parse(url).path))
+  if url.include? '://'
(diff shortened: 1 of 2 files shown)


features/step_definitions/net.rb
```diff
@@ -63,27 +63,36 @@
...
-    else
-      raise "Failed to download file from #{url} with HTTP status #{@result[:exitstatus]}"
+      retries -= 1
     end
-    retries -= 1
+  else
+    filename = File.basename(url)
+    FileUtils.cp(url,"#{filename}")
+
+    teardown_add {
```
pruan-rht, line 93: @liangxia we probably don't need to teardown since the project will be purged after the scenario is run.

liangxia: @pruan-rht  The test file is in the client side where our oc runs.

pruan-rht:
@liangxia would it be possible to get a list of the failed scenarios for this issue before mergeing this PR, so we don't forget which ones are real failures instead of bandaging the issue?

liangxia:
@pruan-rht   We can get that by `grep -r 'download a file from.*BushSlicer' ../verification-tests/`

pruan-rht:
@liangxia I think we can close this one

PR #932, opened 2020-04-09
Fix 'Given I download a file from'

barleyer:
Since the `openshift-qe/v3-testfiles` has been moved to this repo and some steps need to modify the download files,I do some changes on `Given I download a file from` defination.

### DIFF
--- a/features/step_definitions/net.rb
+++ b/features/step_definitions/net.rb
@@ -78,6 +78,12 @@ Given /^I download a file from "(.+?)"(?: into the "(.+?)" dir)?$/ do |url, dl_p
       @result[:file_name] = file_name
       @result[:abs_path] = File.absolute_path(file_name)
       break
+    elsif File.exist?(url) 
+      file_name = File.basename(url)
+      FileUtils.cp(url,file_name)
+      @result[:file_name] = file_name 
+      @result[:abs_path] = File.absolute_path(file_name)
+      break
     elsif @result[:exitstatus] >= 500 && retries > 0
       # we will wait for retry
     else


barleyer:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3/84119/console

barleyer:
@xingxingxia @liangxia please take a look!

features/step_definitions/net.rb
```diff
@@ -78,6 +78,12 @@
       @result[:file_name] = file_name
       @result[:abs_path] = File.absolute_path(file_name)
       break
+    elsif File.exist?(url)
```
xingxingxia, line 81: Space at the end of line.

features/step_definitions/net.rb
```diff
@@ -78,6 +78,12 @@
       @result[:file_name] = file_name
       @result[:abs_path] = File.absolute_path(file_name)
       break
+    elsif File.exist?(url) 
+      file_name = File.basename(url)
```
xingxingxia, line 82: In this elsif condition, should cover all the other `dl_path` logic as well.

xingxingxia:
As I know, this PR wants to fix cases which fail with failure at `When I download a file from "<%= BushSlicer::HOME %>/features/tierN/testdata/..."` AND need to modify the file content.
I don't object the fix which would keep `I download a file from "<%= BushSlicer::HOME %>/features/tierN/testdata/..."` in feature files and just revise the underlying step definition, as long as the mergers @liangxia @akostadinov agree. No more other comment.

akostadinov:
Honestly I don't really like the idea to hack the download step into something that copies local files too (unless local file is presented by a `file://` url). But if step is to be reused, then check for local file needs to be before the download loop. It is not needed to fail download and only then figure out that there is such a local file.

akostadinov:
Let me elaborate. I don't like reusing the step because Gherkin is supposed to be readable.
Ideally we would have a step `test file "some/path" is present`. Then no matter where we put test files in the future, we will only have one place to change.
I can write such step if team is ready to update existing usage.

barleyer:
I only have one concern I hope when we do some changes on architecture next time,please evaluate the advantages and disadvantages , especially for the owners of many cases that need to be changed.

akostadinov:
@barleyer , I assumed many people wanted to move the test files and that's why it was done. I was definitely very reluctant to do so for concerns about performance and backward compatibility.
I don't know if we have any track about taking this decision (except the pull request itself) as I didn't participate directly. But I'd encourage any discussion about making such processes more transparent and smooth.

liangxia:
/hold
Check more detain in OCPQE-52

akostadinov:
```
rg -l "I download a file from.*HOME.*?/testdata/" | xargs sed -i -e 's#I download a file from.*HOME.*/testdata/#I obtain test data file "#'
```

https://github.com/openshift/cucushift/pull/7617
https://github.com/openshift/verification-tests/pull/971

Hope didn't break anything but can't promise if download step was used in some obscure way.

PR #933, opened 2020-04-10
Fix case ID

liangxia:
OCP-21812 - [bz1609112][egressIP] Should remove the egressIP from the array if it was not being used
OCP-15473 - [egressIP] The related iptables/openflow rules will be removed once the egressIP gets removed from netnamespace

Multiple test cases pointing to the same scenario (not scenario outline) does not look correct to me.

I'm removing the "wrong" ID OCP-15473 here based on the test case title, and moved OCP-15473 to not automated in Polarion.
@huiran0826  Please take follow up action for the test cases in Polarion.
/cc @akostadinov @pruan-rht

### DIFF
--- a/features/networking/egress-ip.feature
+++ b/features/networking/egress-ip.feature
@@ -113,7 +113,7 @@ Feature: Egress IP related features
     """
 
   # @author huirwang@redhat.com
-  # @case_id OCP-21812 OCP-15473
+  # @case_id OCP-21812
   # @bug_id  1609112
   @admin
   @destructive


huiran0826:
@liangxia This auto cases just combine steps in the cases OCP-21812 and OCP-15473.  I did this to avoid frequently add/remove egress ips.  The steps for OCP-15473 checking  iptables/openflow  are already in the script.

akostadinov:
@huiran0826 , if one test case is a subset of another, we disable the subset test in Polarion and leave the other one only. Otherwise would end up running same test twice or having one of the test cases executed and the other not.

I understand why you may want them described separately. But in reality we try to stay more efficient as tests take a lot of time anyway.

huiran0826:
@akostadinov  Thanks for the comments. 
So doe that mean we cannot use one auto test case script to track two or more cases in Polarion?

akostadinov:
@huiran0826 , correct, unless the scenario is a Scenario Outline and each test case has a separate example.

PR #934, opened 2020-04-10
Clean-up 3.x/4.0 web rules

liangxia:
Remove 3.x web rules from master branch, which is for 4.x, should be safe as we still have that in v3 branch. With less files, we are faster at download/clone and script parse.
Also removed 4.0 as it is not a valid version now.

/cc @yapei @SSshahan @xiaocwan @yanpzhan 
/cc @akostadinov @pruan-rht


akostadinov:
This feels great!

akostadinov:
615 rules files!

yapei:
Thank you very much

PR #936, opened 2020-04-10
Bug fix for OCP-16485

qinpingli:
From OCP4.4, the ds version is apps/v1


duanwei33:
Seems all 4.x support apps/v1. Does testfile have same branch with test script now? If yes, can we update testfile directly for simple, and it can be used for other test script if needed?

qinpingli:
@duanwei33 Thanks for your review, I submitted a new patch, please help review again.

PR #939, opened 2020-04-13
fix machine annotations don't match the output

sunzhaohua2:
This pr will fix the failer `pattern not found: (?-mix:Annotations:\s+<none>) (RuntimeError)` in 4.5 env,  this failed beacause in 4.5  machines have annotations by default  in order to support autoscaler scale from zero. @jhou @miyadav Please help to take a look? Tested on 4.4 and 4.5 env.
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/84393/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/84392/console

### DIFF
--- a/features/machine/autoscaler.feature
+++ b/features/machine/autoscaler.feature
@@ -103,7 +103,7 @@ Feature: Cluster Autoscaler Tests
       | resource | machineset              |
       | name     | <%= machine_set.name %> |
     Then the step should succeed
-    And the output should match "Annotations:\s+<none>"
+    And the output should not match "autoscaling.openshift.io/machineautoscaler"
 
   # @author zhsun@redhat.com
   # @case_id OCP-22102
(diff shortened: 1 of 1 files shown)


features/machine/autoscaler.feature
```diff
@@ -142,7 +142,7 @@ Feature: Cluster Autoscaler Tests
       | resource | machineset                       |
       | name     | <%= cb.machineset_clone_22102 %> |
     Then the step should succeed
-    And the output should match "Annotations:\s+<none>"
+    And the output should not match "autoscaling.openshift.io/machineautoscaler"
```
jhou1, line 145: No need to run `describe` twice, instead do the following

```
    When I run the :describe admin command with:
      | resource | machineautoscaler |
      | name     | maotest0          |
    Then the step should succeed
    And the output should match "Name:\s+<%= cb.machineset_clone_22102_2 %>"
    And the output should not match "autoscaling.openshift.io/machineautoscaler"
```

features/machine/autoscaler.feature
```diff
@@ -174,7 +174,7 @@ Feature: Cluster Autoscaler Tests
       | resource | machineset                         |
       | name     | <%= cb.machineset_clone_22102_2 %> |
     Then the step should succeed
-    And the output should match "Annotations:\s+<none>"
+    And the output should not match "autoscaling.openshift.io/machineautoscaler"
```
jhou1, line 177: Same here, the test run `describe` 3 times, which is time consuming, run it once and use the `match` clause 3 times.

PR #944, opened 2020-04-14
use admin for checking OpenShiftSDN pods

rbbratta:
In step:
Given /^I wait for the networking components of the node to become ready$/ do

we don't use the admin user to check the pods so we fail with

```
Error from server (Forbidden): pods "sdn-cxjsw" is forbidden: User "testuser-0" cannot get resource "pods" in API group "" in the namespace "openshift-sdn"
```

Normally we switch to the admin user for these tests, but in the cases when we don't
we should use the correct user in the step.


rbbratta:
http://pastebin.test.redhat.com/854811

PR #945, opened 2020-04-14
OCP-26035: increase unidle check timeout to 30 seconds.

rbbratta:
Timeouts >= 8 seconds work, but less than that the service
doesn't seem to unidle in time.


rbbratta:
https://bugzilla.redhat.com/show_bug.cgi?id=1785360

PR #949, opened 2020-04-15
Modify operators.rb to judge whether it's OCP4.1 since no Upgradeable

yuhui-12:
Modify operators.rb to judge whether it's OCP-4.1 since no Upgradeable condition in OCP-4.1

### DIFF
--- a/features/step_definitions/operators.rb
+++ b/features/step_definitions/operators.rb
@@ -28,9 +28,13 @@ end
 
 Given /^the status of condition Upgradeable for marketplace operator as expected$/ do
   ensure_admin_tagged
-  actual_status = cluster_operator('marketplace').condition(type: 'Upgradeable', cached: false)['status']
-  status = 'True'
   cluster_version = cluster_version('version').channel.split('-')[1]
+  if cluster_version == "4.1"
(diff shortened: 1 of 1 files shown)


features/step_definitions/operators.rb
```diff
@@ -28,9 +28,13 @@
 
 Given /^the status of condition Upgradeable for marketplace operator as expected$/ do
   ensure_admin_tagged
-  actual_status = cluster_operator('marketplace').condition(type: 'Upgradeable', cached: false)['status']
-  status = 'True'
   cluster_version = cluster_version('version').channel.split('-')[1]
+  if cluster_version == "4.1"
+    actual_status = 'True'
```
pruan-rht, line 33: @yuhui-12, I don't see this variable being used anywhere in the step, a typo perhaps?

yuhui-12: Hi @pruan-rht ,
Thanks a lot for your comments. The variable will be used in line 59.
  unless status == actual_status

yuhui-12:
Paste the running log here.
http://pastebin.test.redhat.com/856652

PR #954, opened 2020-04-16
rules added for OCP-19727

yapei:
@yanpzhan  @xiaocwan  Can you help review? Thanks


lib/openshift/role.rb
```diff
@@ -0,0 +1,7 @@
+require 'openshift/project_resource'
```
yapei, line 1: added to support `Given admin ensures "list-pod" role is deleted from the "default" project after scenario`

lib/rules/web/admin_console/4.2/common_ui_elements.xyaml
```diff
@@ -88,6 +88,11 @@ click_on_resource_name:
     selector:
       xpath: //a[contains(@class,'co-resource-item__resource-name') and text()='<item>']
     op: click
+choose_project_from_dropdown_list:
```
yapei, line 91: this rule is needed but missing in 4.2, add it back

lib/rules/web/admin_console/4.2/filters.xyaml
```diff
@@ -6,6 +6,14 @@ set_filter_strings:
   - selector:
       <<: *filter_text_ele
     op: send_keys <filter_text>
+set_strings_in_dropdown_filter_box:
```
yapei, line 9: dropdown filter box has different definition then list filter box

PR #956, opened 2020-04-16
more update for ocp 10284 

pruan-rht:
previously tested with client 4.2 and pass.  But when I switched to client version 4.3.x and 4.4.x, it failed due to label mismatch.

### DIFF
--- a/features/storage/add_update_delete_volume.feature
+++ b/features/storage/add_update_delete_volume.feature
@@ -76,7 +76,6 @@ Feature: Add, update remove volume to rc/dc and --overwrite option
       | resource | dc/mydb  |
       | action   | --remove |
       | name     | v1       |
-      | confirm  | false    |
     Then the step should succeed
     And I wait for the resource "pod" named "<%= pod.name %>" to disappear
     And a pod becomes ready with labels:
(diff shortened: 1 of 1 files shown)


pruan-rht:
@akostadinov @liangxia @qinpingli  PTAL.  I have tested it with oc version 4.2, 4.3 and 4.4 and the test finally passed for all of them.

features/storage/add_update_delete_volume.feature
```diff
@@ -68,19 +68,19 @@ Feature: Add, update remove volume to rc/dc and --overwrite option
...
+      | deploymentconfig=mydb |
     When I execute on the pod:
       | grep | opt2 | /proc/mounts |
     Then the step should succeed
     When I run the :set_volume client command with:
       | resource | dc/mydb  |
       | action   | --remove |
       | name     | v1       |
-      | confirm  |          |
+      | confirm  | false    |
```
liangxia, line 79: I just not sure why the initial version adds confirm while it defaults to false,
--confirm=false: If true, confirm that you really want to remove multiple volumes

pruan-rht: @liangxia we can remove it.

pruan-rht: @liangxia ok, I just remove that line completely.

liangxia:
LGTM

PR #957, opened 2020-04-16
Align to standard aws credential file format

wjiangjay:
Make framework can use standard credentials file like following:
```
[default]
aws_access_key_id = xxxx
aws_secret_access_key = xxx
```

And then the creds can be shared with other AWS SDKs.


corehello:
@akostadinov @pruan-rht help review this, thx

pruan-rht:
BTW, since the .awscred file is in the ini file format, we could've use the gem `inifile` to parse it...the downside is that it requires another gem which is probably an overkill for this fix.

PR #958, opened 2020-04-16
Add rules for OCP-27590

xiaocwan:
@yapei Plz review, thanks!
Will add rules on 4.4 as well if you look good.

### DIFF
--- a/lib/rules/web/admin_console/4.5/goto.xyaml
+++ b/lib/rules/web/admin_console/4.5/goto.xyaml
@@ -1,6 +1,7 @@
 # projects
 goto_projects_list_page:
   url: k8s/cluster/projects
+  action: wait_box_loaded
 goto_project_status:
   url: status/ns/<project>
 goto_resource_page_under_default_project:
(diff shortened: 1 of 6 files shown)


lib/rules/web/admin_console/4.5/goto.xyaml
```diff
@@ -88,7 +89,7 @@ goto_deployment_page:
   url: k8s/ns/<project_name>/deployments
 goto_one_deployment_page:
   url: k8s/ns/<project_name>/deployments/<deploy_name>
-
+  action: wait_box_loaded
```
xiaocwan, line 92: Sometimes env is very slow to render the content, need a wait here just after goto. Refer to [the failed log](https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/84838/console) without this rule added here. Failed in next step Click action.

lib/rules/web/admin_console/4.5/goto.xyaml
```diff
@@ -1,6 +1,7 @@
 # projects
 goto_projects_list_page:
   url: k8s/cluster/projects
+  action: wait_box_loaded
```
xiaocwan, line 4: This is not related to the current case, but better to add it to sync with 4.4.

PR #963, opened 2020-04-17
Fix an upgrade scenario

liangxia:
/cc @geliu2016 @kasturinarra @pruan-rht


geliu2016:
LGTM, thanks @liangxia

PR #968, opened 2020-04-17
Wait default-route appearing for image-registry svc

xiuwang:
After patch to enable default route, the-route need times to show up depend on the cluster performance.

### DIFF
--- a/features/step_definitions/registry.rb
+++ b/features/step_definitions/registry.rb
@@ -375,7 +375,8 @@ end
 # Generate default route
 Given /^I enable image-registry default route$/ do
   ensure_admin_tagged
-#  ensure_destructive_tagged
+  org_proj_name = project(generate: false).name rescue nil
+
   step 'I run the :patch admin command with:', table(%{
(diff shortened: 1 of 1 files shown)


wewang58:
LGTM

features/step_definitions/registry.rb
```diff
@@ -375,21 +375,24 @@
 # Generate default route
 Given /^I enable image-registry default route$/ do
   ensure_admin_tagged
-#  ensure_destructive_tagged
```
liangxia, line 378: Changing registry should be destructive.

xiuwang: @liangxia just enable route, it doesn't infect others.

xiuwang: @liangxia if we add `ensure_destructive_tagged`, then we need add destructive label for all cases which need this step. But some ones are actually not destructive cases

xiuwang:
@liangxia Could you help review this change again, thx

PR #971, opened 2020-04-17
special step to obtain test data files

akostadinov:
@barleyer @xingxingxia @liangxia @pruan-rht


akostadinov:
Merging immediately to avoid merge conflicts and fix many scenarios. Tested locally.

pruan-rht:
looks reasonable.

PR #976, opened 2020-04-21
Update: Fix OCP-13386/13387/13388

duanwei33:
Update to fix OCP-13386/13387/13388
1. Remove default sc to avoid impacting other cases.
2. Remove datastore, case failed for vsphere6.7 env.

[Pass log ](https://privatebin-it-iso.int.open.paas.redhat.com/?7018c3bad86bd8ce#D+mHtoGjIoiUCG+2KIfi0xLblnxfkMLjuZjz4pNeIgI=)


duanwei33:
@qinpingli Could you also have a look?

PR #977, opened 2020-04-21
rules added for OCP-29198

yapei:
@yanpzhan  @xiaocwan  Can you help review?


lib/rules/web/admin_console/4.5/common_ui_elements.xyaml
```diff
@@ -442,7 +442,7 @@ wait_box_loaded:
   element:
     selector:
       css: .loading-box__loaded
-    timeout: 20
+    timeout: 120
```
yapei, line 445: since `wait_box_loaded` is reused in many places, we need more time on operator hub page to wait for all operators loaded

PR #978, opened 2020-04-21
Auto test OCP-12930

weliang1:
Auto test OCP-12930

Test log:
https://privatebin-it-iso.int.open.paas.redhat.com/?054484217cd21488#l7m4xqz5l6bHY7M0TQ3+nYv8EpMZJ05zH+O1KnOqCqM=

### DIFF
--- a/features/networking/multicast.feature
+++ b/features/networking/multicast.feature
@@ -138,3 +138,135 @@ Feature: testing multicast scenarios
       | <%= cb.pod1ip %> : joined (S,G) = (*, 232.43.211.234), pinging |
       | <%= cb.pod1ip %> : multicast, xmt/rcv/%loss = 5/0/100%         |
 
+    # @author weliang@redhat.com
+  # @case_id OCP-12930
+  @admin
+  Scenario: Same multicast groups can be created in multiple tenant
+    Given I have a project
+    And evaluation of `project.name` is stored in the :proj1 clipboard
+    When I run the :create client command with:
+      | f | <%= BushSlicer::HOME %>/testdata/networking/multicast-rc.json |
+    Then the step should succeed
+    Given 3 pods become ready with labels:
+      | name=mcast-pods |
+    And evaluation of `pod(0).ip` is stored in the :proj1pod1ip clipboard
+    And evaluation of `pod(0).name` is stored in the :proj1pod1 clipboard
+    And evaluation of `pod(1).ip` is stored in the :proj1pod2ip clipboard
+    And evaluation of `pod(1).name` is stored in the :proj1pod2 clipboard
+    And evaluation of `pod(2).ip` is stored in the :proj1pod3ip clipboard
+    And evaluation of `pod(2).name` is stored in the :proj1pod3 clipboard
+    # Enable multicast in proj1
+    When I run the :annotate admin command with:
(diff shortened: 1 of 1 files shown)


features/networking/multicast.feature
```diff
@@ -138,3 +138,135 @@ Feature: testing multicast scenarios
...
+  @admin
+  Scenario: Same multicast groups can be created in multiple tenant
+    Given I have a project
+    And evaluation of `project.name` is stored in the :proj1 clipboard
+    When I run the :create client command with:
+      | f | <%= BushSlicer::HOME %>/testdata/networking/multicast-rc.json |
+    Then the step should succeed
+    Given 3 pods become ready with labels:
+      | name=mcast-pods |
+    And evaluation of `pod(0).ip` is stored in the :proj1pod1ip clipboard
```
pruan-rht, line 152: @weliang1 how about just save the pod object itself instead of spawning a bunch of clipboards?  This will affect how you call the `cb` though.

features/networking/multicast.feature
```diff
@@ -138,3 +138,135 @@ Feature: testing multicast scenarios
...
+      | exec_command_arg | omping -m 239.255.254.24 -c 5 -T 10 <%= cb.proj2pod1ip %> <%= cb.proj2pod2ip %> <%= cb.proj2pod3ip %> > /tmp/proj2pod3.log |
+    Then the step should succeed
+
+    # Ensure proj2pod3 interface join to the multicast group 239.255.254.24
+    When I execute on the "<%= cb.proj2pod3 %>" pod:
+      | netstat | -ng |
+    Then the step should succeed
+    And the output should match:
+      | eth0\s+1\s+239.255.254.24 |
+    Given 10 seconds have passed
```
pruan-rht, line 263: @weliang1 can we try not to have hard-coded sleep?

weliang1: @pruan-rht Fix to use pulling 10 seconds. Thanks!

features/networking/multicast.feature
weliang1: @pruan-rht Good suggestion, just wonder could you point me any existing scripts which use the way you mentioned above? So I can do the same way in this script. Thanks!

weliang1: @pruan-rht Could you point me any script example using the way you mentioned above? Thanks!

pruan-rht: @weliang1 `And evaluation of `pod.first` is stored in the :proj1_pod clipboard`

weliang1: @pruan-rht  The way of pod(0).ip, pod(0).name are used in existing multicast automation cases : https://github.com/openshift/verification-tests/blob/master/features/networking/multicast.feature#L17
To keep the multicast scripting consistent and easy to debug in the future, we can use same way as above scripts used, how do you think?

rbbratta: I guess one other reason to store the whole pod is that we might want to start using `pod.ip_url` or other pod methods for automagic IPv6 support.

weliang1: @rbbratta We need write the new test cases for ipv6 multicast testing. I suggest to keep same scripting style as our existing multicast cases.

PR #983, opened 2020-04-23
always use multus image as oc debug image

rbbratta:

so we can use can existing image, e.g. multus image or any other

Add the `OPENSHIFT_ENV_OCP_HOST_DEBUG_IMAGE` option to disable
using the multus image by default

### DIFF
--- a/lib/host.rb
+++ b/lib/host.rb
@@ -857,6 +857,7 @@ module BushSlicer
       exec_opts = {}
 
       # override image, arg order matters, image needs to go before --
+      # needed until https://bugzilla.redhat.com/show_bug.cgi?id=1728135 is fixed
       unless node.env.opts[:host_debug_image]
            mpods = Pod.get_labeled("app=multus", user: node.env.admin, project: Project.new(name: "openshift-multus", env: node.env), quiet: true)
            exec_opts[:image] = mpods.first.container(name: "kube-multus").spec.image
--- a/lib/host.rb
+++ b/lib/host.rb
@@ -854,20 +854,31 @@ module BushSlicer
         commands = ["chroot", "/host/", "bash", "-c", commands_to_string(commands)]
       end
 
+      exec_opts = {}
+
+      # override image, arg order matters, image needs to go before --
+      unless node.env.opts[:host_debug_image]
+           mpods = Pod.get_labeled("app=multus", user: node.env.admin, project: Project.new(name: "openshift-multus", env: node.env), quiet: true)
+           exec_opts[:image] = mpods.first.container(name: "kube-multus").spec.image
+      end
+
+      exec_opts.merge!(
+          {
+              resource: "node/#{node.name}",
+              n: service_project.name,
+              oc_opts_end: "",
+              exec_command_arg: commands,
+              _stdin: opts[:stdin],
+              _stdout: opts[:stdout]
+          }
+      )
+
       @exec_lock.synchronize {
         # note this will block until timeout if command does not exist remotely
         # TODO: check debug pod status in the background to avoid freeze (WRKLDS-99)
         # note2: exit status is always 0 (WRKLDS-98)
         # note3: stdin and stderr come together (WRKLDS-110)
-        node.env.admin.cli_exec(
-          :debug,
-          resource: "node/#{node.name}",
-          n: service_project.name,
-          oc_opts_end: "",
-          exec_command_arg: commands,
-          _stdin: opts[:stdin],
-          _stdout: opts[:stdout]
-        )
+        node.env.admin.cli_exec(:debug, **exec_opts)
       }
     end
 
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -325,7 +325,7 @@
      :output: --output=<value>
      :template: --template=<value>
 :debug:
-  :cmd: oc debug --image=quay.io/openshifttest/storage@sha256:a05b96d373be86f46e76817487027a7f5b8b5f87c0ac18a246b018df11529b40 <global_options>
+  :cmd: oc debug <global_options>
   :options:
      :c: -c <value>
      :dry_run: --dry-run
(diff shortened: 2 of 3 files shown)


rbbratta:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/85949/console

rbbratta:
should fix disconnected cluster tests.

anuragthehatter:
@pruan-rht @akostadinov Can we merge this? This not only fixes disconnected cluster tests but also helpful in case we have quay outages. We don't want to be bottle necked by just quay.io/redhat-support-tools image during ```oc debug``` operation.Thanks @rbbratta for investigating this.

lib/host.rb
```diff
@@ -854,20 +854,36 @@ def exec_raw(*commands, **opts)
...
+        if @result[:success]
+          # we could cache this but then we would have to re-get after upgrade
+          opts[:image] = @result[:response].strip
+        end
+      end
+
+      exec_opts = {}
+      # override image, arg order matters, image needs to go before --
+      if opts[:image]
+        exec_opts[:image] = opts[:image]
```
pruan-rht, line 868: @rbbratta could be replaced with `exec_opts[:image] ||= opts[:image]`

rbbratta: very cool, I was not aware of `||=`, but I don't think it matches in this case.  `a ||=b`  seems to roughly approximate `a || a = b`, but we want `b || a = b`

```
opts[:image] || exec_opts[:image] = opts[:image]
```

Had to read this to figure it out:
http://www.rubyinside.com/what-rubys-double-pipe-or-equals-really-does-5488.html

lib/host.rb
```diff
@@ -854,20 +854,36 @@ def exec_raw(*commands, **opts)
         commands = ["chroot", "/host/", "bash", "-c", commands_to_string(commands)]
       end
 
+      unless node.env.opts[:host_debug_image] == "support-tools"
```
akostadinov, line 857: `if a !+ b` will read better here. `unless` is harder to understand.

In either case I would assume you want to set this multus image only in case `:host_debug_image` is not set to anything.

So perhaps here we should have:
```
unless node.env.opts[:host_debug_image]
```

lib/host.rb
```diff
@@ -854,20 +854,36 @@ def exec_raw(*commands, **opts)
...
+        @result = node.env.admin.cli_exec(:get, n:"openshift-multus", resource: "pod", l: "app=multus", output: 'jsonpath={.items[0].spec.containers[?(@.name=="kube-multus")].image}')
+        if @result[:success]
+          # we could cache this but then we would have to re-get after upgrade
+          opts[:image] = @result[:response].strip
+        end
```
akostadinov, line 862: Maybe better use our object model here. Something like
```
mpods = Pod.get_labeled("app=multus", user: node.env.admin, project: Project.new(name: "openshift-multus", env: node.env))
opts[:image] = mpods.first.container(name: "kube-multus").spec.image.name
```

But not sure which one is more readable for others. wdyt?

akostadinov: The opt name should be `debug_image` or something like that though. `image` sounds too generic for the `Host` object.

lib/host.rb
akostadinov: Or to avoid extra calls when we already have it
```
unless self[:debug_image] || node.env.opts[:host_debug_image]
```

akostadinov: Wait, if we have `node.env.opts[:host_debug_image]`, then `self[:debug_image]` should be set to that value.

akostadinov: Unless I misunderstood your intention.

rbbratta: If we are using the multus image by default I wasn't sure if we wanted to allow the user to override the image to anything so I didn't want to use environment variable to set any random debug image, only to default back to the default `oc debug` image, whatever version of `support-tools` is hard-coded in the `oc` binary.

the plan was:

| HOST_DEBUG_IMAGE | image |
| --- | --- |
| nil | use multus |
| non-nil | use default `oc debug` image, current behavior  |

One problem with using the `multus` image is that is may disappear after an upgrade when a new multus image is loaded.  We could cache the multus value and try to invalidate the cache after an upgrade, but that seems to complex, so I decided to just re-query the current multus image each time we run.

lib/host.rb
rbbratta: I tried using `Pod.get_labeled` but couldn't figure out the `Project` or `user` setting and gave up.  I'll try again.

will also try `opts[:debug_image]`

rbbratta: err I guess we don't need `opts[:debug_image]` that was from an old patch where I wanted to set the debug image using a optional step arg passed all the way down to `Host.exec_raw`

I'll remove the use of `opts`.

rbbratta:
stress-tested  http://pastebin.test.redhat.com/868454

anuragthehatter:
Apparently PR #1030 has taken care of it hard coded way but i suggest we put the multus image option as well in ```else``` with PR #1030 code. I can see it beneficial in case we of quay outages which seems to be happening frequently. WDYGT

akostadinov:
/lgtm
agreed
also this allows `oc debug` tests to function same as in the past and test the cluster default behavior. Which IMO is important to make sure debug functionality works OOB.
Thank you.

akostadinov:
Honestly I think we need to push for this `oc debug` [issue](https://bugzilla.redhat.com/show_bug.cgi?id=1728135) fixed instead of having such workarounds. But this is much better than hardcoding the thing in the rules file. And at least `oc debug` test cases should fail to allow us track it's progress.

akostadinov:
We should consider some form of caching btw. I wonder if we can tag it somewhere so it is never removed?

rbbratta:
from the BZ 
> If you don't specify any image, we default to imagestream tools:latest being present in openshift namespace.
Only if that fails we reach out for support-tools from redhat's registry.

This suggests we could just tag the multus image as `tools:latest` and let `oc debug` do the work itself.

PR #986, opened 2020-04-23
Add web rules for OCP-24307

xiaocwan:
Add for both 4.4 and 4.5
@yanpzhan Please review, thanks!


lib/rules/web/admin_console/4.4/goto.xyaml
```diff
@@ -225,6 +227,9 @@ goto_catalog_source_page:
 # quota
 goto_quotas_page:
   url: k8s/ns/<project_name>/resourcequotas
+goto_one_quotas_page:
```
yapei, line 230: `goto_one_quota_page`

PR #993, opened 2020-04-24
Fix OCP-19675 rules on 4.5 changed

xiaocwan:
4.5 updated filter bar for each resource list page using PF4, it came from this User Story: https://issues.redhat.com/browse/CONSOLE-2134
Now the step needs an extra Click action to check the filter categories text.
PASSed on latest 4.5 with version 4.5.0-0.nightly-2020-04-25-170442
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/87060/console 
@yanpzhan Please help review. Thanks

### DIFF
--- a/lib/rules/web/admin_console/4.3/filters.xyaml
+++ b/lib/rules/web/admin_console/4.3/filters.xyaml
@@ -98,3 +98,6 @@ check_item_in_table:
   elements:
   - selector:
       xpath: //tr[@data-test-rows='resource-row']//*[text()='<item>']
+
+# Below is PF4 filter bar related rules (since OCP4.5)
+open_filter_dropdown_list: {}
--- a/lib/rules/web/admin_console/4.4/filters.xyaml
+++ b/lib/rules/web/admin_console/4.4/filters.xyaml
@@ -97,3 +97,6 @@ clear_all_filters:
   - selector:
       xpath: //div/button[starts-with(text(),'Clear')]
     op: click
+
+# Below is PF4 filter bar related rules (since OCP4.5)
+open_filter_dropdown_list: {}
--- a/lib/rules/web/admin_console/4.5/filters.xyaml
+++ b/lib/rules/web/admin_console/4.5/filters.xyaml
@@ -107,3 +107,10 @@ clear_all_filters:
   - selector:
       xpath: //div/button[starts-with(text(),'Clear')]
     op: click
+
+# Below is PF4 filter bar related rules (since OCP4.5)
+open_filter_dropdown_list:
+  elements:
+  - selector:
+      xpath: //button[@class='pf-c-dropdown__toggle' and @data-test-id='filter-dropdown-toggle']
+    op: click


lib/rules/web/admin_console/4.3/filters.xyaml
```diff
@@ -98,3 +98,6 @@ check_item_in_table:
   elements:
   - selector:
       xpath: //tr[@data-test-rows='resource-row']//*[text()='<item>']
+
+# Below is PF4 filter bar related rules (since OCP4.5)
+open_filter_dropdown_list: {}
```
liangxia, line 103: The comments stated that the rules exist since 4.5, but we are adding it to 4.3 ? Which one should be believe ?

xiaocwan: @liangxia both 4.3 and 4.4 are same (should use empty rule for open_filter_dropdown_list), 4.5 needs this rule to do an extra Click, the scenario apply to cluster >=4.3

xiaocwan: @liangxia So are we good to merge this PR? 
Thanks in advance!

liangxia: Sounds the comment should use 'since OCP4.3'

xiaocwan: @liangxia No, this rule does not needed at all for 4.3 and 4.4, since 4.5 needs an extra click, we always use empty same-named rule for previous versions. Otherwise, although 4.5 could PASS, 4.3 and 4.4 will failed if there is no same-named rule at all. If you search the empty rule using `{}` , you can find bunches of previous examples. Do I make it clear?

liangxia: If the rule does not needed at all for 4.3 and 4.4, then we should remove it from 4.3 and 4.4.
If it is needed, that mean it starts from OCP4.3.
I'm leaving it to @yapei to decide. I can merge once got her ack.

yanpzhan: @liangxia the rules added for 4.3 and 4.4 is empty, since the scenario script is used for all cluster versions, but page element would change during different versions, so need these empty rules for some versions to keep compatible, otherwise, the scenario will fail. I suggest to remove the comment "# Below is PF4 filter bar related rules (since OCP4.5)", which causes misunderstanding.

xiaocwan: ok. Will ask @yapei

xiaocwan:
@liangxia Since case author intend to remove these steps directly from feature: https://github.com/openshift/cucushift/pull/7638#discussion_r413600409

I think this fix is Low importance to this case and could be close directly, will submit another patch to feature file to remove the step accordingly

PR #997, opened 2020-04-26
fix IPI abort test

sunzhaohua2:
In IPI env, if there is one machineset with replicas=0 the test will be aborted. This PR will fix this issue. As long as there is one machineset with replicas > 0, the test will continue.  @jhou @miyadav Please help to take a look.

### DIFF
--- a/features/step_definitions/machine.rb
+++ b/features/step_definitions/machine.rb
@@ -8,8 +8,13 @@ Given(/^I have an IPI deployment$/) do
   end
 
   machine_sets.each do | machine_set |
-    unless machine_set.ready?[:success]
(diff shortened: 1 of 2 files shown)


jhou1:
Thanks for improving this, I noticed that test fail a lot because cluster gets deployed having some machinesets with replica=0. Maybe installer team made some change recently.

We have one more fix needed here. Can you please also fix step_definitions/machine_set.rb so that we do not pick the machineset with replicas=0?

```
  ensure_admin_tagged
    select { |ms| ms.available_replicas >= 1 }
  cache_resources *machine_sets.shuffle
end
```

jhou1:
Merging because we need this in regression tests asap.

PR #1002, opened 2020-04-27
fix intermittent test failure

lihongyan1:
fix intermittent test failure when integration run
https://issues.redhat.com/browse/OCPQE-231
https://issues.redhat.com/browse/OCPQE-232


lihongyan1:
@juzhao @yapei Please help review

juzhao:
Can the timeout help on the test failure?

lihongyan1:
> Can the timeout help on the test failure?

I suppose so, they are all intermittent issue and occur when ENV is slow in the integration run. By default it check element and wait 10 seconds, if can't find, case will fail. From the screen shot, we can know element is not loaded or loaded after 10 seconds timeout

lihongyan1:
The change is included in PR https://github.com/openshift/verification-tests/pull/1018

PR #1003, opened 2020-04-27
Auto OCP-29477

SSshahan:
@yapei  could you help to review?  Thanks

### DIFF
--- a/lib/rules/web/admin_console/4.5/operator_hub.xyaml
+++ b/lib/rules/web/admin_console/4.5/operator_hub.xyaml
@@ -279,8 +279,4 @@ switch_to_yaml_view:
     content: Create by manually entering YAML or JSON definitions
   action: click_radio_input
   action: check_page_contains
-click_checkbox_from_providertype:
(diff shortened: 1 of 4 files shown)


lib/rules/web/admin_console/4.5/operator_hub.xyaml
```diff
@@ -266,3 +266,9 @@ open_operator_modal:
     keyword: <operator_name>
   action: filter_by_keyword_on_catalog_page
   action: open_first_card_in_overlay
+click_checkbox_from_providertype:
+  params:
+    filter_type_value: providerType
+  action: wait_filter_section_loaded
+  action: choose_one_checkbox
+
```
yanpzhan, line 274: There is an existing rule: click_checkbox_from_provider_type, could you check if it's the same function with "click_checkbox_from_providertype"

liangxia: What's the change for ?

PR #1005, opened 2020-04-28
add rules for OCP-21245

yapei:
@xiaocwan  Can you help review? Thanks

### DIFF
--- a/lib/rules/web/admin_console/4.2/environments.xyaml
+++ b/lib/rules/web/admin_console/4.2/environments.xyaml
@@ -85,4 +85,37 @@ remove_env_vars:
   - selector: *env_var_delete_icon
     type: js
     op: click
-
+check_env_editor_disabled:
+  elements:
+  - selector:
+      xpath: //div[contains(@class,'pairs-list__name-field')]/input[@disabled]
+  - selector:
+      xpath: //div[contains(@class,'pairs-list__value-field')]/input[@disabled]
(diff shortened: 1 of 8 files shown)


lib/rules/web/admin_console/4.2/environments.xyaml
```diff
@@ -85,4 +85,24 @@ remove_env_vars:
...
-
+check_env_editor_disabled:
+  elements:
+  - selector:
+      xpath: //div[contains(@class,'pairs-list__name-field')]/input[@disabled]
+  - selector:
+      xpath: //div[contains(@class,'pairs-list__value-field')]/input[@disabled]
+  - selector:
+      xpath: //div[contains(@class,'pairs-list__value-ro-field')]/input[@disabled]
+check_readonly_message:
```
xiaocwan, line 96: could you help to add this rule as a general rule in common?

```
check_alert_warning_message:
  elements:
  - selector:
      xpath: //*[contains(@class, 'pf-c-alert__title') and contains(text(), '<title>')]
  - selector:
      xpath: //*[contains(@class, 'pf-c-alert__description')]//*[contains(., '<description>')]
```

One of my automating case is adding that rule

xiaocwan: Only if you think this rule is helpful and clear for this case, otherwise I'll add it in my PR

PR #1010, opened 2020-04-28
Making more readable clone machineset step

miyadav:
made the change using - find /home/miyadav/verification/verification-tests -type f -exec sed -i 's/machineset named/machineset and name it/g' {} \;

Ran one of the cases involving changes , it works fine 
            instanceId: i-03e5945954debaed4
            instanceState: running
      kind: List
      metadata:
        resourceVersion: ""
        selfLink: ""
      
      [06:14:26] INFO> Exit Status: 0
    Then the expression should be true> cb.machines.select{|m|m.instance_state == m.annotation_instance_state}.count == cb.machines.count     # features/step_definitions/common.rb:133
      [06:14:26] INFO> === After Scenario: Verify all machine instance-state should be consistent with their providerStats.instanceState ===
      [06:14:26] INFO> Shell Commands: rm -r -f -- /root/workdir/9d63297830b3-
      
      [06:14:26] INFO> Exit Status: 0
      [06:14:26] INFO> === End After Scenario: Verify all machine instance-state should be consistent with their providerStats.instanceState ===

1 scenario (1 passed)
3 steps (3 passed)
0m11.560s
[06:14:26] INFO> === At Exit ===

### DIFF
--- a/features/machine/autoscaler.feature
+++ b/features/machine/autoscaler.feature
@@ -11,7 +11,7 @@ Feature: Cluster Autoscaler Tests
     Given I store the number of machines in the :num_to_restore clipboard
     And admin ensures node number is restored to "<%= cb.num_to_restore %>" after scenario
 
-    Given I clone a machineset named "machineset-clone-28108"
+    Given I clone a machineset and name it "machineset-clone-28108"
 
     # Create clusterautoscaler
     Given I use the "openshift-machine-api" project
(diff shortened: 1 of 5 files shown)


jhou1:
@miyadav thank you for taking this, would you help on the cucushift repo as well?

jhou1:
And squash the commits, thanks

miyadav:
@jhou1  ,done

liangxia:
> would you help on the cucushift repo as well?(after this is merged)

Instead of doing them separately, I would suggest change them at the same time, and we can merge them in batch. Or else those scenario in cucushift using this step will fail as the step definition has changed there.

jhou1:
Good point! Yes we should merge them together when they are both ready.

miyadav:
@jhou1 ,@liangxia - PTAL

PR #1015, opened 2020-04-28
update ingress source OCP-18789

quarterpin:
The test case gets executed properly with this change applied. The successful v3 runner task link is as below:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/87646/console

@lihongan - Request for a peer review.


lihongan:
The port `8080` is used in OCP 3.9 and before,  it was replaced by `27017` after fixing a bug in 3.10. So we can replace the test data and use it directly in master branch for now.

lihongan:
@pruan-rht PTAL, thanks

PR #1017, opened 2020-04-28
step to fetch ERB from URL is invalid due to change of location of th…

pruan-rht:
…e testdata.  Need to merge with the corresponding pr from cucushift repo

### DIFF
--- a/features/step_definitions/cli.rb
+++ b/features/step_definitions/cli.rb
@@ -111,7 +111,7 @@ end
 # 1. download file from URL
 # 2. load it as an ERB file with the cucumber scenario variables binding
 # 3. runs `oc create` command over the resulting file
-When /^I run oc create( as admin)? over ERB test file: #{QUOTED}$/ do |admin, file_path|
+When /^I run oc create( as admin)? over ERB test file: (.*)$/ do |admin, file_path|
   step %Q|I obtain test data file "#{file_path}"|
   file_path = cb.test_file
   # overwrite with ERB loaded content
--- a/features/step_definitions/cli.rb
+++ b/features/step_definitions/cli.rb
@@ -111,17 +111,17 @@ end
 # 1. download file from URL
 # 2. load it as an ERB file with the cucumber scenario variables binding
 # 3. runs `oc create` command over the resulting file
-When /^I run oc create( as admin)? over ERB URL: #{HTTP_URL}$/ do |admin, url|
-  step %Q|I download a file from "#{url}"|
-
+When /^I run oc create( as admin)? over ERB test file: #{QUOTED}$/ do |admin, file_path|
+  step %Q|I obtain test data file "#{file_path}"|
+  file_path = cb.test_file
   # overwrite with ERB loaded content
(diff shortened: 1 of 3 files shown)


pruan-rht:
dependent PR https://github.com/openshift/cucushift/pull/7661

features/step_definitions/helper_services.rb
```diff
@@ -708,7 +708,7 @@
   })
   step 'the step should succeed'
   # cluster-capacity as a target pod
-  step "I run oc create over ERB URL: #{ENV['BUSHSLICER_HOME']}/testdata/infrastructure/cluster-capacity/cluster-capacity-pod.yaml"
+  step 'I run oc create over ERB test file: "/infrastructure/cluster-capacity/cluster-capacity-pod.yaml"'
```
liangxia, line 711: Perhaps use `#{BushSlicer::HOME}/testdata/infrastructure/cluster-capacity/cluster-capacity-pod.yaml` ?

pruan-rht: @liangxia, we don't need to do that anymore, within the step, we are calling the step `I obtain test data file`

liangxia: In case we need to update the location of BushSlicer::HOME in the future, we can update them in batch. Or else, we can miss those ones.

pruan-rht: @liangxia the [`I obtain test data file` step](https://github.com/openshift/verification-tests/blob/master/features/step_definitions/file.rb#L3-L4) is already doing that, so if we were to add what you suggested, it will lead to invalid path

pruan-rht: @liangxia this PR should help with a cleared error message for debugging https://github.com/openshift/verification-tests/pull/1025

akostadinov:
/lgtm
thank you!

features/step_definitions/helper_services.rb
akostadinov: To clarify `test file` here means that a test file from a standard set of test files. If we some day decide to store test files in some other repo again, we can change in one place and download the files instead of copying them.

features/step_definitions/cli.rb
```diff
@@ -111,17 +111,17 @@
 # 1. download file from URL
 # 2. load it as an ERB file with the cucumber scenario variables binding
 # 3. runs `oc create` command over the resulting file
-When /^I run oc create( as admin)? over ERB URL: #{HTTP_URL}$/ do |admin, url|
-  step %Q|I download a file from "#{url}"|
-
+When /^I run oc create( as admin)? over ERB test file: #{QUOTED}$/ do |admin, file_path|
```
akostadinov, line 114: In the old version we didn't have the quotes around the file. I think we should keep the convention. Which is use quotes if whatever thing is in the middle of step or use non-quoted in case that is last argument. Here we have `step name: param` this we do not quote the `param`. Instead of quoted, pls use `(.*)`.

pruan-rht: ok, updated... PTAL

PR #1023, opened 2020-04-29
update routing images to use quay

lihongan:
@xiuwang @quarterpin Please take a look. Thanks.


quarterpin:
> @xiuwang @quarterpin Please take a look. Thanks.

Since we have quay.io moving test pod images into it will be good. 

/lgtm

PR #1025, opened 2020-04-29
Print out all of the actual tried paths

pruan-rht:
if the user input the correct path like `<%= BushSlicer::HOME %>/features/tierN/testdata/daemon/daemonset.yaml` into the method, the error message currently prints the correct path which is misleading.  We should print the actual paths that were tried instead.   

>     Given I obtain test data file "<%= BushSlicer::HOME %>/features/tierN/testdata/templates/OCP-12856/push-generic-build-args.json" # features/step_definitions/file.rb:1
>       could not find test file in '["/Users/pruan/workspace/BushSlicer/testdata//Users/pruan/workspace/BushSlicer/features/tierN/testdata/templates/OCP-12856/push-generic-build-args.json", "/Users/pruan/workspace/BushSlicer/features/tierN/testdata//Users/pruan/workspace/BushSlicer/features/tierN/testdata/templates/OCP-12856/push-generic-build-args.json"]' (RuntimeError)

### DIFF
--- a/features/step_definitions/file.rb
+++ b/features/step_definitions/file.rb
@@ -3,15 +3,16 @@ Given /^I obtain test data file #{QUOTED}$/ do |path|
     "#{BushSlicer::HOME}/testdata",
     "#{BushSlicer::HOME}/features/tierN/testdata",
   ]
-
+  tried_paths = []
   found = accepted_roots.any? do |root|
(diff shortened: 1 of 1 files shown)


akostadinov:
Got you.
/lgtm

PR #1029, opened 2020-04-30
add permanent image as default value

zhouying7780:
@qinpingli @liangxia  Could you please help review ? thanks.

### DIFF
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -285,7 +285,7 @@
      :output: --output=<value>
      :template: --template=<value>
 :debug:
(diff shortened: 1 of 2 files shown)


anuragthehatter:
@zhouying7780 Why storage image was used instead of support tools (post uploading on quay)? I believe we should strictly use support tools image as a first priority and others as second.

akostadinov:
I personally don't like this change. It hides the issue from test suite, then when I want to do manual steps, I'll have to suffer.

akostadinov:
And it doesn't test the default tools image for clusters where it is accessible. So we are no longer making sure default tools image continues to work.

PR #1033, opened 2020-04-30
Moving Cleanup steps to the appropriate locations

anuragthehatter:
Fixing few code smells here which are causing "Test execution finished prematurely" errors in nightly runs. Small change and can be merged asap to avoid futrther issues @pruan-rht 
cc @weliang1 @rbbratta @zhaozhanqi @huiran0826 

When we create a bridge type net-attach-def, a bridge interface is created on the node only when a pod using that def gets created. The bridge interface doesn't gets created immediately post net-attach-def creation. So the bridge cleanup should definitely be after pod creation step.

Right now ,such tests are throwing premature failure errors due to the fact that say, a net-attach def is successfully created but test fails at pod creation step. However the cleanup is defined post  net-attach-def. So, as we never reached to a pod creation step and hence there is no bridge interface created, automation will throw error saying can't delete bridge interface as no such device found.

e.g. http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/04/28/21:05:49/Pods_can_communicate_each_other_with_same_vlan_tag/console.html

Re-ran logs: http://pastebin.test.redhat.com/860870

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -615,11 +615,6 @@ Feature: Multus-CNI related scenarios
       | n | <%= project.name %>                                                                                                                               |
     Then the step should succeed
     
-    #Clean-up required to erase bridge interfaces created due to above net-attach-def
-    Given I register clean-up steps:
-    """
-    the bridge interface named "mybridge" is deleted from the "<%= cb.nodes[0].name %>" node
-    """  
     #Creating first pod in vlan 100
     When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
       | ["metadata"]["name"]                                      | pod1-vlan100            |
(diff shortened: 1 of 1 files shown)


features/networking/multus.feature
```diff
@@ -711,9 +702,9 @@ Feature: Multus-CNI related scenarios
     
     #Creating first pod in vlan 100
     When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
-      | ["metadata"]["name"] | pod1-vlan100 |
-      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"]| bridgevlan100 |
-      | ["spec"]["nodeName"]                                       | <%= cb.nodes[0].name %> |
+      | ["metadata"]["name"] 					  | pod1-vlan100            |
```
pruan-rht, line 705: @anuragthehatter minor nit-pick...can you please algin the table?

anuragthehatter: Done. Thanks!

weliang1:
@anuragthehatter Suggest to move first Clean-up step from line 631 to line 648, the logic will be  do Clean-up after you create both pod1-vlan100 and pod2-vlan100 in cb.nodes[0].name.
Otherwise, LGTM

anuragthehatter:
> @anuragthehatter Suggest to move first Clean-up step from line 631 to line 648, the logic will be do Clean-up after you create both pod1-vlan100 and pod2-vlan100 in cb.nodes[0].name.
> Otherwise, LGTM

Yep agree, should be done in both scenarios. It will be good as per scenario guidelines as well. Fixed it. Thanks!

PR #1035, opened 2020-04-30
OCP-24488: debug VLAN errors by dumping info before

rbbratta:
We keep getting automation failures because of existing VLANs
before the test.

Dump the VLAN information before we run the test.
To do this we have to pick a node first, then put the pod
on that node.

Failures:

```
/sbin/bridge\ vlan\ show
port	vlan ids
cni0	 1 PVID Egress Untagged

bridge3	 1 PVID Egress Untagged

veth9056d1f1	 1 PVID Egress Untagged
```

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -13,7 +13,7 @@ Feature: Multus-CNI related scenarios
     Given I have a project
     Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml"
     When I run oc create as admin over "macvlan-bridge.yaml" replacing paths:
-      | ["metadata"]["namespace"] | <%= project.name %>                                                                                                                                                                                                                                                                    |    
+      | ["metadata"]["namespace"] | <%= project.name %>                                                                                                                                                                                                                                                                    |
       | ["spec"]["config"]        | '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "<%= cb.default_interface %>","mode": "bridge", "ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100", "rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1" } }' |
     Then the step should succeed
 
@@ -74,8 +74,8 @@ Feature: Multus-CNI related scenarios
     Given I have a project
     Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/macvlan-private.yaml"
     When I run oc create as admin over "macvlan-private.yaml" replacing paths:
-      | ["metadata"]["namespace"] | <%= project.name %> |
-      | ["spec"]["config"]| '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "<%= cb.default_interface %>","mode": "private", "ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100", "rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1" } }' |    
+      | ["metadata"]["namespace"] | <%= project.name %>                                                                                                                                                                                                                                                                     |
+      | ["spec"]["config"]        | '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "<%= cb.default_interface %>","mode": "private", "ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100", "rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1" } }' |
     Then the step should succeed
 
     # Create the first pod which consumes the macvlan custom resource
@@ -104,9 +104,9 @@ Feature: Multus-CNI related scenarios
     # Create the second pod which consumes the macvlan cr
     Given I obtain test data file "networking/multus-cni/Pods/1interface-macvlan-private.yaml"
     When I run oc create over "1interface-macvlan-private.yaml" replacing paths:
-       | ["metadata"]["name"]              | macvlan-private-secondpod   |      
-       | ["spec"]["nodeName"]              | "<%= cb.target_node %>"     |
-       | ["spec"]["containers"][0]["name"] | macvlan-private-secondpod   |
+      | ["metadata"]["name"]              | macvlan-private-secondpod |
+      | ["spec"]["nodeName"]              | "<%= cb.target_node %>"   |
+      | ["spec"]["containers"][0]["name"] | macvlan-private-secondpod |
     Then the step should succeed
     And the pod named "macvlan-private-secondpod" becomes ready
 
@@ -133,8 +133,8 @@ Feature: Multus-CNI related scenarios
     Given I have a project
     Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/macvlan-vepa.yaml"
     When I run oc create as admin over "macvlan-vepa.yaml" replacing paths:
-      | ["metadata"]["namespace"] | <%= project.name %> |
-      | ["spec"]["config"]| '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "<%= cb.default_interface %>","mode": "vepa", "ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100", "rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1" } }' |      
+      | ["metadata"]["namespace"] | <%= project.name %>                                                                                                                                                                                                                                                                  |
+      | ["spec"]["config"]        | '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "<%= cb.default_interface %>","mode": "vepa", "ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100", "rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1" } }' |
     Then the step should succeed
 
     # Create the first pod which consumes the macvlan custom resource
@@ -188,13 +188,13 @@ Feature: Multus-CNI related scenarios
     # Make sure that the multus is enabled
     Given the master version >= "4.1"
     And the multus is enabled on the cluster
-    Given the default interface on nodes is stored in the :default_interface clipboard    
+    Given the default interface on nodes is stored in the :default_interface clipboard
     And evaluation of `node.name` is stored in the :target_node clipboard
     And an 4 character random string of type :hex is stored into the :nic_name clipboard
 
     # Prepare the net link on the node which will be attached to the pod
     When I run command on the "<%= cb.target_node %>" node's sdn pod:
-	    | sh | -c | ip link add <%= cb.nic_name %> link <%= cb.default_interface %> type macvlan mode bridge |
+      | sh | -c | ip link add <%= cb.nic_name %> link <%= cb.default_interface %> type macvlan mode bridge |
     Then the step should succeed
     Given I register clean-up steps:
     """
@@ -207,7 +207,7 @@ Feature: Multus-CNI related scenarios
     Given I have a project
     Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/host-device.yaml"
     When I run oc create as admin over "host-device.yaml" replacing paths:
-      | ["metadata"]["namespace"] | <%= project.name %> |
+      | ["metadata"]["namespace"] | <%= project.name %>                                                              |
       | ["spec"]["config"]        | '{"cniVersion": "0.3.0", "type": "host-device", "device": "<%= cb.nic_name %>"}' |
     Then the step should succeed
     # Create the first pod which consumes the host-device custom resource
@@ -258,20 +258,20 @@ Feature: Multus-CNI related scenarios
     Given the master version >= "4.1"
     And the multus is enabled on the cluster
     Given the default interface on nodes is stored in the :default_interface clipboard
-    And evaluation of `node.name` is stored in the :target_node clipboard    
+    And evaluation of `node.name` is stored in the :target_node clipboard
     # Create the net-attach-def via cluster admin
     Given I have a project
     Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml"
     When I run oc create as admin over "macvlan-bridge.yaml" replacing paths:
-      | ["metadata"]["namespace"] | <%= project.name %> |
-      | ["spec"]["config"]| '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "<%= cb.default_interface %>","mode": "bridge", "ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100", "rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1" } }' |      
+      | ["metadata"]["namespace"] | <%= project.name %>                                                                                                                                                                                                                                                                    |
+      | ["spec"]["config"]        | '{ "cniVersion": "0.3.0", "type": "macvlan", "master": "<%= cb.default_interface %>","mode": "bridge", "ipam": { "type": "host-local", "subnet": "10.1.1.0/24", "rangeStart": "10.1.1.100", "rangeEnd": "10.1.1.200", "routes": [ { "dst": "0.0.0.0/0" } ], "gateway": "10.1.1.1" } }' |
     Then the step should succeed
 
     # Create the pod which consumes multiple macvlan custom resources
     Given I obtain test data file "networking/multus-cni/Pods/2interface-macvlan-macvlan.yaml"
     When I run oc create over "2interface-macvlan-macvlan.yaml" replacing paths:
       | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | macvlan-bridge, macvlan-bridge |
-      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |
+      | ["spec"]["nodeName"]                                       | "<%= cb.target_node %>"        |
     Then the step should succeed
     And a pod becomes ready with labels:
       | name=two-macvlan-pod |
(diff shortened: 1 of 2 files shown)


features/networking/multus.feature
```diff
@@ -424,10 +424,14 @@ Feature: Multus-CNI related scenarios
       | n | <%= project.name %>                                                                                                                              |
     Then the step should succeed
     #Creating no-vlan pod absorbing above net-attach-def
+    Given I store all worker nodes to the :nodes clipboard
+    And CNI vlan info is obtained on the "<%= cb.nodes[0].name %>" node
+    # force pod onto the node we already got CNI vlan information for
     When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
       | ["metadata"]["name"] | pod-novlan |
```
pruan-rht, line 431: @rbbratta while you are at it, pls align the tables in PR.  thx!

anuragthehatter:
Thanks @rbbratta for proactive on this. On the flip side, i did some dig on this today and apparently CRIO runtime also creates a bridge on each node so its evident that ```bridge vlan show``` will list this. I believe this is not an issue but this scenario needs to be adjusted to not react if it sees cni0 present. Will confirm this with dev tomorrow to make sure.

```
# cat /etc/cni/net.d/
100-crio-bridge.conf       200-loopback.conf          87-podman-bridge.conflist  
sh-4.4# cat /etc/cni/net.d/100-crio-bridge.conf 
{
    "cniVersion": "0.3.1",
    "name": "crio-bridge",
    "type": "bridge",
    "bridge": "cni0",  <<<<<<<<<<<<
    "isGateway": true,
    "ipMasq": true,
    "hairpinMode": true,
    "ipam": {
        "type": "host-local",
        "routes": [
            { "dst": "0.0.0.0/0" },
            { "dst": "xxxx:xxx::1/24" }
        ],
        "ranges": [
            [{ "subnet": "x.x.x.x/16" }],
            [{ "subnet": "xxxx:xxx::/24" }]
        ]
    }
}
```

rbbratta:
we could do a `bridge vlan show` before and after and check the delta.   Could do a hash of all the VLANs and then a set difference, or some other diff operation.

anuragthehatter:
> we could do a `bridge vlan show` before and after and check the delta. Could do a hash of all the VLANs and then a set difference, or some other diff operation.

@rbbratta Sounds good. Adding pre debug is good idea unless its dumping numerous lines but i know in this case it would be few like <10. We can definitely do that and see what we get in next catch. Meanwhile we might get more info from dev which might require to crank this case a bit. Thank you, and can you also post logs to make sure it doesn't break anywhere?

weliang1:
> > we could do a `bridge vlan show` before and after and check the delta. Could do a hash of all the VLANs and then a set difference, or some other diff operation.
> 
> @rbbratta Sounds good. Adding pre debug is good idea unless its dumping numerous lines but i know in this case it would be few like <10. We can definitely do that and see what we get in next catch. Meanwhile we might get more info from dev which might require to crank this case a bit. Thank you, and can you also post logs to make sure it doesn't break anywhere?

Agree, dumping net information from debug steps can help QE to analysis the root cause of failures

features/networking/multus.feature
rbbratta: fixed.   

there were lots of whitespace problems.  we should do a `sed -i -e 's/[[:space:]]\+$//g'` or similar to clean all the trailing whitespace.

is it possible to do a linting step with the PR robot?  I've done travis CI with github once, a while ago.

rbbratta:
because we need to pick a node this PR triggers a `oc get nodes` which adds ~2100 lines of log output.  I'd have to change the env.nodes stuff to add `:quiet` to change that.

`Given I store the ready and schedulable workers in the :nodes clipboard` step triggers to `oc get nodes`

anuragthehatter:
> because we need to pick a node this PR triggers a `oc get nodes` which adds ~2100 lines of log output. I'd have to change the env.nodes stuff to add `:quiet` to change that.
> 
> `Given I store the ready and schedulable workers in the :nodes clipboard` step triggers the `oc get nodes`

@rbbratta Confirmed by CRIO team (and manually tested), ```cni0``` will be created when podman is run directly so its likely that non-sdn tests ran before this one caused cni0 creation on the node and hence we see 3 interfaces. I would suggest you to also change the logic in this same PR for output to ```at least``` contains

```bridge3	 1 PVID Egress Untagged```

```veth9056d1f1	 1 PVID Egress Untagged```

**Comment by anuragthehatter** [comment]
Also i don't think we now need any explicitly debugging to be added post knowing the CRIO reason above so may be if we still want to dump info, we can just dump ```bridge vlan show``` prior. WDYT

**Comment by anuragthehatter** [comment]
> > because we need to pick a node this PR triggers a `oc get nodes` which adds ~2100 lines of log output. I'd have to change the env.nodes stuff to add `:quiet` to change that.
> > `Given I store the ready and schedulable workers in the :nodes clipboard` step triggers the `oc get nodes`
> 
> @rbbratta Confirmed by CRIO team (and manually tested), `cni0` will be created when podman is run directly so its likely that non-sdn tests ran before this one caused cni0 creation on the node and hence we see 3 interfaces. I would suggest you to also change the logic in this same PR for output to `at least` contains
> 
> `bridge3 1 PVID Egress Untagged`
> 
> `veth9056d1f1 1 PVID Egress Untagged`
> 
> So may be just grep bridge3 and veth with 2 times 1 PVIDs

@rbbratta May be we can merge this soon. Keep popping up..heh. Thanks

**Comment by rbbratta** [comment]
rebased 

http://pastebin.test.redhat.com/873139

**Comment by rbbratta** [comment]
rebased
http://pastebin.test.redhat.com/878644

**Comment by rbbratta** [comment]
@anuragthehatter needs another LGTM after rebase

File: features/networking/multus.feature [hunk 2]
```diff
@@ -500,33 +504,33 @@ Feature: Multus-CNI related scenarios
...
     #Patching simplemacvlan config in network operator config CRD
     Given I have a project
     Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with:
 	    | {"spec":{"additionalNetworks":[{"name":"test-macvlan-case3","namespace":"<%= project.name %>","simpleMacvlanConfig":{"ipamConfig":{"staticIPAMConfig":{"addresses": [{"address":"10.128.2.100/23","gateway":"10.128.2.1"}]},"type":"static"},"master":"<%= cb.default_interface %>","mode":"bridge"},"type":"SimpleMacvlan"}]}} |
     #Cleanup for bringing CRD to original
     Given I register clean-up steps:
     """
-    Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with: 
+    Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with:
     | {"spec":{"additionalNetworks": null}} |
```
pruan-rht, line 531: indentation.

features/networking/multus.feature
```diff
@@ -985,7 +989,7 @@ Feature: Multus-CNI related scenarios
     #Cleanup for bringing CRD to original
     Given I register clean-up steps:
     """
-    as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with: 
+    as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with:
     | {"spec":{"additionalNetworks": null}} |
```
pruan-rht, line 993: indentation

anuragthehatter:
> > > because we need to pick a node this PR triggers a `oc get nodes` which adds ~2100 lines of log output. I'd have to change the env.nodes stuff to add `:quiet` to change that.
> > > `Given I store the ready and schedulable workers in the :nodes clipboard` step triggers the `oc get nodes`
> > 
> > 
> > @rbbratta Confirmed by CRIO team (and manually tested), `cni0` will be created when podman is run directly so its likely that non-sdn tests ran before this one caused cni0 creation on the node and hence we see 3 interfaces. I would suggest you to also change the logic in this same PR for output to `at least` contains
> > `bridge3 1 PVID Egress Untagged`
> > `veth9056d1f1 1 PVID Egress Untagged`
> > So may be just grep bridge3 and veth with 2 times 1 PVIDs
> 
> @rbbratta May be we can merge this soon. Keep popping up..heh. Thanks

@rbbratta Did you commit the fix for above point? Or if i have missed it somehow. We need to change logic for output check to check at least 2 times ```1PVID``` and ```untagged``` to counter that CNI0 issue. Right now we are strictly checking thise 2 pmtrs.

rbbratta:
Okay well it turns out the `bridge -j vlan show` can output JSON, so we can grab the JSON from before and after and do an array difference in ruby to find the added bridges, then we can exactly count the PVID 1 Untagged VLANs in the JSON output.

`   '/sbin/bridge\ -j\ vlan\ show`

before:
```JSON 
      [{"ifname":"cni0","vlans":[{"vlan":1,"flags":["PVID","Egress Untagged"]}]}]
```

after:
```
      [{"ifname":"cni0","vlans":[{"vlan":1,"flags":["PVID","Egress Untagged"]}]},{"ifname":"bridge3","vlans":[{"vlan":1,"flags":["PVID","Egress Untagged"]}]},{"ifname":"veth52c43afc","vlans":[{"vlan":1,"flags":["PVID","Egress Untagged"]}]}]
```

The comparison code could probably better rubified better but I did:
```ruby
  added_bridges = cb[clip_b] - cb[clip_a]
  num_vlans = added_bridges.count {|b|
    c = b["vlans"].count {|v|
      v["vlan"] == pvid && v["flags"].include?("PVID") && v["flags"].include?(mode)
    }
    c > 0
  }
  if num_vlans != expected_vlans.to_i
    raise "Not enough bridge VLANS of #{pvid} and mode #{mode}"
  end
```

http://pastebin.test.redhat.com/879053

features/networking/multus.feature
rbbratta: Okay, I let RubyMine fixup all the tables, hopefully that should fix it.

features/networking/multus.feature
rbbratta: Done

features/networking/multus.feature
rbbratta: Done

features/step_definitions/networking.rb
```diff
@@ -725,14 +727,31 @@
...
+Given /^the number of bridge PVID (\d+) #{QUOTED} VLANs added between the #{SYM} and #{SYM} clipboards is (\d+)$/ do |pvid, mode, clip_a, clip_b, expected_vlans|
+  pvid = pvid.to_i
+  added_bridges = cb[clip_b] - cb[clip_a]
+  logger.info("added_bridges: #{added_bridges}")
+  num_vlans = added_bridges.count {|b|
+    c = b["vlans"].count {|v|
+      v["vlan"] == pvid && v["flags"].include?("PVID") && v["flags"].include?(mode)
+    }
+    c > 0
+  }
```
rbbratta, line 744: this is the PVID diff check algorithm, needs review.  @anuragthehatter 

we take the added bridges and count all the vlans that match the `PVID` and have flags `"Egress Untagged"`

rbbratta:
@pruan-rht @anuragthehatter  PTAL

features/networking/multus.feature
```diff
@@ -464,9 +469,9 @@ Feature: Multus-CNI related scenarios
     #Entering into corresponding no eot make sure No VLAN ID information shown for secondary interface
     Given CNI vlan info is obtained on the "<%= cb.pod.node_name %>" node
     Then the step should succeed
-    And the output should contain 2 times:
-      | 1 PVID   |
-      | ntagged |
+    And evaluation of `@result[:parsed]` is stored in the :vlans_after clipboard
+    Given the number of bridge PVID 1 "Egress Untagged" VLANs added between the :vlans_before and :vlans_after clipboards is 2
+
```
rbbratta, line 474: new step

features/step_definitions/networking.rb
anuragthehatter: Thanks Ross for the efforts. lgtm except we want to make sure we strictly don't check ```"Egress untagged"``` but just ```"ntagged```. The reason were explained in few of teh old PRs as some platforms exhibits small/capital U and private clusters or external http proxy one's tends to add ```Egress```` so we definitely shouldn't check ```Egress and U/u```

features/step_definitions/networking.rb
```diff
@@ -725,14 +727,31 @@
   ensure_admin_tagged
   node = node(node_name)
   host = node.host
-  @result = host.exec_admin("/sbin/bridge vlan show")
+  @result = host.exec_admin("/sbin/bridge -j vlan show")
```
anuragthehatter, line 730: Cool stuff. I didn't know that

features/networking/multus.feature
anuragthehatter: Thanks , Ross. We should strictly check ```ntagged```  only to make compatibility across various platforms as per reasoning in other comment. Can you also pastebin logs? Seems like that one got are empty.

features/step_definitions/networking.rb
rbbratta: This shouldn't be possible in the JSON output based on the source from iproute2

https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/bridge/vlan.c#n421
```c
static void print_vlan_flags(__u16 flags)
{
	if (flags == 0)
		return;

	open_json_array(PRINT_JSON, "flags");
	if (flags & BRIDGE_VLAN_INFO_PVID)
		print_string(PRINT_ANY, NULL, " %s", "PVID");

	if (flags & BRIDGE_VLAN_INFO_UNTAGGED)
		print_string(PRINT_ANY, NULL, " %s", "Egress Untagged");
	close_json_array(PRINT_JSON, NULL);
}
```

rbbratta: Well I'll check more, maybe 4.5 is diff

On 4.6 we are 
```
sh-4.4# rpm -qf /usr/sbin/bridge
iproute-5.3.0-1.el8.x86_64
```
Code is the same in v5.3.0

https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/bridge/vlan.c?h=v5.3.0&id=18e631bd4b1269cdc32abd09dc3cd4a3c209e8e2#n447

rbbratta: Looks like it was fixed here

2018-09-10 11:48:06 -0700
https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/commit/bridge/vlan.c?h=v5.3.0&id=0f36267485e30099a4f735c3aadfa58b5efa1918

first added in v4.19.0.

rbbratta: 4.5.0-0.nightly-2020-07-25-031342
```bash
sh-4.4# rpm -qf `which bridge`
iproute-5.3.0-1.el8.x86_64
```

I have a fix using `ntagged`, just need to test

anuragthehatter: Thanks. Nice dissection on iproute, cool

rbbratta:
4.6 OVN
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-07-29-21-56-15.log 

4.5 SDN
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-07-29-22-05-06.log

features/step_definitions/networking.rb
rbbratta: The next step is to see if `jq` is installed everywhere.

Looks like some scripts are using it
https://github.com/openshift/machine-config-operator/pull/1860/files#diff-8197d7b27eb62b1ba2b9b44460cd3238R17

```bash
      iface=$(ip -j route show default | jq -r '.[0].dev')
```

rbbratta:
rebased

http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-09-14-20-32-36.log

rbbratta:
@pruan-rht @anuragthehatter PTAL

rbbratta:
rebased
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-10-20-15-48-00.log

pruan-rht (commented):
@rbbratta this step is used throughout the framework...  What is the reason for the modification?

rbbratta:
> @rbbratta this step is used throughout the framework... What is the reason for the modification?

For delimiters, negated character classes are more proper than non-greedy quantifiers.  See https://github.com/openshift/verification-tests/pull/1279

Must have noticed it when changing
```ruby
    And evaluation of `@result[:parsed]` is stored in the :vlans_after clipboard
```

I removed it for now.

rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-10-23-13-23-45.log

features/step_definitions/networking.rb
```diff
@@ -1,3 +1,5 @@
+require 'json'
```
pruan-rht, line 1: @rbbratta don't think you need to import `json`  you can use the `yaml` lib instead which is a superset of JSON

features/step_definitions/networking.rb
```diff
@@ -698,8 +700,26 @@
   ensure_admin_tagged
   node = node(node_name)
   host = node.host
-  @result = host.exec_admin("/sbin/bridge vlan show")
+  @result = host.exec_admin("/sbin/bridge -j vlan show")
   raise "Failed to execute bridge vlan show command" unless @result[:success]
+  @result[:parsed] = JSON.load @result[:stdout].to_s
```
pruan-rht, line 705: @rbbratta I think you can just use `YAML.load` here.

pruan-rht: @rbbratta also, I don't think you need to `.to_s` call

rbbratta:
re-tested
```

[18:15:00] INFO> Shell Commands: oc debug  --kubeconfig=/home/rbrattai/Documents/deployments/119923/workdir/ocp4_admin.kubeconfig -n tests-rbrattai-rbrattai --image=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:8f4882cff3c2f9521215eac681c5abda42876e3e955431c1387fb457940b8344 node/ip-10-0-152-71.ca-central-1.compute.internal -- chroot /host/ bash -c cd\ \'/tmp/workdir/rbrattai-rbrattai\''
'/sbin/bridge\ -j\ vlan\ show

[{"ifname":"bridge3","vlans":[{"vlan":1,"flags":["PVID","Egress Untagged"]}]},{"ifname":"veth846e046a","vlans":[{"vlan":1,"flags":["PVID","Egress Untagged"]}]}]

STDERR:
Starting pod/ip-10-0-152-71ca-central-1computeinternal-debug ...
To use host binaries, run `chroot /host`

Removing debug pod ...


[18:15:02] INFO> Exit Status: 0

[18:15:02] INFO> added_bridges: [{"ifname"=>"bridge3", "vlans"=>[{"vlan"=>1, "flags"=>["PVID", "Egress Untagged"]}]}, {"ifname"=>"veth846e046a", "vlans"=>[{"vlan"=>1, "flags"=>["PVID", "Egress Untagged"]}]}]
```

pruan-rht (commented):
/lgtm

PR #1036, opened 2020-05-01
fixed scenarios post few bugs

anuragthehatter:
Fixed OCP-21793 due to bug https://bugzilla.redhat.com/show_bug.cgi?id=1827375. We should not create net-attach-def in default namespace now. 
Also fixed OCP-24466, we added some debug steps to it earlier to see arp entries and seems like it fails because 192.168.1.1 was already assigned in openstack cluster due to NAT'ing used there. ALl nodes seems ot be in 192.168 subnet and we use same in test http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/04/30/13:07:05/CNO_manager_macvlan_configured_manually_with_DHCP/console.html
I am simply now using rare subnet like 87.8.x.x which shouldn't have any problem.
cc @rbbratta @weliang1 
Re-ran logs, OCP-24466: http://pastebin.test.redhat.com/861157
OCP-21793: http://pastebin.test.redhat.com/861657

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -598,7 +598,7 @@ Feature: Multus-CNI related scenarios
       | namespace isolation |
       | violat              |
     """
-    
+
   # @author anusaxen@redhat.com
   # @case_id OCP-24490
   @admin
@@ -908,7 +908,7 @@ Feature: Multus-CNI related scenarios
     And I run commands on the host:
       | ip link add mvlanp0 type vxlan id 100 remote <%= cb.mastr_inf_address %> dev <%= cb.workr_inf_name %> dstport 14789 |
       | ip link set up mvlanp0                                                                                              |
-      | ip a add 87.87.8.20/24 dev mvlanp0                                                                                  |
+      | ip a add 192.18.0.10/15 dev mvlanp0                                                                                 |
     Then the step should succeed
     #Cleanup for deleting worker interface
     Given I register clean-up steps:
@@ -920,7 +920,7 @@ Feature: Multus-CNI related scenarios
     And I run commands on the host:
       | ip link add mvlanp0 type vxlan id 100 remote <%= cb.workr_inf_address %> dev <%= cb.mastr_inf_name %> dstport 14789 |
       | ip link set up mvlanp0                                                                                              |
-      | ip a add 87.87.8.10/24 dev mvlanp0                                                                                  |
+      | ip a add 192.18.0.20/15 dev mvlanp0                                                                                 |
     Then the step should succeed
     #Cleanup for deleting master interface
     Given I register clean-up steps:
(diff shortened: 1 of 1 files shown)


anuragthehatter:
I mean we can create net-attach-defs in default namespaces but its not allowed to use as a reference during pod creation in other namespaces

features/networking/multus.feature
```diff
@@ -600,7 +598,7 @@ Feature: Multus-CNI related scenarios
       | namespace isolation |
       | violat              |
     """
- 
+
```
rbbratta, line 601: maybe a `sed -i -e 's/[[:space:]]\+$//g'`  to clean up trailing whitespace.  

or if vim,  `:%s/\s\+$//g`  and maybe `set list` to show the trailing whitespace.
```
set list
set listchars=tab:»\ ,trail:¬
```

features/networking/multus.feature
```diff
@@ -910,7 +908,7 @@ Feature: Multus-CNI related scenarios
     And I run commands on the host:
       | ip link add mvlanp0 type vxlan id 100 remote <%= cb.mastr_inf_address %> dev <%= cb.workr_inf_name %> dstport 14789 |
       | ip link set up mvlanp0                                                                                              |
-      | ip a add 192.168.1.2/24 dev mvlanp0                                                                                 |
+      | ip a add 87.87.8.20/24 dev mvlanp0                                                                                  |
```
rbbratta, line 911: also trailing whitespace

features/networking/multus.feature
```diff
@@ -922,7 +920,7 @@ Feature: Multus-CNI related scenarios
     And I run commands on the host:
       | ip link add mvlanp0 type vxlan id 100 remote <%= cb.workr_inf_address %> dev <%= cb.mastr_inf_name %> dstport 14789 |
       | ip link set up mvlanp0                                                                                              |
-      | ip a add 192.168.1.1/24 dev mvlanp0                                                                                 |
+      | ip a add 87.87.8.10/24 dev mvlanp0                                                                                  |
```
rbbratta, line 923: I think there are other non-routable blocks.

could we use 169.254.0.0/16?, or maybe the TEST-NET-3 reserved 203.0.113.0/24?

https://en.wikipedia.org/wiki/Reserved_IP_addresses

rbbratta: 198.18.0.0/15 is reserved for Benchmarking, that sounds exciting.  :smiley:

anuragthehatter: Nice. yeah, The bench-marking one really sounds exciting. Re-ran to make sure http://pastebin.test.redhat.com/862033. Also these tests will be soon replaced by multus ```wherabouts``` which will be used to assign dhcp addresses so just matter of short time we need to keep these existing tests running

weliang1: @anuragthehatter Multus whereabouts DHCP is supported in v4.5 and backport will be in v4.4 soon. Compare the way this case using DHCP server, whereabouts is much easy to use, also we do not have formal release docs to describe how to setup the DHCP server as this test case tested, in the future, we may think about to disable this testing if automation fail again.

anuragthehatter:
Hey guys, any more comments?

weliang1:
/LGTM

FYI: https://bugzilla.redhat.com/show_bug.cgi?id=1827375 is covered by https://github.com/openshift/verification-tests/pull/1037 now

features/networking/multus.feature
```diff
@@ -574,19 +574,17 @@ Feature: Multus-CNI related scenarios
     # Make sure that the multus is enabled
     Given the multus is enabled on the cluster
     # Create the net-attach-def via cluster admin
```
liangxia, line 576: The comment should be removed as we are switch to a new approach.

anuragthehatter: Thanks. Fixed!

PR #1038, opened 2020-05-06
fixes for logging 4.5

QiaolingTang:
Update the scenarios to match the logging 4.5 changes, @anpingli PTAL

### DIFF
--- a/lib/openshift/service.rb
+++ b/lib/openshift/service.rb
@@ -102,7 +102,7 @@ module BushSlicer
       ports.each do | p |
         port = p['port'] if p['name'] == name
       end
-      raise "Could not find port with name #{name}, does the name really exist?" unless port != nil
+      raise "Could not find port with name #{name}, does the name really exist?" if port.nil?
       return port
     end
 
--- a/lib/openshift/service.rb
+++ b/lib/openshift/service.rb
@@ -96,6 +96,16 @@ module BushSlicer
       return node_port
     end
 
+    def port(user: nil, name:, cached: true, quiet: false)
+      port = nil
+      ports = self.ports(user: user, cached: cached, quiet: quiet)
+      ports.each do | p |
+        port = p['port'] if p['name'] == name
+      end
+      raise "Could not find port with name #{name}, does the name really exist?" unless port != nil
+      return port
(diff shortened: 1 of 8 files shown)


lib/openshift/service.rb
```diff
@@ -96,6 +96,15 @@ def node_port(user: nil, port:, cached: true, quiet: false)
       return node_port
     end
 
+    def port(user: nil, name:, cached: true, quiet: false)
+      port = nil
+      ports = self.ports(user: user, cached: cached, quiet: quiet)
+      ports.each do | p |
+        port = p['port'] if p['name'] == name
+      end
+      return port
```
pruan-rht, line 105: @QiaolingTang should we always get back a valid port instead of nil?  Do we need to raise an error at the end of `ports.each`, there is not a match...in case where user gave an name of the port that's incorrect.  WDYT?

QiaolingTang: > @QiaolingTang should we always get back a valid port instead of nil?
For me, yes.

> Do we need to raise an error at the end of `ports.each`, there is not a match...in case where user gave an name of the port that's incorrect. WDYT?
I'll update it, thanks a lot.

QiaolingTang: @pruan-rht I updated it , PTAL.

```
[1] pry(#<BushSlicer::DefaultWorld>)> 
[2] pry(#<BushSlicer::DefaultWorld>)> service('fluentd').port(name: 'metric')
RuntimeError: Could not find port with name metric, does the name really exist?
from /Users/tangqiaoling/github.com/verification-tests/lib/openshift/service.rb:105:in `port'
[3] pry(#<BushSlicer::DefaultWorld>)> service('fluentd').port(name: 'metrics')
=> 24231
[4] pry(#<BushSlicer::DefaultWorld>)> 
```

pruan-rht: @QiaolingTang you can do `if port.nil?` instead of `unless port != nil`

QiaolingTang: Done

pruan-rht: /lgtm

PR #1042, opened 2020-05-07
Update: Modify iscsi/nfs image url for disconnect env

duanwei33:
Did some test for iscsi/nfs pass.

### DIFF
--- a/features/step_definitions/helper_services.rb
+++ b/features/step_definitions/helper_services.rb
@@ -9,8 +9,9 @@ Given /^I have a NFS service in the(?: "([^ ]+?)")? project$/ do |project_name|
   end
 
   step %Q/SCC "privileged" is added to the "system:serviceaccounts:<%= project.name %>" group/
+  step 'I obtain test data file "storage/nfs/nfs-server.yaml"'
   step %Q{I run the :create client command with:}, table(%{
-    | f | #{ENV['BUSHSLICER_HOME']}/testdata/storage/nfs/nfs-server.yaml |
+    | f | <%= BushSlicer::HOME %>/testdata/storage/nfs/nfs-server.yaml |
   })
   step %Q/the step should succeed/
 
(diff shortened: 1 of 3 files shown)


features/step_definitions/helper_services.rb
```diff
@@ -484,7 +484,7 @@
       res = host.exec_admin(*setup_commands)
       raise "iSCSI initiator setup commands error" unless res[:success]
     end
-    @result = admin.cli_exec(:create, n: _project.name, f: 'https://raw.githubusercontent.com/openshift-qe/docker-iscsi/master/iscsi-target.json')
+    @result = admin.cli_exec(:create, n: _project.name, f: "#{ENV['BUSHSLICER_HOME']}/testdata/storage/iscsi/iscsi-target.json")
```
liangxia, line 487: We can use `BushSlicer::HOME` to replace `ENV['BUSHSLICER_HOME']`

duanwei33: Sure, let's follow the professional guidance :) 
Yes, I checked the other file, BushSlicer::HOME is more used, I updated the another one in this file.

features/step_definitions/helper_services.rb
```diff
@@ -10,7 +10,7 @@
 
   step %Q/SCC "privileged" is added to the "system:serviceaccounts:<%= project.name %>" group/
   step %Q{I run the :create client command with:}, table(%{
-    | f | #{ENV['BUSHSLICER_HOME']}/testdata/storage/nfs/nfs-server.yaml |
+    | f | <%= BushSlicer::HOME %>/testdata/storage/nfs/nfs-server.yaml |
```
akostadinov, line 13: @liangxia , actually to make sure that whenever we decide to keep test data, we can quickly update all places, even if it is in a separate repo again, it is much better to do like 
```
  step 'I obtain test data file "testdata/storage/nfs/nfs-server.yaml"'
  step %Q{I run the :create client command with:}, table(%{
    | f | <%= BushSlicer::HOME %>/testdata/storage/nfs/nfs-server.yaml |
  })

duanwei33:
@liangxia @akostadinov PTAL

PR #1057, opened 2020-05-08
Revert "remove invalid creds"

akostadinov:
One cannot create requests knowing access key only. Also this key has been invalidated. Thus removing to keep the good example.

Reverts openshift/verification-tests#1051


pruan-rht:
/lgtm
That's fine...the creds have been revoked already

PR #1062, opened 2020-05-09
Fix get_version from cluster version

liangxia:
/cc @akostadinov @pruan-rht 

The command returns an hash, while we expect a string.
Also, the version does not match `/^[\d.]+$/`, e.g 4.5.0-0.nightly-2020-05-08-015855, so updated the check.

Thank you @zhouying7780 for reporting the issue.

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -267,8 +267,8 @@ module BushSlicer
     # helper parser
     def parse_version(ver_str)
       ver = ver_str.sub(/(^v|^openshift-clients-)/,"")
-      if ver !~ /^[\d.]+$/
-        raise "version '#{ver}' does not match /^[\d.]+$/"
+      if ver !~ /^\d[.]\d+\b.*$/
+        raise "version '#{ver}' does not match /^\d[.]\d+\b.*$/"
       end
       ver = ver.split(".").reject(&:empty?).map(&:to_i)
       [ver[0], ver[1]].map(&:to_i)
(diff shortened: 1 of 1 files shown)


lib/environment.rb
```diff
@@ -267,8 +267,8 @@ def system_docker_repo
     # helper parser
     def parse_version(ver_str)
       ver = ver_str.sub(/(^v|^openshift-clients-)/,"")
-      if ver !~ /^[\d.]+$/
-        raise "version '#{ver}' does not match /^[\d.]+$/"
+      if ver !~ /^[\d][.][\d].*$/
+        raise "version '#{ver}' does not match /^[\d][.][\d].*$/"
```
akostadinov, line 271: What do you think about `/^\d[.]\d+\b.*$/`? This will require one digit, then dot, then any number of digits, then a word boundary followed by anything.

lib/environment.rb
```diff
@@ -267,8 +267,8 @@ def system_docker_repo
...
       end
       ver = ver.split(".").reject(&:empty?).map(&:to_i)
```
akostadinov, line 273: maybe we need to change this as well if we want to support version of the format `4.5-whatever`. If we are sure that we get a dot or end of string after the second number, then doesn't matter.

akostadinov: actually ignore, it doesn't matter. 
```
[7] pry(main)> "32-d243".to_i
=> 32
```

lib/environment.rb
liangxia: @akostadinov Updated.

PR #1065, opened 2020-05-11
update rules for kibana

QiaolingTang:
I change the kibana rules directory because I need new rules for the logging 4.5.

@yapei @anpingli PTAL, thanks.


yanpzhan:
@QiaolingTang you update doesn't have effect to ui console, so it's ok for us.

PR #1067, opened 2020-05-11
Update image

kasturinarra:
Update image so the case can be tested in disconnected cluster


kasturinarra:

PR #1068, opened 2020-05-11
Fix auto case OCP_10696

kasturinarra:
change run to create_deploymentconfig command.


kasturinarra:
@zhouying7780 , can you please help review ? Thanks !!

PR #1072, opened 2020-05-12
Remove unnecessary version check

liangxia:
In the earlier 3.x (3.4 and earlier), we use major version 3 for OSE, while major version is 1 for origin. So we need the logic to deal with the specific versions. They are not needed now (as we are in 4.x and 3.4 and earlier reached EOL).

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -291,12 +291,10 @@ module BushSlicer
       # obtained = user.rest_request(:version)
       # if obtained[:request_opts][:url].include?("/version/openshift") &&
       #     !obtained[:success]
-        # seems like pre-3.3 version, lets hardcode to 3.2
-        # After bug 1692670 fix, will recover hardcode to '3.2'.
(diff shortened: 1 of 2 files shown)


lib/environment.rb
```diff
@@ -318,22 +316,17 @@ def version_cmp(version, user:)
...
 
-      norm_version = @major_version == '1' ? '3' : @major_version
-      if ( major == '1' )
-        major = '3'
-      end
-
-      # presently handle only major ver `4`, `3`, `1` for OCP and OKD
-      bad_majors = [@major_version, major] - [1, 3, 4]
+      # presently handle only major ver `4`, `3` for OCP and OKD
+      bad_majors = [@major_version, major] - [3, 4]
```
akostadinov, line 320: I'm not sure we need this check anymore. I'd say, lets remove it altogether for the sake of simplicity.

liangxia: @akostadinov  Updated.

PR #1084, opened 2020-05-13
fix ElementClickInterceptedError

yapei:


lib/rules/web/admin_console/4.5/browser_opts.yml
```diff
@@ -21,6 +21,16 @@
...
+          when /pf-c-form__actions/
+            logger.info "Scrolling from action button out of view"
+            scroll_by = browser.element(css: ".pf-c-form__actions").height
+            scroll_by += browser.element(css: ".ocs-radio-group-field").height
+            browser.execute_script("window.scrollBy(0,-#{scroll_by})")
+          when /ocs-form-footer/
+            logger.info "Scrolling footer out of view"
+            scroll_by = browser.element(css: ".ocs-form-footer").height
+            scroll_by += browser.element(css: ".ocs-radio-group-field").height
+            browser.execute_script("window.scrollBy(0,-#{scroll_by})")
```
akostadinov, line 33: If this is a footer, shouldn't we scroll down instead of scroll up? Perhaps remove the `-` sign. 

btw do we still need `nav-pf-secondary-nav` and `navbar-pf-vertical` scrolling?

akostadinov: Another thing is that the offending element is `ocs-form-footer ocs-form-footer__sticky co-m-btn-bar`. So we only need to scroll by it only. No need to scroll by `.ocs-radio-group-field`.

lib/rules/web/admin_console/4.5/browser_opts.yml
```diff
@@ -21,6 +21,16 @@
             scroll_by = browser.element(css: ".navbar.navbar-pf-vertical").height
             scroll_by += browser.element(css: ".project-bar").height
             browser.execute_script("window.scrollBy(0,-#{scroll_by})")
+          when /pf-c-form__actions/
+            logger.info "Scrolling from action button out of view"
+            scroll_by = browser.element(css: ".pf-c-form__actions").height
+            scroll_by += browser.element(css: ".ocs-radio-group-field").height
```
akostadinov, line 27: I think we shouldn't need to scroll by `.ocs-radio-group-field` here too.

lib/rules/web/admin_console/4.5/browser_opts.yml
```diff
@@ -21,6 +21,16 @@
...
+            browser.execute_script("window.scrollBy(0,-#{scroll_by})")
```
akostadinov, line 28: I didn't see this in the log but is it a header or a footer? If it is a footer then scrolling down makes more sense.

yapei:
@akostadinov  I tried several times but it still doesn't work. Here's the latest jenkins log #92817

akostadinov:
superseded by #1125

It is an intrusive change so maybe needs a little more testing. Also the fix itself is a little bit wild but I hope it will work for a wide variety of use cases. Anyway, having access to the element itself inside the hook gives a lot more opportunity for navigating the page correctly.

PR #1091, opened 2020-05-13
Add cmd sleep

kasturinarra:


lib/rules/cli/4.1.yaml
```diff
@@ -186,6 +186,7 @@
     :restart: --restart=<value>
     :save_config: --save-config=<value>
     :schedule: --schedule=<value>
+    :sleep:  -- sleep <value>
```
pruan-rht, line 189: @kasturinarra should it be `--sleep=<value>`

lib/rules/cli/4.4.yaml
```diff
@@ -186,6 +186,7 @@
     :restart: --restart=<value>
     :save_config: --save-config=<value>
     :schedule: --schedule=<value>
+    :sleep:  -- sleep <value>
```
pruan-rht, line 189: ditto...

lib/rules/cli/4.1.yaml
kasturinarra: It should not have '=',  the command which i wanted to run is "oc create cronjob <name> <imagename> --schedule='* * * * *' -- sleep 180

pruan-rht: @kasturinarra if `:sleep` is not part of the `oc command` then this shouldn't be part of the yaml file.  We only add options that are specific to the `oc` client command

kasturinarra:
closing this due to https://github.com/openshift/verification-tests/pull/1095

PR #1095, opened 2020-05-14
update cronjob lib and ocp-17515

zhouying7780:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1290/console
@kasturinarra  Please help review , thanks

### DIFF
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -177,11 +177,14 @@
     :from_literal: --from-literal=<value>
     :name: <value>
 :create_cronjob:
-  :cmd: oc create cronjob <name>
+  :cmd: oc create cronjob <name> <global_options> <options>
   :options:
     :allow_missing_template_keys: --allow-missing-template-keys=<value>
     :dry_run: --dry-run=<value>
+    :exec_command: <value>
+    :exec_command_arg: <value>
     :image: --image=<value>
+    :oc_opts_end: --
     :output: --output <value>
     :restart: --restart=<value>
     :save_config: --save-config=<value>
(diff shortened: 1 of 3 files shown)


kasturinarra:
@zhouying7780 can we make sure that it works with 44 as well ? Thanks !!

zhouying7780:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/92853/console

@kasturinarra  works well for ocp4.4 , please help review , thanks

lib/rules/cli/4.1.yaml
```diff
@@ -177,9 +177,10 @@
     :from_literal: --from-literal=<value>
     :name: <value>
 :create_cronjob:
-  :cmd: oc create cronjob <name>
```
pruan-rht, line 180: @zhouying7780 I don't think you need to add `<global_options>` since they are global, all rules will have access to those.

akostadinov: Perhaps he needs argument ordering where command follows any other options. Like with `exec`. Correct?

zhouying7780: @pruan-rht @akostadinov yes , just need to argument the ordering.

zhouying7780:
@liangxia  Please help review , thanks.

zhouying7780:
@liangxia Please help review and merge , thanks

lib/rules/cli/4.1.yaml
liangxia: Any conclusion now ?

lib/rules/cli/4.1.yaml
```diff
@@ -177,9 +177,10 @@
...
+  :cmd: oc create cronjob <name> <global_options> <options>
   :options:
     :allow_missing_template_keys: --allow-missing-template-keys=<value>
+    :cmd: <value>
```
liangxia, line 183: Perhaps we can use `oc_opts_end`

zhouying7780:
@liangxia Have resolved all the comments , please help review again, thanks

lib/rules/cli/4.4.yaml
```diff
@@ -208,12 +211,15 @@
     :output: --output <value>
     :template: --template=<value>
 :create_job:
-  :cmd: oc create job <name>
+  :cmd: oc create job <name> <global_options> <options>
```
liangxia, line 214: No need the change in this line now.
Also please sort the options in alphanumeric.

zhouying7780: No , must order the global_options and options . because if without the order '--kubeconfig' will not take effect .

zhouying7780:
@liangxia have sort the options in alphanumeric, please help review thanks .

PR #1096, opened 2020-05-14
Multicast update

weliang1:
Change to run all the multicast cases from verification-tests:

1. Copy all multicast cases from cucushift to verification-tests, now all the multicast cases can be run from verification-tests
2. Update YAML path to use <%= BushSlicer::HOME %>/testdata/networking/multicast-rc.json
3. Add more steps in OCP-12966 to wait old pods are deleted in default name before creating the new pods
4. Remove "Given the env is using multitenant or networkpolicy network" from old cases
5. No need to change old steps for those old test cases.

Test log:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1318/console

### DIFF
--- a/features/networking/multicast.feature
+++ b/features/networking/multicast.feature
@@ -142,6 +142,8 @@ Feature: testing multicast scenarios
   # @case_id OCP-12930
   @admin
   Scenario: Same multicast groups can be created in multiple tenant
+    Given the env is using multitenant or networkpolicy network
+    
     Given I have a project
     And evaluation of `project.name` is stored in the :proj1 clipboard
     When I run the :create client command with:
@@ -279,6 +281,8 @@ Feature: testing multicast scenarios
   @admin
   @destructive
   Scenario: pods in default project should not be able to receive multicast traffic from other tenants
+    Given the env is using multitenant or networkpolicy network
+    
     # create multicast testing pod in one project
     Given I have a project
     And evaluation of `project.name` is stored in the :proj1 clipboard
@@ -356,6 +360,8 @@ Feature: testing multicast scenarios
   # @case_id OCP-12928
   @admin
   Scenario: pods should be able to join multiple multicast groups at same time
+    Given the env is using multitenant or networkpolicy network
+    
     # create some multicast testing pods in the project
     Given I have a project
     And evaluation of `project.name` is stored in the :proj1 clipboard
@@ -449,6 +455,8 @@ Feature: testing multicast scenarios
   # @case_id OCP-12929
   @admin
   Scenario: pods should not be able to receive multicast traffic from other pods in different tenant
+    Given the env is using multitenant or networkpolicy network
+    
     # create some multicast testing pods in one project
     Given I have a project
     And evaluation of `project.name` is stored in the :proj1 clipboard
@@ -528,6 +536,8 @@ Feature: testing multicast scenarios
   @admin
   @destructive
   Scenario: pods in default project should be able to receive multicast traffic from other default project pods
+    Given the env is using multitenant or networkpolicy network
+    
     # enable multicast and create testing pods
     Given I switch to cluster admin pseudo user
     And I use the "default" project
--- a/features/networking/multicast.feature
+++ b/features/networking/multicast.feature
@@ -277,6 +277,7 @@ Feature: testing multicast scenarios
   # @author hongli@redhat.com
   # @case_id OCP-12931
   @admin
+  @destructive
   Scenario: pods in default project should not be able to receive multicast traffic from other tenants
     # create multicast testing pod in one project
     Given I have a project
@@ -525,6 +526,7 @@ Feature: testing multicast scenarios
   # @author hongli@redhat.com
   # @case_id OCP-12966
   @admin
+  @destructive
   Scenario: pods in default project should be able to receive multicast traffic from other default project pods
     # enable multicast and create testing pods
     Given I switch to cluster admin pseudo user
@@ -533,6 +535,12 @@ Feature: testing multicast scenarios
       | netnamespace.network.openshift.io/multicast-enabled=true |
     Then the step should succeed
 
+    And I wait up to 120 seconds for the steps to pass:
+    """
+    And all existing pods die with labels:
+    | name=mcast-pods |
+    """
+ 
     Given admin ensures "mcast-rc" rc is deleted after scenario
     When I run the :create client command with:
       | f | <%= BushSlicer::HOME %>/testdata/networking/multicast-rc.json |
--- a/features/networking/multicast.feature
+++ b/features/networking/multicast.feature
@@ -273,3 +273,328 @@ Feature: testing multicast scenarios
       | <%= cb.proj2pod1ip %>.*multicast, xmt/rcv/%loss = 5/5/0%                |
       | <%= cb.proj2pod2ip %>.*multicast, xmt/rcv/%loss = 5/5/0%                |
     """
+
+  # @author hongli@redhat.com
+  # @case_id OCP-12931
+  @admin
+  Scenario: pods in default project should not be able to receive multicast traffic from other tenants
+    # create multicast testing pod in one project
+    Given I have a project
+    And evaluation of `project.name` is stored in the :proj1 clipboard
+    When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/multicast-rc.json" replacing paths:
+      | ["spec"]["replicas"] | 1 |
+    Then the step should succeed
+    Given 1 pods become ready with labels:
+      | name=mcast-pods |
+    And evaluation of `pod.ip` is stored in the :proj1_podip clipboard
+    And evaluation of `pod.name` is stored in the :proj1_pod clipboard
+
+    # enable multicast for the netnamespace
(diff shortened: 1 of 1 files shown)


anuragthehatter:
@weliang1 If its just one step, I suggest to modify it like following and this shall solve all the problems. We can discuss this in sync. If we do following modification in master branch (as it points to 4.x) of [https://github.com/openshift/verification-tests/blob/master/features/step_definitions/networking.rb], that should solve it

```
Given /^the env is using multitenant or networkpolicy network$/ do
  if env.version_ge("4.1", user: user)
    logger.info("Skipping this check as plugin is networkpolicy by default on 4.x")
  else
    step 'the env is using one of the listed network plugins:', table([["multitenant","networkpolicy"]])
 end
end
```

weliang1:
> @weliang1 If its just one step, I suggest to modify it like following and this shall solve all the problems. We can discuss this in sync. If we do following modification in master branch (as it points to 4.x) of [https://github.com/openshift/verification-tests/blob/master/features/step_definitions/networking.rb], that should solve it
> 
> ```
> Given /^the env is using multitenant or networkpolicy network$/ do
>   if env.version_ge("4.1", user: user)
>     logger.info("Skipping this check as plugin is networkpolicy by default on 4.x")
>   else
>     step 'the env is using one of the listed network plugins:', table([["multitenant","networkpolicy"]])
>  end
> end
> ```

Actually scripts need "Given the env is using multitenant or networkpolicy network" step, because multicast is not supported in openshift-ovs-subnet, I will add "Given the env is using multitenant or networkpolicy network" step back

weliang1:
Let's focus on reviewing above changes in this PR, I will update the steps to support multicast in OVN in next PR.

features/networking/multicast.feature
```diff
@@ -273,3 +275,344 @@ Feature: testing multicast scenarios
       | <%= cb.proj2pod1ip %>.*multicast, xmt/rcv/%loss = 5/5/0%                |
       | <%= cb.proj2pod2ip %>.*multicast, xmt/rcv/%loss = 5/5/0%                |
     """
+
+  # @author hongli@redhat.com
```
pruan-rht, line 279: @weliang1 you submitted the PR but the `@author is honglin@rehdat.com`??

features/networking/multicast.feature
```diff
@@ -273,3 +275,344 @@ Feature: testing multicast scenarios
...
+    Then the step should succeed
+    And the output should contain:
+      | joined (S,G) = (*, 232.43.211.234), pinging |
+      | multicast, xmt/rcv/%loss = 5/0/100%         |
+      
+  # @author hongli@redhat.com
+  # @case_id OCP-12928
+  @admin
+  Scenario: pods should be able to join multiple multicast groups at same time
+    Given the env is using multitenant or networkpolicy network
```
pruan-rht, line 363: @weliang1 I would actually rewrite this step and absorb the `or` logic within the step instead of having the user write `multitenant or networkpolicy`

features/networking/multicast.feature
```diff
@@ -273,3 +275,344 @@ Feature: testing multicast scenarios
...
+    Given 1 pods become ready with labels:
+      | name=mcast-pods |
+    And evaluation of `pod.ip` is stored in the :proj2_podip clipboard
+    And evaluation of `pod.name` is stored in the :proj2_pod clipboard
+
+    # enable multicast for the netnamespace
+    When I run the :annotate admin command with:
+      | resource     | netnamespace    |
+      | resourcename | <%= cb.proj2 %> |
+      | overwrite    | true            |
```
pruan-rht, line 494: @weliang1 table `|` alignment

features/networking/multicast.feature
```diff
@@ -273,3 +275,344 @@ Feature: testing multicast scenarios
...
+    Then the step should succeed
+
+    # check the omping result on pod in second project
+    Given I use the "<%= cb.proj2 %>" project
+    When I run the :exec background client command with:
+      | pod              | <%= cb.proj2_pod %> |
+      | oc_opts_end      |                     |
+      | exec_command     | sh                  |
+      | exec_command_arg | -c                  |
+      | exec_command_arg | omping -c 5 -T 10 <%= cb.proj1_podip %> <%= cb.proj2_podip %> > /tmp/p2.log |
```
pruan-rht, line 518: @weliang1 table | alignment

features/networking/multicast.feature
```diff
@@ -273,3 +275,344 @@ Feature: testing multicast scenarios
...
+    Given the env is using multitenant or networkpolicy network
+    
+    # enable multicast and create testing pods
+    Given I switch to cluster admin pseudo user
+    And I use the "default" project
+    Given admin adds and overwrites following annotations to the "default" netnamespace:
+      | netnamespace.network.openshift.io/multicast-enabled=true |
+    Then the step should succeed
+
+    And I wait up to 120 seconds for the steps to pass:
```
pruan-rht, line 548: @weliang1, don't need to wrap around the step as there's a time-out within that step already.

features/networking/multicast.feature
```diff
@@ -273,3 +275,344 @@ Feature: testing multicast scenarios
...
+
+    # ensure interface join to the multicast group
+    When I execute on the "<%= cb.pod3 %>" pod:
+      | netstat | -ng |
+    Then the step should succeed
+    And the output should match:
+      | eth0\s+1\s+232.43.211.234 |
+
+    # check the result on third pod and should received 5 multicast packets from other pods
+    Given 10 seconds have passed
```
pruan-rht, line 610: @weliang1 is there a way to do this w/o the hard-coded wait time?

features/networking/multicast.feature
```diff
@@ -273,3 +275,344 @@ Feature: testing multicast scenarios
...
+      | exec_command_arg | omping -c 5 -T 10 <%= cb.proj1_podip %> <%= cb.proj2_podip %> > /tmp/p2.log |
+    Then the step should succeed
+
+    When I execute on the "<%= cb.proj2_pod %>" pod:
+      | netstat | -ng |
+    Then the step should succeed
+    And the output should match:
+      | eth0\s+1\s+232.43.211.234 |
+
+    Given 10 seconds have passed
```
pruan-rht, line 527: @weliang1 try to avoid hard-coded sleep if possible.

features/networking/multicast.feature
weliang1: > @weliang1 you submitted the PR but the `@author is honglin@rehdat.com`??

Here are the what this PR does:
1. Copy all multicast cases from cucushift to verification-tests, now all the multicast cases can be run from verification-tests
2.Update YAML path to use <%= BushSlicer::HOME %>/testdata/networking/multicast-rc.json
3. Add more steps in OCP-12966 to wait old pods are deleted in default name before creating the new pods

So you will see @author is still honglin@rehdat.com for those old test cases. There are no new test cases created in the PR

features/networking/multicast.feature
weliang1: @pruan-rht The step "Given the env is using multitenant or networkpolicy network" is an existing step in each old test cases in both cucushift and verification-tests, no new changes did in this PR for this step.
https://github.com/openshift/cucushift/blob/master/networking/multicast.feature#L7
https://github.com/openshift/verification-tests/blob/master/features/networking/multicast.feature#L7

features/networking/multicast.feature
weliang1: @pruan-rht Above test steps are new added and really needed. The reason is after combine all the multicast together and run them all, this case always fail. After debugging and find out the pods with name=mcast-pods are Terminated but not clean up, those pods are created in default namespace in previous case OCP-12931. For these two test cases, pods are created in default namespace, deleting the pods are take a while, without the steps to check if the pods are really clean up, this case always fail.

features/networking/multicast.feature
zhaozhanqi: @weliang1 is any special reason copy these scenario to verification-tests repo?

weliang1: > @weliang1 is any special reason copy these scenario to verification-tests repo?

@zhaozhanqi The main reason to run all multicast case from one repo is to update, maintain and debug those cases easily, we really do not want spread same feature scripts in two repo locations. We discussed this in our sync up meeting several weeks ago, the multicast is upstream feature and it is fine to run all multicast cases from verfication-tests rep

anuragthehatter: @weliang1 I can presume following implications if we move these tests to public repo

- You need to generate two PRs , against master and v3 branch of verification-tests repo. 
  This PR points to master only.
- To keep  ```Given /^the env is using multitenant or networkpolicy network``` compatible, it 
  either needs a bypass logic for master branch PR like 
  https://github.com/openshift/verification-tests/pull/1096#issuecomment-628899201 
  or it can be removed from networking.rb in ```master``` branch and hence removed from 
  multicast scenarios as well in `master` branch PR
- Update all multicast scripts in Polarion to point to right location after it s moved

weliang1:
To keep thing simple, I would suggest to continue the way multicast scripts are doing now.
Just close this PR.

PR #1105, opened 2020-05-15
Create the pod in the correct node

weliang1:
Fix the issue to let the pod created in the node where other step get default interface name from, then multus pod creation will use correct interface name.

Test log:
https://privatebin-it-iso.int.open.paas.redhat.com/?a731e366cfb491a7#4PrHxwOeX/2POxx9QhtybwVEvm0VR0xfqj6Bn9d7Z/s=

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -1186,6 +1186,7 @@ Feature: Multus-CNI related scenarios
     # Make sure that the multus is enabled
     Given the multus is enabled on the cluster
     Given the default interface on nodes is stored in the :default_interface clipboard
+    And evaluation of `node.name` is stored in the :target_node clipboard
     # Create the net-attach-def via cluster admin
     Given I have a project
     When I run oc create as admin over "<%= BushSlicer::HOME %>/testdata/networking/multus-cni/NetworkAttachmentDefinitions/whereabouts-macvlan.yaml" replacing paths:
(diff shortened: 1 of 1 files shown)


anuragthehatter:
@weliang1 The change looks good however at line 1216 and 1237 , we cannot define static IPs because net-attach-def has a defined a range ```range": "192.168.22.100/24"``` an any IP can be allocated to the pods at any time. So this test will break intermttentally. WDYT. May be just grep ```192.168.22```

weliang1:
> @weliang1 The change looks good however at line 1216 and 1237 , we cannot define static IPs because net-attach-def has a defined a range `range": "192.168.22.100/24"` and any IP can be allocated to the pods at any time. So this test will break intermttentally. WDYT. May be just grep `192.168.22`

@anuragthehatter Very good catch. Right now whereabouts always allocate 192.168.22.100 and 192.168.22.101 to two pods when using 192.168.22.100/24. Update 192.168.22.100/24 to 192.168.22.100/31 will guarantee the whereabout will only assign pods to have 192.168.22.100 and 192.168.22.101, then my verification steps using static 192.168.22.100 and 192.168.22.101 will always pass.

anuragthehatter:
Sounds good.

/LGTM

weliang1:
> > @weliang1 The change looks good however at line 1216 and 1237 , we cannot define static IPs because net-attach-def has a defined a range `range": "192.168.22.100/24"` and any IP can be allocated to the pods at any time. So this test will break intermttentally. WDYT. May be just grep `192.168.22`
> 
> @anuragthehatter Very good catch. Right now whereabouts always allocate 192.168.22.100 and 192.168.22.101 to two pods when using 192.168.22.100/24. Update 192.168.22.100/24 to 192.168.22.100/31 will guarantee the whereabout will only assign pods to have 192.168.22.100 and 192.168.22.101, then my verification steps using static 192.168.22.100 and 192.168.22.101 will always pass.

PR #1106, opened 2020-05-15
Imagepruner

pruan-rht:


lib/openshift/image_pruner_imageregistry_operator_openshift_io.rb
```diff
@@ -0,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class ImagePrunerImageregistryOperatorOpenshiftIo < ClusterResource
+    RESOURCE = 'imagepruner.imageregistry.operator.openshift.io'
```
akostadinov, line 5: I expect this still needs to be `imagepruners.imageregistry.operator.openshift.io`

pruan-rht: @akostadinov interestingly it works without it...but I updated it anyway.   PTAL.  Thanks!

akostadinov: @pruan-rht , presently doing proper plural of the item should not break often because it would only be used with API calls. But I have not seen us do API calls for a long time. Still better to have it right in case we start doing that.

PR #1113, opened 2020-05-18
Use quay.io

jhou1:
Image is mirrored to quay, now updating automation to remove reference in docker.io.
@sunilcio @weinliu @lyman9966 PTAL.


akostadinov:
/lgtm

@pruan-rht , same about alternative architectures here too. Would be very helpful if you can ask guys what is needed if anything so these scenarios to work on alternative architectures.

pruan-rht:
@akostadinov yeah, I'm following up with ART to see what's up.

PR #1114, opened 2020-05-18
Add create deployment for kubectl

kasturinarra:
Adding create deployment for kubectl

### DIFF
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -186,7 +186,6 @@
     :restart: --restart=<value>
     :save_config: --save-config=<value>
     :schedule: --schedule=<value>
-    :sleep:  -- sleep <value>
     :template: --template=<value>
     :validate: --validate=<value>
 :create_deployment:
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -186,6 +186,7 @@
     :restart: --restart=<value>
     :save_config: --save-config=<value>
     :schedule: --schedule=<value>
+    :sleep:  -- sleep <value>
     :template: --template=<value>
     :validate: --validate=<value>
 :create_deployment:
--- a/lib/rules/cli/4.4.yaml
+++ b/lib/rules/cli/4.4.yaml
@@ -186,7 +186,6 @@
     :restart: --restart=<value>
     :save_config: --save-config=<value>
     :schedule: --schedule=<value>
-    :sleep:  -- sleep <value>
     :template: --template=<value>
     :validate: --validate=<value>
 :create_deployment:
--- a/lib/rules/cli/4.4.yaml
+++ b/lib/rules/cli/4.4.yaml
@@ -186,6 +186,7 @@
     :restart: --restart=<value>
     :save_config: --save-config=<value>
     :schedule: --schedule=<value>
+    :sleep:  -- sleep <value>
     :template: --template=<value>
     :validate: --validate=<value>
 :create_deployment:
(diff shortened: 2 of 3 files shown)


kasturinarra:

kasturinarra:

using kubectl 4.5: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1345/console
using kubectl 4.4: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1346/console

zhouying7780:
@liangxia Could you please review and merge , thanks .

lib/rules/cli/kubectl-4.1.yaml
```diff
@@ -145,6 +145,17 @@
...
     :from_file: --from-file=<value>
     :from_literal: --from-literal=<value>
+:create_deployment:
+  :cmd: kubectl create deployment <name>
+  :options:
+    :allow_missing_template_keys: --allow-missing-template-keys=<value>
+    :dry_run: --dry-run=<value>
+    :generator: --generator=<value>
+    :image: --image=<value>
+    :output: --output <value>
```
pruan-rht, line 155: @kasturinarra should it be `--output=<value>`?

kasturinarra: @pruan-rht I took create deloyment which already exist in oc-4.1.yaml and did similar one for kubectl, i think this should be fine.

pruan-rht: @kasturinarra your assumption is that the 4.1 yaml is correct in the first place.  Although it works, but in reality if you look at the help file it shows that a `=` is present like other commands.  

> Create a deployment with the specified name.
> 
> Aliases:
> deployment, deploy
> 
> Usage:
>   oc create deployment NAME --image=image [--dry-run] [flags]
> 
> Examples:
>   # Create a new deployment named my-dep that runs the busybox image.
>   oc create deployment my-dep --image=busybox
> 
> Options:
>       --allow-missing-template-keys=true: If true, ignore any errors in templates when a field or map key is missing in
> the template. Only applies to golang and jsonpath output formats.
>       --dry-run=false: If true, only print the object that would be sent, without sending it.
>       --generator='': The name of the API generator to use.
>       --image=[]: Image name to run.
>   -o, --output='': Output format. One of:
> json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file.
>       --save-config=false: If true, the configuration of current object will be saved in its annotation. Otherwise, the
> annotation will be unchanged. This flag is useful when you want to perform kubectl apply on this object in the future.
>       --template='': Template string or path to template file to use when -o=go-template, -o=go-template-file. The
> template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
>       --validate=false: If true, use a schema to validate the input before sending it

pruan-rht: @kasturinarra I would suggest to add the `=` and correct it going forward to be consistent with the other options which all have the `=`.

kasturinarra: @pruan-rht changed as per your comments here, please help review, thanks !!

lib/rules/cli/4.1.yaml
```diff
@@ -186,6 +186,7 @@
     :restart: --restart=<value>
     :save_config: --save-config=<value>
     :schedule: --schedule=<value>
+    :sleep:  -- sleep <value>
```
pruan-rht: @kasturinarra should be `--sleep=<value>`?

lib/rules/cli/4.4.yaml
```diff
@@ -186,6 +186,7 @@
     :restart: --restart=<value>
     :save_config: --save-config=<value>
     :schedule: --schedule=<value>
+    :sleep:  -- sleep <value>
```
pruan-rht: @kasturinarra should be `--sleep=<value>`?

PR #1116, opened 2020-05-18
check inf first and then perform operation

anuragthehatter:
Quick fix to avoid few test execution premature errors we have been seeing few times or rarely due to this particular step. In test env like all-in-one (node acting as both worker and master). a test interface or bridge is tried to be deleted couple of times and hence tests exit prematurely when it tries to delete an already deleted interface or bridge. This will solve this issue. Thanks
re-ran the tests to make sure
@rbbratta @zhaozhanqi @weliang1 @huiran0826

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -733,7 +733,7 @@ Given /^the bridge interface named "([^"]*)" is deleted from the "([^"]*)" node$
   ensure_admin_tagged
   node = node(node_name)
   host = node.host
-  @result = host.exec_admin("/sbin/ip link delete #{bridge_name}")
+  @result = host.exec_admin("if ip addr show  #{bridge_name};then ip link delete #{bridge_name};else echo #{bridge_name} does not exist on this node;fi")
   raise "Failed to delete bridge interface" unless @result[:success]
 end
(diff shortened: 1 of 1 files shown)


features/step_definitions/networking.rb
```diff
@@ -733,7 +733,7 @@
   ensure_admin_tagged
   node = node(node_name)
   host = node.host
-  @result = host.exec_admin("/sbin/ip link delete #{bridge_name}")
+  @result = host.exec_admin("if ip addr show  #{bridge_name};then ip link delete #{bridge_name};else echo #{bridge_name} does not exist on this node;fi")
```
pruan-rht, line 736: @anuragthehatter make make it kind of hard to read...would it be possible to pull bash script out, assign it a variable and then call that instead?

anuragthehatter: Thanks Peter, Fixed. Not seeing any premature exists errors even if inf doesn't exist on a node https://privatebin-it-iso.int.open.paas.redhat.com/?a7333e1ec1d3b598#DfdLk89I4fMQ3rx6CKcd36kgC5Csj7REN1LA1JNKFoA=

PR #1124, opened 2020-05-19
45 rules fixes

yapei:


lib/rules/web/admin_console/4.5/projects.xyaml
```diff
@@ -44,9 +44,10 @@ send_delete_string:
     op: send_keys <resource_name>
 create_project_from_get_started_instru:
   params:
-    button_text: Create Project
+    button_text: Create a new project
```
yapei, line 47: The button text changed

lib/rules/web/admin_console/4.5/goto.xyaml
```diff
@@ -6,6 +6,7 @@ goto_project_status:
   url: status/ns/<project>
 goto_resource_page_under_default_project:
   url: k8s/ns/default/<resource_url_name>
+  action: wait_box_loaded
```
yapei, line 9: added `wait_box_loaded` to make sure page is loaded completely before going next step otherwise we will see errors in Runner-v3 #92866

PR #1126, opened 2020-05-19
GCE IaaS connector support

akostadinov:


akostadinov:
/assign @qinpingli 

This is not full support but covers initial requirement for GCE. Feel welcome to improve implementation to use API for getting credentials instead of accessing HTTP API from node.

Also welcome are implementations for the rest of the providers.

qinpingli:
@akostadinov Thank you! GCE part lgtm.

PR #1129, opened 2020-05-20
Fix OCP-23030

wewang58:
In some versions, there's no some resources, so fix it. here's [pass log](http://pastebin.test.redhat.com/866680)
@openshift/devexp-qe please check it, thanks

### DIFF
--- a/features/registry/registry.feature
+++ b/features/registry/registry.feature
@@ -80,23 +80,22 @@ Feature: Testing registry
       | resource_name | image-registry |
       | o             | yaml           |
     Then the step should succeed
-    And the output should contain:
-      | name: system:registry                      |
-      | resource: clusterroles                     |
-      | name: registry-registry-role               |
-      | resource: clusterrolebindings              |
-      | name: registry                             |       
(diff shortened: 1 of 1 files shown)


features/registry/registry.feature
```diff
@@ -80,23 +80,22 @@ Feature: Testing registry
...
+      | (resource: serviceaccounts)?                  |
+      | (name: image-registry-certificates)?          |
+      | (resource: configmaps)?                       |
+      | (name: image-registry-private-configuration)? |
+      | (resource: secrets)?                          | 
+      | (name: image-registry)?                       |
+      | (resource: services)?                         |
+      | (name: node-ca)?                              |
+      | (resource: daemonsets)?                       |
+      | (resource: deployments)?                      |
```
liangxia, line 98: No matter those resources exist or not, the scenario will pass when using like above. Then what's the meaning of leaving them there ?

wewang58: some versions it's not include these, but older versions will had these.

PR #1131, opened 2020-05-20
Auto OCP-15473

huiran0826:
Originally I combined OCP-21812 and  OCP-15473 into one auto script case.
Due to PR https://github.com/openshift/verification-tests/pull/933, we cannot use one auto test case script to track two or more cases in Polarion .  So move the scripts for OCP-15473 out from OCP-21812. 

https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1384/console

### DIFF
--- a/features/networking/egress-ip.feature
+++ b/features/networking/egress-ip.feature
@@ -145,38 +145,7 @@ Feature: Egress IP related features
       | bash | -c | ip address show <%= cb.interface %> |
     Then the step should succeed
     And the output should contain "<%= cb.valid_ip %>"
-    """
-
-    #Check related iptables added
-    When I run command on the "<%= node.name%>" node's sdn pod:
-      | bash | -c | iptables-save \| grep "<%= cb.valid_ip %>" |
-    Then the step should succeed
-    And the output should contain:
-      | OPENSHIFT-MASQUERADE      |
-      | OPENSHIFT-FIREWALL-ALLOW  |
-
-    #check related openflow added
-    When I run command on the "<%= node.name%>" node's sdn pod:
-      | bash | -c | ovs-ofctl dump-flows br0 -O OpenFlow13 \| grep table=100 |
-    Then the step should succeed
-    And the output should contain "reg0=0x"
-
-    #Remove egress ip from namespace
-    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
-      | {"egressIPs": null} |
-
-    # Check related iptables and openlows removed
-    And I wait up to 30 seconds for the steps to pass:
-    """
-    When I run command on the "<%= node.name%>" node's sdn pod:
-      | bash | -c | iptables-save \| egrep "OPENSHIFT-FIREWALL-ALLOW\|OPENSHIFT-MASQUERADE" |
-    Then the step should succeed
-    And the output should not contain:
-      |  <%= cb.valid_ip %> |
-    When I run command on the "<%= node.name%>" node's sdn pod:
-      | bash | -c | ovs-ofctl dump-flows br0 -O OpenFlow13 \| grep table=100 |
-    Then the step should succeed
-    And the output should not contain "reg0=0x"
+    And the output should not contain "<%= cb.newip %>"
     """
 
   # @author huirwang@redhat.com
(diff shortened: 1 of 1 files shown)


zhaozhanqi:
@huiran0826 
could you check what's this scenario `Should remove the egressIP from the array if it was not being used` is for? should not show the `new_ip` in `ip a`. if so. please add step to check there is no `new_ip` when check `ip a`.

huiran0826:
> @huiran0826
> could you check what's this scenario `Should remove the egressIP from the array if it was not being used` is for? should not show the `new_ip` in `ip a`. if so. please add step to check there is no `new_ip` when check `ip a`.

@zhaozhanqi   Good catch, fixed it.

zhaozhanqi:
The steps LGTM
BTW,  those two cases are `destructive`? I did not see any operation will affect other project or cluster.

huiran0826:
> The steps LGTM
> BTW, those two cases are `destructive`? I did not see any operation will affect other project or cluster.

As if the two egressIP cases run in parellel, it may affect each other. It is better to let them run one by one.

zhaozhanqi:
> > The steps LGTM
> > BTW, those two cases are `destructive`? I did not see any operation will affect other project or cluster.
> 
> As if the two egressIP cases run in parellel, it may affect each other. It is better to let them run one by one.

hmm, I don't think they will affect each other since they are using different namespaces. however, if they affect each other. I doubt if it should be a bug.

huiran0826:
> > > The steps LGTM
> > > BTW, those two cases are `destructive`? I did not see any operation will affect other project or cluster.
> > 
> > 
> > As if the two egressIP cases run in parellel, it may affect each other. It is better to let them run one by one.
> 
> hmm, I don't think they will affect each other since they are using different namespaces. however, if they affect each other. I doubt if it should be a bug.

1. Firstly for each egressIP cases, after case finish, there is a clean up step, will remove the egressIPs from that node. So if two cases run in parellel, egressIPs will be randomly patched to some nodes, maybe on same node. In this case, if the one case finish early and remove the egressIPs from the node, may cause another failing.

2. For case in this PR OCP-15473, it will check ovs openflow.  It really needs the env to be clean. If some other case patch egressIPs or remove egressIPs on the same node, may cause the checking failing.

zhaozhanqi:
> > > > The steps LGTM
> > > > BTW, those two cases are `destructive`? I did not see any operation will affect other project or cluster.
> > > 
> > > 
> > > As if the two egressIP cases run in parellel, it may affect each other. It is better to let them run one by one.
> > 
> > 
> > hmm, I don't think they will affect each other since they are using different namespaces. however, if they affect each other. I doubt if it should be a bug.
> 
> 1. Firstly for each egressIP cases, after case finish, there is a clean up step, will remove the egressIPs from that node. So if two cases run in parellel, egressIPs will be randomly patched to some nodes, maybe on same node. In this case, if the one case finish early and remove the egressIPs from the node, may cause another failing.
> 2. For case in this PR OCP-15473, it will check ovs openflow.  It really needs the env to be clean. If some other case patch egressIPs or remove egressIPs on the same node, may cause the checking failing.
Do you mean the clear up step will clear all `egressip` on node even if the other case is using the egressip?

huiran0826:
> > > > > The steps LGTM
> > > > > BTW, those two cases are `destructive`? I did not see any operation will affect other project or cluster.
> > > > 
> > > > 
> > > > As if the two egressIP cases run in parellel, it may affect each other. It is better to let them run one by one.
> > > 
> > > 
> > > hmm, I don't think they will affect each other since they are using different namespaces. however, if they affect each other. I doubt if it should be a bug.
> > 
> > 
> > 
> > 1. Firstly for each egressIP cases, after case finish, there is a clean up step, will remove the egressIPs from that node. So if two cases run in parellel, egressIPs will be randomly patched to some nodes, maybe on same node. In this case, if the one case finish early and remove the egressIPs from the node, may cause another failing.
> > 2. For case in this PR OCP-15473, it will check ovs openflow.  It really needs the env to be clean. If some other case patch egressIPs or remove egressIPs on the same node, may cause the checking failing.
> 
> Do you mean the clear up step will clear all `egressip` on node even if the other case is using the egressip?

Yes, our register clean up steps is: 
oc patch --type=merge hostsubnet node_name -p '{"egressIPs":null}'.

zhaozhanqi:
> > > > > > The steps LGTM
> > > > > > BTW, those two cases are `destructive`? I did not see any operation will affect other project or cluster.
> > > > > 
> > > > > 
> > > > > As if the two egressIP cases run in parellel, it may affect each other. It is better to let them run one by one.
> > > > 
> > > > 
> > > > hmm, I don't think they will affect each other since they are using different namespaces. however, if they affect each other. I doubt if it should be a bug.
> > > 
> > > 
> > > 
> > > 1. Firstly for each egressIP cases, after case finish, there is a clean up step, will remove the egressIPs from that node. So if two cases run in parellel, egressIPs will be randomly patched to some nodes, maybe on same node. In this case, if the one case finish early and remove the egressIPs from the node, may cause another failing.
> > > 2. For case in this PR OCP-15473, it will check ovs openflow.  It really needs the env to be clean. If some other case patch egressIPs or remove egressIPs on the same node, may cause the checking failing.
> > 
> > 
> > Do you mean the clear up step will clear all `egressip` on node even if the other case is using the egressip?
> 
> Yes, our register clean up steps is:
> oc patch --type=merge hostsubnet node_name -p '{"egressIPs":null}'.

ah. I see.  I checked the step `Given /^the valid egress IP is added to the#{OPT_QUOTED} node` will also replace the egressip.  just thought this will not remove other iptables and openflow of namespace.
thanks for checking this. /lgtm

PR #1145, opened 2020-05-21
Forcing scenario to execute on specific node only

anuragthehatter:
Apparently this case depicted few failures when it ran on mixed node clusters. The reason is uniform cluster like with all RHCOS nodes, they share same master interface naming convention but if we add a RHEL worker to an existing uniform cluster, that might come up with different master interface name. We want to make sure we check IPVLAN functionality on a specific node only and that node master interface should be used in further configurations. Just a small change to alleviate CI failures.
@pruan-rht @weliang1 @rbbratta @zhaozhanqi @huiran0826 

Re-ran Logs: https://privatebin-it-iso.int.open.paas.redhat.com/?27da8e5f3fd5c146#TU8Y4xibWlmjhxQKjjwFIFHZqSfJ4o0bN5JPOIv3RPI=

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -1134,7 +1134,6 @@ Feature: Multus-CNI related scenarios
     # Make sure that the multus is enabled
     Given the multus is enabled on the cluster
     And the default interface on nodes is stored in the :default_interface clipboard
-    And evaluation of `node.name` is stored in the :target_node clipboard
     #Storing default interface mac address for comparison later with pods macs
     Given I run commands on the host:
       | ip addr show <%= cb.default_interface %> |
@@ -1151,9 +1150,9 @@ Feature: Multus-CNI related scenarios
     #Creating various pods and making sure their mac matches to default inf and they get unique IPs assigned
     #Creating pod1 absorbing above net-attach-def
     When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
-      | ["metadata"]["name"]                                       | pod1                    |
-      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | myipvlan76              |
-      | ["spec"]["nodeName"]                                       | <%= cb.target_node %>   |
+      | ["metadata"]["name"]                                       | pod1             |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | myipvlan76       |
+      | ["spec"]["nodeName"]                                       | <%= node.name %> |
     Then the step should succeed
     And the pod named "pod1" becomes ready
     When I execute on the pod:
@@ -1165,9 +1164,9 @@ Feature: Multus-CNI related scenarios
     
     #Creating pod2 absorbing above net-attach-def
     When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
-      | ["metadata"]["name"]                                       | pod2                  |
-      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | myipvlan76            |
-      | ["spec"]["nodeName"]                                       | <%= cb.target_node %> |
+      | ["metadata"]["name"]                                       | pod2             |
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | myipvlan76       |
+      | ["spec"]["nodeName"]                                       | <%= node.name %> |
     Then the step should succeed
     And the pod named "pod2" becomes ready
     When I execute on the pod:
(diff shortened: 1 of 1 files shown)


features/networking/multus.feature
```diff
@@ -1154,7 +1154,7 @@ Feature: Multus-CNI related scenarios
     When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
       | ["metadata"]["name"]                                       | pod1                    |
       | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | myipvlan76              |
-      | ["spec"]["nodeName"]                                       | <%= cb.nodes[0].name %> |
+      | ["spec"]["nodeName"]                                       | "<%= cb.target_node %>" |
```
pruan-rht, line 1157: @anuragthehatter do we need the `"`?

features/networking/multus.feature
```diff
@@ -1168,7 +1168,7 @@ Feature: Multus-CNI related scenarios
     When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/multus-cni/Pods/generic_multus_pod.yaml" replacing paths:
       | ["metadata"]["name"]                                       | pod2                    |
       | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | myipvlan76              |
-      | ["spec"]["nodeName"]                                       | <%= cb.nodes[0].name %> |
+      | ["spec"]["nodeName"]                                       | "<%= cb.target_node %>" |
```
pruan-rht, line 1171: @anuragthehatter do we need the `"`?

features/networking/multus.feature
```diff
@@ -1135,17 +1135,17 @@ Feature: Multus-CNI related scenarios
     Given the multus is enabled on the cluster
     And I store all worker nodes to the :nodes clipboard
```
huiran0826, line 1136: Do we still need this step?

features/networking/multus.feature
anuragthehatter: Nope, taken care. Thx

features/networking/multus.feature
anuragthehatter: Thats redundant. Taken care. thx

features/networking/multus.feature
```diff
@@ -1133,19 +1133,18 @@ Feature: Multus-CNI related scenarios
   Scenario: Create pod with Multus ipvlan CNI plugin	
     # Make sure that the multus is enabled
     Given the multus is enabled on the cluster
-    And I store all worker nodes to the :nodes clipboard
     And the default interface on nodes is stored in the :default_interface clipboard
+    And evaluation of `node.name` is stored in the :target_node clipboard
```
rbbratta, line 1137: side-question:  I'm not sure if we always need to use a clipboard?  I guess if we implicitly change nodes in some steps, but if we don't change nodes can we use `<%= node.name %>` directly?

Does it work to do:
```
| ["spec"]["nodeName"]                                       | <%= node.name %>   |
```

features/networking/multus.feature
```diff
@@ -1178,7 +1177,7 @@ Feature: Multus-CNI related scenarios
     And evaluation of `@result[:response].match(/\h+:\h+:\h+:\h+:\h+:\h+/)[0]` is stored in the :pod2_net1_mac clipboard
     And the expression should be true> cb.pod2_net1_mac==cb.default_interface_mac
     And the expression should be true> !(cb.pod2_net1_ip==cb.pod1_net1_ip)
-
+
```
rbbratta, line 1180: please trim whitespace

features/networking/multus.feature
```diff
@@ -1133,19 +1133,18 @@ Feature: Multus-CNI related scenarios
   Scenario: Create pod with Multus ipvlan CNI plugin	
     # Make sure that the multus is enabled
     Given the multus is enabled on the cluster
-    And I store all worker nodes to the :nodes clipboard
     And the default interface on nodes is stored in the :default_interface clipboard
+    And evaluation of `node.name` is stored in the :target_node clipboard
     #Storing default interface mac address for comparison later with pods macs
-    Given I use the "<%= cb.nodes[0].name %>" node
-    And I run commands on the host:
+    Given I run commands on the host:
```
weliang1, line 1139: Before this line, should script still need the step of Given I use the "<%= cb.target_node %>" node? This step will get MAC from the host, and the rest script will check pod's MAC will be same as the MAC from the cb.target_node. Hopefully the host here is just the cb.target_node, otherwise the MAC checking will fail.

features/networking/multus.feature
anuragthehatter: Yep, we can definitely use node.name here as the behavior is driven here on a single node only. We generally save ```cb``` in case there is too much detours on various nodes which is clearly not  in this case. Thanks!

features/networking/multus.feature
anuragthehatter: @weliang1 This is fine. In this case ```cb.nodes[0].name``` will be the host only. If you are using these steps back to back without changing any course, your ```node.name``` will always be your host. Fore more info
Ref: https://github.com/openshift/verification-tests/blob/master/lib/openshift/node.rb#L27

PR #1151, opened 2020-05-22
ignore STDERR when running node commands

akostadinov:


akostadinov:
/assign @qinpingli 

Please check if this fix the issue you have seen. Not sure why it didn't manifest for me.

PR #1153, opened 2020-05-25
add debug step to isolate the auth issue

lihongan:
This step is for further debugging the `Error getting bearer token` issue during login procedure.
The new added step is harmless and won't block the test execution but just print the result of HTTP.GET of console url before starting the login with normal user.  

Example:
```
   Given I have a project                                                                # features/step_definitions/project.rb:7
      [02:06:29] INFO> HTTP GET https://console-openshift-console.apps.hongli-pl212.qe.gcp.devcluster.openshift.com
      [02:06:35] INFO> HTTP GET took 5.942 sec: 200 OK | text/html 3372 bytes
      
      [02:06:35] INFO> HTTP GET https://api.hongli-pl212.qe.gcp.devcluster.openshift.com:6443/.well-known/oauth-authorization-server
      [02:06:41] INFO> HTTP GET took 5.707 sec: 200 OK | application/json 672 bytes
      
      [02:06:41] INFO> HTTP GET testuser-46@https://oauth-openshift.apps.hongli-pl212.qe.gcp.devcluster.openshift.com/oauth/authorize
      [02:06:47] INFO> HTTP GET took 5.925 sec: 503 Service Unavailable
      [02:06:47] ERROR> #<RuntimeError: Error getting bearer token, see log>
```
The example output can tell that Auth has issue but console is still accessible.

@xingxingxia @jhou1 @pruan-rht please help take a look. thanks

### DIFF
--- a/lib/login.rb
+++ b/lib/login.rb
@@ -11,6 +11,15 @@ module BushSlicer
     # @param [String] user the username we want token for
     # @return [String]
     def new_token_by_password(user:, password:, env:)
+      # just print HTTP GET result of console route for further debugging
+      # should check if console route accessible when reporting "Error getting bearer token"
+      # if console_url is accessible then it is likely Auth issue
+      # if console_url is inaccessible then likely network related issue including ingress, snd or platform flake
+      console_url = env.api_endpoint_url.delete_suffix(':6443').gsub(/api/, 'console-openshift-console.apps')
+      opts = {:url => console_url, :method => "GET" }
+      opts[:proxy] = env.client_proxy if env.client_proxy
+      debug_res = Http.request(**opts)
+
       # try challenging client auth
       res = oauth_bearer_token_challenge(
         server_url: env.api_endpoint_url,


lihongan:
see also: https://issues.redhat.com/browse/OCPQE-526

and here is an example output without the changes:
```
GivenI have a project ==>@  features/step_definitions/project.rb:7
01:51:56 INFO> HTTP GET https://api.qeci-4709.qe.devcluster.openshift.com:6443/.well-known/oauth-authorization-server
01:51:56 INFO> HTTP GET took 0.016 sec: 200 OK | application/json 651 bytes
01:51:56 INFO> HTTP GET testuser-16@https://oauth-openshift.apps.qeci-4709.qe.devcluster.openshift.com/oauth/authorize
01:51:56 INFO> HTTP GET took 0.013 sec: 503 Service Unavailable
01:51:56 ERROR> #<RuntimeError: Error getting bearer token, see log>
```

xingxingxia:
@'ing @rhpmali for knowing

lihongan:
Any comments?  cc @rhpmali @akostadinov

rhpmali:
Adding debug line here is really good.
lgtm

lib/login.rb
```diff
@@ -11,6 +11,15 @@ module LoginIncl
...
     # @return [String]
     def new_token_by_password(user:, password:, env:)
+      # just print HTTP GET result of console route for further debugging
+      # should check if console route accessible when reporting "Error getting bearer token"
+      # if console_url is accessible then it is likely Auth issue
+      # if console_url is inaccessible then likely network related issue including ingress, snd or platform flake
+      console_url = env.api_endpoint_url.delete_suffix(':6443').gsub(/api/, 'console-openshift-console.apps')
+      opts = {:url => console_url, :method => "GET" }
+      opts[:proxy] = env.client_proxy if env.client_proxy
+      debug_res = Http.request(**opts)
```
pruan-rht, line 21: @lihongan you defined this `debug_res` but it's not used anywhere...not a big deal, but just so you aware.

akostadinov:
@lihongan @pruan-rht , I think this adds a lot of useless debug output for every login attempt. Doesn't normal return status from getting `api_endpoint_url` tell you what issue is. If it is timeout or name issue or whatever or just bad password?

Do you have any example where console output was needed?

lihongan:
@akostadinov I think my example output has told that why `api_endpoint_url` is not enough for locating the issue. Let me try to explain it clearer.
1. The connection to API and ingress are very different, they are using two different Load Balancing and different ports and different path inside the cluster. Take AWS for example, API uses NLB (port 6443) that redistributes the traffic to all Master nodes, but ingress uses ELB (port 80/443) that redistributes the traffic to all nodes. In AWS Route 53 there are also have two records each for API and ingress (*.apps) that pointing to corresponding LB.
2. The console, auth and monitoring (prometheus, grafana...) components expose their service by route, so the connections go through the ingress path (not API).
3. Both example outputs show that the API ( `api_endpoint_url`) is accessible but auth route return 503. 

API (the url is api.<cluster_name>.<base_domain>)
```
01:51:56 INFO> HTTP GET https://api.qeci-4709.qe.devcluster.openshift.com:6443/.well-known/oauth-authorization-server
01:51:56 INFO> HTTP GET took 0.016 sec: 200 OK | application/json 651 bytes
```
Auth (the url is *.apps.<cluster_name>.<base_domain>)
```
01:51:56 INFO> HTTP GET testuser-16@https://oauth-openshift.apps.qeci-4709.qe.devcluster.openshift.com/oauth/authorize
01:51:56 INFO> HTTP GET took 0.013 sec: 503 Service Unavailable
```
4. Considering the `Error getting bearer token` message, it is also hard to tell it is auth problem or ingress path issue. So I'd like to check console route to isolate the issue. That's just the purpose of this PR.
5. How to analysis the output: 
    now we have three HTTP outputs: console, api and auth. If three failed then might be network connectivity, DNS, slave flake. if both console and auth failed then might be ingress issue, if only one failed then just need to check that components. So I don't think it is useless.
6. I'd like to have the PR running for some days and to try to get more info during the CI execution, then I will revert it later.
7. Yes, it is added to every login attempt, but won't cost too much time (equivalent to a retry)

Anyway, all comments are welcome, if anyone has more ideas to resolve the `Error getting bearer token` issue, or give more clear info to debug the issue and reduce our effort to check it again and again.

akostadinov:
@lihongan , Your explanation makes sense to me. But to avoid log clutter and performance hit, I'd suggest 

1. open console route only in case API or auth failed (for shorter code you can put the web console query code in a separate method or use `ensure` block inside the method)
2. instead of `env.api_endpoint_url.delete_suffix(':6443').gsub(/api/, 'console-openshift-console.apps')` use `env.web_console_url`.
3. a little nicer to use `Http.get` method instead of `Http.request` method

With the above points implemented you can leave code permanently in repo.

lihongan:
@akostadinov Thank you for you valuable comments and totally agree with you about the item 1 and 3. I'm thinking about updating them.
For the second one, indeed I've tried to use `env.web_console_url` and `admin_console_url`, but the former is same to `env.api_endpoint_url`. The latter is the console route that I really want but it needs more steps to get it from `openshift-console` project, more details see:
https://github.com/openshift/verification-tests/blob/8e8ce110c5ed79545bff696f3ed78ac1d7a74751/lib/environment.rb#L173-L189

akostadinov:
@lihongan ,  extra steps is no issue because they will execute only if login fails in the first place. The bigger issue is actually that in case you can't access API, you can't also run the extra steps :)

But maybe in your code it would be worth checking whether `evn.opts[:admin_console_url]` has been set and do `gsub` only in case it wasn't.

lihongan:
@akostadinov I'm trying to address each of your comments in the https://github.com/openshift/verification-tests/pull/1201, could you please take a look?
Thanks.

PR #1154, opened 2020-05-25
added support for installation of metering via OperatorHub UI

pruan-rht:


pruan-rht:
@yapei @akostadinov @liangxia PTAL.  Thanks!

lib/rules/web/admin_console/4.5/catalog.xyaml
```diff
@@ -129,13 +129,17 @@ click_subscribe_button:
...
+    button_text: Install
+  action: click_button_text
 subscribe_using_default_params:
   action: wait_box_loaded
   action: click_subscribe_button
 set_custom_channel_and_subscribe:
   action: wait_box_loaded
   action: choose_update_channel
-  action: click_subscribe_button
+  action: click_install_button
```
yapei, line 142: in 4.5 `click_subscribe_button` just click on a button with text `Install`

```
click_subscribe_button:
  params:
    button_text: Install
  action: click_button_text
```

pruan-rht: @yapei hmm...in my branch that's not defined.  I guess it must've been udpated before I started this.

pruan-rht: @yapei, yeah that's the case, my orignal branch didn't have that.  Thanks for pointing it out, I'll revet that change
```
click_subscribe_button:
  params:
    button_text: Subscribe
  action: click_button_text
```

features/step_definitions/metering.rb
```diff
@@ -1,23 +1,27 @@
 # helper step for metering scenarios
 require 'ssl'
+
 # For metering automation, we just install once and re-use the installation
 # unless the scenarios are for testing installation
 # checks that metering service is already installed
-Given /^metering service has been installed successfully(?: using (ansible|shell script|OLM))?$/ do |method|
+# shell => ./bin/hack.sh
```
liangxia, line 7: Is it still valid ?

liangxia: Then perhaps we need to update line 10, or else we are not able to use it.

pruan-rht: @liangxia we plan to phase that `hack.sh` script out.  That's why I took it out as an option.

liangxia: Or we should remove this line as well ?  It is confusing that we leave it there but removed it from line 10

pruan-rht: line removed... PTAL again. thx!

features/step_definitions/metering.rb
```diff
@@ -27,19 +30,17 @@
...
-    cb[:metering_namespace] = project(namespace)
     metering_name = metering_config.raw_resource['metadata']['name']
   end
 
   # change project context
-  unless cb.metering_namespace
+  unless metering_config(namespace).exists?
     case method
     when "shell script"
       step %Q/metering service is installed using shell script/
```
liangxia, line 40: Those lines should be removed as well.

pruan-rht: ok, updated.

PR #1161, opened 2020-05-26
fix oc new-app issue for console cases

yapei:


lib/rules/cli/4.1.yaml
```diff
@@ -573,6 +573,14 @@
...
     :strategy: --strategy=<value>
     :template: --template=<value>
+:new_app_as_dc:
+  :cmd: oc new-app
+  :options:
+    :code: --code=<value>
+    :image_stream: --image-stream=<value>
+    :name: --name=<value>
+    :source_spec: <value>
+    :app_repo: <value>
```
liangxia, line 583: @yapei  Could you sort them in alphanumeric order ?

yapei: sure, let me update

PR #1162, opened 2020-05-26
run cmd deprecated deploymentconfig, change to use dc template.

xiuwang:
Passed locally http://pastebin.test.redhat.com/868496
In 4.5, run cmd deprecated deploymentconfig, change to use dc template.

### DIFF
--- a/testdata/templates/ssh-git/ssh-git-dc.yaml
+++ b/testdata/templates/ssh-git/ssh-git-dc.yaml
@@ -1,0 +1,27 @@
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
+metadata:
+  name: git-server
+spec:
+  replicas: 1
+  selector:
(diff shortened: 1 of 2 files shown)


testdata/templates/ssh-git/ssh-git-dc.yaml
```diff
@@ -0,0 +1,27 @@
...
+    type: Rolling
+  template:
+    metadata:
+      labels:
+        name: git-server
+      name: git-server
+    spec:
+      containers:
+      - image: aosqe/ssh-git-server-openshift
+        imagePullPolicy: Always
```
liangxia, line 19: Perhaps we should pull image only when it does not exist ?

xiuwang: @liangxia Updated, please help review again, thanks

PR #1167, opened 2020-05-27
fix OCP-27609

sunzhaohua2:
Fix OCP-27609. Because it couldn't provision instance successful,  and the machine couldn't be deleted. Finally will finish prematurely. Like error below:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/97009/consoleFull

```
      [03:52:05] INFO> === End After Scenario: Scaling a machineset with providerSpec.publicIp set to true ===

1 scenario (1 passed)
8 steps (8 passed)
11m19.901s
```
@jhou1 @miyadav Help to take a look.


miyadav:
Thanks @sunzhaohua2 , good catch 
/lgtm

PR #1171, opened 2020-05-28
helper function to get all master_nodes for a cluster

pruan-rht:
works in tandum with https://github.com/openshift/cucushift/pull/7801, which depends on this PR

### DIFF
--- a/features/step_definitions/node.rb
+++ b/features/step_definitions/node.rb
@@ -603,6 +603,11 @@ Given /^ssh key for accessing nodes is copied to(?: the #{QUOTED} directory in)?
   raise "Error syncing files over to '#{pod.name}' pod '#{@result[:stderr]}'" unless @result[:success]
 end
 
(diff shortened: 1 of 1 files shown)


features/step_definitions/node.rb
```diff
@@ -603,6 +603,11 @@
   raise "Error syncing files over to '#{pod.name}' pod '#{@result[:stderr]}'" unless @result[:success]
 end
 
+Given /^I store all master nodes to the#{OPT_SYM} clipboard$/ do |cb_name|
```
liangxia, line 606: We have `Given /^I store the( schedulable| ready and schedulable)? (node|master|worker)s in the#{OPT_SYM} clipboard$/ do |state, role, cbname|`

pruan-rht: ok, didn't know that existed.  I'll just close the PR then.

PR #1173, opened 2020-05-28
Fix cronjob lib for 4.5

kasturinarra:
Fix cronjob lib for ocp4.5

### DIFF
--- a/lib/rules/cli/4.5.yaml
+++ b/lib/rules/cli/4.5.yaml
@@ -177,11 +177,14 @@
     :from_literal: --from-literal=<value>
     :name: <value>
 :create_cronjob:
-  :cmd: oc create cronjob <name>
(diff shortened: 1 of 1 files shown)


kasturinarra:
@zhouying7780  can you please help review, thanks !!

akostadinov:
please also update rule 4.1 file in case this is not already done

liangxia:
> please also update rule 4.1 file in case this is not already done

@akostadinov  We already have that in 4.1 and 4.4
```
$ grep -r 'create cron' *
lib/rules/cli/4.1.yaml:  :cmd: oc create cronjob <name> <global_options> <options>
lib/rules/cli/4.4.yaml:  :cmd: oc create cronjob <name> <global_options> <options>
lib/rules/cli/4.5.yaml:  :cmd: oc create cronjob <name>
```

PR #1175, opened 2020-05-28
Fix scenario name for OCP-31200

kasturinarra:
Fix scenario name for OCP-31200 so that it picks up the right test case


kasturinarra:
@zhouying7780  please help review, thanks !!

https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/98122/console

PR #1176, opened 2020-05-28
remove oc debug default image for 4.5

rbbratta:
use the  multus cluster image logic so we don't have to access quay.io

see #983

### DIFF
--- a/lib/rules/cli/4.5.yaml
+++ b/lib/rules/cli/4.5.yaml
@@ -319,7 +319,7 @@
      :output: --output=<value>
      :template: --template=<value>
 :debug:
-  :cmd: oc debug --image quay.io/openshifttest/storage@sha256:a05b96d373be86f46e76817487027a7f5b8b5f87c0ac18a246b018df11529b40 <global_options>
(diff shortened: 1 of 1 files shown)


rbbratta:
http://pastebin.test.redhat.com/869885

akostadinov:
@rbbratta , I see your comment in #983 - `tools:latest being present in openshift namespace`. Can we instead require whoever installs cluster to create that image stream? Or we do it if it doesn't exist? wdyt?

/lgtm

rbbratta:
> @rbbratta , I see your comment in #983 - `tools:latest being present in openshift namespace`. Can we instead require whoever installs cluster to create that image stream? Or we do it if it doesn't exist? wdyt?

I tried, but there is an operator somewhere that resets the imagestreamtag.  

I'll keep investigating, but we probably can't fight the operator.

`oc tag  -n openshift $(oc get pods -n openshift-multus -l app=multus -o jsonpath='{.items[0].spec.containers[?(@.name=="kube-multus")].image}') tools:latest`

PR #1181, opened 2020-05-29
Ensure every provisioned machine is deleted in test tear down

jhou1:
Even though we did our best effort to wait for machine to be deleted, occasionally a machine would escape the wait logic because its deletionTimestamp isn't added timely.

Previously: When a machine has a deletionTimestamp, wait for the machine to be removed in tear down. Works most of the time, problematic when deletionTimestamp isn't added timely before we invoke `machine.deleting?` method.

Changes: Compare the list of original machine names with list of current machine names, the `current - original` is the list of machines to be deleted. Admin ensures every of them is deleted. 

With this change test tear down could properly clean up every machine provisioned.

### DIFF
--- a/features/step_definitions/machine.rb
+++ b/features/step_definitions/machine.rb
@@ -57,23 +57,24 @@ end
 
 Then(/^admin ensures machine number is restored after scenario$/) do
   ensure_admin_tagged
-  cb.all_machines = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api"))
-  machine_orig_num = cb.all_machines.length
+  machine_names_orig = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api")).map { | m | m.name }
 
   teardown_add {
-    machines_waiting_delete = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api")).select { | machine |
-      machine.deleting?
-    }
+    machines = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api"))
 
-    machines_waiting_delete.each do | machine |
-      step %Q{I use the "openshift-machine-api" project}
-      step %Q{I wait for the resource "machine" named "#{machine.name}" to disappear within 1200 seconds}
-      step %Q{the step should succeed}
+    # compare current macine names and original machine names, new names are waiting delete
+    machine_names_current = machines.map { | m | m.name }
+    machine_names_waiting_del = machine_names_current - machine_names_orig
+    return if machine_names_waiting_del.empty?
+
+    machines_waiting_delete = []
+    machines.each do | machine |
+      machines_waiting_delete << machine if machine_names_waiting_del.include?(machine.name)
     end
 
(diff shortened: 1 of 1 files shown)


jhou1:
@sunzhaohua2 @miyadav PTAL, tested and works well.

features/step_definitions/machine.rb
```diff
@@ -57,23 +57,24 @@
 
 Then(/^admin ensures machine number is restored after scenario$/) do
   ensure_admin_tagged
-  cb.all_machines = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api"))
-  machine_orig_num = cb.all_machines.length
+  machine_names_orig = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api")).map { | m | m.name }
```
akostadinov, line 60: a little shortening can be done here with:
```
BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api")).map(&:name)
```
This is a shorter way to tell call `#name` method of the object. And only works when parameters are not needed.

features/step_definitions/machine.rb
```diff
@@ -57,23 +57,24 @@
...
-      machine.deleting?
-    }
+    machines = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api"))
 
-    machines_waiting_delete.each do | machine |
-      step %Q{I use the "openshift-machine-api" project}
-      step %Q{I wait for the resource "machine" named "#{machine.name}" to disappear within 1200 seconds}
-      step %Q{the step should succeed}
+    # compare current macine names and original machine names, new names are waiting delete
+    machine_names_current = machines.map { | m | m.name }
```
akostadinov, line 66: same

akostadinov:
@jhou1 , nice update.

I changed with a little shorter syntax and added a check for missing machines at end of scenario. But did not test the changes. Feel free to update if you don't like something. And merge at your discretion.

sunzhaohua2:
@jhou1  Thank you. There is another situation we need to consider, sometimes some machines stuct in Provisioning phase or phase is empty due to network issue or bug, these machines will stuct in Deleting phase when we want to delete them, this also could cause finish prematurely. Such as https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/99098/console. I tested OCP-27609 using your fixes. While this case is running, I created a machine with wrong credentialsSecret, the machine's phase will be empty due to bug https://bugzilla.redhat.com/show_bug.cgi?id=1805639 and couldn't be deleted. 
I think we can add a verification, if the machine has a deletionTimestamp, we first delete the machine finalizers to ensure that all machines can be deleted. But I have no idea how to add it now.

jhou1:
@sunzhaohua2 Thanks for the comment. This PR helps us delete every machine that should be deleted. If a machine is stuck and can't be deleted at all, that is usually caused by system/config error or production bug. We can discuss that in a separate PR or thread.

jhou1:
Thank @akostadinov for helping. I tested this today, works very well. I'm going to merge this based on above comments.

PR #1187, opened 2020-06-01
fix proxy environment injection and fix warnings

akostadinov:


zhouying7780:
@akostadinov  thanks very much!

PR #1190, opened 2020-06-02
merger OCP-31315 into OCP-28108

sunzhaohua2:
I want to merger OCP-31315 into OCP-28108, and I will delete case OCP-31315. @jhou1 @miyadav Please help to review, thanks.
 OCP-31315: DeletionCandidateOfClusterAutoscaler and ToBeDeletedByClusterAutoscaler taints should be deleted when min nodes is reached
OCP-28108: Cluster could automatically scale up and scale down with ClusterAutoscaler deployed

```
     [08:50:49] INFO> === End After Scenario: Cluster should automatically scale up and scale down with clusterautoscaler deployed ===

1 scenario (1 passed)
25 steps (25 passed)
13m21.889s
```

### DIFF
--- a/features/machine/autoscaler.feature
+++ b/features/machine/autoscaler.feature
@@ -51,6 +51,17 @@ Feature: Cluster Autoscaler Tests
     """
     Then the machineset should have expected number of running machines
 
+    # Check autoscaler taints are deleted when min node is reached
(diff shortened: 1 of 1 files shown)


miyadav:
May be you can mention in comments , the extra scenario it is validating , as the test case will be removed , other than that 

/lgtm

PR #1192, opened 2020-06-02
Automate OCP-31331

duanwei33:
Currently, iscsi provisioner doesn’t work smoothly during upgrade, this case is only performed on BareMetal before upgrade and after upgrade (v.s. OCP-23501):
1. The cluster operator storage should be in the correct status before and after the upgrade.
2. The cluster operator storage operator version should match the current cluster version after upgrade

[pass log](https://privatebin-it-iso.int.open.paas.redhat.com/?efbbf8b6ed07e308#zlCHYPHk07qHFuHZi+q8sFzaiilIWxZwOqBPok4/cCc=)

### DIFF
--- a/features/upgrade/storage/upgrade.feature
+++ b/features/upgrade/storage/upgrade.feature
@@ -139,3 +139,32 @@ Feature: Storage upgrade tests
       | ls | /mnt/storage |
     Then the output should contain:
       | test-upgrade |
+
+  # @author wduan@redhat.com
+  # @case_id OCP-31331
+  @upgrade-prepare
+  @users=upuser1,upuser2
+  @admin
+  Scenario: Cluster operator storage should be in correct status after upgrade - prepare
+    Given I switch to cluster admin pseudo user
+    # Check cluster operator storage should be in correct status
+    Given the expression should be true> cluster_operator('storage').condition(type: 'Progressing')['status'] == "False"
+    Given the expression should be true> cluster_operator('storage').condition(type: 'Available')['status'] == "True"
+    Given the expression should be true> cluster_operator('storage').condition(type: 'Degraded')['status'] == "False"
+    Given the expression should be true> cluster_operator('storage').condition(type: 'Upgradeable')['status'] == "True"
+
+  # @author wduan@redhat.com
+  @upgrade-check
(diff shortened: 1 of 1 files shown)


features/upgrade/storage/upgrade.feature
```diff
@@ -139,3 +139,32 @@ Feature: Storage upgrade tests
       | ls | /mnt/storage |
     Then the output should contain:
       | test-upgrade |
+
+  # @author wduan@redhat.com
+  # @case_id OCP-31331
+  @upgrade-prepare
+  @users=upuser1,upuser2
+  @admin
+  Scenario: Cluster operator storage should be in correct status after upgrade
```
pruan-rht, line 148: @duanwei33 I think you mean `before upgrade`, right?

features/upgrade/storage/upgrade.feature
```diff
@@ -139,3 +139,32 @@ Feature: Storage upgrade tests
...
+    Given I switch to cluster admin pseudo user
+    # Check cluster operator storage should be in correct status
+    Given the expression should be true> cluster_operator('storage').condition(type: 'Progressing')['status'] == "False"
+    Given the expression should be true> cluster_operator('storage').condition(type: 'Available')['status'] == "True"
+    Given the expression should be true> cluster_operator('storage').condition(type: 'Degraded')['status'] == "False"
+    Given the expression should be true> cluster_operator('storage').condition(type: 'Upgradeable')['status'] == "True"
+
+  # @author wduan@redhat.com
+  @upgrade-check
```
pruan-rht, line 157: missing case ID?

features/upgrade/storage/upgrade.feature
duanwei33: @pruan-rht  I followed the upgrade case syntax, all added actually one case, a half part is for prepare data and the other is for check after the upgrade, but this is one caseid. But thanks for the reminder, I do miss one keyword(prepare) in line 148.

features/upgrade/storage/upgrade.feature
duanwei33: Same one, they are the same case.

duanwei33:
@qinpingli PTAL

PR #1196, opened 2020-06-03
Add option --docker-image to new_app_as_dc

chao007:
Since `oc new app` is changed and `new_app_as_dc` is added.
So adding option --docker-image to `new_app_as_dc`

### DIFF
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -584,6 +584,7 @@
   :options:
     :app_repo: <value>
     :code: --code=<value>
+    :docker_image: --docker-image=<value>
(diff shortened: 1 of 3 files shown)


chao007:
Testing log is linked under https://github.com/openshift/cucushift/pull/7795

lib/rules/cli/4.1.yaml
```diff
@@ -585,6 +585,7 @@
     :app_repo: <value>
     :code: --code=<value>
     :image_stream: --image-stream=<value>
+    :docker_image: --docker-image=<value>
```
liangxia, line 588: @chao007 Please sort them.

PR #1206, opened 2020-06-04
Using taints to prevent pods from scheduling to test nodes

jhou1:
By adding taints, only our pods with toleration can schedule to the nodes we had scaled up. No other pods can schedule there and this prevents unexpected failures in other teams' tests. For example even though our tests run in destructive mode daemonset and statefulset can still schedule pods to our nodes. This pr prevents that.

Tested and it's working well. @sunzhaohua2 @miyadav PTAL

### DIFF
--- a/features/step_definitions/machine_set.rb
+++ b/features/step_definitions/machine_set.rb
@@ -64,6 +64,8 @@ Given(/^I clone a machineset and name it "([^"]*)"$/) do | ms_name |
   new_spec["metadata"]["name"] = ms_name
   new_spec["spec"]["selector"]["matchLabels"]["machine.openshift.io/cluster-api-machineset"] = ms_name
   new_spec["spec"]["template"]["metadata"]["labels"]["machine.openshift.io/cluster-api-machineset"] = ms_name
+  # Adding taints to machineset so that pods without toleration can not schedule to the nodes we provision
+  new_spec["spec"]["template"]["spec"]["taints"] = [{"effect": "NoSchedule","key": "mapi","value": "mapi_test"}]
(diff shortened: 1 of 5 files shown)


features/step_definitions/machine_set.rb
```diff
@@ -1,7 +1,8 @@
 Given(/^I pick a random machineset to scale$/) do
   ensure_admin_tagged
+  # MachineSets with replicas=0 or have taints are intentionally not picked
   machine_sets = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api")).
-    select { |ms| ms.available_replicas >= 1 }
+    select { |ms| ms.available_replicas >= 1 && ms.taints.nil? }
```
akostadinov, line 5: Isn't overwriting them better than trying to find MS without taints? Leaving this as only a question for future consideration. I don't mind this as long as it is working for you.

akostadinov: I  mean LGTM on my end.

jhou1: Yes, I could overwrite it, thank you.

miyadav:
@jhou1  , so other monitoring related pods/daemonsets  will be able to get scheduled on these ?

jhou1:
> so other monitoring related pods/daemonsets will be able to get scheduled on these ?

They can't schedule on these, unless they have the tolerations.

miyadav:
/lgtm

@jhou1  , as you suggested this wont impact other alerts related tests being created , only  new nodes will have taints ..it helps us to run destructive tests without worrying about impact to other tests , thanks ..

PR #1207, opened 2020-06-04
Fix OCP-31241

wewang58:
Some versions are not include info "probe failed", so updated it
[Pass log](https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/100644/console)
@openshift/devexp-qe ptal, thanks

### DIFF
--- a/features/cli/oc_set_probe.feature
+++ b/features/cli/oc_set_probe.feature
@@ -196,9 +196,8 @@ Feature: oc_set_probe.feature
       | resource | pod              |
       | l        | deployment=mysql |
     Then the output should match:
-      | Readiness      |
-      | tcp-socket :33 |
-      | probe failed   |
+      | Readiness       |
+      | tcp-socket :33  |
     """
 
   # @author wewang@redhat.com


features/cli/oc_set_probe.feature
```diff
@@ -196,9 +196,9 @@ Feature: oc_set_probe.feature
       | resource | pod              |
       | l        | deployment=mysql |
     Then the output should match:
-      | Readiness      |
-      | tcp-socket :33 |
-      | probe failed   |
+      | Readiness        |
+      | tcp-socket :33   |
+      | (probe failed)?  |
```
xiuwang, line 201: one more space

wewang58: Good eyes, updated.

features/cli/oc_set_probe.feature
```diff
@@ -196,9 +196,9 @@ Feature: oc_set_probe.feature
       | resource | pod              |
       | l        | deployment=mysql |
     Then the output should match:
-      | Readiness      |
-      | tcp-socket :33 |
-      | probe failed   |
+      | Readiness       |
+      | tcp-socket :33  |
+      | (probe failed)? |
```
liangxia, line 201: @wewang58  I guess `probe failed` exist in old versions, and does not exist in new versions. And this PR tries to fix that.

But it looks strange to me. Let me explain it a little.
Situation 1. Let's assume the string `probe failed` exist in all versions.
As it exist in all version, and it matches the output, it's a PASS.

Situation 2. Let's assume the string `probe failed` does not exist in all versions.
As we use `(xxx)?`, it means the string is optional, it's a PASS.

Situation 3. Let's assume the string `probe failed` exist in old versions, and does not exist in new versions.
As a combination of S1 and S2, we still got a PASS.

Situation 4. Let's assume the string `probe failed` exist in new versions, and does not exist in old versions.
As a combination of S1 and S2, we still got a PASS.

In total, no matter the string exist or not exist in any version, we will got a PASS. Then it's a meaningless test.
We can fix it with version check/compare, or simply remove it.

wewang58: @liangxia  removed it, please check it again, thanks

PR #1209, opened 2020-06-04
Auto OCP-19615

huiran0826:
@zhaozhanqi @weliang1 @anuragthehatter @rbbratta 

https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1542/console

### DIFF
--- a/features/networking/egress-ingress.feature
+++ b/features/networking/egress-ingress.feature
@@ -300,3 +300,49 @@ Feature: Egress-ingress related networking scenarios
     When I execute on the pod:
       | curl | --head | www.test.com |
     Then the step should succeed
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-19615
+  @admin
+  Scenario: Iptables should be updated with correct endpoints when egress DNS policy was used
+    Given I have a project
+    When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/list_for_pods.json" replacing paths:
+      | ["items"][0]["spec"]["replicas"] | 1 |
+    Then the step should succeed
+    And 1 pods become ready with labels:
+      | name=test-pods |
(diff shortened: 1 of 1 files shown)


features/networking/egress-ingress.feature
```diff
@@ -300,3 +300,50 @@ Feature: Egress-ingress related networking scenarios
...
+  @admin
+  Scenario: Iptables should be updated with correct endpoints when egress DNS policy was used
+    Given I have a project
+    When I run oc create over "<%= BushSlicer::HOME %>/testdata/networking/list_for_pods.json" replacing paths:
+      | ["items"][0]["spec"]["replicas"] | 1 |
+    Then the step should succeed
+    And 1 pods become ready with labels:
+      | name=test-pods |
+    Given I use the "test-service" service
+    And evaluation of `service.ip(user: user)` is stored in the :service_ip clipboard
```
rbbratta, line 315: I was going to suggest `service.ip_url()` for future IPv6 support, but we only have `ip_url` method for Pod,
but we do have `service.url()` method, should we be using that instead of hard-coding port `27017` below?

rbbratta: see also https://github.com/openshift/verification-tests/pull/1210/files

liangxia: Above two line can be shorted as
```
And evaluation of `service("test-service").ip` is stored in the :service_ip clipboard`
```

rbbratta: #1210 was merged so we can do

```
And evaluation of `service("test-service").url` is stored in the :service_url clipboard`

...

| /usr/bin/curl | -k | <%= cb.service_url %> |

```

huiran0826: Thanks @rbbratta, updated it.

PR #1210, opened 2020-06-05
service: add IPv6 URL brackets and fix return

rbbratta:
check if we get IPv6 address and encode in brackets

Needs testing on cluster with IPv6 services

### DIFF
--- a/lib/openshift/service.rb
+++ b/lib/openshift/service.rb
@@ -66,8 +66,9 @@ module BushSlicer
     # @note call without parameters only when props are loaded
     def url(user: nil, cached: true, quiet: false)
       ip = self.ip(user: user, cached: cached, quiet: quiet)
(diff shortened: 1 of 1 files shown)


rbbratta:
looking for `oc get svc` output on IPv6 cluster to verify the IPv6 address is raw.

rbbratta:
confirmed that IPv6 address is raw.
```
            "spec": {
                "clusterIP": "fd02::1",
                "ports": [
                    {
                        "name": "https",
                        "port": 443,
                        "protocol": "TCP",
                        "targetPort": 6443
                    }
                ],
                "sessionAffinity": "None",
                "type": "ClusterIP"
            },
```

rbbratta:
http://pastebin.test.redhat.com/872524

PR #1218, opened 2020-06-08
Add labels parameter to new_app_as_dc cmd

qinpingli:


akostadinov:
Why use `new_app_as_dc` rule? I don't see any significant difference from the normal `new_app` rule?

akostadinov:
Please check my question before we can merge.

akostadinov:
Actually now I see what you mean. Apparently rules for 4.5 and earlier differ. Please bring them in sync and we can merge. Namely `--as-deployment-config`  option is missing from rules for older versions.

qinpingli:
> Actually now I see what you mean. Apparently rules for 4.5 and earlier differ. Please bring them in sync and we can merge. Namely `--as-deployment-config` option is missing from rules for older versions.

@akostadinov the parameter `--as-deployment-config` is not supported in the older versions. So, I did not change the rules, just add `--label` option to all versions for `new_app_as_dc` cmd.

lib/rules/cli/4.1.yaml
```diff
@@ -589,6 +589,7 @@
     :name: --name=<value>
     :source_spec: <value>
     :strategy: --strategy=<value>
+    :labels: --labels=<value>
```
liangxia, line 592: Please sort the options.

features/cli/oc_volume.feature
```diff
@@ -78,18 +78,17 @@ Feature: oc_volume.feature
...
-      | o              | yaml      |
+      | resource       | :false                                                   |
+      | resource_name  | dc/mydc                                                  |
+      | o             | jsonpath={.spec.template.spec.containers[*].volumeMounts} |
+    Then the step should succeed
+    And the output should contain 1 times:
+	    | name:secret |
+
+    When I run the :get client command with:
+      | resource       | :false                                                   |
```
liangxia, line 89: Why do we need this ?

liangxia:
Since version 4.5, `new-app` switched from creating `deploymentconfig` to `deployment`, and added an option `--as-deployment-config`.
For teams that are not testing `new-app`, but just need to create some resources via `new-app`, they created a new command `new_app_as_dc` to work across all versions. In 4.1~4.4, `new_app_as_dc` is the same as `new_app`, while in 4.5, `new_app_as_dc` means `new_app --as-deployment-config`

features/cli/oc_volume.feature
qinpingli: In the original version, we checked the output for dc and rc together, it's unstable, so I checked them separately. One for dc, one for rc.

Looks there are some format issues, fixed them.

features/cli/oc_volume.feature
```diff
@@ -78,18 +78,17 @@ Feature: oc_volume.feature
     Then the step should succeed
 
     When I run the :get client command with:
-      | resource       | :false    |
-      | resource_name  | dc/mydc   |
-      | resource_name  | rc/mydc-1 |
-      | o              | yaml      |
+      | resource       | :false                                                    |
+      | resource_name  | dc/mydc                                                   |
```
liangxia, line 82: We can use `| resource | dc/mydc |`

qinpingli: Thanks, done.

PR #1220, opened 2020-06-09
Use BushSlicer::HOME

liangxia:
/cc @pruan-rht @akostadinov


akostadinov:
Actually we need  to use step `/^I obtain test data file #{QUOTED}$/`. Still this is better than before.

pruan-rht:
That will probably be more time consuming as that will be another line to be added and change the current lines that has the HOME rereference.

akostadinov:
I guess so. But still accessing data files in abstracted way is worth it. Another option is to write a method in `World` so that it is another `sed`. Maybe better to do that way.

liangxia:
@akostadinov @pruan-rht  Switch to step `/^I obtain test data file #{QUOTED}$/` in https://github.com/openshift/verification-tests/pull/1227

PR #1225, opened 2020-06-10
change utility name and fix thread error

pruan-rht:


pruan-rht:
@akostadinov @liangxia PTAL...thx!

PR #1226, opened 2020-06-10
Fix for stability

kasturinarra:
In the previous code it waits for the pod to disappear within 30 seconds,  but the pod sometimes is taking more than 30 seconds to disappear so fixing it.


kasturinarra:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/102931/console

PR #1227, opened 2020-06-10
Use relative path for test files

liangxia:
/cc @pruan-rht @akostadinov

### DIFF
--- a/features/build/stibuild.feature
+++ b/features/build/stibuild.feature
@@ -2,8 +2,9 @@ Feature: stibuild.feature
   # @author xiuwang@redhat.com
   Scenario Outline: Trigger s2i/docker/custom build using additional imagestream
     Given I have a project
+    Given I obtain test data file "templates/<template>"
     And I run the :new_app client command with:
-      | file | <%= BushSlicer::HOME %>/testdata/templates/<template> |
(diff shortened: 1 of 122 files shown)


features/build/stibuild.feature
```diff
@@ -94,8 +95,9 @@ Feature: stibuild.feature
     When a pod becomes ready with labels:
       | deploymentconfig=git |
       | deployment=git-2     |
+    Given I obtain test data file "build/httpd-ex.git"
```
xiuwang, line 98: testdata/build/httpd-ex.git is a dir,then I met this error
    `Given I obtain test data file "build/httpd-ex.git" # features/step_definitions/file.rb:1`
      `Is a directory - read (Errno::EISDIR)`

Could I change this back?

akostadinov: Let's create a step `I obtain test data dir ...`. Do you think you can copy existing step and make it work with a directory? Or do you want dx team do it?

akostadinov: @xiuwang ^^

xiuwang: @akostadinov I will do that, thanks for advice

PR #1166, opened 2020-05-26
Updated scenario ocp 13065 upgrade auto

sunilcio:
Updated scenario ocp 13065 upgrade auto

PR #1230, opened 2020-06-10
ocp 13065 upgrade auto

sunilcio:
ocp 13065 upgrade auto

### DIFF
--- a/lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
+++ b/lib/rules/web/admin_console/4.1/common_ui_elements.xyaml
@@ -79,6 +79,44 @@ set_input_value:
       xpath: //input[@id='<input_field_id>']
     op: send_keys <input_value>
     type: input
+set_label_input:
+  elements:
+  - selector: &label_input
+      xpath: (//label[contains(text(),'<label_text>')]/following::input[1] | //label[contains(text(),'<label_text>')]//input[1])[1]
+    op: clear
+  - selector:
+      <<: *label_input
+    op: send_keys <input_value>
+set_label_input_simple:
+  elements:
+  - selector: &label_input_value
+      xpath: //label[text()='<label_text>']/..//input
+    op: clear
+    type: input
+  - selector:
+      <<: *label_input_value
+    op: send_keys <input_value>
+    type: input
+set_placeholder_input:
+  elements:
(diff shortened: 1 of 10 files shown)


sunilcio:
@liangxia  @pruan-rht PTAL
I was updating my last PR (#1166) but got mixed up with other vm. Created this one with your suggested changes.
Case runs correctly with the changes http://pastebin.test.redhat.com/873865

features/upgrade/node/scc.feature
```diff
@@ -0,0 +1,22 @@
...
+
+  # @author sunilc@redhat.com
+  @upgrade-prepare
+  @admin
+  Scenario: Upgrade - Seccomp part of SCC policy should be kept and working after upgrade - prepare
+    Given I switch to cluster admin pseudo user
+    When I run the :new_project client command with:
+      | project_name | node-upgrade-scc |
+    And I use the "node-upgrade-scc" project
+    When I run the :create client command with:
```
pruan-rht, line 11: @sunilcio please use the step `Given I obtain...` to retrieve the testdata.  Please refer #1227 as example of how to do it.

jhou1:
```
Given I obtain test data file "testdata/node/scc.yaml"
When I run the :create client command with:
  | f | testdata/node/scc.yaml |
```

features/upgrade/node/scc.feature
```diff
@@ -0,0 +1,22 @@
+Feature: Seccomp part of SCC policy should be kept and working after upgrade	
+
+  # @author sunilc@redhat.com
+  @upgrade-prepare
+  @admin
+  Scenario: Upgrade - Seccomp part of SCC policy should be kept and working after upgrade - prepare
+    Given I switch to cluster admin pseudo user
+    When I run the :new_project client command with:
+      | project_name | node-upgrade-scc |
+    And I use the "node-upgrade-scc" project
```
liangxia, line 10: SCC is a cluster resource. I don't think we need the project here as it's not used actually.

features/upgrade/node/scc.feature
liangxia: By the way, creating scc should be destructive, as scc can significantly affect a lot of scenarios.

PR #1233, opened 2020-06-11
Auto OCP-29709

xiuwang:
@openshift/devexp-qe Please help review, thx

### DIFF
--- a/features/upgrade/samples_operator/upgrade.feature
+++ b/features/upgrade/samples_operator/upgrade.feature
@@ -77,3 +77,99 @@ Feature: image-registry operator upgrade tests
     Given the expression should be true> cluster_operator('image-registry').condition(type: 'Available')['status'] == "True"
     Given the expression should be true> cluster_operator('image-registry').condition(type: 'Progressing')['status'] == "False"
     Given the expression should be true> cluster_operator('image-registry').condition(type: 'Degraded')['status'] == "False"
+
+  @upgrade-prepare
+  @users=upuser1,upuser2
+  @admin
+  Scenario: Upgrade imagestream from old version which not implement node credentials - prepare 
+    Given I switch to cluster admin pseudo user
+    When I run the :extract admin command with:
+      | resource  | secret/pull-secret |
+      | namespace | openshift-config   | 
+      | to        | /tmp               |
+      | confirm   | true               |
+    Then the step should succeed
+    When I run the :new_project client command with:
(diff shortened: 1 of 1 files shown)


features/upgrade/samples_operator/upgrade.feature
```diff
@@ -37,3 +37,99 @@ Feature: image-registry operator upgrade tests
...
+      | name        | pj-secret                                |
+      | from_file   | .dockerconfigjson=/tmp/.dockerconfigjson |
+      | type        | kubernetes.io/dockerconfigjson           |
+    Then the step should succeed
+    When I run the :import_image client command with:
+      | from             | registry.redhat.io/rhscl/ruby-25-rhel7:latest |
+      | confirm          | true                                          |
+      | image_name       | imagestream1:latest                           |
+      | reference-policy | local                                         |
+    Then the step should succeed
```
wewang58, line 67: The project is include pull secret, still import failed?

xiuwang: @wewang58 this is a upgrade prepare case on 4.4 which is not include the pull secret.

features/upgrade/samples_operator/upgrade.feature
```diff
@@ -37,3 +37,99 @@ Feature: image-registry operator upgrade tests
...
+    And the output should not contain:
+      | error: Import failed |
+    When I run the :import_image client command with:
+      | from             | registry.redhat.io/rhscl/ruby-25-rhel7:latest |
+      | confirm          | true                                          |
+      | image_name       | imagestream2:latest                           |
+    Then the step should succeed
+    And the output should not contain:
+      | error: Import failed |
```
akostadinov, line 76: Do we need string checks like this? They are unreliable check IMO and only make scenario longer.

I'd suggest to remove this and all other similar string checks unless you have seen actual value from them.

akostadinov: other than that lgtm

xiuwang: @akostadinov Import_image cmd return succeed whenever the image import succeed or fail. In result we need check if there are import failure to confirm the image import situation.

akostadinov: @xiuwang , this deserves a bugzilla. A failed command should return a non-zero status. I'll merge but this is a bug.

PR #1235, opened 2020-06-11
[WIP] Fix format

liangxia:
/cc @akostadinov @pruan-rht 

Add the shell script in https://github.com/openshift/verification-tests/pull/1234, this PR is generated with the script via command `tools/formatter.sh $(find . -name *.feature)`

### DIFF
--- a/features/build/buildconfig.feature
+++ b/features/build/buildconfig.feature
@@ -67,10 +67,10 @@ Feature: buildconfig.feature
   Scenario Outline: Build with images pulled from private repositories
     Given I have a project
     When I run the :create_secret client command with:
-     | name        | pull                                                                            |
-     | secret_type | generic                                                                         |
-     | from_file   | .dockercfg=<%= expand_private_path(conf[:services, :docker_hub, :dockercfg]) %> | 
-     | type        | kubernetes.io/dockercfg                                                         |
+      | name        | pull                                                                            |
+      | secret_type | generic                                                                         |
+      | from_file   | .dockercfg=<%= expand_private_path(conf[:services, :docker_hub, :dockercfg]) %> |
+      | type        | kubernetes.io/dockercfg                                                         |
     Then the step should succeed
     Given I obtain test data file "build/<template>"
     When I run the :create client command with:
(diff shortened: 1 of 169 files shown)


features/build/buildconfig.feature
```diff
@@ -79,30 +79,30 @@ Feature: buildconfig.feature
     And the "ruby-sample-build-1" build was created
     Then the "ruby-sample-build-1" build completes
 
-    Examples:
-      | template                              |
-      | tc479541/test-buildconfig-s2i.json    | # @case_id OCP-11474
```
liangxia, line 84: Definitely, this is not something we want.

pruan-rht:
@liangxia nice trick to use `cucumber --dry-run`... but I'm not sure if some of the resulting output is what we want in regards to comments formats...  @akostadinov WDYT?

features/build/buildconfig.feature
```diff
@@ -79,30 +79,30 @@ Feature: buildconfig.feature
...
+    Examples: 
+      | template                           |
+      | tc479541/test-buildconfig-s2i.json |
```
liangxia, line 84: Definitely, this is not something we want.

liangxia: Workarounds,
-  We skip those feature files which contains `Scenario Outline`.
- We change the format to
```
    Examples: 
      | template                           |
      # @case_id OCP-11474
      | tc479541/test-buildconfig-s2i.json |
      # @case_id OCP-xxx
      | tc479541/test-buildconfig-xxx.json |
```

rbbratta:
we could copy  https://github.com/cucumber/cucumber-ruby/blob/v2.4.0/lib/cucumber/formatter/pretty.rb#L47   and modify the indent.

```
                                     FORMAT can also be the fully qualified class name of
                                     your own custom formatter. If the class isn't loaded,
                                     Cucumber will attempt to require a file with a relative
                                     file name that is the underscore name of the class name.
                                     Example: --format Foo::BarZap -> Cucumber will look for
                                     foo/bar_zap.rb. You can place the file with this relative
                                     path underneath your features/support directory or anywhere
                                     on Ruby's LOAD_PATH, for example in a Ruby gem.
```

features/build/buildconfig.feature
akostadinov: How can we skip scenario outlines?
Changing comment location might be tricky and is uglier IMO. Both are acceptable though.

Or might be easier to just develop out own formatter :) I dunno.

pruan-rht: having the comment line in the same place is clearer to me.

liangxia:
> we could copy https://github.com/cucumber/cucumber-ruby/blob/v2.4.0/lib/cucumber/formatter/pretty.rb#L47 and modify the indent.
> 
> ```
> FORMAT can also be the fully qualified class name of
> your own custom formatter. If the class isn't loaded,
> Cucumber will attempt to require a file with a relative
> file name that is the underscore name of the class name.
> Example: --format Foo::BarZap -> Cucumber will look for
> foo/bar_zap.rb. You can place the file with this relative
> path underneath your features/support directory or anywhere
> on Ruby's LOAD_PATH, for example in a Ruby gem.
> ```

Tried,
adding `pretty.rb` to `features/support/ocp/test.rb`
adding `pretty.rb` to `features/support/ocp/test_abc.rb`
adding `pretty.rb` to `features/support/test.rb`
Got,
```
cannot load such file -- ocp/test_abc
Error creating formatter: Ocp::TestAbc (LoadError)
```
```
cannot load such file -- test
Error creating formatter: Test (LoadError)
```

akostadinov:
Do you change class name and module path to match directory path? Sounds like these need to match.

PR #1236, opened 2020-06-11
Updated OCP-13065 upgrade auto

sunilcio:
Updated OCP-13065 upgrade auto


sunilcio:
@pruan-rht @liangxia hi, needed to create this new one as I reverted my old PR commit and it got closed. Sorry for confusion :) 

I updated the retrieve test data steps. Also this this is not destructive as we are creating new scc which others will not use and will not affect them.

PR #1238, opened 2020-06-11
NB DB scenarios

anuragthehatter:
Keep Adding more scenarios related to NB DB to this PR
@rbbratta @zhaozhanqi @huiran0826 @weliang1 
Logs OCP-29954 : https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1595/console

### DIFF
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -17,25 +17,75 @@ Feature: OVN related networking scenarios
     And evaluation of `pod(1).name` is stored in the :pod2_name clipboard
     #Checking whether Kube API data is synced on OVN NB db which in this case are couple of pods
     Given I store the ovnkube-master "north" leader pod in the clipboard
-    And evaluation of `pod.ip_url` is stored in the :OVN_NB_leader_IP clipboard
-    And evaluation of `pod.node_name` is stored in the :OVN_NB_leader_node clipboard
+    And evaluation of `pod.ip_url` is stored in the :ovn_nb_leader_ip clipboard
+    And evaluation of `pod.node_name` is stored in the :ovn_nb_leader_node clipboard
     And admin executes on the pod:
-      | bash | -c | ovn-nbctl -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt --db ssl:<%= cb.OVN_NB_leader_IP %>:9641 list logical_switch_port |
+      | bash | -c | ovn-nbctl list logical_switch_port |
     Then the step should succeed
     And the output should contain:
       | <%= cb.pod1_name %> |
       | <%= cb.pod2_name %> |
     #Simulating a NB db crash
-    Given I use the "<%= cb.OVN_NB_leader_node %>" node
+    Given I use the "<%= cb.ovn_nb_leader_node %>" node
     And I run commands on the host:
       | pkill -f OVN_Northbound |
     And admin waits for all pods in the "openshift-ovn-kubernetes" project to become ready up to 120 seconds
     #Making sure the pod entries are synced again when NB db is re-created
     Given I store the ovnkube-master "north" leader pod in the clipboard
-    And evaluation of `pod.ip_url` is stored in the :New_OVN_NB_leader_IP clipboard
+    And evaluation of `pod.ip_url` is stored in the :new_ovn_nb_leader_ip clipboard
     Given admin executes on the pod:
-      | bash | -c | ovn-nbctl -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt --db ssl:<%= cb.New_OVN_NB_leader_IP %>:9641 list logical_switch_port |
+      | bash | -c | ovn-nbctl list logical_switch_port |
     Then the step should succeed
     And the output should contain:
       | <%= cb.pod1_name %> |
       | <%= cb.pod2_name %> |
+
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-30055
+  @admin
+  @destructive
+  Scenario: OVN DB should be updated correctly if a resource only exist in Kube API but not in OVN NB db
+    Given the env is using "OVNKubernetes" networkType
+    # Now scale down CNO pod to 0 and makes sure dhcp pods still running and erase additionalnetworks config from CNO
+    Given admin uses the "openshift-network-operator" project
+    And I run the :scale admin command with:
+      | resource | deployment       |
+      | name     | network-operator |
+      | replicas | 0                |
+    Then the step should succeed
+    Given I register clean-up steps:
+    """
+    Given admin uses the "openshift-network-operator" project
+    And I run the :scale admin command with:
+      | resource | deployment       |
+      | name     | network-operator |
+      | replicas | 1                |
+    Then the step should succeed
+    """
+    When I run the :delete admin command with:
+      | object_type       | ds                       |
+      | object_name_or_id | ovnkube-master           |
+      | n                 | openshift-ovn-kubernetes |
+    Then the step should succeed
+    Given I have a project
+    And I obtain test data file "networking/pod-for-ping.json"
+    When I run the :create client command with:
+      | f | pod-for-ping.json |
+    Then the step should succeed
+    #Now scale down CNO pod to 1 and check whether hello-pod is synced to NB db
+    Given I run the :scale admin command with:
+      | resource | deployment       |
+      | name     | network-operator |
+      | replicas | 1                |
+    Then the step should succeed
+    #A minimum wait for 10 seconds is tested to reflect CNO deployment to be effective which will then re-spawn ovn pods
+    Given 10 seconds have passed
+    And admin waits for all pods in the "openshift-ovn-kubernetes" project to become ready up to 60 seconds
+    #Checking whether Kube API data is synced on OVN NB db which in this case is a test-pod created in earlier steps
+    Given I store the ovnkube-master "north" leader pod in the clipboard
+    And admin executes on the pod:
+      | bash | -c | ovn-nbctl list logical_switch_port |
+    Then the step should succeed
+    And the output should contain:
+      | hello-pod |
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -63,12 +63,7 @@ Feature: OVN related networking scenarios
       | replicas | 1                |
     Then the step should succeed
     """
-    When I run the :delete admin command with:
-      | object_type       | ds                       |
-      | object_name_or_id | ovnkube-master           |
-      | n                 | openshift-ovn-kubernetes |
-    Then the step should succeed
-    Given admin uses the "openshift-ovn-kubernetes" project
+    And admin ensures "ovnkube-master" ds is deleted from the "openshift-ovn-kubernetes" project
     And admin executes existing pods die with labels:
       | app=ovnkube-master | 
     Given I have a project
@@ -126,13 +121,9 @@ Feature: OVN related networking scenarios
       | replicas | 1                |
     Then the step should succeed
     """
-    When I run the :delete admin command with:
-      | object_type       | ds                       |
-      | object_name_or_id | ovnkube-master           |
-      | n                 | openshift-ovn-kubernetes |
-    Then the step should succeed
+    And admin ensures "ovnkube-master" ds is deleted from the "openshift-ovn-kubernetes" project
     And admin executes existing pods die with labels:
-      | app=ovnkube-master | 
+      | app=ovnkube-master |
     And I ensures "hello-pod" pod is deleted from the "<%= cb.hello_pod_project %>" project
     #Now scale up CNO pod to 1 and check whether hello-pod status is synced to NB db means it should not present in the DB
     Given admin uses the "openshift-network-operator" project
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -47,7 +47,7 @@ Feature: OVN related networking scenarios
   @destructive
   Scenario: OVN DB should be updated correctly if a resource only exist in Kube API but not in OVN NB db
     Given the env is using "OVNKubernetes" networkType
-    # Now scale down CNO pod to 0 and makes sure dhcp pods still running and erase additionalnetworks config from CNO
+    # Now scale down CNO pod to 0
     Given admin uses the "openshift-network-operator" project
     And I run the :scale admin command with:
       | resource | deployment       |
@@ -79,8 +79,8 @@ Feature: OVN related networking scenarios
       | name     | network-operator |
       | replicas | 1                |
     Then the step should succeed
-    #A minimum wait for 10 seconds is tested to reflect CNO deployment to be effective which will then re-spawn ovn pods
-    Given 10 seconds have passed
+    #A minimum wait for 30 seconds is tested to reflect CNO deployment to be effective which will then re-spawn ovn pods
+    Given 30 seconds have passed
     And admin waits for all pods in the "openshift-ovn-kubernetes" project to become ready up to 60 seconds
     #Checking whether Kube API data is synced on OVN NB db which in this case is a test-pod created in earlier steps
     Given I store the ovnkube-master "north" leader pod in the clipboard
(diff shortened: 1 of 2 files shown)


features/networking/ovn.feature
```diff
@@ -0,0 +1,41 @@
...
+    Given 2 pods become ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(0).name` is stored in the :pod1_name clipboard
+    And evaluation of `pod(1).name` is stored in the :pod2_name clipboard
+    #Checking whether Kube API data is synced on OVN NB db which in this case are couple of pods
+    Given I store the ovnkube-master "north" leader pod in the clipboard
+    And evaluation of `pod.ip_url` is stored in the :OVN_NB_leader_IP clipboard
+    And evaluation of `pod.node_name` is stored in the :OVN_NB_leader_node clipboard
+    And admin executes on the pod:
+      | bash | -c | ovn-nbctl -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt --db ssl:<%= cb.OVN_NB_leader_IP %>:9641 list logical_switch_port |
```
rbbratta, line 23: If we can select a container to run on and use the `nbdb` container, then we wouldn't need to use the SSL options.

e.g.

`oc exec -c nbdb ovnkube-master-4cjwf -- ovn-nbctl list logical_switch_port`

but unfortunately none of the pod execute commands let you specify a container.  I was going to look into adding this somehow.

features/networking/ovn.feature
```diff
@@ -0,0 +1,41 @@
...
+    When I run the :create client command with:
+      | f | list_for_pods.json |
+    Then the step should succeed
+    Given 2 pods become ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(0).name` is stored in the :pod1_name clipboard
+    And evaluation of `pod(1).name` is stored in the :pod2_name clipboard
+    #Checking whether Kube API data is synced on OVN NB db which in this case are couple of pods
+    Given I store the ovnkube-master "north" leader pod in the clipboard
+    And evaluation of `pod.ip_url` is stored in the :OVN_NB_leader_IP clipboard
```
pruan-rht, line 20: @anuragthehatter variable and symbol guide line for ruby https://github.com/rubocop-hq/ruby-style-guide#snake-case-for-symbols-methods-and-variables states that we should use snake_case...probably need to de-uppercase for all of these variables.

features/networking/ovn.feature
rbbratta: This works in a quick test

```diff
diff --git a/features/networking/ovn.feature b/features/networking/ovn.feature
index d7ea10ef..2a148407 100644
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -19,8 +19,8 @@ Feature: OVN related networking scenarios
     Given I store the ovnkube-master "north" leader pod in the clipboard
     And evaluation of `pod.ip_url` is stored in the :OVN_NB_leader_IP clipboard
     And evaluation of `pod.node_name` is stored in the :OVN_NB_leader_node clipboard
-    And admin executes on the pod:
-      | bash | -c | ovn-nbctl -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt --db ssl:<%= cb.OVN_NB_leader_IP %>:9641 list logical_switch_port |
+    And admin executes on the pod "nbdb" container:
+      | bash | -c | ovn-nbctl  list logical_switch_port |
     Then the step should succeed
     And the output should contain:
       | <%= cb.pod1_name %> |
@@ -33,8 +33,8 @@ Feature: OVN related networking scenarios
     #Making sure the pod entries are synced again when NB db is re-created
     Given I store the ovnkube-master "north" leader pod in the clipboard
     And evaluation of `pod.ip_url` is stored in the :New_OVN_NB_leader_IP clipboard
-    Given admin executes on the pod:
-      | bash | -c | ovn-nbctl -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt --db ssl:<%= cb.New_OVN_NB_leader_IP %>:9641 list logical_switch_port |
+    Given admin executes on the pod "nbdb" container:
+      | bash | -c | ovn-nbctl list logical_switch_port |
     Then the step should succeed
     And the output should contain:
       | <%= cb.pod1_name %> |
diff --git a/features/step_definitions/pod.rb b/features/step_definitions/pod.rb
index a98c32b1..9e6a2b75 100644
--- a/features/step_definitions/pod.rb
+++ b/features/step_definitions/pod.rb
@@ -213,7 +213,7 @@ end

 # args can be a table where each cell is a command or an argument, or a
 #   multiline string where each line is a command or an argument
-When /^(I execute|admin executes) on the(?: "(.+?)")? pod:$/ do |by, pod_name, raw_args|
+When /^(I execute|admin executes) on the(?: "(.+?)")? pod(?: "(.+?)" container)?:$/ do |by, pod_name, container, raw_args|
   _user = by.split.first == "admin" ? admin : user
   if raw_args.respond_to? :raw
     # this is table, we don't mind dimentions used by user
@@ -223,7 +223,7 @@ When /^(I execute|admin executes) on the(?: "(.+?)")? pod:$/ do |by, pod_name, r
     args = raw_args.split("\n").map(&:strip)
   end

-  @result = pod(pod_name).exec(*args, as: _user)
+  @result = pod(pod_name).exec(*args, as: _user, container: container)
 end

 # wrapper around  oc logs, keep executing the command until we have an non-empty response
```

rbbratta: err, it seems to work on the default `northd` container as well.  I guess as long as we run on the master we don't need to use the SSL.


```bash
$ oc exec  ovnkube-master-4cjwf -- ovn-nbctl list logical_switch_port
Defaulting container name to northd.
Use 'oc describe pod/ovnkube-master-4cjwf -n openshift-ovn-kubernetes' to see all of the containers in this pod.
_uuid               : b5f38583-6461-4d27-93eb-a6027f8c28ee
addresses           : [dynamic]
dhcpv4_options      : []
dhcpv6_options      : []

```

anuragthehatter: Ahh, thats right.Thanks. Sticking to leader pod will alleviate the use of certs to be used via other pods. Make sense. Also #1245 would make sense in future use cases for sure where containers need to leveraged for execution.

features/networking/ovn.feature
anuragthehatter: oops, thats right. Apparently I put some emphasis on networkType (OVN) and its database(NB)

anuragthehatter:
Added OCP-30055 as well. [Logs](https://privatebin-it-iso.int.open.paas.redhat.com/?b7cbdb94a4809880#qNGyQWaxJ41ITHzG55s2ngKRD7iazCEUz4BHYjbJ3ro=) here

features/networking/ovn.feature
```diff
@@ -0,0 +1,147 @@
...
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-30055
+  @admin
+  @destructive
+  Scenario: OVN DB should be updated correctly if a resource only exist in Kube API but not in OVN NB db
+    Given the env is using "OVNKubernetes" networkType
+    # Now scale down CNO pod to 0
+    Given admin uses the "openshift-network-operator" project
+    And I run the :scale admin command with:
```
zhaozhanqi, line 52: do we need to scale the network operator to 0 ? is there any difference if only delete the `ovnkube-master ` pods and make it recreated ?

features/networking/ovn.feature
```diff
@@ -0,0 +1,147 @@
...
+    Then the step should succeed
+    #A recommended wait for 30 seconds is tested to reflect CNO deployment to be in effect which will then re-spawn ovn pods
+    Given 30 seconds have passed
+    And admin waits for all pods in the "openshift-ovn-kubernetes" project to become ready up to 60 seconds
+    Given I store the ovnkube-master "north" leader pod in the clipboard
+    And admin executes on the pod:
+      | bash | -c | ovn-nbctl list logical_switch_port |
+    Then the step should succeed
+    #making sure here that hello-pod absense is properly synced
+    And the output should not contain:
```
zhaozhanqi, line 146: hmm,  Could you help understand what's the scenario is for? 
normally the user will use deployment/ds/rc to deploy the pods.   So I think the pod will sync back to db and working well  after the ovn recreated.

features/networking/ovn.feature
zhaozhanqi: ok, thanks for your explanation.  I thought they should be have same behaviour (with deleting cno) and (without deleting cno).
anyways. if you already confirmed this with Dev. that's ok for me.

features/networking/ovn.feature
zhaozhanqi: ok, thanks

features/networking/ovn.feature
```diff
@@ -0,0 +1,147 @@
...
+    And admin executes on the pod:
+      | bash | -c | ovn-nbctl list logical_switch_port |
+    Then the step should succeed
+    And the output should contain:
+      | <%= cb.pod1_name %> |
+      | <%= cb.pod2_name %> |
+    #Simulating a NB db crash
+    Given I use the "<%= cb.ovn_nb_leader_node %>" node
+    And I run commands on the host:
+      | pkill -f OVN_Northbound |
```
rbbratta, line 31: normally `SIGKILL` or `SIGSEGV` would be more appropriate, because `SIGTERM` is usually handled.

It looks like the only signal handler in the OVS source is for `SIGSEV` which prints out a backtrace, but I'm not sure.

https://github.com/openvswitch/ovs/commit/e2ed6fbeb18cba1235c6a260b754f11ecc9d4cb0#diff-37c472540fb42eff036bcb7639176986R212

probably safer to do a `pkill --SIGKILL` to be sure OVS isn't handling anything.

or is the test already working okay?

features/networking/ovn.feature
rbbratta: scaling to zero makes sense.

If we do need to do a more fine-grained pause, we could find the exact processes and `pkill -STOP`  which freezes the process in the kernel scheduler without killing it, then `pkill -CONT` to resume.

For the northd leader election lock inactivity timer BZ-1834473, I tested with `pkill -STOP` and `pkill -CONT` to check the inactivity timeout, so at some point we will have a test for that.

features/networking/ovn.feature
```diff
@@ -0,0 +1,147 @@
...
+      | resource | deployment       |
+      | name     | network-operator |
+      | replicas | 1                |
+    Then the step should succeed
+    """
+    When I run the :delete admin command with:
+      | object_type       | ds                       |
+      | object_name_or_id | ovnkube-master           |
+      | n                 | openshift-ovn-kubernetes |
+    Then the step should succeed
```
rbbratta, line 70: Do we need a `ensure deleted` step for the `ovnkube-master` pods?  It seems like we should be certain they are terminated.

features/networking/ovn.feature
```diff
@@ -0,0 +1,147 @@
...
+    Given I have a project
+    And I obtain test data file "networking/pod-for-ping.json"
+    When I run the :create client command with:
+      | f | pod-for-ping.json |
+    Then the step should succeed
+    #Now scale down CNO pod to 1 and check whether hello-pod is synced to NB db
```
rbbratta, line 76: s/scale down/scale up/

features/networking/ovn.feature
```diff
@@ -0,0 +1,147 @@
...
+      | replicas | 1                |
+    Then the step should succeed
+    """
+    When I run the :delete admin command with:
+      | object_type       | ds                       |
+      | object_name_or_id | ovnkube-master           |
+      | n                 | openshift-ovn-kubernetes |
+    Then the step should succeed
+    And I ensures "hello-pod" pod is deleted from the "<%= cb.hello_pod_project %>" project
+    #Now scale down CNO pod to 1 and check whether hello-pod status is synced to NB db means it should not present in the DB
```
rbbratta, line 131: s/scale down/scale up/

features/networking/ovn.feature
```diff
@@ -0,0 +1,147 @@
...
+      | resource | deployment       |
+      | name     | network-operator |
+      | replicas | 1                |
+    Then the step should succeed
+    """
+    When I run the :delete admin command with:
+      | object_type       | ds                       |
+      | object_name_or_id | ovnkube-master           |
+      | n                 | openshift-ovn-kubernetes |
+    Then the step should succeed
```
rbbratta, line 129: same question, should we ensure the pods are terminated?

features/networking/ovn.feature
anuragthehatter: @rbbratta yea, thats a good suggestion for processes but as we know CNO runs as container so we cant really leverage that here. But this can be used for NB db which runs as a process. But again IMO pkill would make sense as we are simulating a intentional NB db crash here which should be quickly resilient.

features/networking/ovn.feature
anuragthehatter: @rbbratta Yep, I have been testing it with pkill with no issues. So seems good to me for now.

features/networking/ovn.feature
anuragthehatter: @rbbratta IMO, Kube API response should be enough to make sure that. The positive response from KUBE API will let 70 to pass. Pods not terminated even if ds is deleted would be a deployment team test/bug they already must be running similar checks.

features/networking/ovn.feature
anuragthehatter: @rbbratta IMO, Kube API response should be enough to make sure that. The positive response from KUBE API will let 129 to pass.Pods not terminated even if ds is deleted would be a deployment team test/bug they already must be running similar checks.

rbbratta: I though I saw the delete ds return before and then the pods in Terminating state, so I think it might take a few seconds before the pods finish terminating.  It might not be an issue, but I think there could be race.

anuragthehatter: @rbbratta No worries. Added a step to make sure that. had to modify ```existing pods die with labels:``` step a bit to not break existing scenarios and support admin capabilities as well for accessing system projects.

pruan-rht: @anuragthehatter you can replace the call with the step `Given admin ensure...`

anuragthehatter: Ah right. Fixed it.Thanks and re-ran to make sure

features/networking/ovn.feature
```diff
@@ -0,0 +1,145 @@
...
+    """
+    Given admin uses the "openshift-network-operator" project
+    And I run the :scale admin command with:
+      | resource | deployment       |
+      | name     | network-operator |
+      | replicas | 1                |
+    Then the step should succeed
+    """
+    And admin ensures "ovnkube-master" ds is deleted from the "openshift-ovn-kubernetes" project
+    And admin executes existing pods die with labels:
```
pruan-rht, line 125: @anuragthehatter this step doesn't sound right to me.

anuragthehatter: @pruan-rht Existing step ```all existing pods die with labels``` doesn't support admin capabilities so added ```all | admin executes``` to allow flexibility to poke in system projects as well. I also had an option to switch to pseudo user and execute this non-admin'ly but guess the change will be def. useful in future. Also it won't break existing test which uses this step as its ```all | admin executes``` now.

Line 124 is deleting ds associated with ovnkube-master pods and 125 is making sure those all pods are gone successfully

anuragthehatter: @pruan-rht Let me know if any further questions

pruan-rht: have to regressed it against existing scenarios that use the `all` existing pods die...`?

anuragthehatter: Ya i did that initially and regressed again a case likehttps://github.com/openshift/verification-tests/blob/master/features/networking/multus.feature#L366. No issues 
Logs: https://privatebin-it-iso.int.open.paas.redhat.com/?13d0c61305f04132#u2ZwRH/SPHBSpqeCSHi/mlKB+JGPUQnGApuU0un2Fr0=

anuragthehatter:
/assign @pruan-rht 

Please hep merging it if no more questions!

anuragthehatter:
Bump!

pruan-rht (commented):
/lgtm

PR #1243, opened 2020-06-15
OCP-18482: Modify testdata file locations

quarterpin:
In order to effectively utilize the `Obtain` expression, soft links to the relevant test data files have been added to ensure the test cases get executed properly. In the current state, without this change, the `Examples` expression basically passes the raw values for the variables in the steps which cause `file/directory not found` error to occur when fetching the test data files and the execution fails. 

Due to some transient environment issues as of writing, I will share the V3-runner test with the modified test case, shortly.

### DIFF
--- a/features/routing/rate-limit.feature
+++ b/features/routing/rate-limit.feature
@@ -22,7 +22,7 @@ Feature: Testing haproxy rate limit related features
     When I run the :create client command with:
       | f | <service> |
     Then the step should succeed
-    Given I obtain test data file "routing/<route>"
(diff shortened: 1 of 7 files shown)


quarterpin:
Succesful v3-runner after the changes applied is as below: 
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/104539/console

@lihongan , request for a peer review.

lihongan:
/lgtm
thank you @quarterpin

PR #1248, opened 2020-06-16
45 rule fixes

yapei:


lib/rules/web/admin_console/4.5/common_ui_elements.xyaml
```diff
@@ -427,12 +427,7 @@ uncheck_checkbox:
 check_link_and_text:
   element:
     selector:
-      xpath: //a[contains(.,'<text>')]
-  scripts: 
-    - command: |
-        var encodedhref = document.querySelector("a[data-test-id='<text>']").href
-        return decodeURIComponent(encodedhref).endsWith("<link_url>")
-      expect_result: true
+      xpath: //a[contains(.,'<text>') and contains(@href,'<link_url>')]
```
yapei, line 430: revert back the changes

lib/openshift/subscription.rb
```diff
@@ -8,7 +8,7 @@ def current_csv(user: nil, quiet: false, cached: false)
     end
 
     def ready?(user:, quiet: false, cached: false)
-       res = get(user: user, cached: cached, quiet: quiet)
+       res = get(user: user, quiet: quiet)
```
yapei, line 11:
```
    def get(user: nil, quiet: false)
      user = default_user(user)
```
only two parameters are accepted for `get`

yapei:
@liangxia  Can you help merge if it looks good to you?

lib/openshift/subscription.rb
liangxia: There is one parameter unused in `ready`.

yapei: @liangxia  Updated and got passed jenkins log at #104672

PR #1249, opened 2020-06-16
Add obtain test dir

xiuwang:

Passed locally http://pastebin.test.redhat.com/875383

### DIFF
--- a/features/build/stibuild.feature
+++ b/features/build/stibuild.feature
@@ -4,7 +4,7 @@ Feature: stibuild.feature
     Given I have a project
     Given I obtain test data file "templates/<template>"
     And I run the :new_app client command with:
-      | file | <template> |
+      | file | tc498848-s2i.json |
     Then the step should succeed
     And the "sample-build-1" build was created
     When I run the :cancel_build client command with:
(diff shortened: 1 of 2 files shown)


features/build/stibuild.feature
```diff
@@ -4,7 +4,7 @@ Feature: stibuild.feature
     Given I have a project
     Given I obtain test data file "templates/<template>"
     And I run the :new_app client command with:
-      | file | <template> |
+      | file | tc498848-s2i.json |
```
akostadinov, line 7: this change doesn't look right

Other than that looks good.

xiuwang: The original snenario is a multiple one, now other cases have removed from 4.x, only this case is left. I don't want to change the automation field in polarion , just update this case instead

akostadinov: but few lines above you have `I obtain test data file "templates/<template>"` which depends on the parameter. It doesn't make sense to me. Updating polarion is just calling a simple script 
https://mojo.redhat.com/docs/DOC-1125811

Although I don't see why you'd need to change.

akostadinov: Ok, I've got it. Not ideal but anyway.

PR #1256, opened 2020-06-17
Remove undefined step from the script

kasturinarra:
Replace step Given I wait for the pod named "hooks-3-deploy" to die with Given I wait for the "hooks-3-deploy" pod to die as cucumber says the step is undefined.


kasturinarra:
@zhouying7780  please help review, thanks !!

https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/104808/console

PR #1263, opened 2020-06-19
update OCP-25706

huiran0826:
Remove -9 in kill command, to just use kill which terminates the process the normal way.

### DIFF
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -224,7 +224,7 @@ Feature: SDN related networking scenarios
       | bash | -c | pgrep ovs-vswitchd |
     Then the step should succeed
     When I run command on the "<%= cb.node_name %>" node's sdn pod:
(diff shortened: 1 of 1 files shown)


features/networking/sdn.feature
```diff
@@ -224,7 +224,7 @@ Feature: SDN related networking scenarios
       | bash | -c | pgrep ovs-vswitchd |
     Then the step should succeed
     When I run command on the "<%= cb.node_name %>" node's sdn pod:
-      | bash | -c | pgrep ovs-vswitchd \| xargs kill -9 |
+      | bash | -c | pgrep ovs-vswitchd \| xargs kill |
```
rbbratta, line 227: could we use `pkill`?    `pkill ovs-vswitchd`

huiran0826: Thanks @rbbratta, updated it!

PR #1264, opened 2020-06-19
Remove case id from upgrade-prepare

jhou1:
This should fix the could not find scenario problem.

/cc @sunilcio @lyman9966


liangxia:
The script field in Polarion needs update, too

jhou1:
> The script field in Polarion needs update, too

Yes, thanks for the heads up, almost forgot this one.

PR #1265, opened 2020-06-19
remove case id from upgrade-prepare

barleyer:
This should fix the could not find scenario problem.

/cc: @zhouying7780


liangxia:
The script field in Polarion needs update, too.

PR #1268, opened 2020-06-19
automate case OCP-32002

QiaolingTang:
Adding automation scenario for https://bugzilla.redhat.com/show_bug.cgi?id=1823305. 
@anpingli @yanpzhan PTAL
Runner-v3 log: https://url.corp.redhat.com/99eb278

### DIFF
--- a/features/logging/kibana.feature
+++ b/features/logging/kibana.feature
@@ -145,3 +145,40 @@ Feature: Kibana related features
     When I run the :check_log_count web action
     Then the step should succeed
     """
+
+  # @author qitang@redhat.com
+  # @case_id OCP-32002
+  @admin
+  @destructive
+  @commonlogging
+  Scenario: Kibana logout function should log off user
+    Given the master version < "4.5"
+    Given I switch to the first user
+    Given I create a project with non-leading digit name
+    Given evaluation of `project.name` is stored in the :proj_name clipboard
+    Given I obtain test data file "logging/loggen/container_json_log_template.json"
+    When I run the :new_app client command with:
+      | file | container_json_log_template.json |
+    Then the step should succeed
+    Given a pod becomes ready with labels:
+      | run=centos-logtest,test=centos-logtest |
+    Given the first user is cluster-admin
+    And I use the "openshift-logging" project
+    And I wait for the "project.<%= cb.proj_name %>" index to appear in the ES pod with labels "es-node-master=true"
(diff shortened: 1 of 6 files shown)


features/logging/kibana.feature
```diff
@@ -145,3 +145,40 @@ Feature: Kibana related features
...
+    And I perform the :kibana_find_index_pattern web action with:
+      | index_pattern_name | .operations.* |
+    Then the step should succeed
+    When I perform the :logout_kibana web action with:
+      | kibana_url | https://<%= cb.kibana_url %> |
+    Then the step should succeed
+    Given 10 seconds have passed
+    When I access the "https://<%= cb.kibana_url %>" url in the web browser
+    Then the step should succeed
+    When I perform the :login_kibana web action with:
```
anpingli, line 180: the bug needs to cover the point  ' the user must log in again once logout'.  For all auth types, the user should be redirected to the OAuth page.   For automatical auth type(Google, GitHub) once the user clicks the auth type, it will be redirected to Google or GitHub to finish the authorization.    For none automatical auth type(HTPASSWORD, LDAP), it should prompt the user to enter username/ password.
Could we cover the none automatical auth type be redirected to the re-login page?

QiaolingTang: I believe all the steps are in the action [`login_sequence`](https://github.com/openshift/verification-tests/blob/master/lib/rules/web/console/base/login.xyaml#L7), I already used it in https://github.com/openshift/verification-tests/pull/1268/files#diff-43a5548c1984e71bf1630e52a75df78bR6. I don't know if we could test login with Google IDP or Github IDP. If we can't, then the case will fail at https://github.com/openshift/verification-tests/pull/1268/files#diff-1f0bdbeeb96ac98d1a73878de51047adR169.

QiaolingTang: No matter what type of IDP we're testing, we must log into Kibana first, if we can't log in, then checking if the page could be redirected to the re-login page or not doesn't make sense.

QiaolingTang: @yanpzhan Could you please help to confirm if we could test with Google/Github/Gitlab IDP ?

yanpzhan: We could not use Google/Github/Gitlab IDP when run console automation, the console is different from old console( version 3.x) which could use token to run web automation.

PR #1269, opened 2020-06-22
Remove pip and minio

liangxia:
pip and minio are added in https://github.com/openshift/verification-tests/pull/987

We also do install python3, pip and several other packages via pip in oc4x dockerfiles, so the decision is that we move them (python3 and so on from oc4x image, and pip, minio from cucushift-public image) into cucushift-base image, to reduce some duplicate effort and building resource/time.


liangxia:
I have made the changes in cucushift-base:latest and cucushift:ocxx, and verified they are working.

Another benefit is that the image size of oc4x decreased by 30% +

PR #1276, opened 2020-06-23
Fix restore steps for scheduler

kasturinarra:
Cases fail if run immediately after the scheduler operator cases, fixing the same using the PR so that no failures are seen.

### DIFF
--- a/features/step_definitions/scheduler.rb
+++ b/features/step_definitions/scheduler.rb
@@ -24,5 +24,30 @@ Given /^the #{QUOTED} scheduler CR is restored after scenario$/ do |name|
     opts = {resource: 'scheduler', resource_name: name, p: patch_json, type: 'merge' }
     @result = _admin.cli_exec(:patch, **opts)
     raise "Cannot restore scheduler: #{name}" unless @result[:success]
+    timeout = 300
+    wait_for(timeout) do
+      @result = admin.cli_exec(:get, resource: "clusteroperators", resource_name: "kube-scheduler", o: "jsonpath={.status.conditions[?(.type == \"Progressing\")].status}")
+      if @result[:response] == "True"
+        break
+      end
+    end
+    wait_for(timeout) do
+      @result = admin.cli_exec(:get, resource: "clusteroperators", resource_name: "kube-scheduler", o: "jsonpath={.status.conditions[?(.type == \"Progressing\")].status}")
+      if @result[:response] == "False"
+        break
+      end
+    end
+    wait_for(timeout) do
+      @result = admin.cli_exec(:get, resource: "clusteroperators", resource_name: "kube-scheduler", o: "jsonpath={.status.conditions[?(.type == \"Degraded\")].status}")
+      if @result[:response] == "False"
+        break
+      end
+    end
+    wait_for(timeout) do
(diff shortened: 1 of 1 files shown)


kasturinarra:

https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1626/console

features/step_definitions/scheduler.rb
```diff
@@ -24,5 +24,15 @@
     opts = {resource: 'scheduler', resource_name: name, p: patch_json, type: 'merge' }
     @result = _admin.cli_exec(:patch, **opts)
     raise "Cannot restore scheduler: #{name}" unless @result[:success]
+    conditions= cluster_operator("kube-scheduler").conditions
+    expected = {"Degraded"=>"False", "Progressing"=>"False", "Available"=>"True"}
+    sleep 300
```
pruan-rht, line 29: @kasturinarra please don't use hardcoded sleep

kasturinarra: @pruan-rht agree, could you please  give me some pointers on how to come across this ?

pruan-rht: @kasturinarra do you mean how to avoid the hard sleep?  Please refer to https://github.com/openshift/verification-tests/blob/master/lib/openshift/pod.rb#L268 for an example.  Basically you use the `wait_for` call which will exit as soon as the condition check is meet.

kasturinarra: @pruan-rht i tried using wait_for with the below steps [1] but i always end up seeing premature failures, so can we keep it this way for now ?

[1] http://pastebin.test.redhat.com/879302

kasturinarra: @pruan-rht could you please help review again ? Thanks !!

kasturinarra:
@zhouying7780 could you please help review again ? Thanks !!

zhouying7780:
@liangxia Could you please help review and merge ? thanks .

liangxia:
@pruan-rht  Could you check if current version address your concern ?

features/step_definitions/scheduler.rb
```diff
@@ -24,5 +24,30 @@
...
     raise "Cannot restore scheduler: #{name}" unless @result[:success]
+    timeout = 300
+    wait_for(timeout) do
+      @result = admin.cli_exec(:get, resource: "clusteroperators", resource_name: "kube-scheduler", o: "jsonpath={.status.conditions[?(.type == \"Progressing\")].status}")
+      if @result[:response] == "True"
+        break
+      end
+    end
+    wait_for(timeout) do
+      @result = admin.cli_exec(:get, resource: "clusteroperators", resource_name: "kube-scheduler", o: "jsonpath={.status.conditions[?(.type == \"Progressing\")].status}")
```
pruan-rht, line 35: @kasturinarra this line is the same as line 29.  Typo perhaps?

kasturinarra: @pruan-rht not a typo actually, it is the same line but checking for two different conditions, one for true and another for false.  Let me explain you here.

Once the scheduler cluster is patched it takes sometime for the kube-scheduler to come into progressing = True and if we do not wait for the kube-scheduler progressing status to be true  then script will move further and it will fail with out proper restoration of scheduler, that is the reason for having it twice.

PR #1245, opened 2020-06-15
pod: allow specifying container in pod to execute on

rbbratta:
For OVN we sometimes need to execute on a specific container
in the pod in order to directly connect to the database over
unix socket and avoid SSL network connections.

PR #1278, opened 2020-06-23
OCPQE-933: Fix bad regexp replacement

rbbratta:
Incorrect constant replacement without testing.

### DIFF
--- a/features/step_definitions/pod.rb
+++ b/features/step_definitions/pod.rb
@@ -162,7 +162,7 @@ end
 # Called without the 'regardless...' parameter ir checks that pod reaches a
 #   ready status, then somehow dies. With the parameter it just makes sure
 #   pod os not there regardless of its current status.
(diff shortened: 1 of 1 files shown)


akostadinov:
@rbbratta , @pruan-rht , this should break existing steps. What was the issue fixed?

rbbratta:
> @rbbratta , @pruan-rht , this should break existing steps. What was the issue fixed?

The original patch was wrong
https://github.com/openshift/verification-tests/pull/1245/files#diff-4dca78f22294466e0439bcec8a262ce3L167
```diff
-Given /^I wait for the pod(?: named "(.+)")? to die( regardless of current status)?$/ do |name, ignore_status|
+Given /^I wait for the pod#{OPT_QUOTED} to die( regardless of current status)?$/ do |name, ignore_status|
```

I missed the `named` arg and though `OPT_QUOTED` would work, but it won't because of the `named` arg.

PR #1279, opened 2020-06-23
use negated character class in quoted regex

rbbratta:
It looks like some cases were using greedy matching and other cases we
used the constant QUOTED with non-greedy (lazy) matching.

A better alternative is to use a negated character class containing the
double-quote character.  This should obviate the need for back-tracking
and non-greedy matching and is the preferred method for matching inside
delimiters.

We could also use negative lookahead if the delimiter is more than one
character and not suitable for a negated character class

See discussion in _Mastering Regular Expressions, 3rd Edition by Jeffrey E.F. Friedl (O'Reilly)_

Chapter 4.  Paragraph "Problems of Greediness"


rbbratta:
untested

rbbratta:
reference: https://www.oreilly.com/library/view/mastering-regular-expressions/0596528124/ch04.html

features/step_definitions/_constants.rb
```diff
@@ -1,11 +1,11 @@
...
 HTTP_URL = '(https?:\/\/.+)'
 USER = 'the( \\S+)? user'
-NUMBER = '([0-9]+|<%=.+?%>)'
-WORD = '(\w+|<%=.+?%>)'
+NUMBER = '([0-9]+|<%=(?!%>).+?%>)'
+WORD = '(\w+|<%=(?!%>).+?%>)'
 OPT_WORD = "(?: #{WORD})?"
 # use for regular expression
-RE='/(.+?)/'
+RE='/([^/\n]+)/'
```
akostadinov, line 11: we should allow `/` in regular expressions

features/step_definitions/_constants.rb
```diff
@@ -1,11 +1,11 @@
 OPT_SYM = '(?: :(\S+))?'
 SYM = ':(\S+)'
-OPT_QUOTED = '(?: "(.+?)")?'
-QUOTED = '"(.+?)"'
+OPT_QUOTED = '(?: "([^"\n]+)")?'
+QUOTED = '"([^"\n]+)"'
```
akostadinov, line 4: we can't have `\n` anyway. The `"` character being forbidden will prevent us from using expressions between `<%= some code with quotes %>`

akostadinov:
Could you paste relevant parts of "Problems of Greediness" or explain a little what issue you see with current usage? It's interesting to know.

features/step_definitions/_constants.rb
rbbratta: I'm not sure how, given that `/` is the delimiter.  The non-greedy `.+?` will stop at the first `/`. 

```ruby
>> "/asdf/".match('/([^/\n]+)/')
=> #<MatchData "/asdf/" 1:"asdf">
>> "/as/df/".match('/([^/\n]+)/')
=> #<MatchData "/as/" 1:"as">
>> "/asdf/".match('/(.+?)/')
=> #<MatchData "/asdf/" 1:"asdf">
>> "/as/df/".match('/(.+?)/')
=> #<MatchData "/as/" 1:"as">
```

rbbratta:
Lazy requires back-tracking, it is better to eliminate unwanted states directly.


https://www.oreilly.com/library/view/mastering-regular-expressions/0596528124/ch04.html

It starts with the section on Lazy quantifiers.

### The Essence of Greediness, Laziness, and Backtracking ###
The lesson of the preceding section is that it makes no difference whether there are greedy or lazy components to a regex; an overall match takes precedence over an overall non-match. This includes taking from what had been greedy (or giving to what had been lazy) if that’s what is required to achieve a match, because when a “local failure” is hit, the engine keeps going back to the saved states (retracing steps to the piles of bread crumbs), trying the untested paths. _**Whether greedily or lazily, every possible path is tested before the engine admits failure.**_

The order that the paths are tested is different between greedy and lazy quantifiers (after all, that’s the whole point of having the two!), but in the end, if no match is to be found, it’s known only after testing every possible path.

If, on the other hand, there exists just one plausible match, both a regex with a greedy quantifier and one with a lazy quantifier find that match, although the series of paths they take to get there may be wildly different. In these cases, selecting greedy or lazy doesn’t influence what is matched, but merely how long or short a path the engine takes to get there (which is an efficiency issue, the subject of Chapter 6).

Finally, if there is more than one plausible match, understanding greediness, laziness, and backtracking allows you to know which is selected.

features/step_definitions/_constants.rb
rbbratta: `.` doesn't match `\n`, but negated classes can match `\n`.

I don't think we can have quotes within quotes either.

```ruby
>> '"asdf"'.match('"(.+?)"')
=> #<MatchData "\"asdf\"" 1:"asdf">
>> '"as"d"ff"'.match('"(.+?)"')
=> #<MatchData "\"as\"" 1:"as">
>> '"asdf"'.match('"([^"\n]+)"')
=> #<MatchData "\"asdf\"" 1:"asdf">
>> '"as"d"f"'.match('"([^"\n]+)"')
=> #<MatchData "\"as\"" 1:"as">
```

rbbratta:
I'll try to come up with more proper test cases for the regex.

features/step_definitions/_constants.rb
akostadinov: It will try that, but then the rest of the step will likely not match, so there will be a backtrack where the `/` is used as part of the RE. 
```
[13] pry(main)> 'asd /first re with/ slach/ and /second re with / slash/ yo' =~ /asd \/(.+?)\/ and \/(.+?)\/ yo/
=> 0
```

features/step_definitions/_constants.rb
akostadinov: About quotes, see my reply about slashes. wrt `\n` it can match depending on RE flags. In either case, I believe gherkin syntax will not allow new lines anyway so I think the handling of `\n` is just making REs harder to read.

Do you see any place where you can insert `\n` where we don't want it to match?

rbbratta:
not needed

PR #1280, opened 2020-06-23
add step to get cloudcredential based on the cluster's platform

pruan-rht:


akostadinov:
Hi, I would like to see an actual scenario that needs credentials to see what implementation will make sense.

pruan-rht:
@akostadinov the usage various...for `metering` the use case is to store the report data of to a 3rd party storage (s3-bucket, gce/azure storage).  The limitation of using this approach is that the cluster has to be the same provider as the storage if we were to get the creds from the cluster.  Hmm...that kind of kill the argument of flexibility of 3rd party storage as we will be forcing the cluster and off-cluster storage to be the same.  I think the class & method would still be useful though.  But for the case of `metering`, we may have to end up storing the creds in `private` or use `git-crypt`.  WDYT?

akostadinov:
I understood. How exactly are you going to use the credentials is the question. I doubt you use in raw format.

pruan-rht:
@akostadinov my use usage is ```apiVersion: v1
kind: Secret
metadata:
  name: "my-aws-secret"
data:
  aws-access-key-id: <base64_encoded_id>
  aws-secret-access-key: <base64_encoded_key> ```

akostadinov:
Ok, can you submit this change together with your actual use case?

akostadinov:
btw I used your intelligence to update GCE in #1308. I think this feature would better go somehow in `iaas.rb`. But still I want to see the actual code user before merging.

akostadinov:
@pruan-rht , can you point me at the place where credentials are used? Seems to be in some other step but this is not visible in that scenario.

pruan-rht:
@akostadinov it's used within this step 
 https://github.com/openshift/cucushift/blob/master/metering/storage.feature#L94
which is subbed into here https://github.com/openshift/cucushift/blob/master/testdata/metering/secrets/s3.yaml

pruan-rht:
@akostadinov bump...PTAL.  Thx!

akostadinov:
I'm not completely sure such low-level step could be abused in the future. But lets see.

/lgtm

PR #1281, opened 2020-06-24
Fix OCP-22618: namespaces "aos-qe-ci" not found

liangxia:


akostadinov:
Are you sure guys this is not a bug in the command? Why is parameter NAMESPACE ignored?

liangxia:
Report https://bugzilla.redhat.com/show_bug.cgi?id=1875324, but closed as not a bug.

akostadinov:
Is this possibly a duplicate of https://bugzilla.redhat.com/show_bug.cgi?id=1906121 ?
If you explain the bug as it is explained in the 4.7 issue, then perhaps it will be fixed.

liangxia:
There are differences for the two bugs.
For 1906121, when we test a test case, it works for 4.6, but not for 4.7, so it's a regression for 4.7
For 1875324, when we test a test case, if it works, then it works for all version; if it does not work, it does not work for all versions. So it is not a regression.

For 1906121, there are no namespace in config context, even after user run command new-project.
For 1875324, there are no default namespace in config context for the kubeconfig which is generated by installer.
So when automation script is using pseudo admin, and do not specify/create project, it will use SA's project aos-qe-ci, then it will show "namespace not found".
And for this specific PR, it is passing namespace to `-p`, ` -p, --param=[]: Specify a key-value pair (eg. -p FOO=BAR) to set/override a parameter value in the template`, the namespace is used to override the value in the template, it does not pass the namespace to the process/create command (It needs `-n` to pass the namespace to the proces/create command).

PR #1290, opened 2020-06-25
metering deploy refactor to use sharedPVC as default

pruan-rht:
1. HDFS storage is for development and not production ready.  Change the default storage to use NFS/sharedPVC approach.
2. added support to install metering using OLM via CLI (quicker than from UI)
3. added support methods
4. testdata templates and actual examples.
run log: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1639/console

### DIFF
--- a/lib/openshift/metering_config.rb
+++ b/lib/openshift/metering_config.rb
@@ -10,5 +10,9 @@ module BushSlicer
       raw_resource(user: user, cached: cached, quiet: quiet).dig('spec', 'storage', 'hive', 'type')
     end
 
+    def namespace(user: nil, cached: true, quiet: false)
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('metadata', 'namespace')
+    end
+
   end
 end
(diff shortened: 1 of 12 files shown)


pruan-rht:
@akostadinov PTAL

lib/openshift/metering_config.rb
```diff
@@ -10,5 +10,9 @@ def hive_type(user: nil, cached: true, quiet: false)
       raw_resource(user: user, cached: cached, quiet: quiet).dig('spec', 'storage', 'hive', 'type')
     end
 
+    def namespace(user: nil, cached: true, quiet: false)
```
akostadinov, line 13: This method is redundant. All `ProjectResource` classes have the `project` method. To get the name you do `obj.project.name`

akostadinov: other than that lgtm

pruan-rht: @akostadinov ok, I took that method out.  Please merge is there are no more issues.

lib/openshift/package_manifest.rb
```diff
@@ -2,5 +2,31 @@ module BushSlicer
   class PackageManifest < ClusterResource
     RESOURCE = "packagemanifests"
```
akostadinov, line 3: this needs to be full API name but it seems to be from before

PR #1291, opened 2020-06-26
exit gracefully without throwing exception[WIP]

miyadav:
It is related to - https://issues.redhat.com/browse/OCPQE-946

### DIFF
--- a/features/step_definitions/machine.rb
+++ b/features/step_definitions/machine.rb
@@ -3,6 +3,9 @@ Given(/^I have an IPI deployment$/) do
   machines = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api"))
   if machines.length == 0
     raise "Not an IPI deployment, there are no machines"
+ #TODO
+    #Exit gracefully without giving RuntimeError like below
+    #Not an IPI deployment, there are no machines (RuntimeError)
   end
 
   machines.each do | machine |


features/step_definitions/machine.rb
```diff
@@ -3,6 +3,9 @@
   machines = BushSlicer::Machine.list(user: admin, project: project("openshift-machine-api"))
   if machines.length == 0
     raise "Not an IPI deployment, there are no machines"
+ #TODO
+    #Exit gracefully without giving RuntimeError like below
+    #Not an IPI deployment, there are no machines (RuntimeError)
```
akostadinov, line 8: what do you mean with "gracefully"? How more graceful it can be conpared to now?

miyadav: I mean since it is not an IPI deployment , we just exit execution without giving RuntimeError , instead , simply a message like "This enviroment is not suitable for the scaling tests"
I am not sure if there is any other way , to make sure that the tests are not even selected for UPI envs

akostadinov: They shouldn't be selected in the first place. If we mark as skipped which is possible, then how do we report them to the test case management system? They are neither passed, nor failed. And we only have blocked status which may break reporting.

jhou1: @miyadav These tests should not run on the UPI clusters in the first place. If it's an UPI cluster the test fails anyway, so we fail it early. We need to use Polarion to prevent such tests from being executed on UPI clusters, so if it happens we have to check how the run is created and remove the test cases from the run because we only want passed or failed status finally.

miyadav: @jhou1  ok , will check

miyadav:
Since this will not be needed as it is managed by yaml run template , which is fixed now , details updated in JIRA

PR #1301, opened 2020-06-30
Removing test data file as dup

kasturinarra:
Removing this file and updating the image in cucushift/testdata/scheduler/pod_with_nodename.json so that it could run in disconnected env.


liangxia:
/lgtm

Verified the file is only used in cucushift, it's safe to remove it from verification-tests.
```
$ grep -r 'pod_with_nodename' *
Binary file features/tierN/.git/index matches
features/tierN/scheduler/operator.feature:    Given I obtain test data file "scheduler/pod_with_nodename.json"
features/tierN/scheduler/operator.feature:    When I run oc create over "pod_with_nodename.json" replacing paths:
```

PR #1128, opened 2020-05-20
OCP-29741 upgrade auto

barleyer:
@xingxingxia please take a look!

PR #1304, opened 2020-07-01
Change the title of the prepare test case

barleyer:
Change the title of the prepare test case to consistent with  the title of  the upgrade-check case !


xingxingxia:
/lgtm
Sorry missed this issue when reviewing #1128 . Suggest provide at least one time of pass Runner-v3 jobs to avoid such issue.

PR #1305, opened 2020-07-01
fix jira OCPQE-984

pruan-rht:
run log https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1688/console


pruan-rht:
@akostadinov @liangxia PTAL, thanks!

PR #1306, opened 2020-07-01
Update the step to avoid choosing a master node

qinpingli:
Before the update,  the node chosen might be a master node, in OCP4.x, the pod without tolerations created in this project can not be scheduled successfully.

### DIFF
--- a/features/step_definitions/project.rb
+++ b/features/step_definitions/project.rb
@@ -164,7 +164,8 @@ Given /^admin creates a project with a random schedulable node selector$/ do
     | project_name  | #{project_name}       |
     | node_selector | #{project_name}=label |
     })
-  step %Q/I select a random node's host/
+  step %Q/I store the ready and schedulable workers in the :nodes clipboard/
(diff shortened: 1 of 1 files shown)


features/step_definitions/project.rb
```diff
@@ -164,7 +164,8 @@
     | project_name  | #{project_name}       |
     | node_selector | #{project_name}=label |
     })
-  step %Q/I select a random node's host/
+  step %Q/I store the ready and schedulable workers in the :nodes clipboard/
+  step %Q/I use the "<%= cb.nodes[0].name %>" node/
```
pruan-rht, line 168: @qinpingli since the original step selects a random node, I think you should do the same with your work around.  Use the `sample` method instead of always selecting the first element

qinpingli: @pruan-rht Thank you, good catch! I removed the step `I use the <node.name> node`.

qinpingli:
@pruan-rht  Please help take a look again, thx.

PR #1307, opened 2020-07-01
IPAM thrash test for SDN-892 and a test file

anuragthehatter:
A thrash test for ovnkube master IPAM as discussed with dev. ovnkube master is now allocating mac/addresses to the pods instead of northd allocation approach to improve scale/performance. The test will ennsure that each pod will get IP( means Running) during rapid creation/deletion of pods

This test will take 15 minutes to run. Lets discuss if it make sense to include in daily CI runs
cc @zhaozhanqi @rbbratta @huiran0826 @weliang1


features/networking/ovn.feature
```diff
@@ -0,0 +1,25 @@
+Feature: OVN related networking scenarios
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-32205
+  @admin
+  Scenario: Thrashing ovnkube master IPAM allocator by creating and deleting various pods on a specific node
+    Given the env is using "OVNKubernetes" networkType
```
rbbratta, line 7: It looks like this step is the only step that requires admin?  Would this test also work on SDN?  Then we could remove the step and not require admin maybe.  Maybe it's not worth it.

features/networking/ovn.feature
```diff
@@ -0,0 +1,25 @@
...
+    And I store all worker nodes to the :nodes clipboard
+    And I have a project
+    Given I obtain test data file "networking/generic_test_pod_with_replica.yaml"
+    When I run the steps 10 times:
+    """
+    When I run oc create over "generic_test_pod_with_replica.yaml" replacing paths:
+      | ["spec"]["replicas"]                     | 25                      |
+      | ["spec"]["template"]["spec"]["nodeName"] | <%= cb.nodes[0].name %> |
+    Then the step should succeed
+    And 25 pods become ready with labels:
```
zhaozhanqi, line 17: not sure all platform can schedule 25 pods in one node at same times and make them running.  some of them may be 'pending' if resource is not enough.

testdata/networking/generic_test_pod_with_replica.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
zhaozhanqi, line 1: do we still need to this file? remembered we have one list_for_pods.json

features/networking/ovn.feature
anuragthehatter: @rbbratta This is OVN specific as scenario title suggests.

features/networking/ovn.feature
anuragthehatter: I guess we can mark this test as low priority, not to be run everytime. 25*10 (250) is chosen to check IP address recycling by IPAM on a node.

weliang1: @anuragthehatter look like the  the scenario is Thrashing ovnkube master IPAM allocator, after pods getting 10 ip addresses each time, do we have any way to check if those ip addressed are from ovnkube master IPAM allocator?

For SDN, we can check pods' ip addresses allocated from CNI IPAM:
sh-4.2# ls -l /var/lib/cni/networks/openshift-sdn
total 44
-rw-r--r--. 1 root root 64 Jul  9 14:00 10.130.2.10
-rw-r--r--. 1 root root 64 Jul  9 14:00 10.130.2.11
-rw-r--r--. 1 root root 64 Jul  9 13:48 10.130.2.2
-rw-r--r--. 1 root root 64 Jul  9 13:48 10.130.2.3
-rw-r--r--. 1 root root 64 Jul  9 16:15 10.130.2.78
-rw-r--r--. 1 root root 64 Jul  9 16:15 10.130.2.79
-rw-r--r--. 1 root root 64 Jul  9 16:15 10.130.2.80
-rw-r--r--. 1 root root 64 Jul  9 16:15 10.130.2.81
-rw-r--r--. 1 root root 64 Jul  9 16:15 10.130.2.82
-rw-r--r--. 1 root root 64 Jul  9 16:15 10.130.2.83
-rw-r--r--. 1 root root 11 Jul  9 16:15 last_reserved_ip.0
sh-4.2#

testdata/networking/generic_test_pod_with_replica.yaml
anuragthehatter: let me take a look. I guess we might be able to reuse that. The last time i tried is it creates a service as well and when we tries to re-create pods, it sborts with error sayign svc already exist.

features/networking/ovn.feature
anuragthehatter: let me check with dev. It seems black box to me for OVN but not sure at the moment

testdata/networking/generic_test_pod_with_replica.yaml
zhaozhanqi: aha, get it. there is a service in list_for_pods.json.   then it's not a problem

features/networking/ovn.feature
anuragthehatter: @weliang1 Confirmed with Aniket, it seems internal to OVN northd and there is no way to populate IPs like above in OpenshiftSDN. So seems black box for now. Yea, this test just make sure that any pod should not stuck in pending state with IP allocation error while ongoing thrashing activity

weliang1: @anuragthehatter Thanks for confirming this. Better to have a way to check allocated IP as SDN does in the future.
/LGTM

akostadinov:
Just one concern - shouldn't such kind of tests be part of SVT test suites? SVT team deals with perf/scale/reliability testing. I'm ok to include it but if we have such concerns, better work with SVT team to add that.

anuragthehatter:
Added OCP-32184 as well https://privatebin-it-iso.int.open.paas.redhat.com/?014d16dc271fb8a3#5MH+EfKi8yrAQ7ebFOLUOyYKNgkOhWQWnmUSxnWPpg0=

features/networking/ovn.feature
anuragthehatter: Confirmed with dev. Even 5 pods creation deletion would be sufficient to check IPAM allocator resiliency. So we will now only create 5 pods but in loop of 10. So it will be only 5 pods at a time. Should be a good functional test to keep. cc @akostadinov

anuragthehatter:
/assign @pruan-rht 

Please help merging it!

anuragthehatter:
Bump!

pruan-rht (commented):
/lgtm

features/networking/ovn.feature
```diff
@@ -143,3 +143,51 @@ Feature: OVN related networking scenarios
...
+    Given I have a pod-for-ping in the project
+    Then evaluation of `pod.ip` is stored in the :hello_pod_ip clipboard
+    When I execute on the pod:
+       | bash | -c | ip a show eth0 |
+    Then the step should succeed
+    And evaluation of `@result[:response].match(/\h+:\h+:\h+:\h+:\h+:\h+/)[0]` is stored in the :hello_pod_mac clipboard
+  
+    Given I store the ovnkube-master "north" leader pod in the clipboard
+    And evaluation of `pod.ip_url` is stored in the :ovn_nb_leader_ip clipboard
+    And evaluation of `pod.node_name` is stored in the :ovn_nb_leader_node clipboard
```
rbbratta, line 186: do we need these two lines?  doesn't look like we use `:ovn_nb_leader` clipboards.

features/networking/ovn.feature
```diff
@@ -143,3 +143,51 @@ Feature: OVN related networking scenarios
...
+    And admin executes on the pod:
+      | bash | -c | ovn-nbctl list logical_switch_port \| grep "hello-pod" -C 10 |
+    Then the step should succeed
+    #Make sure addresses doesn;t say dynamic but show ip and mac assigned to the hello-pod and dynamic_addresses field should be empty
```
rbbratta, line 190: I think style likes a space after `#`

```
# Make sure addresses don't say dynamic
```

features/networking/ovn.feature
anuragthehatter: Taken care, Thanks. Seems like i was doing some experiment and left it there

PR #1309, opened 2020-07-02
auto some egressIP cases

huiran0826:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1700/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1702/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1703/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1704/console

Note: Case OCP-25640 failed due to https://bugzilla.redhat.com/show_bug.cgi?id=1762580

@zhaozhanqi  @weliang1 @anuragthehatter @rbbratta

### DIFF
--- a/features/networking/egress-ip.feature
+++ b/features/networking/egress-ip.feature
@@ -232,3 +232,142 @@ Feature: Egress IP related features
     Then the step should succeed
     And the output should not contain "reg0=0x"
     """
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-19973
+  @admin
+  @destructive
+  Scenario: The egressIP should still work fine after the node or network service restarted
+    Given the valid egress IP is added to the node
+    And I have a project
+    And I have a pod-for-ping in the project
+
+    # add the egress ip to the project
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+
+    When I execute on the pod:
+      | curl | -s | --connect-timeout | 5 | ifconfig.me |
+    Then the step should succeed
+    And the output should contain "<%= cb.valid_ip %>"
+
+    #Restart the network service
+    Given I restart the network components on the node
+    And I wait up to 120 seconds for the steps to pass:
+    """
+    When I execute on the pod:
+      | curl | -s | --connect-timeout | 5 | ifconfig.me |
+    Then the step should succeed
+    And the output should contain "<%= cb.valid_ip %>"
+    """
+
+    # Reboot the node which patched egressIP
+    Given the host is rebooted and I wait it up to 600 seconds to become available
+    When I execute on the pod:
+      | curl | -s | --connect-timeout | 5 | ifconfig.me |
+    Then the step should succeed
+    And the output should contain "<%= cb.valid_ip %>"
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-15998
+  @admin
+  Scenario Outline: Invalid egressIP should not be acceptable
+    Given I select a random node's host
+    When I run the :patch admin command with:
+      | resource      | hostsubnet                     |
+      | resource_name | <%= node.name %>               |
+      | p             | {"egressIPs": <invalid_ip> }   |
+      | type          | merge                          |
+    Then the step should fail
+    And the output should contain "Invalid JSON Patch"
+
+    Examples:
+      | invalid_ip              |
+      | fe80::5054:ff:fedd:3698 |
+      | a.b.c.d                 |
+      | 10.10.10.-1             |
+      | 10.0.0.1/64             |
+      | 10.1.1/24               |
+      | A008696                 |
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-25694
+  @admin
+  @destructive
+  Scenario: Random outages with egressIP
+    Given I store the schedulable workers in the :nodes clipboard
+    And the valid egress IP is added to the "<%= cb.nodes[0].name %>" node
+    And I have a project
+    Given I obtain test data file "networking/aosqe-pod-for-ping.json"
+    When I run oc create over "aosqe-pod-for-ping.json" replacing paths:
+      | ["spec"]["nodeName"] | "<%= cb.nodes[1].name %>" |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | name=hello-pod |
+
+    # add the egress ip to the project
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+
+    # Access some external host that is not responding
+    When I execute on the pod:
+      | nc | -v | -w2 | 8.8.8.8 | 23 |
+    Then the step should fail
+    And the output should contain "Operation timed out"
+
+    #Check the sdn logs on pod node
+    Given I use the "<%= pod.node_name %>" node
+    When I get the networking components logs of the node since "90s" ago
+    And the output should not contain "may be offline"
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-25640
+  @admin
+  @destructive
+  Scenario: Should be able to access to the service's externalIP with egressIP
+    Given I have a project
+    Given I store the schedulable nodes in the :nodes clipboard
+    And the Internal IP of node "<%= cb.nodes[0].name %>" is stored in the :hostip clipboard
+
+    # Create additional network through CNO
+    Given as admin I successfully merge patch resource "networks.config.openshift.io/cluster" with:
+      | {"spec":{"externalIP":{"policy":{"allowedCIDRs":["<%= cb.hostip %>/24"]}}}} |
+
+    # Clean-up required to erase above externalIP policy after testing done
+    Given I register clean-up steps:
+    """
+    Given as admin I successfully merge patch resource "networks.config.openshift.io/cluster" with:
+      | {"spec":{"externalIP":{"policy":{"allowedCIDRs": null}}}} |
+    """
+
+    # Create a svc with externalIP
+    Given I switch to the first user
+    Given I obtain test data file "networking/externalip_service1.json"
+    And I wait up to 300 seconds for the steps to pass:
+    """
+    When I run oc create over "externalip_service1.json" replacing paths:
+      | ["spec"]["externalIPs"][0] | <%= cb.hostip %> |
+    Then the step should succeed
+    """
+
+    # Create a pod
+    Given I obtain test data file "routing/caddy-docker.json"
+    When I run the :create client command with:
+      | f | caddy-docker.json |
+    Then the step should succeed
+    And the pod named "caddy-docker" becomes ready
+
+    # Patch egressIP to the node
+    Given the valid egress IP is added to the node
+
+    # Patch egressIP to a new project
+    Given I create a new project
+    And I have a pod-for-ping in the project
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+
+    # Curl externalIP:portnumber from project patched egressIP should pass
+    When I execute on the pod:
+      | /usr/bin/curl | --connect-timeout | 10 | <%= cb.hostip %>:27017 |
+    Then the output should contain:
+      | Hello-OpenShift-1 http-8080 |
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -589,7 +589,7 @@ Given /^I store a random unused IP address from the reserved range to the#{OPT_S
 
   validate_ip = IPAddr.new(reserved_range).to_range.to_a.shuffle.each { |ip|
     @result = step "I run command on the node's ovs pod:", table(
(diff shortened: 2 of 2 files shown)


features/networking/egress-ip.feature
```diff
@@ -232,3 +232,127 @@ Feature: Egress IP related features
...
+    When I get the networking components logs of the node since "90s" ago
+    And the output should not contain "may be offline"
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-25640
+  @admin
+  @destructive
+  Scenario: Should be able to access to the service's externalIP with egressIP
+    Given I have a project
+    And SCC "privileged" is added to the "system:serviceaccounts:<%= project.name %>" group
```
zhaozhanqi, line 313: just confirmed, still need 'priviledged' scc for 4.x version @weliang1

features/networking/egress-ip.feature
```diff
@@ -232,3 +232,127 @@ Feature: Egress IP related features
...
+    Given I register clean-up steps:
+    """
+    Given as admin I successfully merge patch resource "networks.config.openshift.io/cluster" with:
+      | {"spec":{"externalIP":{"policy":{"allowedCIDRs": null}}}} |
+    """
+
+    # Create a svc with externalIP
+    Given I switch to the first user
+    Given I obtain test data file "networking/externalip_service1.json"
+    And I wait up to 300 seconds for the steps to pass:
```
zhaozhanqi, line 331: need to wait up 300 seconds for creating svc?

huiran0826: @weliang1  There is a bug, right?

weliang1: @huiran0826 yes, the bug is not fixed yet: https://bugzilla.redhat.com/show_bug.cgi?id=1793099

features/networking/egress-ip.feature
weliang1: During my automating externalIP scripts,  running the scripts always cause random failure without line 313, not sure what's the behavior right now. @huiran0826 you may try to run your script without that line

features/networking/egress-ip.feature
huiran0826: @weliang1  Thanks!

features/networking/egress-ip.feature
huiran0826: @weliang1  Removed the line and rerun it, currently seems there is no difference.

features/networking/egress-ip.feature
```diff
@@ -232,3 +232,126 @@ Feature: Egress IP related features
...
+    Then the step should fail
+    And the output should contain "Invalid JSON Patch"
+
+    Examples:
+      | invalid_ip              |
+      | fe80::5054:ff:fedd:3698 |
+      | a.b.c.d                 |
+      | 10.10.10.-1             |
+      | 10.0.0.1/64             |
+      | 10.1.1/24               |
```
rbbratta, line 280: Slightly off-topic but have we ever tested the alternate IPv4 representations?

> For example, the quad-dotted IP address 192.0.2.235 represents the 32-bit decimal number 3221226219, which in hexadecimal format is 0xC00002EB. This may also be expressed in dotted hex format as 0xC0.0x00.0x02.0xEB, or with octal byte values as 0300.0000.0002.0353.

https://en.wikipedia.org/wiki/IPv4#Address_representations

features/networking/egress-ip.feature
```diff
@@ -232,3 +232,126 @@ Feature: Egress IP related features
...
+    And the output should not contain "may be offline"
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-25640
+  @admin
+  @destructive
+  Scenario: Should be able to access to the service's externalIP with egressIP
+    Given I have a project
+    Given I store the schedulable nodes in the :nodes clipboard
+    And the Internal IP of node "<%= cb.nodes[0].name %>" is stored in the :hostip clipboard
```
rbbratta, line 314: Will this work on IPv6 dual-stack?  Or maybe all egressIP tests are IPv4 for now?

huiran0826: Currently egressIP only works on openshfit-sdn  for IPv4 for now.

features/networking/egress-ip.feature
huiran0826: It  doesn't support other format of IP address, it will show the warning  " egressIPs in body should match '^(([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$",  I can add one example data here. Thanks Ross.

huiran0826:
@zhaozhanqi @rbbratta @weliang1 @anuragthehatter  Any other comments? If no, can you put lgtm? Thanks!

features/networking/egress-ip.feature
```diff
@@ -232,3 +232,127 @@ Feature: Egress IP related features
...
+    And the pod named "caddy-docker" becomes ready
+
+    # Patch egressIP to the node
+    Given the valid egress IP is added to the node
+
+    # Patch egressIP to a new project
+    Given I create a new project
+    And I have a pod-for-ping in the project
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+     | {"egressIPs": ["<%= cb.valid_ip %>"]} |
```
liangxia, line 352: Indention issue.

features/networking/egress-ip.feature
liangxia: We are hiding bugs in this way.  We should let it fail given there are bugs affecting this scenario.

huiran0826: Agree with you! We should not hide the bug.
But actually from the comments in bug https://bugzilla.redhat.com/show_bug.cgi?id=1793099, it is still not decided it is a valid bug or the time is reasonable.   I think we can keep the time here for now and for externalIP service scenario, there are other cases tracked to test it.

For this case it mainly to test egressIP which also has one bug open  which cause the case failing and we did not workaround it just let it fail.

features/networking/egress-ip.feature
huiran0826: Thanks, fixed it.

features/networking/egress-ip.feature
```diff
@@ -232,3 +232,127 @@ Feature: Egress IP related features
...
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+
+    When I execute on the pod:
+      | curl | -s | --connect-timeout | 5 | ifconfig.me |
+    Then the step should succeed
+    And the output should contain "<%= cb.valid_ip %>"
+
+    # Reboot the node which patched egressIP
+    Given the host is rebooted and I wait it up to 600 seconds to become available
```
weliang1, line 255: If cluster has three nodes, and egressIP is added to the node-A, line 255 is to reboot the host/node, I assume you code make sure the host is just the node-A, just double check

weliang1: The Scenario of this case is "The egressIP should still work fine after the node or network service restarted", how about add one more step to just restart network service without rebooting the node? WDYT

features/networking/egress-ip.feature
```diff
@@ -232,3 +232,127 @@ Feature: Egress IP related features
...
+      | 10.1.1/24               |
+      | A008696                 |
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-25694
+  @admin
+  @destructive
+  Scenario: Random outages with egressIP
+    Given I store the masters in the :nodes clipboard
+    And the valid egress IP is added to the "<%= cb.nodes[0].name %>" node
```
weliang1, line 289: Does this step is to add egressIP to master node? I assume the egressIP is needed to add to work nodes

features/networking/egress-ip.feature
weliang1: @huiran0826 after checking your steps again, you do not need line 331 to sleep 300 seconds.  We will hit bug #1793099 only when running two externalIP automation cases together, the first case do not need sleep 300s, but the second case will update externalIP policy to use new allow/deny CIDR, this updating process right now need at least 220 seconds to accept the new creating service.
For your testing here, you only has one externalIP configuration, I assume without sleeping 300 seconds, this case can still pass

features/networking/egress-ip.feature
```diff
@@ -232,3 +232,127 @@ Feature: Egress IP related features
...
+    Given I create a new project
+    And I have a pod-for-ping in the project
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+
+    # Curl externalIP:portnumber from project patched egressIP should pass
+    When I execute on the pod:
+      | /usr/bin/curl | --connect-timeout | 10 | <%= cb.hostip %>:27017 |
+    Then the output should contain:
+      | Hello-OpenShift-1 http-8080 |
```
weliang1, line 358: egressIP testing in this case is trick, even curl passed here, but without sniffing the packets, you do not know the IP source address is from the added egressIP or nodeIP itself

features/networking/egress-ip.feature
huiran0826: Thanks @weliang1,  any suggestion what time I should use?

features/networking/egress-ip.feature
huiran0826: Okay, will add it.

features/networking/egress-ip.feature
huiran0826: Yes, it is intended to add the egressIP to the master node. That insure the test pod can be on different node from the egressIP node.

features/networking/egress-ip.feature
huiran0826: hmm, this case is from customer's issue : https://bugzilla.redhat.com/show_bug.cgi?id=1762580. The issue is with this configuration, the curl failed(It should use the egressIP, without egressIP configuration that means use the nodeIP, it succeeded).
When it configured with egressIP, and if the traffic access the node IP or  even the outside website, source IP should be translated to egressIP.  In 3.11, we installed the ipecho image on the node then curl $node_ip:port to get the IP source address. 
In 4.x auto, we cannot install the this image directly on the node, so use curl ifconfig.me to replace. we have other cases to check this point.
The logic never change.  Thanks!

features/networking/egress-ip.feature
weliang1: @huiran0826 The deny/allow policy for externalIP is for cluster, but the svc configuration is for project. Agree, we do not know if other externalIP cases will run before this case.  We use 300 seconds to continue creating the svc until creation pass, if now other externalIP case before your case, your svc creation will success on first try. You need to keep this step.

features/networking/egress-ip.feature
weliang1: > Yes, it is intended to add the egressIP to the master node. That insure the test pod can be on different node from the egressIP node.

Should egressIP only be configured on worker node not master node? I do not think egressIP on the master is a real user case

features/networking/egress-ip.feature
weliang1: > If cluster has three nodes, and egressIP is added to the node-A, line 255 is to reboot the host/node, I assume you code make sure the host is just the node-A, just double check

@huiran0826 Could you double check this?

features/networking/egress-ip.feature
huiran0826: Thanks!

features/networking/egress-ip.feature
huiran0826: I have checked this, and it is sure use the selected node.

features/networking/egress-ip.feature
huiran0826: I can change it, but I don't think there is difference to configure it on master node or worker node. And generally vsphere resource is limited, from my personal testing, I generally deploy one worker node there.

features/networking/egress-ip.feature
huiran0826: Added the step for "restart network service" and local run passed. Thanks!

features/networking/egress-ip.feature
huiran0826: For real user case, using worker nodes make sense because the lots of will be redirected to the egressIP nodes. But for our testing purpose, it is just one simple curl command. I think it is okay to configure it on master node.

weliang1: For the testing results, you are right we may not see the difference, but master nodes and worker nodes take different roles in the cluster, same as the testing for egressrouter which should be deployed in the worker nodes not master nodes, even you can do this way.

I can image if this test case fail in the future, the dev will ask QE to reproduce the problem by changing to deploy the egreesIP in the worker nodes. In order to avoid wasting time in the future, I would suggest to update the step to use worker node now. WDYT?

huiran0826: @weliang1 Fixed it, thanks!

huiran0826:
@liangxia Can you help to merger? All the comments have been taken care.

PR #1313, opened 2020-07-02
OCP-26014 Pod readiness check for OVN, fix case id

rbbratta:


anuragthehatter:
Right on time. It was in my todo before couple days and been carried along. Thanks for taking care of it @rbbratta 

/LGTM

PR #1315, opened 2020-07-03
corner cases fix for metering steps

pruan-rht:
run log: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1718/console

@akostadinov @liangxia PTAL, it should fix some of the failures.  Thx!


pruan-rht:
@yapei I've updated it to follow the master version for the channel to select to use in OLM install.  PTAL

PR #1252, opened 2020-06-16
SDN-655 SystemOVS: update tests to use system OVS

rbbratta:
restarting the OVS pod should do nothing because
OVS is running as a systemd unit on the host

switch back to some of the <3.10 code that used system OVS already

PR #1319, opened 2020-07-06
exec_raw: remove extra bash -c

rbbratta:
While swtiching from running OVS in a Pod to running
OVS on the host we need to run similar commands on host or pod,
but unfortunately when running in a pod we use `bash -c` to
do inline shell.  When running a command with `bash -c` in `oc debug`
on the host the command is already wrapped with a `chroot /host` and a
`bash -c cd #{workdir}` and the command is line split inside the `bash -c`

For example in the same test we have the two different commands

      | iptables -S \| wc -l |
    When I run command on the node's sdn pod:
      | bash | -c | iptables -S \| wc -l |

Call trace:

`host.exec_admin` calls
```ruby
def exec_admin(*commands, **opts)
  exec_as(:admin, *commands, **opts)
```

 which checks if "root" and calls itself again with `nil` user

```ruby
if self[:user] == "root"
  return exec_as(nil, *commands, **opts)
```

then prepends `cd #{workdir}`

```
when nil, self[:user]
  # perform blind exec in workdir
  return exec_raw("cd '#{workdir}'", commands, **opts)
```

which prepends a chroot and a bash -c in `OCDebugAccessibleHost`

```ruby
unless opts[:single]
  commands = ["chroot", "/host/", "bash", "-c", commands_to_string(commands)]
```

Thus we end up with an invalid command because the inner bash -c is mangled.

```bash
   [01:02:47] INFO> Shell Commands: oc debug  node/compute.internal -- chroot /host/ bash -c cd\ \'/tmp/workdir/rbrattai-rbrattai\''
   'bash'
   '-c'
   'pgrep\ ovs-vswitchd
   bash: line 2: -c: command not found
   1272
```

Current proposed solution is to remove `bash -c` in host.exec_admin calls

### DIFF
--- a/lib/host.rb
+++ b/lib/host.rb
@@ -848,9 +848,19 @@ module BushSlicer
       node.env.service_project
     end
 
+    def remove_bash_c(*commands)
+      flat = commands.flatten
+      flat.reject.with_index { |c, i|
+        (flat[i] == "bash" and (flat.length >= i + 1) and flat[i + 1] == "-c") or
+            (flat[i] == "-c" and (i - 1 >= 0) and flat[i - 1] == "bash")
+      }
+    end
+
     # @note execute commands without special setup
     def exec_raw(*commands, **opts)
       unless opts[:single]
+        # remove prepended `bash -c` because we already have a 'bash -c'
+        commands = remove_bash_c(*commands)
         commands = ["chroot", "/host/", "bash", "-c", commands_to_string(commands)]
       end
 


rbbratta:
This wasn't the problem, although it is annoying to have the double `bash -c` it should work.

The issue was passing `table.raw` to `host.exec_admin` instead of the proper way using `table.raw.flatten.join(" ")`

akostadinov:
Will try to reproduce. Thanks.

rbbratta:
If we have host OVS we use `host.exec_admin(table.raw.flatten.join(" "))` otherwise we run the command in the pod with `pod.exec(table.raw)`

https://github.com/openshift/verification-tests/pull/1252/files#diff-b4003f8a21fcd4d7140fd9ca8d001ac7R861

```ruby
Given /^I run cmds on all ovs pods:$/ do | table |
  ensure_admin_tagged
  network_cmd = table.raw

  # If we have host ovs don't use the pods
  host_ovs = false
  env.nodes.each do |n|
    if BushSlicer::Platform::SystemdService.enabled?("openvswitch.service", n.host)
      logger.info("environment using systemd to launch openvswitch")
      host_ovs ||= true
      @result = n.host.exec_admin(network_cmd.flatten.join(" "))
```

I don't think we need this. so I'm inclined to close this for now.

akostadinov:
It is alright to have different semantics in `pod.exec` and `host.exec`. What I wanted to understand and ultimately didn't have time to try is whether there are commands that you cannot run with one of the approaches.

Your initial message sounds like some commands are not possible. And this is what I'd like to see fixed (if there is such thing).

Could you confirm whether you see a use case that fails with one of the methods?

akostadinov:
btw `host.exec_admin(table.raw.flatten.join(" "))` looks a little unsafe. Could you show me a scenario that presentlty does not work well and the way you want it to work?

rbbratta:
I haven't tried nested `bash -c ''` commands.  The nested single quotes should break things.

I tweaked the steps so that we don't have to use `bash -c`. We tend to default to using `bash -c`  for lots of `oc exec` commands.
The original step was 
```
      | bash | -c | pkill ovs-vswitchd |
```
https://github.com/openshift/verification-tests/pull/1252/files#diff-cc61b046e4707280e9620f66a294b2dfR221

The `bash -c` is redundant for the `pkill`.   But either way I now need to run the `pkill` table row either in the pod or on the host.  To run the `pkill` on the host I need to convert the table row to a single string.

I don't think `flatten` will work, instead I was going to try looping over the rows and `.join(" ")` each row, then passing `host_exec` an array of strings, because `commands_to_string` does a `flatten.join("\n")`

rbbratta:
@akostadinov  `:single` won't work because it doesn't add the `chroot` that we need.  Since we need the `chroot` and the `bash -c` we will need to collapse all the remaining args into a single string.  `commands_to_string` joins with newlines, so it won't work to flatten and join, we need a new method maybe to `.flatten.join(" ")`

```ruby
      unless opts[:single]
        commands = ["chroot", "/host/", "bash", "-c", commands_to_string(commands)]
      end
```
https://github.com/openshift/verification-tests/blob/master/lib/host.rb#L853

akostadinov:
@rbbratta , looking at `OCDebugAccessibleHost` it looks to me that we are somehow breaking conventions presently. I think that 
```
if opts[:raw]
  # leave unmodified commands
elsif opts[:single]
  commands = ["chroot", "/host/", commands]
else
  commands = ["chroot", "/host/", "bash", "-c", commands_to_string(commands)]
end
```

WDYT?

akostadinov:
The idea is that normally we run on host when using chroot. If in some rare occasion we need to run outside the chroot, then we have this option. But it wont be the norm if ever needed. Other than that convention is that `:single` and non-single should run exactly the same except that in one case command elements are taken, in the other case (multiple) shell commands are taken.

rbbratta:
@akostadinov I guess the real issue is using `host.exec_admin` because it always adds a `"cd '#{workdir}'"`

```ruby
    def exec_as(user, *commands, **opts)
      case user
      when nil, self[:user]
        # perform blind exec in workdir
        return exec_raw("cd '#{workdir}'", commands, **opts)
      when :admin
        # try to use sudo
        # in the future we may use `properties` for different methods
        # we may also allow choosing shell
        # TODO: are we admins?
        if self[:user] == "root"
          return exec_as(nil, *commands, **opts)
```
https://github.com/openshift/verification-tests/blob/master/lib/host.rb#L465

We always `cd` into the workdir, for some presumably good reason, and since we are cd-ing we have to use compound commands, and thus always `bash -c` with `chroot`.  I'm not sure the history of `exec_as`, `exec_admin`, vs. `exec_raw`.  

Maybe I should switch to `host.exec_raw`

Or we push the `cd '#{workdir}'` with a flag down to `OCDebugAccessibleHost.exec_raw` and adjust the command somehow.

akostadinov:
@rbbratta , did we discuss that this is unnecessary or we just forgot about this? Please remind me.

PR #1325, opened 2020-07-07
fix uninstall via Operatorhub error due to missing rule for 4.3, also…

pruan-rht:
… more corner case failures fix


pruan-rht:
@akostadinov @liangxia @yapei  PTAL.  Thx!

akostadinov:
I'm not much aware of these steps, maybe @yapei should review it.

yanpzhan:
/lgtm
Yapei is on SL. The update is ok from my side.

PR #1328, opened 2020-07-07
Revert "use 'open_uri_redirections' gem instead to handle redirect"

akostadinov:
Reverts openshift/verification-tests#1324

we shouldn't use deprecated APIs. From error message they recommend using `URI.open` instead of `Kernel.open`. The additional gem seems to patch this but this is not a reliable long term solution. Please try using `URI.open` as suggested by the original error message.


akostadinov:
@pruan-rht FYI, please try the above. We should be conservative adding new gem dependencies because this requires update of some systems and needs to be coordinated. Also the fewer dependencies the better.

PR #1329, opened 2020-07-08
Make OCP-12144 test more robust

jhou1:
The test flakes because it assumes after 25 seconds the pod is removed from the project, which is usually not true. @sunilcio @lyman9966 @weinliu PTAL

### DIFF
--- a/features/cli/oc_delete.feature
+++ b/features/cli/oc_delete.feature
@@ -116,24 +116,20 @@ Feature: oc_delete.feature
     Given I obtain test data file "pods/graceful-delete/10.json"
     When I run the :create client command with:
       | f | 10.json |
+    Then the step should succeed
     Given the pod named "grace10" becomes ready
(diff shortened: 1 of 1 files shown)


features/cli/oc_delete.feature
```diff
@@ -116,24 +116,25 @@ Feature: oc_delete.feature
...
-    And the output should contain "Terminating"
-    Given 25 seconds have passed
+    Then the output should contain "Terminating"
+    And I wait for the steps to pass:
+    """
     When I get project pods
-    Then the step should succeed
-    And the output should not contain "Terminating"
+    Then the output should not contain "Terminating"
     And the output should not contain "Running"
```
liangxia, line 136: How about use `And the pod named "xxx" does not exist` to replace above 3 lines ?

jhou1: Is there such step definition?

PR #1330, opened 2020-07-08
use RestClient.get which handles re-direct

pruan-rht:
@akostadinov PTAL, address the revert introducted in https://github.com/openshift/verification-tests/pull/1328

### DIFF
--- a/tools/find_image.rb
+++ b/tools/find_image.rb
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 
 require 'commander'
-require 'open-uri'
+require 'rest-client'
 require 'nokogiri'
 require 'thread'
 
@@ -33,14 +33,14 @@ class QueryPayload
           :url => "https://openshift-release.svc.ci.openshift.org/"
         raise "missing query string" if options.query.nil?
         print("Querying #{options.url} against #{options.build_type}\n")
-        doc = Nokogiri::HTML(open(options.url).read)
+        doc = Nokogiri::HTML(RestClient.get(options.url).to_s)
         links = doc.xpath("//a[contains(text(), '#{options.build_type}')]")
 
         target_links = links.map { |l|  options.url + l.attributes['href'].value }
         threads = []
         target_links.each do |link|
           threads << Thread.new(link) do |i|
-            doc = Nokogiri::HTML(open(link).read)
+            doc = Nokogiri::HTML(RestClient.get(link).to_s)
               if doc.to_s.include? options.query
                 puts "Pattern '#{options.query}' found in payload: #{link}"
               end


tools/find_image.rb
```diff
@@ -33,14 +33,14 @@ def run
           :url => "https://openshift-release.svc.ci.openshift.org/"
         raise "missing query string" if options.query.nil?
         print("Querying #{options.url} against #{options.build_type}\n")
-        doc = Nokogiri::HTML(open(options.url).read)
+        doc = Nokogiri::HTML(RestClient.get(options.url).to_s)
```
akostadinov, line 36: Did you try `URI.open` instead of `RestClient`.

pruan-rht: @akostadinov I don't think that suggested text is correct as URI does not have a public method that's named `open` https://ruby-doc.org/stdlib-2.5.1/libdoc/uri/rdoc/URI.html

akostadinov: `require 'open-uri'` makes this method available. Try it.

pruan-rht: @akostadinov that worked, but it's requires more work than simply using the lib `rest-client` which is part of the [Gemfile](https://github.com/openshift/verification-tests/blob/master/Gemfile#L5).  For example, the following is equivalent of `RestClient.get(options.url).to_s`

```addr = "https://openshift-release.svc.ci.openshift.org/"
uri = URI.parse(addr)
tries = 3
begin
  res = uri.open(redirect: false)
rescue OpenURI::HTTPRedirect => redirect
  binding.pry
  uri = redirect.uri # assigned from the "Location" response header
  retry if (tries -= 1) > 0
  raise
end
doc = open(res).read
```

akostadinov: Have you tried that redirection doesn't work with `URI.open(addr)`? For me redirections worked. Not sure I tried exactly your URL.

pruan-rht: @akostadinov, nope, that didn't work ```[1] pry(main)> require 'open-uri'
=> true
[2] pry(main)> URI.open('https://openshift-release.svc.ci.openshift.org/')
RuntimeError: redirection forbidden: https://openshift-release.svc.ci.openshift.org/ -> http://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/
from /Users/pruan/.rvm/rubies/ruby-2.6.6/lib/ruby/2.6.0/open-uri.rb:239:in `open_loop'```

akostadinov: This is actually a security issue on the registry side. HTTPS redirects to a HTTP URL. So user can't trust that what they get is legitimate. It strips off HTTPS protection. We need to report this to the URL administrator.

akostadinov: `URI.open` does support redirection between HTTPS URLs. As well from HTTP->HTTPS since Ruby 2.4. But it correctly in my opinion refuses to redirect back to HTTP.

pruan-rht: the final re-direct does point to https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/. so I'm wondering if that `URI.open` can ignore the intermediate `http` url.  Anyway, what's the objection of using `RestClient.get` to get around the issue while whatever team is responsible notified and the fix is implemented and merged?

pruan-rht:
closing this PR because one https://github.com/openshift/release/pull/10186/files is merged the original code should work.  In the meantime, the work around is to specify the final url by adding the option `--url=<final_url>` to the command line.

PR #1337, opened 2020-07-09
oc new-app  generate deployment(>4.4) and deploymentConfig,this is for Compatible with different version OC

barleyer:
oc new-app  generate deployment(>4.4) and deploymentConfig,this is for Compatible with different version OC

https://github.com/openshift/oc/pull/355

@xingxingxia @rhpmali  PTAL


xingxingxia:
/lgtm
@rhpmali , if having time, let's try to avoid the fix PRs' first&major review pending days somewhat long, there is chance without more prioritized work occupying time.

PR #1339, opened 2020-07-09
fixes to test result reporting and logging

akostadinov:
Test html log with the changes:

when after hook failed:
http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/07/09/23:31:33/Scale_Elasticsearch_nodes_by_nodeCount_2-_3-_4_in_clusterlogging/console.html

when everything succeeded:
http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/07/09/23:31:33/upgrade_OLM/console.html


pruan-rht:
@akostadinov looks good, in the failed log, there's a line ```Then the expression should be true> false``` that could benefit from some CSS TLC>

PR #1349, opened 2020-07-14
node: allow restore node from YAML at any time

rbbratta:
sometimes we need to restore the node in the middle
of the test, so modify the step to work immediately


rbbratta:
needed for https://github.com/openshift/cucushift/pull/7950

rbbratta:
This is bad, see https://github.com/openshift/cucushift/pull/7950#discussion_r454631378

closing

PR #1350, opened 2020-07-14
show systemctl status after restart

rbbratta:
Try to debug systemd restart failures by
running systemctl status after systemctl restart


lib/platform/systemd_service.rb
```diff
@@ -111,6 +111,8 @@ def restart(**opts)
         result = host.exec_admin("systemctl restart #{name}")
         results.push(result)
         unless result[:success]
+          # we should always check status if restart fails
+          host.exec_admin("systemctl status #{name}", quiet: false)
```
akostadinov, line 115: maybe add the `-l` option would be useful?

PR #1352, opened 2020-07-14
Add nodocs to yum install

liangxia:
/cc @pruan-rht @akostadinov @zhouying7780 

```
  Installing : autoconf-2.69-11.el7.noarch                                27/77 
install-info: No such file or directory for /usr/share/info/autoconf.info
```

### DIFF
--- a/tools/install_os_deps.sh
+++ b/tools/install_os_deps.sh
@@ -24,7 +24,7 @@ elif [ "$(os_type)" == "rhel6" ]; then
     additional_deps=install_rvm_if_ruby_is_outdated
     need_bundler_from_gem=1
 elif [ "$(os_type)" == "rhel7" ] || [ "$(os_type)" == "centos7" ]; then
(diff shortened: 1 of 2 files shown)


tools/install_os_deps.sh
```diff
@@ -24,7 +24,7 @@ elif [ "$(os_type)" == "rhel6" ]; then
     additional_deps=install_rvm_if_ruby_is_outdated
     need_bundler_from_gem=1
 elif [ "$(os_type)" == "rhel7" ] || [ "$(os_type)" == "centos7" ]; then
-    cmd="yum install -y --setopt=skip_missing_names_on_install=False"
+    cmd="yum install -y --setopt=skip_missing_names_on_install=False,tsflags=nodocs"
```
akostadinov, line 27: Does this flag apply to RHEL8 too?

liangxia: Tested on RHEL7, so only added it to RHEL7. I'll add it to RHEL8 once i got a RHEL8 to test it.

PR #1355, opened 2020-07-14
Initiating Windows Container automation

anuragthehatter:
Basic test to cover windows and linux pod connectivity

Logs: https://privatebin-it-iso.int.open.paas.redhat.com/?ac1d54fa09f51fe7#42BCNJXHOiPThet/pMM0gK0YoEgfnla5LUscDvDte4I=

### DIFF
--- a/features/networking/ovn_winc.feature
+++ b/features/networking/ovn_winc.feature
@@ -1,0 +1,56 @@
+Feature: OVNKubernetes Windows Container related networking scenarios
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-26360
+  @admin
+  Scenario: Ensure linux and window pod comunication on Hybrid cluster with window container
+    Given the env has hybridOverlayConfig enabled
+    And the env is using windows container
+    Given I have a project
+    And I obtain test data file "networking/list_for_pods.json"
+    When I run the :create client command with:
+      | f | list_for_pods.json |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | name=test-pods |
+    And evaluation of `pod` is stored in the :linux_pod clipboard
+    
+    Given I use the "test-service" service
+    And evaluation of `service.ip(user: user)` is stored in the :linux_service_ip clipboard
+    # Run level needs to set to 1 for windows pod to be spawned in a test project
+    When I run the :label admin command with:
+      | resource | namespace                |
+      | name     | <%= project.name %>      |
+      | key_val  | openshift.io/run-level=1 |
+    Then the step should succeed
+    Given I obtain test data file "networking/windows_pod_and_service.yaml"
+    When I run the :create client command with:
+      | f | windows_pod_and_service.yaml |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | app=win-webserver |
+    And evaluation of `pod` is stored in the :windows_pod clipboard
+    
+    Given I use the "win-service" service
+    And evaluation of `service.ip(user: user)` is stored in the :windows_service_ip clipboard
+    #Checking Service communication across pods
+    When I execute on the "<%= cb.linux_pod.name %>" pod:
+      | curl | -s | --connect-timeout | 2 | <%= cb.windows_service_ip %>:27018 |
+    Then the step should succeed
+    And the output should contain "Windows Container Web Server" 
+    
+    When I execute on the "<%= cb.windows_pod.name %>" pod:
+      | curl | -s | --connect-timeout | 2 | <%= cb.linux_service_ip %>:27017 |
+    Then the step should succeed
+    And the output should contain "Hello OpenShift" 
+    
+    #Checking network communication across pods
+    When I execute on the "<%= cb.linux_pod.name %>" pod:
+      | curl | -s | --connect-timeout | 2 | <%= cb.windows_pod.ip %>:80 |
+    Then the step should succeed
+    And the output should contain "Windows Container Web Server" 
+    
+    When I execute on the "<%= cb.windows_pod.name %>" pod:
+      | curl | -s | --connect-timeout | 2 | <%= cb.linux_pod.ip %>:8080 |
+    Then the step should succeed
+    And the output should contain "Hello OpenShift" 
--- a/features/networking/ovn_winc.feature
+++ b/features/networking/ovn_winc.feature
@@ -3,7 +3,7 @@ Feature: OVNKubernetes Windows Container related networking scenarios
   # @author anusaxen@redhat.com
   # @case_id OCP-26360
   @admin
-  Scenario: Ensure linux and window pod comunication on Hybrid cluster with window container
+  Scenario: Ensure Pods and Service communication across window and linux nodes
     Given the env has hybridOverlayConfig enabled
     And the env is using windows container
     Given I have a project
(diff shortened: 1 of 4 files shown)


akostadinov:
lgtm, let team merge after review.

features/networking/ovn_winc.feature
```diff
@@ -0,0 +1,56 @@
...
+      | name=test-pods |
+    And evaluation of `pod` is stored in the :linux_pod clipboard
+    
+    Given I use the "test-service" service
+    And evaluation of `service.ip(user: user)` is stored in the :linux_service_ip clipboard
+    # Run level needs to set to 1 for windows pod to be spawned in a test project
+    When I run the :label admin command with:
+      | resource | namespace                |
+      | name     | <%= project.name %>      |
+      | key_val  | openshift.io/run-level=1 |
```
zhaozhanqi, line 24: `openshift.io/run-level=1` is must for project? not sure this is by design or bug. see the similar bug https://bugzilla.redhat.com/show_bug.cgi?id=1830497

features/networking/ovn_winc.feature
```diff
@@ -0,0 +1,56 @@
...
+    Given I obtain test data file "networking/windows_pod_and_service.yaml"
+    When I run the :create client command with:
+      | f | windows_pod_and_service.yaml |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | app=win-webserver |
+    And evaluation of `pod` is stored in the :windows_pod clipboard
+    
+    Given I use the "win-service" service
+    And evaluation of `service.ip(user: user)` is stored in the :windows_service_ip clipboard
```
rbbratta, line 35: could we use `service.url` which should include the port and also quote the address for IPv6

features/step_definitions/networking.rb
```diff
@@ -797,6 +797,21 @@
   raise "the networkType is not #{network_type}" unless @result[:response] == network_type
 end
 
+Given /^the env is using windows container$/ do
+  ensure_admin_tagged
+  _admin = admin
+  @result = _admin.cli_exec(:get, resource: "nodes",show_label:true)
```
rbbratta, line 803: I think there should be a space after commas  `"nodes", show_label:true)`

features/networking/ovn_winc.feature
rbbratta: hadn't thought about IPv6 dual-stack in the context of Windows Containers, is it supported?

features/networking/ovn_winc.feature
```diff
@@ -0,0 +1,56 @@
...
+    And the output should contain "Hello OpenShift" 
+    
+    #Checking network communication across pods
+    When I execute on the "<%= cb.linux_pod.name %>" pod:
+      | curl | -s | --connect-timeout | 2 | <%= cb.windows_pod.ip %>:80 |
+    Then the step should succeed
+    And the output should contain "Windows Container Web Server" 
+    
+    When I execute on the "<%= cb.windows_pod.name %>" pod:
+      | curl | -s | --connect-timeout | 2 | <%= cb.linux_pod.ip %>:8080 |
```
rbbratta, line 54: similarly `pod.ip_url` will quote for IPv6

features/step_definitions/networking.rb
rbbratta: er, well we could use the `env.nodes` object

```ruby
  is_windows = env.nodes.any? { |n| n.labels["kubernetes.io/os"] == "windows"}
  raise "env doesn't have any windows nodes" unless is_windows
```

features/step_definitions/networking.rb
```diff
@@ -797,6 +797,21 @@
...
+  raise "env doesn't have any windows container" unless @result[:response].include? "kubernetes.io/os=windows"
```
rbbratta, line 804: s/windows container/windows nodes/

features/networking/ovn_winc.feature
```diff
@@ -0,0 +1,56 @@
...
+    And I obtain test data file "networking/list_for_pods.json"
+    When I run the :create client command with:
+      | f | list_for_pods.json |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | name=test-pods |
+    And evaluation of `pod` is stored in the :linux_pod clipboard
+    
+    Given I use the "test-service" service
+    And evaluation of `service.ip(user: user)` is stored in the :linux_service_ip clipboard
```
liangxia, line 19: No need the `(user: user)` part.

features/networking/ovn_winc.feature
```diff
@@ -0,0 +1,56 @@
...
+      | name     | <%= project.name %>      |
+      | key_val  | openshift.io/run-level=1 |
+    Then the step should succeed
+    Given I obtain test data file "networking/windows_pod_and_service.yaml"
+    When I run the :create client command with:
+      | f | windows_pod_and_service.yaml |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | app=win-webserver |
+    And evaluation of `pod` is stored in the :windows_pod clipboard
```
liangxia, line 32: We can save `pod.name` to the clipborad, so we can use `cb.windows_pod` in below steps.

features/networking/ovn_winc.feature
anuragthehatter: Yea, its a bug and will be addressed by a web hook. Either we can make this change later or keep this PR open. WDYT. Either way this script will help me check this functionality in regular basis

features/networking/ovn_winc.feature
anuragthehatter: Windows nodes have no plan to support IPv6 for now at least till 4.8 or up. I would keep this as is as we might need different ways to handle ipv6 later specifically for windows node which is still not clear to us.

features/networking/ovn_winc.feature
anuragthehatter: Same reason

features/step_definitions/networking.rb
anuragthehatter: @rbbratta yea but we intentionally want this because it will dump nodes info with kube version. Most of the time i noticed windows container are one version behind wrt to linux. So it will be helpful in debugging and bugs reporting. I believe env.nodes won't print the nodes ```oc get nodes --show-labels``` info.

anuragthehatter: It will actually print all nodes yaml with each node printing 500 lines so I guess we still favor short and sweet output with only 6-8 lines instead of 3000+ lines.

features/networking/ovn_winc.feature
rbbratta: we should be consistent for Linux curl usages at least.

anuragthehatter: Ross, Can you elaborate? Are you concerned about time outs?

features/networking/ovn_winc.feature
rbbratta: I guess it is more a question of `curl` syntax on Windows, which should be governed by RFC, but as it should be the same `curl` code base I would expect `curl` to act the same on Windows and Linux.

Looks like it is defined by RFC
https://docs.microsoft.com/en-us/windows/win32/wininet/ip-version-6-support

Just tested Windows 10 with official `curl-7.71.1-win64-mingw` and  `curl.exe http://[fe80::1:1%18]/` works

anuragthehatter: Ah I understand your confusion now. Yea so both linux and win pod honor same syntax. Would have been nicely represented if ipv6 was involved. Its just Ipv4 here with the pod.ip or service.ip clipbaord, not the host url so guess it should be fine and by default i believe,curl <ip_address> adds http as prefix if none defined.

features/networking/ovn_winc.feature
zhaozhanqi: I think we can make this change later after the bug is fixed since it's not a big issue.

PR #1358, opened 2020-07-16
Fix undefined steps which is caused by typo

liangxia:
/cc @pruan-rht @anuragthehatter


anuragthehatter:
Have this change in #1355 as well. Anyways we can merge this one first to avoid failures. Thanks

PR #1360, opened 2020-07-16
Auotmate OCP-28757

weliang1:
Auotmate OCP-28757: Establish pod to pod SCTP connections

Test log: https://privatebin-it-iso.int.open.paas.redhat.com/?2689abc01f193f67#CN81T3gdCEElTAE7hOkvPYv0WBGCodhxsifOQWlOb50=

Please help to review the first SCTP test case, I will automate the rest cases after get your good comments,  thanks!

### DIFF
--- a/features/networking/sctp.feature
+++ b/features/networking/sctp.feature
@@ -1,0 +1,55 @@
+Feature: SCTP related scenarios
+
+  # @author weliang@redhat.com
+  # @case_id OCP-28757
+  @admin
+  @destructive
+  Scenario: Establish pod to pod SCTP connections
+    Given I store the workers in the :workers clipboard
+    And I store the number of worker nodes to the :num_workers clipboard
+    And the Internal IP of node "<%= cb.workers[1].name %>" is stored in the :worker1_ip clipboard
+    
+    Given I install machineconfigs load-sctp-module
+    And I wait up to 800 seconds for the steps to pass:
+    """
+    Given I check load-sctp-module in <%= cb.num_workers %> workers
+    """
+    
+    Given I have a project
+    Given I obtain test data file "networking/sctp/sctpserver.yaml"
+    When I run oc create as admin over "sctpserver.yaml" replacing paths:
+      | ["metadata"]["namespace"] | <%= project.name %>       |
+      | ["spec"]["nodeName"]      | <%= cb.workers[0].name %> |
+    Then the step should succeed
+    And the pod named "sctpserver" becomes ready
+    Then evaluation of `pod.ip` is stored in the :serverpod_ip clipboard
+
+    Given I obtain test data file "networking/sctp/sctpclient.yaml"
+    When I run oc create as admin over "sctpclient.yaml" replacing paths:
+      | ["metadata"]["namespace"] | <%= project.name %>       |
+      | ["spec"]["nodeName"]      | <%= cb.workers[1].name %> |
+    Then the step should succeed
+    And the pod named "sctpclient" becomes ready
+   
+    # sctpserver pod start to wait for sctp traffic
+    When I run the :exec background admin command with:
+      | pod              | sctpserver          |
+      | namespace        | <%= project.name %> |
+      | oc_opts_end      |                     |
+      | exec_command     | sh                  |
+      | exec_command_arg | -c                  |
+      | exec_command_arg | nc -l 30102 --sctp  |     
+    Then the step should succeed
+
+    # sctpclient pod start to send sctp traffic
+    When I run the :exec background admin command with:
+      | pod              | sctpclient                                                       |
+      | namespace        | <%= project.name %>                                              |
+      | oc_opts_end      |                                                                  |
+      | exec_command     | sh                                                               |
+      | exec_command_arg | -c                                                               |
+      | exec_command_arg | echo test-openshift \| nc -v <%= cb.serverpod_ip %> 30102 --sctp | 
+    Then the step should succeed
+    And the output should contain:
+      | Connected to <%= cb.serverpod_ip %>:30102 |
+      | 15 bytes sent                             |
--- a/features/networking/sctp.feature
+++ b/features/networking/sctp.feature
@@ -32,7 +32,7 @@ Feature: SCTP related scenarios
     And the pod named "sctpclient" becomes ready
    
     # sctpserver pod start to wait for sctp traffic
-    When I run the :exec background admin command with:
+    When I run the :exec background client command with:
       | pod              | sctpserver          |
       | namespace        | <%= project.name %> |
       | oc_opts_end      |                     |
@@ -42,7 +42,7 @@ Feature: SCTP related scenarios
     Then the step should succeed
 
     # sctpclient pod start to send sctp traffic
-    When I run the :exec background admin command with:
+    When I run the :exec client command with:
       | pod              | sctpclient                                                       |
       | namespace        | <%= project.name %>                                              |
       | oc_opts_end      |                                                                  |
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1092,3 +1092,30 @@ Given /^ptp config daemon is ready$/ do
     | app=linuxptp-daemon |
   })
 end
+
+Given /^I install machineconfigs load-sctp-module$/ do
+  ensure_admin_tagged
+  _admin = admin
+  @result = _admin.cli_exec(:get, resource: "machineconfigs", output: 'jsonpath={.items[?(@.metadata.name=="load-sctp-module")].metadata.name}')
+  if @result[:response] != "load-sctp-module"
+    @result = _admin.cli_exec(:create, f: "#{BushSlicer::HOME}/testdata/networking/sctp/load-sctp-module.yaml")
+    raise "Failed to install load-sctp-module" unless @result[:success]
+  end
+end
+
+Given /^I check load-sctp-module in #{NUMBER} workers$/ do | workers_num |
+  ensure_admin_tagged
+  _admin = admin
+  $i = 0
+  $num = Integer(workers_num)
+  while $i < $num  do
+    step %Q{I run the :debug admin command with:}, table(%{
+      | resource     | node/<%= cb.workers[#$i].name %> |
+      | oc_opts_end  |                                  |
+      | exec_command | lsmod                            |
+    })
+    step %Q/the step should succeed/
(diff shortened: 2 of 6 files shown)


anuragthehatter:
@weliang1 This is a destructive case

weliang1:
> @weliang1 This is a destructive case

@anuragthehatter Just forgot that, fixed it now,  thanks!

features/networking/sctp.feature
```diff
@@ -0,0 +1,59 @@
...
+  @destructive
+  Scenario: Establish pod to pod SCTP connections		
+    Given I store the workers in the :workers clipboard
+    And the Internal IP of node "<%= cb.workers[1].name %>" is stored in the :worker1_ip clipboard
+    
+    Given I install load-sctp-module
+    # Installing sctp module will reboot all the nodes and it will more than 10 minutes
+    And I wait up to 800 seconds for the steps to pass:
+    """
+    Given 60 seconds have passed
```
zhaozhanqi, line 15: I think better add the following steps into `Given /^I install load-sctp-module/ do` . otherwise if you run the second `sctp` case in future.  it will always wait 60 seconds even if the the server already load sctp module. 
```
   Given 60 seconds have passed 
    When I run the :get admin command with:
      |resource | nodes |
    Then the outputs should not contain "SchedulingDisabled"
```

anuragthehatter: It would be also better if we add a condition check in ```Given I install load-sctp-module``` like if module is already installed on the cluster, bypass this step. @weliang1 Do you think for every sctp test, loading sctp module from scratch is necessary?

features/networking/sctp.feature
```diff
@@ -0,0 +1,59 @@
...
+      | exec_command_arg | nc -l 30102 --sctp  |     
+    Then the step should succeed
+    Given 10 seconds have passed
+
+    # sctp client pod start to send traffic 
+    When I run the :exec background admin command with:
+      | pod              | sctpclient                                                         |
+      | namespace        | <%= project.name %>                                                |
+      | oc_opts_end      |                                                                    |
+      | exec_command     | sh                                                                 |
```
anuragthehatter, line 53: Please revisit indentation.

features/networking/sctp.feature
```diff
@@ -0,0 +1,59 @@
...
+    Given I store the workers in the :workers clipboard
+    And the Internal IP of node "<%= cb.workers[1].name %>" is stored in the :worker1_ip clipboard
+    
+    Given I install load-sctp-module
+    # Installing sctp module will reboot all the nodes and it will more than 10 minutes
+    And I wait up to 800 seconds for the steps to pass:
+    """
+    Given 60 seconds have passed 
+    When I run the :get admin command with:
+      |resource | nodes |
```
anuragthehatter, line 17: Missing space before ```resource```

features/networking/sctp.feature
```diff
@@ -0,0 +1,59 @@
+Feature: SCTP related scenarios
+
+  # @author weliang@redhat.com
+  # @case_id OCP-28757
+  @admin
+  @destructive
+  Scenario: Establish pod to pod SCTP connections		
+    Given I store the workers in the :workers clipboard
+    And the Internal IP of node "<%= cb.workers[1].name %>" is stored in the :worker1_ip clipboard
```
anuragthehatter, line 9: This step seems redundant

features/networking/sctp.feature
```diff
@@ -0,0 +1,59 @@
...
+    # sctp server pod start to wait for sctp traffic
+    When I run the :exec background admin command with:
+      | pod              | sctpserver          |
+      | namespace        | <%= project.name %> |
+      | oc_opts_end      |                     |
+      | exec_command     | sh                  |
+      | exec_command_arg | -c                  |
+      | exec_command_arg | nc -l 30102 --sctp  |     
+    Then the step should succeed
+    Given 10 seconds have passed
```
anuragthehatter, line 46: Weibin, I don't think so its needed?

features/networking/sctp.feature
```diff
@@ -0,0 +1,59 @@
...
+
+    # sctp client pod start to send traffic 
+    When I run the :exec background admin command with:
```
anuragthehatter, line 49: Hmm, background for server might make sense but its not required for client. Anyways this is okay.

features/step_definitions/networking.rb
```diff
@@ -1008,3 +1008,13 @@
   end 
   logger.info "The multicast is enable in the #{project_name} project"
 end
+
+Given /^I install load-sctp-module/ do
+  ensure_admin_tagged
+  _admin = admin
+  @result = _admin.cli_exec(:get, resource: "machineconfigs", output: "jsonpath={.items[10].metadata.name}")
```
rbbratta, line 1015: `items[10]` won't work forever.  We should be able to match the array element based on the `metadata.name`

We have to use single quotes, or escape the double quotes
```go
'jsonpath={.items[?(@.metadata.name=="load-sctp-module")].metadata.name}'
```
should work

features/networking/sctp.feature
weliang1: > I think better add the following steps into `Given /^I install load-sctp-module/ do` . otherwise if you run the second `sctp` case in future. it will always wait 60 seconds even if the the server already load sctp module.
> 
> ```
>    Given 60 seconds have passed 
>     When I run the :get admin command with:
>       |resource | nodes |
>     Then the outputs should not contain "SchedulingDisabled"
> ```

@zhaozhanqi Fix to have more accurate step to load and check sctp mode, 

    Given I install machineconfigs load-sctp-module   
    And I wait up to 800 seconds for the steps to pass:
    """
    Given I check load-sctp-module in <%= cb.num_workers %> workers
    """

features/step_definitions/networking.rb
anuragthehatter: Static indices are discouraged, yep

features/networking/sctp.feature
weliang1: > It would be also better if we add a condition check in `Given I install load-sctp-module` like if module is already installed on the cluster, bypass this step. @weliang1 Do you think for every sctp test, loading sctp module from scratch is necessary?

@anuragthehatter Yes, we have that step in Given I install machineconfigs load-sctp-module. Any time the first SCTP test case need load sctp module, the rest case do not need after checking if the node has the sctp module already.

features/networking/sctp.feature
weliang1: This line is deleted in the new version.

features/networking/sctp.feature
weliang1: This line is deleted in the new version.

features/step_definitions/networking.rb
weliang1: @rbbratta Fixed it, thanks for good tip!

features/networking/sctp.feature
weliang1: @anuragthehatter New version has three new steps about node's information, do you still think they are redundant? 
 Given I store the workers in the :workers clipboard
    And I store the number of worker nodes to the :num_workers clipboard
    And the Internal IP of node "<%= cb.workers[1].name %>" is stored in the :worker1_ip clipboard

features/networking/sctp.feature
weliang1: @anuragthehatter Could you please check if you still see the indentation between line 46 and 51 in new commit? Thanks!

weliang1:
New test log with need to load sctp module:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1974/console

New test log without need to load sctp module:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1975/console

features/networking/sctp.feature
anuragthehatter: Looks good, Thanks Weibin!

features/networking/sctp.feature
anuragthehatter: Ah ok. Is it expected that we should check sctp module load on each worker node even if machineconfigs says sctp module exist? Also as i understand its a module and will be synced across all other nodes so checking on one node might be sufficient? hence reduce runtime.

anuragthehatter: Would you like to have ```Given I check load-sctp-module``` code under ```Given I install machineconfigs``` step. So it could be a just one step which will install and may not install if already installed. WIll leave that upto you.

anuragthehatter: Also line 8 and 9 should come under ```Given /^I check load-sctp-module``` step else this step always needs to accompany those two steps so better keep'em under it.WDYT

weliang1: > Ah ok. Is it expected that we should check sctp module load on each worker node even if machineconfigs says sctp module exist? Also as i understand its a module and will be synced across all other nodes so checking on one node might be sufficient? hence reduce runtime.

The whole process is like this: 1. run load sctp module, execution command finish 2. within 60 seconds, all the nodes start to reboot one by one. 3. After each node's rebooting, checking sctp module in each node is to make sure that the testing failure is not because the sctp module is not loaded in the node.
So it is expected that we should check sctp module load on each worker node.

anuragthehatter: > > Ah ok. Is it expected that we should check sctp module load on each worker node even if machineconfigs says sctp module exist? Also as i understand its a module and will be synced across all other nodes so checking on one node might be sufficient? hence reduce runtime.
> 
> The whole process is like this: 1. run load sctp module, execution command finish 2. within 60 seconds, all the nodes start to reboot one by one. 3. After each node's rebooting, checking sctp module in each node is to make sure that the testing failure is not because the sctp module is not loaded in the node.
> So it is expected that we should check sctp module load on each worker node.

Thanks for clarifications, Weibin. Yea good to keep that since nodes are being restarted

weliang1: Actually I missed those steps in the first commit, thanks @pruan-rht for this accurate way.

features/networking/sctp.feature
weliang1: @anuragthehatter Update to not run nc in the background and run it as client not admin

weliang1:
Any more comments for this PR?

features/networking/sctp.feature
anuragthehatter: Ah was not needed. This was trivial. Anyways, thanks

weliang1:
@pruan-rht @liangxia

PR #1363, opened 2020-07-17
fix ocp-13022

lyman9966:
I extend the waiting time of machineconfigpool check from 20mins to 33 mins. Because one master needs about 9-10mins to finish kubeletconfig sync and restart according to UpgradeCI log(Though it only cost about 5 mins per master in manual test), so I leave 11 mins per master and it's should be secure.

### DIFF
--- a/features/upgrade/node/node-upgrade.feature
+++ b/features/upgrade/node/node-upgrade.feature
@@ -20,7 +20,7 @@ Feature: Node components upgrade tests
     Then the output should match:
       | master.*False\\s+True\\s+False |
     """
-    And I wait up to 1200 seconds for the steps to pass:
+    And I wait up to 1980 seconds for the steps to pass:
     """
     When I run the :get client command with:
       | resource | machineconfigpool |
--- a/features/upgrade/node/node-upgrade.feature
+++ b/features/upgrade/node/node-upgrade.feature
@@ -16,16 +16,18 @@ Feature: Node components upgrade tests
     And I wait up to 30 seconds for the steps to pass:
     """
     When I run the :get client command with:
-      | resource | machineconfigpool |
+      | resource      | machineconfigpool |
+      | resource_name | master            |
     Then the output should match:
-      | master.*False\\s+True\\s+False |
+      | .*False\\s+True\\s+False |
     """
     And I wait up to 1980 seconds for the steps to pass:
     """
     When I run the :get client command with:
-      | resource | machineconfigpool |
+      | resource      | machineconfigpool |
+      | resource_name | master            |
     Then the output should match:
-      | master.*True\\s+False\\s+False |
+      | .*True\\s+False\\s+False |
     """
     Given I store the masters in the :masters clipboard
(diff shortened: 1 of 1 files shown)


lyman9966:
@jhou1 @sunilcio @weinliu Can you help to have a  look ? Thanks

jhou1:
I suggest to make the following updates as well

The steps:
```
      | resource | machineconfigpool |
    Then the output should match:
      | master.*False\\s+True\\s+False |
```
can be updated with
```
      | resource | machineconfigpool |
      | resource_name | master |
    Then the output should match:
      | .*False\\s+True\\s+False |
```

There are several occurrence similar to this block. By adding resource_name, we compress the output of the runner, and reading logs will be easier.

jhou1:
What do you think if we only update the machineconfigpool and don't verify the output in prepare? Leave the sync to the machineconfig daemon and we only verify it in upgrade-check. This can reduce the execution time. @lyman9966 @sunilcio @weinliu @liangxia

sunilcio:
@lyman9966 @jhou1 one issue I see if we do not verify during prepare is, if the upgraded is started before machineconfigpool is rolled out successfully, that would cause issues during upgrade. Or if any other case from other component is dependent on machineconfigpool or node status, they may report error.

lyman9966:
@jhou1 , I agree with @sunilcio , we will bring more uncertainties and risks to upgrade process if we are not sure the machineconfigpool is rolled out successfully. 
But if the machineconfigpool status can be checked before starting upgrade,  I think it is acceptable. Then someone need to modify upgrade script and we need to discuss with team if anyone's prepare script is dependent on machineconfigpool status and node status.

lyman9966:
> I suggest to make the following updates as well
> 
> The steps:
> 
> ```
> When I run the :get client command with:
>       | resource | machineconfigpool |
>     Then the output should match:
>       | master.*False\\s+True\\s+False |
> ```
> 
> can be updated with
> 
> ```
> When I run the :get client command with:
>       | resource | machineconfigpool |
>       | resource_name | master |
>     Then the output should match:
>       | .*False\\s+True\\s+False |
> ```
> 
> There are several occurrence similar to this block. By adding resource_name, we compress the output of the runner, and reading logs will be easier.

@jhou1 Yeah, it's a good advice , thanks

features/upgrade/node/node-upgrade.feature
```diff
@@ -16,16 +16,18 @@ Feature: Node components upgrade tests
     And I wait up to 30 seconds for the steps to pass:
     """
     When I run the :get client command with:
-      | resource | machineconfigpool |
+      | resource      | machineconfigpool |
+      | resource_name | master            |
     Then the output should match:
-      | master.*False\\s+True\\s+False |
+      | .*False\\s+True\\s+False |
```
akostadinov, line 22: I even don't know what this check should mean by reading the scenario. Also human readable strings and order of elements may change. Better write a special step to check what is needed here. 

I will not block merging as this is same as before. But consider updating it when possible.

PR #1364, opened 2020-07-17
update for deployer pod collect logs

zhouying7780:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/118013/console

@kasturinarra Please help review , thank.

### DIFF
--- a/features/step_definitions/pod.rb
+++ b/features/step_definitions/pod.rb
@@ -249,6 +249,18 @@ Given /^I collect the deployment log for pod "(.+)" until it disappears$/ do |po
   @result  = res_cache
 end
 
+Given /^I collect the deployment log for pod "(.+)" until it becomes :([^\s]*?)$/ do |name, status|
+  opts = {resource_name: name}
+  timeout = 15 * 60   # just put a timeout so we don't hang there indefintely
+  podstatus = pod(name).wait_till_status(status.to_sym, user, timeout)
+  unless podstatus[:success]
+    logger.error(podstatus[:response])
+    raise "pod #{name} didn't become #{status}"
+  end
+
+  @result = user.cli_exec(:logs, **opts)
+end
+
 # pod_info is the user pod, for example.... deployment-example
 # the step will do 'docker ps | grep deployment-example' to filter out a target
 # TODO: cri-o is not implemented yet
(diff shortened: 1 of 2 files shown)


features/step_definitions/pod.rb
```diff
@@ -249,6 +249,18 @@
   @result  = res_cache
 end
 
+Given /^I collect the deployment log for pod "(.+)" until it becomes :([^\s]*?)$/ do |name, status|
+  opts = {resource_name: name}
+  timeout = timeout ? Integer(timeout) : 15 * 60   # just put a timeout so we don't hang there indefintely
+  podstatus = pod(name).wait_till_status(status.to_sym, user, timeout)
+  unless podstatus[:success]
+    logger.error(podstatu[:response])
+    raise "#{pod.name} pod was never ready"
```
kasturinarra, line 258: Should we say pod was never ready or pod never succeeded as in the above test script i see that we are already checking for pod ready status before collecting the logs, thanks !!

zhouying7780: @kasturinarra  "succeeded" just a parametes, we could also input the "running"  or "failed" status. then also confused for debug .

kasturinarra: okay !!

features/step_definitions/pod.rb
```diff
@@ -249,6 +249,18 @@
   @result  = res_cache
 end
 
+Given /^I collect the deployment log for pod "(.+)" until it becomes :([^\s]*?)$/ do |name, status|
+  opts = {resource_name: name}
+  timeout = timeout ? Integer(timeout) : 15 * 60   # just put a timeout so we don't hang there indefintely
```
liangxia, line 254: Perhaps `timeout = 15 * 60` as we do not define timeout in the step name.

kasturinarra: @zhouying7780 could you please check and update this, so that lxia could merge the patch as it is causing one of the test to fail ? Thanks !!

zhouying7780:
@liangxia  Please help review again, have resolved the comment . thanks.

features/step_definitions/pod.rb
liangxia: raise "pod #{name} didn't become #{status}"

kasturinarra:
@liangxia could you please help review and provide if any comments ? one of the test is failing due to this not being merged, thanks !!

features/step_definitions/pod.rb
liangxia: Any update for this ?  By the way, found an typo here `podstatu`

zhouying7780:
@liangxia  Please help review , have resolved all the comments.

PR #1365, opened 2020-07-17
Some manual cases automation

anuragthehatter:
Added OCP-27333 and will keep adding more next week.

Logs: http://pastebin.test.redhat.com/885594


anuragthehatter:
OCP-24694: https://privatebin-it-iso.int.open.paas.redhat.com/?7e8cf2924decd424#9ABiuo36OGWmrnz9U73zPO4vrjyYRhYqrL9LPhaV0O0=

features/step_definitions/networking.rb
```diff
@@ -790,6 +790,25 @@
   logger.info "Subnet range for the primary interface on the node is stored in the #{cb_name} clipboard."
 end
 
+Given /^the node's MTU value is stored in the#{OPT_SYM} clipboard$/ do |cb_node_mtu|
+  ensure_admin_tagged
+  cb_node_mtu = "mtu" unless cb_node_mtu
+  @result = admin.cli_exec(:get, resource: "network.operator", output: "jsonpath={.items[*].spec.defaultNetwork.type}")
```
zhaozhanqi, line 796: how about using `oc debug node/xxx chroot /host ip route show default` to get the mtu, then we do not need to check the networkType.

features/networking/service.feature
```diff
@@ -431,3 +431,48 @@ Feature: Service related networking scenarios
...
+    And I run commands on the host:
+      | systemctl start NetworkManager |
+    Then the step should succeed
+    And the node's MTU value is stored in the :mtu_actual_redeployed clipboard
+    And the expression should be true> cb.mtu_actual == cb.mtu_actual_redeployed
+    #The node should get un-taint post this step in few seconds
+    And I run commands on the host:
+      | pkill openshift-sdn |
+    Then the step should succeed
+    #Check if node has removed the taint
```
zhaozhanqi, line 473: Do we need to wait some second to check the node when `pkill openshift-sdn`?

anuragthehatter: Actually that process just get re-spawned immediately as per my testing but your comment gave me an idea to add time loop on next ```oc describe``` step in case..heh

features/step_definitions/networking.rb
anuragthehatter: Ya this was discussed earlier in some cases. Good approach is to get required pmtrs from network pods instead of node as oc debug node tends to throw intermittent stderr issues may be due to auto hiccups causing premature failures. WDYT

features/networking/operator.feature
```diff
@@ -345,3 +345,25 @@ Feature: Operator related networking scenarios
     And all existing pods die with labels:
       | app=dhcp-daemon |
     """
+
+  # @author anusaxen@redhat.com
+  # @case_id OCP-27333
+  @admin
+  @destructive
+  Scenario: Changing mtu in CNO should not be allowed
+  Given the mtu value "1750" is patched in CNO config according to the networkType
```
liangxia, line 354: We use one more level indention for the steps.

liangxia: Do we need to restore the change after scenario ?

features/networking/operator.feature
```diff
@@ -345,3 +345,25 @@ Feature: Operator related networking scenarios
...
+  And admin uses the "openshift-network-operator" project
+  When I run the :get admin command with:
+    | resource | pods                               |
+    | o        | jsonpath={.items[*].metadata.name} |
+  Then the step should succeed
+  And evaluation of `@result[:response]` is stored in the :network_operator_pod clipboard
```
liangxia, line 360: We can use below steps, should be more reliable in case there are multiple pod in the project.
```
And a pod becomes ready with labels:
  | foo=bar |
```
Then call `<%= pod.name %>` in the logs command.

features/networking/operator.feature
anuragthehatter: Restore is being taken care in same step via ```teardown_add``` procedure

features/networking/operator.feature
anuragthehatter: yea this is actually a system project and ```a pod becomes ready with labels``` would be called by a testuser so might not work here.

features/networking/operator.feature
```diff
@@ -345,3 +345,25 @@ Feature: Operator related networking scenarios
...
+  @admin
+  @destructive
+  Scenario: Changing mtu in CNO should not be allowed
+    Given the mtu value "1750" is patched in CNO config according to the networkType
+    And admin uses the "openshift-network-operator" project
+    When I run the :get admin command with:
+      | resource | pods                               |
+      | o        | jsonpath={.items[*].metadata.name} |
+    Then the step should succeed
+    And evaluation of `@result[:response]` is stored in the :network_operator_pod clipboard
```
liangxia, line 360: What will happen if there are multiple pods in the openshift-network-operator project ?
How about use `Given a pod becomes ready with labels:` ?

anuragthehatter: I guess its okay to leave as is @liangxia. ```Openshift-network-operator``` system project is expected to have only one pod in it. We can definitely take ```Given a pod becomes ready with labels:``` but that needs to be accompanied by pseudo admin user step. Let me know WDYT

anuragthehatter: @liangxia ?

anuragthehatter:
Bump!

PR #1366, opened 2020-07-20
fix_etcd_cluster_rb

geliu2016:
@liangxia @pruan-rht @akostadinov this file seems miss require statement at the beginner, pls help to confirm and merge it, thanks


akostadinov:
@geliu2016 , this shouldn't be needed and it needs to be `project_resource` if we have anything at  all.

PR #1367, opened 2020-07-20
properly fix fully qualified resource names cache

akostadinov:
/hold


akostadinov:
/unhold
tested by others

PR #1371, opened 2020-07-21
Enable test for disconnected

jhou1:
This makes sure the image gets mirrored in disconnected env.


jhou1:
@sunilcio @weinliu @lyman9966 PTAL

PR #1379, opened 2020-07-22
OCP_11565_Fix help output issue

weinliu:
@jhou1 @sunilcio @lyman9966 could you help take a look?
log:  http://pastebin.test.redhat.com/886545

### DIFF
--- a/features/cli/oc_help.feature
+++ b/features/cli/oc_help.feature
@@ -1,6 +1,7 @@
 Feature: oc related features
 
   # @author chezhang@redhat.com
+  # @author weinliu@redhat.com
   # @case_id OCP-11565
   Scenario: kubectl secret subcommand - help
     Given I have a project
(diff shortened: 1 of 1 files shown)


sunilcio:
LGTM. However I want to discuss if we can deprecate this case as I see it is just checking oc secret help output which I guess is not adding much value. @jhou1 @lyman9966 hi, also need your opinion.

weinliu:
> LGTM. However I want to discuss if we can deprecate this case as I see it is just checking oc secret help output which I guess is not adding much value. @jhou1 @lyman9966 hi, also need your opinion.

Yes...Before fixing this, I also had a plan B of deprecating it.

jhou1:
+1, this test does not have much value for us.

lyman9966:
+1 I agree with the deprecation plan. It's not so valuable and for the help output may change in future version, it may bring more maintenance cost.

weinliu:
Inactive the case as discussed.

PR #1388, opened 2020-07-24
Default values for Different IaaS - aws/azure/gcp/ 

miyadav:
[10:57:16] INFO> Shell Commands: rm -r -f -- /root/workdir/d63db26e8a09-
    
    [10:57:16] INFO> Exit Status: 0
    [10:57:16] INFO> === End After Scenario: Implement defaulting machineset values for AWS/GCP/Azure/Vsphere, Examples (#1) ===

1 scenario (1 passed)
8 steps (8 passed)

Has flakes

### DIFF
--- a/features/machine/machine.feature
+++ b/features/machine/machine.feature
@@ -225,3 +225,81 @@ Feature: Machine features testing
       | name     | <%= cb.noderef_name %> |
     Then the step should succeed
     And the output should match "node-role.kubernetes.io/infra="
+
+  # @author miyadav@redhat.com
+  # @case_id OCP-32269
+  @admin
+  @destructive
+  Scenario: Implement defaulting machineset values for AWS
+    Given I have an IPI deployment
+    And I switch to cluster admin pseudo user
+    And I use the "openshift-machine-api" project
+    Then admin ensures machine number is restored after scenario
+
+    Given I store the last provisioned machine in the :machine clipboard
+    When evaluation of `machine(cb.machine).aws_ami_id` is stored in the :default_ami_id clipboard
+    And evaluation of `machine(cb.machine).aws_availability_zone` is stored in the :default_availability_zone clipboard
+    Then admin ensures "default-valued-32269" machineset is deleted after scenario
+
+    Given I obtain test data file "cloud/ms-aws/ms_default_values.yaml"
+    When I run oc create over "ms_default_values.yaml" replacing paths:
+      | n                                                                                         | openshift-machine-api                           |
+      | ["spec"]["selector"]["matchLabels"]["machine.openshift.io/cluster-api-cluster"]           | <%= infrastructure("cluster").infra_name %>     |
+      | ["spec"]["selector"]["matchLabels"]["machine.openshift.io/cluster-api-machineset"]        | default-valued-32269                            |
+      | ["spec"]["template"]["metadata"]["labels"]["machine.openshift.io/cluster-api-cluster"]    | <%= infrastructure("cluster").infra_name %>     |
+      | ["spec"]["template"]["spec"]["providerSpec"]["value"]["ami"]["id"]                        | <%= cb.default_ami_id %>                        |
+      | ["spec"]["template"]["spec"]["providerSpec"]["value"]["placement"]["availabilityZone"]    | <%= cb.default_availability_zone %>             |
+      | ["spec"]["template"]["metadata"]["labels"]["machine.openshift.io/cluster-api-machineset"] | default-valued-32269                            |
+    Then the step should succeed
+
+  # @author miyadav@redhat.com
+  # @case_id OCP-33056
+  @admin
+  @destructive
+  Scenario: Implement defaulting machineset values for GCP
+    Given I have an IPI deployment
+    And I switch to cluster admin pseudo user
+    And I use the "openshift-machine-api" project
+    Then admin ensures machine number is restored after scenario
+   
+    Given I store the last provisioned machine in the :machine clipboard
+    When evaluation of `machine(cb.machine).gcp_region` is stored in the :default_region clipboard
+    And evaluation of `machine(cb.machine).gcp_zone` is stored in the :default_zone clipboard
+    Then admin ensures "default-valued-33056" machineset is deleted after scenario
+   
+    Given I obtain test data file "cloud/ms-gcp/ms_default_values.yaml"
+    When I run oc create over "ms_default_values.yaml" replacing paths:
+      | n                                                                                         | openshift-machine-api                           |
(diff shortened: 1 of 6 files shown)


features/machine/machine.feature
```diff
@@ -225,3 +225,32 @@ Feature: Machine features testing
...
+
+  # @author miyadav@redhat.com
+  # @case_id OCP-32269
+  @admin
+  @destructive
+  Scenario Outline: Implement defaulting machineset values for AWS/GCP/Azure/Vsphere
+    Given I have an IPI deployment
+    And I switch to cluster admin pseudo user
+    And I use the "openshift-machine-api" project
+    And admin ensures "default-valued-<path>" machineset is deleted after scenario
```
jhou1, line 237: Please put the delete step after the create step

miyadav: thanks , made a new commit ..

lib/openshift/machine.rb
```diff
@@ -43,6 +43,12 @@ def ready?(user: nil, cached: true, quiet: false)
       instance_state == 'running'
     end
 
+    #location  for Azure machines set by installer 
+    def location(user: nil, cached: true, quiet: false)
```
akostadinov, line 47: This property seems only valid for Azure. Perhaps rename it to `azure_location`. In case all or most  clouds have such property, then I agree to keep it as is and it will return the location depending on which cloud it is in.

lib/openshift/machine.rb
```diff
@@ -43,6 +43,34 @@ def ready?(user: nil, cached: true, quiet: false)
...
+         dig('spec', 'providerSpec', 'value', 'zone')
+    end
+
+    #aws ami-id and availability zone
+    def aws_ami(user: nil, cached: true, quiet: false)
+       raw_resource(user: user, cached: cached ,quiet: quiet).
+         dig('spec', 'providerSpec', 'value', 'ami','id')
+    end
+
+    def aws_availabilityzone(user: nil, cached: true, quiet: false)
```
akostadinov, line 69: we split words for readability - `aws_availability_zone`

lib/openshift/machine.rb
```diff
@@ -43,6 +43,34 @@ def ready?(user: nil, cached: true, quiet: false)
...
+         dig('spec', 'providerSpec', 'value', 'region')
+    end
+
+    def gcp_zone(user: nil, cached: true, quiet: false)
+       raw_resource(user: user, cached: cached ,quiet: quiet).
+         dig('spec', 'providerSpec', 'value', 'zone')
+    end
+
+    #aws ami-id and availability zone
+    def aws_ami(user: nil, cached: true, quiet: false)
```
akostadinov, line 64: this returns both - ami id and availability zone? This is not very good especially if method name is `aws_ami`. I think it should be `aws_ami_id` and return only AMI id, then have another one for zone. But if you want both, call it some meaningful name.

akostadinov: ok, I see two different methods. Actually if method name is clear like `aws_ami_id`, adding a comment can only confuse the meaning. please remove the comments.

features/machine/machine.feature
```diff
@@ -225,3 +225,43 @@ Feature: Machine features testing
...
+      | n                                                                                         | openshift-machine-api                           |
+      | ["spec"]["selector"]["matchLabels"]["machine.openshift.io/cluster-api-cluster"]           | <%= infrastructure("cluster").infra_name %>     |
+      | ["spec"]["selector"]["matchLabels"]["machine.openshift.io/cluster-api-machineset"]        | default-valued-<path>                           |
+      | ["spec"]["template"]["metadata"]["labels"]["machine.openshift.io/cluster-api-cluster"]    | <%= infrastructure("cluster").infra_name %>     |
+      |  <extra_specs_1>                                                                          | <value_1>                                       |
+      |  <extra_specs_2>                                                                          | <value_2>                                       |
+      | ["spec"]["template"]["metadata"]["labels"]["machine.openshift.io/cluster-api-machineset"] | default-valued-<path>                           |
+    Then the step should succeed
+    And admin ensures "default-valued-<path>" machineset is deleted after scenario
```
akostadinov, line 259: clean-up step should be registered just before creating the object. To make sure that even if step partially fails, we will remove machine set

features/machine/machine.feature
```diff
@@ -225,3 +225,80 @@ Feature: Machine features testing
...
+      | n                                                                                         | openshift-machine-api                           |
+      | ["spec"]["selector"]["matchLabels"]["machine.openshift.io/cluster-api-cluster"]           | <%= infrastructure("cluster").infra_name %>     |
+      | ["spec"]["selector"]["matchLabels"]["machine.openshift.io/cluster-api-machineset"]        | default-valued-32269                            |
+      | ["spec"]["template"]["metadata"]["labels"]["machine.openshift.io/cluster-api-cluster"]    | <%= infrastructure("cluster").infra_name %>     |
+      | ["spec"]["template"]["spec"]["providerSpec"]["value"]["ami"]["id"]                        | <%= cb.default_ami_id %>                        |
+      | ["spec"]["template"]["spec"]["providerSpec"]["value"]["placement"]["availabilityZone"]    | <%= cb.default_availability_zone %>             |
+      | ["spec"]["template"]["metadata"]["labels"]["machine.openshift.io/cluster-api-machineset"] | default-valued-32269                            |
+    Then the step should succeed
+    And admin ensures "default-valued-32269" machineset is deleted after scenario
```
akostadinov, line 252: This remains unaddressed in all scenarios

> clean-up step should be registered just before creating the object. To make sure that even if step partially fails, we will remove machine set

miyadav:
@jhou1 @sunzhaohua2   - PTAL 
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/128965/console [aws]
 https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/128931/console  [gcp]
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/128944/console [azure]

jhou1:
In the testfiles, please remove the auto generated fields, like timestamp, generations, managedFields.

miyadav:
@jhou1 @sunzhaohua2  updated testdata files and validated looks good now - PTAL 
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/129396/console

akostadinov:
would be better if you also check that underlying cloud is what we expect in the beginning of scenario. But I will let you do that at a later time because it doesn't seem critical

/lgtm

PR #1391, opened 2020-07-26
need to delete existing meteringconfig if the types are different

pruan-rht:


pruan-rht:
@liangxia @akostadinov PTAL.  Thx!

PR #1392, opened 2020-07-27
match test case based on absolute path

rbbratta:
match test case based on absolute path

Certain test-runners (Rubymine) force the use of the abspath of the
`.feature` files on the cucumber command line.  If the `scenario.file`
is an abspath then testing strict string equaliy will fail.

Instead compare the absolute path with the `automation_file`
relative to `BushSlicer::HOME`

Example:
```
scenario.file = /home/rbrattai/workspace/verification-tests/features/tierN/networking/network-policy.feature
automation_file = features/tierN/networking/network-policy.feature
```

### DIFF
--- a/lib/polarshift/test_case.rb
+++ b/lib/polarshift/test_case.rb
@@ -130,7 +130,7 @@ module BushSlicer
           return false # no need for expensive checks once we are complete
         else
           scenario = ScenarioWrapper.new(test_case)
-          if automation_file == scenario.file &&
+          if File.absolute_path(scenario.file) == File.join(BushSlicer::HOME, automation_file) &&
              automation_scenario == scenario.name
             if automation_args && !scenario.example?
               logger.error "case #{id} mismatch with scenario type, will never run"


akostadinov:
Could you tell me what exactly you are doing such this becomes a problem? When we use the PolarShift matcher, we do not specify any files so I'm not sure what workflow causes an issue.

pruan-rht:
@akostadinov I think @rbbratta ran into this when he tried to run the script from within RubyMine IDE (@rbbratta, pls correct me if I misread your original descriptioin).

rbbratta:
Using RubyMine, I need to specify a path.

https://www.jetbrains.com/help/ruby/cucumber-support.html#cucumber_run_config

http://file.rdu.redhat.com/~rbrattai/logs/rubymine.png

akostadinov:
@rbbratta , I understood, I don't see you using PolarShift mode in the screenshot. Or do I miss it? Asking because the change only affects PolarShift mode.

rbbratta:
env vars

http://file.rdu.redhat.com/~rbrattai/logs/rubymine-env-vars.png

akostadinov:
I am thinking whether using `HOME` constant to check files are relative to repo makes sense here. WDYT? Is it good, unnecessary complication or harmful?

rbbratta:
It looks like `File.identical?` will work. It does a `File.stat` and compares dev and inode.


```C
rb_file_identical_p(VALUE obj, VALUE fname1, VALUE fname2)
{
    struct stat st1, st2;

    if (rb_stat(fname1, &st1) < 0) return Qfalse;
    if (rb_stat(fname2, &st2) < 0) return Qfalse;
    if (st1.st_dev != st2.st_dev) return Qfalse;
    if (st1.st_ino != st2.st_ino) return Qfalse;
    return Qtrue;

}
```

rbbratta:
doing the stat might slow things down a little since it has to touch the disk, but we've already parsed the scenario, so it should be cached.

rbbratta:
well `File.identical?` isn't what we want either.  Yes we should do

```ruby
          if scenario.file == File.join(BushSlicer::HOME, automation_file) &&
```

rbbratta:
err, that won't work if a relative path is provided.   We need an abspath somewhere, now I remember why I never work with relative paths and always normalize everything.

akostadinov:
@rbbratta , I lost track of this, do you still need it?

rbbratta:
Yes, it is still necessary for local Rubymine testing, and IMHO it is slightly more correct to compare abspath.

rbbratta:
There is no log output but it find tests on Runner-v3
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/200259/parameters/

akostadinov:
/lgtm

@liangxia , FYI, in case this causes some trouble, please revert this one.

PR #1396, opened 2020-07-28
auto case OCP-33206

zhaozhanqi:
@anuragthehatter @weliang1 @huiran0826 @rbbratta  PTAL, thanks
logs:
http://pastebin.test.redhat.com/888205

### DIFF
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -392,3 +392,31 @@ Feature: SDN related networking scenarios
       | since         | 30s                            |
     And the output should contain:
       | Not applying unsafe change: invalid configuration |
+
+
+  # @author zzhao@redhat.com
+  # @case_id OCP-33206
+  @admin
+  Scenario: sdn metrics should be using https scheme
+    Given the master version >= "4.6"
+    Given I switch to cluster admin pseudo user
+    When I use the "openshift-sdn" project
+    Then the expression should be true> service_monitor('monitor-sdn').service_monitor_endpoints_spec.first.scheme == 'https'
+
+    When evaluation of `endpoints('sdn-metrics').subsets.first.addresses.first.ip.to_s` is stored in the :sdn_metri_ip clipboard
+    And evaluation of `endpoints('sdn-metrics').subsets.first.ports.first.port.to_s` is stored in the :sdn_metri_port clipboard
+    And evaluation of `cb.sdn_metri_ip + ':' +cb.sdn_metri_port` is stored in the :sdn_metri_endpoint clipboard
+
+    And I use the "openshift-monitoring" project
+    When evaluation of `secret(service_account('prometheus-k8s').get_secret_names.find {|s| s.match('token')}).token` is stored in the :sa_token clipboard
+    When I run the :exec admin command with:
+      | n                | openshift-monitoring |
+      | pod              | prometheus-k8s-0     |
+      | c                | prometheus           |
+      | oc_opts_end      |                      |
+      | exec_command     | sh                   |
+      | exec_command_arg | -c                   |
+      | exec_command_arg | curl -k -H "Authorization: Bearer <%= cb.sa_token %>" https://<%= cb.sdn_metri_endpoint %>/metrics |
+    Then the step should succeed
+    And the output should contain:
+      | openshift_sdn_ovs_flows |
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -46,9 +46,9 @@ end
 Given /^the env is using one of the listed network plugins:$/ do |table|
   ensure_admin_tagged
   plugin_list = table.raw.flatten
-  _admin = admin
+  #_admin = admin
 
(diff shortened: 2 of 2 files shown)


zhaozhanqi:
Just saw OVN also have this feature.  maybe I need to update this for OVN

features/networking/sdn.feature
```diff
@@ -392,3 +392,30 @@ Feature: SDN related networking scenarios
...
+    And I use the "openshift-monitoring" project
+    When evaluation of `secret(service_account('prometheus-k8s').get_secret_names.find {|s| s.match('token')}).token` is stored in the :sa_token clipboard
+    When I run the :exec admin command with:
+      | n                | openshift-monitoring |
+      | pod              | prometheus-k8s-0     |
+      | c                | prometheus           |
+      | oc_opts_end      |                      |
+      | exec_command     | sh                   |
+      | exec_command_arg | -c                   |
+      | exec_command_arg | curl -k -H "Authorization: Bearer <%= cb.sa_token %>" https://<%= cb.sdn_metri_endpoint %>/metrics |
```
rbbratta, line 418: do we need to use `sh -c` in this case?

anuragthehatter: I guess if ```sh -c``` or ```bash -c``` is not specified for containers , the rest of the keywords after curl would be assumed as curl arguments which would fatal the run. Using sh/bash would tie whole command as per my experience. Not sure whats the latest is

anuragthehatter:
> Just saw OVN also have this feature. maybe I need to update this for OVN

Yea, but might be complicated though to make this case compatible to both. We don't have metrics cases for OVN though. You can search metrics staring with ```ovnkube-master``` or ```ovnkube-node``` though

features/networking/sdn.feature
zhaozhanqi: yes. As @anuragthehatter said

features/networking/sdn.feature
```diff
@@ -392,3 +392,30 @@ Feature: SDN related networking scenarios
...
       | Not applying unsafe change: invalid configuration |
+
+
+  # @author zzhao@redhat.com
+  # @case_id OCP-33206
+  @admin
+  Scenario: sdn metrics should be using https scheme
+    Given I switch to cluster admin pseudo user
+    When I use the "openshift-sdn" project
+    Then the expression should be true> service_monitor('monitor-sdn').service_monitor_endpoints_spec.first.scheme == 'https'
```
pruan-rht, line 403: @zhaozhanqi @anuragthehatter pointed out in #1461 that the endpoint names are different for OCP < 4.4.  So if you want this testcase to work for all version of OCP, we need to change the logic a bit.

anuragthehatter: @pruan-rht This is a specific 4.6 feature so no need to implement that logic here. Right @zhaozhanqi ?

pruan-rht: Then perhaps add a filter at the beginning to make sure the ocp version is >= 4.6

zhaozhanqi: yes, this case need version >=4.6.  I added the step >=4.6
Actually. this case cannot be selected for <4.6 by case filter.

features/networking/sdn.feature
rbbratta: the regular table form works in my testing

```cucumber
    And the output should contain "Hello OpenShift"
      | n                | openshift-monitoring                         |
      | pod              | prometheus-k8s-0                             |
      | c                | prometheus                                   |
      | oc_opts_end      |                                              |
      | exec_command     | curl                                         |
      | exec_command_arg | -k                                           |
      | exec_command_arg | -H                                           |
      | exec_command_arg | Authorization: Bearer <%= cb.sa_token %>     |
      | exec_command_arg | https://<%= cb.sdn_metri_endpoint %>/metrics |
```

anuragthehatter: That's seems right. Yeah, passing each flag and its argument as separate under ```exec_command_arg```.  Cool find. Nice!

anuragthehatter: But it seems to be required in such cases where where things are ```; or :``` separated

anuragthehatter: I guess to avoid spending time in figurng out what works or not its better we stick to ```bash -c```

rbbratta: exec only works for direct single command lines as it does use the shell at all.

These are the bash metacharacters:

```shell
       metacharacter
              A character that, when unquoted, separates words.  One of the following:
              |  & ; ( ) < > space tab newline
       control operator
              A token that performs a control function.  It is one of the following symbols:
              || & && ; ;; ;& ;;& ( ) | |& <newline>
```

and any other shell keyword:

```shell
       ! case  coproc  do done elif else esac fi for function if in select then until while { } time [[ ]]
```
Any of these require using `bash`


Once I was bored and read the entire `bash` man page, it is very useful, tons of cool tricks. :smiley:

zhaozhanqi:
I think this PR can be closed since another PR #1461 already cover this, thanks

PR #1398, opened 2020-07-28
Fix failed

kasturinarra:
Reduced the wait time to make sure that the test passes, otherwise test always fails to find hooks-1 dc

### DIFF
--- a/features/cli/deploy.feature
+++ b/features/cli/deploy.feature
@@ -77,7 +77,7 @@ Feature: deployment related features
       | f | deployment1.json |
     Then the step should succeed
     Given I obtain test data file "deployment/updatev1.json"
-    And I wait up to 60 seconds for the steps to pass:
+    And I wait up to 30 seconds for the steps to pass:
     """
     When I run the :replace client command with:
       | f | updatev1.json |


kasturinarra:

Ran with other set of cases with this fix and looks like it passes and need for reducing time was  i looked at the differences when ran in local and via jenkins and below are the differnces and initially there was no wait to replace the updatev1.json file due to error in 4.5 , just reduced the time to see if that passes !!

1. For  local i see that "oc get dc hooks -o yaml" has observedGeneration as "1" or "0"
2. via jenkins i see that "oc get dc hooks -o yaml" has observedGeneration as "2" 

https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/123636/consoleFull - Passed
Other runs where it failed:
=========================
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/123632/consoleFull
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/123611/consoleFull

features/cli/deploy.feature
```diff
@@ -77,7 +77,7 @@ Feature: deployment related features
       | f | deployment1.json |
     Then the step should succeed
     Given I obtain test data file "deployment/updatev1.json"
-    And I wait up to 60 seconds for the steps to pass:
+    And I wait up to 30 seconds for the steps to pass:
```
liangxia, line 80: I doubt this is the root cause of failure.
wait up to xx seconds means we wait xx seconds at most if the situation does not met (then timeout and fail the scenario). If the situation is met earlier, then we just break out the wait loop and continue execution the following steps.

kasturinarra: @liangxia i agree, i need to figure out what is causing this issue and fix it, i will mark this as hold for now !!

zhouying7780: @kasturinarra  I'm also can't figure out why hit this error , but I think we should change test logic to follow:
1) create dc;
2) check the first deploy succeed ;
3) replace the dc;
4) check the second deploy succeed;
5) rollback . 
Is this ok ?

kasturinarra: @zhouying7780 should we just wait for the hooks-1-deploy and its pod to be in completed state before rolling back ? Do you think that will help ?

kasturinarra:
closing the PR

PR #1401, opened 2020-07-29
fix 4.3 and 4.4 to support flexible scroll

yapei:
@yanpzhan  Help review the changes, comparison log 
Before the fix, we can see many `ElementClickInterceptedError` error  [log before the fix](https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/124050/consoleFull)
After the fix, no more `ElementClickInterceptedError` errors [log after the fix](https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/124417/consoleFull)


yapei:
Also added rules for OCP-25803 in 4.3 since `Given the master version >= "4.3"`

PR #1402, opened 2020-07-29
Add some step definations for local storage operator

qinpingli:
@chao007 @duanwei33  Added some step definitions for local storage operator test, could you help review if they're enough.

### DIFF
--- a/features/step_definitions/cluster_configuration/localstorage.rb
+++ b/features/step_definitions/cluster_configuration/localstorage.rb
@@ -51,7 +51,7 @@ Given /^local storage provisioner has been installed successfully$/ do
 
   step %Q/I use the "#{namespace}" project/
   # create localvolume
-  unless local_volume("#{name}").exists?
+  unless local_volume_local_storage_openshift_io("#{name}").exists?
     step %Q|I obtain test data file "storage/localvolume/localvolume.yaml"|
     step %Q/I run oc create over "localvolume.yaml" replacing paths:/, table(%{
       | ["metadata"]["name"]                                   | #{name}      |
@@ -86,6 +86,7 @@ Given /^local storage PVs are created successfully in schedulable workers$/ do
   })
 
   # check PVs are created
+  step %Q/20 seconds have passed/
   @result = BushSlicer::PersistentVolume.wait_for_labeled("storage.openshift.com/local-volume-owner-name=local-storage", user: admin, seconds: 5 * 60)
   if @result[:matching].count != cb.nodes.count
     raise "Create PVs failed, expected #{cb.nodes.count}, got: #{@result[:matching].count}"
--- a/features/step_definitions/cluster_configuration/localstorage.rb
+++ b/features/step_definitions/cluster_configuration/localstorage.rb
@@ -14,7 +14,7 @@ Given /^local storage operator has been installed successfully$/ do
   # create operator group
   step %Q/I use the "#{namespace}" project/
   unless operator_group(name).exists?
-    step %Q/I obtain test data file "storage\/localvolume\/operatorgroup.yaml"/
+    step %Q|I obtain test data file "storage/localvolume/operatorgroup.yaml"|
     step %Q/I run oc create over "operatorgroup.yaml" replacing paths:/, table(%{
       | ["metadata"]["name"]           | #{name}      |
       | ["metadata"]["namespace"]      | #{namespace} |
@@ -24,23 +24,22 @@ Given /^local storage operator has been installed successfully$/ do
   end
 
   # get channel
-  step %Q/evaluation of `cluster_version('version').channel.split('-')[1]` is stored in the :channel clipboard/
+  cb.channel = cluster_version('version').channel.split('-')[1]
 
   # create subscription
   unless subscription("#{name}").exists?
-    step %Q/I obtain test data file "storage\/localvolume\/sub.yaml"/
+    step %Q|I obtain test data file "storage/localvolume/sub.yaml"|
     step %Q/I run oc create over "sub.yaml" replacing paths:/, table(%{
       | ["metadata"]["name"] | #{name}         |
       | ["spec"]["channel"]  | "#{cb.channel}" |
     })
-    raise "Create subscription #{name} failied" unless @result[:success]
+    raise "Create subscription #{name} failed" unless @result[:success]
   end
 
   # check status
   step %Q/a pod becomes ready with labels:/, table(%{
     | name=local-storage-operator |
   })
-  raise "No pod is running for local storage operator" unless @result[:success]
 
   logger.info(" local storage operator is installed successfully in #{namespace} namespace")
 end
@@ -53,7 +52,7 @@ Given /^local storage provisioner has been installed successfully$/ do
   step %Q/I use the "#{namespace}" project/
   # create localvolume
   unless local_volume("#{name}").exists?
-    step %Q/I obtain test data file "storage\/localvolume\/localvolume.yaml"/
+    step %Q|I obtain test data file "storage/localvolume/localvolume.yaml"|
     step %Q/I run oc create over "localvolume.yaml" replacing paths:/, table(%{
       | ["metadata"]["name"]                                   | #{name}      |
       | ["metadata"]["namespace"]                              | #{namespace} |
@@ -65,10 +64,10 @@ Given /^local storage provisioner has been installed successfully$/ do
   # check ds status
   step %Q/I store the nodes in the :nodes clipboard/
   step %Q/#{cb.nodes.count} pods become ready with labels:/, table(%{
-      | app=local-volume-diskmaker-local-storage |
+    | app=local-volume-diskmaker-local-storage |
   })
   step %Q/#{cb.nodes.count} pods become ready with labels:/, table(%{
-      | app=local-volume-provisioner-local-storage |
+    | app=local-volume-provisioner-local-storage |
   })
 
   unless storage_class("#{name}").exists?
@@ -76,7 +75,7 @@ Given /^local storage provisioner has been installed successfully$/ do
   end
 end
 
-Given /^some local storage PVs are created successfully$/ do
+Given /^local storage PVs are created successfully in schedulable workers$/ do
   ensure_admin_tagged
 
   step %Q/I store the schedulable workers in the :nodes clipboard/
@@ -87,9 +86,9 @@ Given /^some local storage PVs are created successfully$/ do
   })
 
   # check PVs are created
-  @result = BushSlicer::PersistentVolume.wait_for_labeled("storage.openshift.com/local-volume-owner-name=local-storage", user: admin, seconds: 2 * 60)
+  @result = BushSlicer::PersistentVolume.wait_for_labeled("storage.openshift.com/local-volume-owner-name=local-storage", user: admin, seconds: 5 * 60)
   if @result[:matching].count != cb.nodes.count
-    raise "Create PVs failed"
+    raise "Create PVs failed, expected #{cb.nodes.count}, got: #{@result[:matching].count}"
   end
 
 end
--- a/features/step_definitions/cluster_configuration/localstorage.rb
+++ b/features/step_definitions/cluster_configuration/localstorage.rb
@@ -1,245 +1,95 @@
-# OCP supports fileystem local storage provisioner from 3.7
-# For some OCP versions, before depolying local storage provisoner,
-# we need enable some feature gates first.
-# ref: doc/storage_automation_spec.adoc
-
-Given /^I deploy local storage provisioner(?: with "([^ ]+?)" version)?$/ do |img_version|
+Given /^local storage operator has been installed successfully$/ do
   ensure_admin_tagged
-  ensure_destructive_tagged
+  namespace = "openshift-local-storage"
+  name = "local-storage"
 
-  namespace = env.local_storage_provisioner_project.name
-  config_hash = env.master_services[0].config.as_hash()
-  imgformat = config_hash["imageConfig"]["format"]
-  img_registry = imgformat.split("\/")[0]
-  ose_version = env.get_version(user: admin)
-  img_version = "v#{ose_version[0]}" unless img_version
-  serviceaccount="local-storage-admin"
-  configmap="local-volume-config"
-  template="local-storage-provisioner"
-  image="#{img_registry}/openshift3/local-storage-provisioner:#{img_version}"
-  path ||="/mnt/local-storage"
-  cmurl = "#{BushSlicer::HOME}/testdata/storage/localvolume/configmap-37.yaml"
-  if env.version_ge("3.10", user: user)
-    cmurl = "#{BushSlicer::HOME}/testdata/storage/localvolume/configmap.yaml"
+  # create namespace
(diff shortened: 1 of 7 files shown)


duanwei33:
The steps look good to me. 
Just one question, if we define some different sub.yaml files for different release+source, it could also support 4.2-4.5?

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,95 @@
...
-      "mkdir -p #{path}/slow/vol2",
-      "mount -t tmpfs svol2 #{path}/slow/vol2",
-      "chcon -R unconfined_u:object_r:svirt_sandbox_file_t:s0 #{path}"
-    ]
-    res = host.exec_admin(*setup_commands)
-    raise "error preaparing subdirs for local storage provisioner" unless res[:success]
+  # create operator group
+  step %Q/I use the "#{namespace}" project/
+  unless operator_group(name).exists?
+    step %Q/I obtain test data file "storage\/localvolume\/operatorgroup.yaml"/
```
pruan-rht, line 17: @qinpingli to bypass escape, you can use something like `step %Q|....|` instead.  You are not limited to use `/` as a step delimiter.

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,95 @@
...
-  @result = admin.cli_exec(add_command, **_opts)
-  step %Q/the step should succeed/
-
-  step %Q/I run the :create admin command with:/, table(%{
-      | f | #{BushSlicer::HOME}/testdata/storage/localvolume/local-block-template.yaml |
-      | n | #{namespace}                                                                                                     |
+  # check ds status
+  step %Q/I store the nodes in the :nodes clipboard/
+  step %Q/#{cb.nodes.count} pods become ready with labels:/, table(%{
+      | app=local-volume-diskmaker-local-storage |
```
pruan-rht, line 68: @qinpingli indentation should only be 2

testdata/storage/localvolume/sub.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: local-storage-operator
+spec:
+  channel: "4.6"
```
pruan-rht, line 6: @qinpingli perhap this should be a variable?  Or are you  patching it in your test?

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,95 @@
...
-  step %Q/I download a file from "#{cmurl}"/
-  cfm = YAML.load(@result[:response])
-  filepath = @result[:abs_path]
-  if env.version_ge("3.10", user: user)
-    cfm["data"]["storageClassMap"].gsub!("/mnt/local-storage", path)
-  else
-    cfm["data"]["local-fast"].gsub!("/mnt/local-storage", path)
-    cfm["data"]["local-slow"].gsub!("/mnt/local-storage", path)
+  # get channel
+  step %Q/evaluation of `cluster_version('version').channel.split('-')[1]` is stored in the :channel clipboard/
```
pruan-rht, line 27: @qinpingli when it's inside a step defintion, you don't need to have the cucumber step overhead.  Probably just `cb.channel = cluster_version('version').channel.split('-')[1]` would be enough.

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,95 @@
...
-  step 'admin ensures "local-block:provisioner-pv-binding" clusterrolebinding is deleted'
-  step %Q/I run the :new_app admin command with:/, table(%{
-      | param    | CONFIGMAP=#{configmap}            |
-      | param    | SERVICE_ACCOUNT=#{serviceaccount} |
-      | param    | NAMESPACE=#{namespace}            |
-      | param    | PROVISIONER_IMAGE=#{image}        |
-      | template | #{template}                       |
-      | n        | #{namespace}                      |
+  step %Q/#{cb.nodes.count} pods become ready with labels:/, table(%{
+      | app=local-volume-provisioner-local-storage |
```
pruan-rht, line 71: indentation

qinpingli:
> The steps look good to me.
> Just one question, if we define some different sub.yaml files for different channel+source, it could also support 4.2-4.5?

Now the channel comes from the cluster version, if the cluster version is 4.3.z, then the step will update the subscription channel to 4.3.

There is no interface to help define the source, if it's needed, let me add it.

testdata/storage/localvolume/sub.yaml
qinpingli: Yes, I updated it in the step with the following code:
```
# get channel
step %Q/evaluation of `cluster_version('version').channel.split('-')[1]` is stored in the :channel clipboard/
step %Q/I run oc create over "sub.yaml" replacing paths:/, table(%{
      | ["metadata"]["name"] | #{name}         |
      | ["spec"]["channel"]  | "#{cb.channel}" |
    })
```

qinpingli:
@pruan-rht  @duanwei33 Thanks for your review. Updated the PR per the comments, please help take a look again.

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,95 @@
...
-      "losetup /dev/loop23 --show #{path}/block-devices/dev1",
-      "losetup /dev/loop24 --show #{path}/block-devices/dev2",
-      "ln -s /dev/loop23 #{path}/block-devices/ldev1",
-      "ln -s /dev/loop24 #{path}/block-devices/ldev2",
-      "chcon -R unconfined_u:object_r:svirt_sandbox_file_t:s0 /dev/loop23 /dev/loop24",
-      "chcon -R unconfined_u:object_r:svirt_sandbox_file_t:s0 #{path}"
-    ]
-    res = host.exec_admin(*setup_commands)
-    raise "error preaparing subdirs for local storage provisioner" unless res[:success]
+  namespace = "openshift-local-storage"
```
pruan-rht, line 50: @qinpingli can the local storage operator be installed in any namespace?

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,95 @@
...
+  unless storage_class("#{name}").exists?
+    raise "Create storageclass #{name} failed"
+  end
+end
 
-  pv_count = 0
-  BushSlicer::PersistentVolume.list(user: admin).each { |pv|
-    pv_count += 1 if pv.name.start_with?("local-pv-") && (pv.local_path&.start_with?("#{path}/block-devices"))
-  }
+Given /^some local storage PVs are created successfully$/ do
```
pruan-rht, line 79: @qinpingli can you pls elaborate why you named the step `some local...`?  As it stands, it's kind of vague as to what `some` means.

features/test/storage.feature
```diff
@@ -23,3 +23,10 @@ Feature: some storage related scenarios
   @admin
   Scenario: Store PV into clipboard
     Given admin stores all persistentvolumes to the clipboard
+
+  @admin
+  Scenario: install lso
```
pruan-rht, line 28: perhaps use thee full name instead to be clear (local-storage operator)

features/step_definitions/cluster_configuration/localstorage.rb
qinpingli: Yes, it can be installed in any namespace, but we don't recommend doing this.

features/step_definitions/cluster_configuration/localstorage.rb
qinpingli: This step is used to prepare PVs for local storage plugin testing. The PV number depends on the cluster schedulable workers account. In this step, one loop device is created on each schedulable work, so the local storage provisioner will create one PV for each loop device.
If we want to test the local storage provisioner, we need another step.

features/test/storage.feature
qinpingli: It's not a test case, just a test code for the new steps. I'll update it.

features/step_definitions/cluster_configuration/localstorage.rb
pruan-rht: ok

features/step_definitions/cluster_configuration/localstorage.rb
pruan-rht: @qinpingli I think we should rename the step... how about `Given local storage PVs are created successfully in schedulable workers`?  WDYT @akostadinov @liangxia

qinpingli: Thanks, updated.

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,95 @@
...
+  cb.channel = cluster_version('version').channel.split('-')[1]
+
+  # create subscription
+  unless subscription("#{name}").exists?
+    step %Q|I obtain test data file "storage/localvolume/sub.yaml"|
+    step %Q/I run oc create over "sub.yaml" replacing paths:/, table(%{
+      | ["metadata"]["name"] | #{name}         |
+      | ["spec"]["channel"]  | "#{cb.channel}" |
+    })
+    raise "Create subscription #{name} failied" unless @result[:success]
```
liangxia, line 36: Typo of "failied"

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,95 @@
...
-  step %Q{I obtain test data file "storage/misc/storageClass.yaml"}
-  sc = YAML.load(@result[:response])
-  filepath = @result[:abs_path]
-  sc["metadata"]["name"] = "local-fast"
-  sc["provisioner"] = "kubernetes.io/no-provisioner"
-  sc["volumeBindingMode"] = "Immediate"
-  File.write(filepath, sc.to_yaml)
-  @result = admin.cli_exec(:create, f: filepath)
-  step %Q/the step should succeed/
+  raise "No pod is running for local storage operator" unless @result[:success]
```
liangxia, line 43: This check is not needed.  As above check will fail itself if the criteria is not met.

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,95 @@
...
+      | if [ ! -e /srv/block-devices/dev1 ]; then dd if=/dev/zero of=/srv/block-devices/dev1 bs=1M count=100 seek=30000; fi |
+      | if [ ! -e /dev/loop23 ]; then losetup /dev/loop23 --show /srv/block-devices/dev1; fi                                |
+  })
 
-  lf_sc = storage_class("block-devices")
-  lf_sc.ensure_deleted(user: admin)
+  # check PVs are created
+  @result = BushSlicer::PersistentVolume.wait_for_labeled("storage.openshift.com/local-volume-owner-name=local-storage", user: admin, seconds: 2 * 60)
+  if @result[:matching].count != cb.nodes.count
+    raise "Create PVs failed"
```
liangxia, line 92: How about `raise "Create PVs failed, expected: xxx, got: yyy"`

features/step_definitions/cluster_configuration/localstorage.rb
qinpingli: Fixed

features/step_definitions/cluster_configuration/localstorage.rb
qinpingli: Fixed.

features/step_definitions/cluster_configuration/localstorage.rb
qinpingli: fixed.

qinpingli:
@liangxia Could you help take a look at this PR again?

lib/openshift/local_volume.rb
```diff
@@ -0,0 +1,8 @@
+require 'openshift/project_resource'
+
+module BushSlicer
+  # represents LocalVolume CRD
+  class LocalVolume < ProjectResource
+    RESOURCE = 'localvolume'
```
akostadinov, line 6: please use full resource path and API name. most likely `localvolumes.openshift.something...`

akostadinov: look at `selfLink` for exact string

features/step_definitions/cluster_configuration/localstorage.rb
```diff
@@ -1,245 +1,94 @@
-# OCP supports fileystem local storage provisioner from 3.7
-# For some OCP versions, before depolying local storage provisoner,
-# we need enable some feature gates first.
-# ref: doc/storage_automation_spec.adoc
-
-Given /^I deploy local storage provisioner(?: with "([^ ]+?)" version)?$/ do |img_version|
+Given /^local storage operator has been installed successfully$/ do
```
akostadinov, line 1: Why not use previous step name, just remove version stuff? This step name doesn't speak well whether installation is performed now or we just check that there is such installation. Old name made clear that we do install now.

akostadinov: Same with the other steps. We better make clear in step name whether we install now or we just check it is installed.

lib/openshift/local_volume.rb
qinpingli: Thanks for the review, updated.

features/step_definitions/cluster_configuration/localstorage.rb
qinpingli: This step is split into 3 steps, one is local storage operator deployment, the second one is local storage provisioner deployment, the third one is creating some PVs.

For each step, it will check if the desired result is fulfilled first, if yes, it will do nothing, else it will do something to reach the desired result.

qinpingli:
@akostadinov any update?

features/step_definitions/cluster_configuration/localstorage.rb
akostadinov: ok

lib/openshift/local_volume.rb
akostadinov: did you forget to push the change? I still see the old string here.

akostadinov:
@qinpingli , sorry for the delay, feel fre to ping me in chat to expedite this PR

qinpingli:
@akostadinov Sorry, I forgot to push the update, pushed it now. Please help review again, thanks!

lib/openshift/local_volume_local_storage_openshift_io.rb
```diff
@@ -0,0 +1,8 @@
+require 'openshift/project_resource'
+
+module BushSlicer
+  # represents LocalVolume CRD
+  class LocalVolumeLocalStorageOpenshiftIo < ProjectResource
+    RESOURCE = 'localvolume.local.storage.openshift.io'
```
akostadinov, line 6: I believe this needs to be `localvolumes.local.storage.openshift.io'
I mean the `s` at the end of RESOURCE. Sorry for delay, please ping me if I fail to respond in a few days!

@pruan-rht , @liangxia , this is good to merge on your end?

PR #1406, opened 2020-07-30
update for yaml format

zhouying7780:
@kasturinarra  Please help review , thanks


akostadinov:
btw this is not changing "YAML format" but is updating the resource YAML to a newer API version. Would be good to use more precise commit messages.

PR #1408, opened 2020-07-31
Auto OCP-18316 OCP-18315 OCP-19785

huiran0826:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1976/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1977/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/1978/console

@anuragthehatter @zhaozhanqi @weliang1 @rbbratta

### DIFF
--- a/features/networking/egress-ip.feature
+++ b/features/networking/egress-ip.feature
@@ -371,3 +371,201 @@ Feature: Egress IP related features
       | /usr/bin/curl | --connect-timeout | 10 | <%= cb.hostip %>:27017 |
     Then the output should contain:
       | Hello-OpenShift-1 http-8080 |
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-18316
+  @admin
+  @destructive
+  Scenario: The egressIPs should work well when re-using the egressIP which is holding by a deleted project
+    Given I store the schedulable workers in the :nodes clipboard
+    And the valid egress IP is added to the "<%= cb.nodes[0].name %>" node
+    And I have a project
+
+    # add the egress ip to the project
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+
+    # delete the project
+    Given admin ensures "<%= project.name %>" project is deleted
+
+    #create a new project
+    Given I create a new project
+    And I have a pod-for-ping in the project
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+
+    # The traffic should be allowed and the source ip is egress ip
+    When I execute on the pod:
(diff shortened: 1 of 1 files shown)


features/networking/egress-ip.feature
```diff
@@ -371,3 +371,107 @@ Feature: Egress IP related features
...
+
+    #create a new project
+    Given I create a new project
+    And I have a pod-for-ping in the project
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+
+    # The traffic should be allowed and the source ip is egress ip
+    When I execute on the pod:
+      | curl | -s | --connect-timeout | 5 | ifconfig.me |
```
rbbratta, line 399: does this flake a lot when the Internet breaks?   If we made this a step we could retry or try alternate sites

It looks like we can simplify the output to just the IP with https://ifconfig.me/ip

It looks like https://ifconfig.co/ip  works as well as a backup.

rbbratta: Looks like https://ifconfig.co/ is open source, so we could host it locally on RH AWS or some such. https://github.com/mpolden/echoip  

Also "Faster than ifconfig.me and has IPv6 support"

huiran0826: Thanks @rbbratta for the suggestion. Actually I have thought this to setup a local VM and install echoip service on it. But this way, we have to maintain a long running VM there that also would cause some problems if the server down.
 Currently I think it is okay to keep it this way.

rbbratta: did we want to use the `/ip` path to just get the `text/plain` simple output.

`curl ifconfig.io/ip`

huiran0826: / # curl ifconfig.io
136.144.52.212

the output also only contains ip, is there something else I miss?

rbbratta: ah, I guess they detect the curl User-Agent and just return the IP, nice.

features/networking/egress-ip.feature
```diff
@@ -371,3 +371,202 @@ Feature: Egress IP related features
...
+  Scenario: Pods will not be affected by the egressIP set on other netnamespace
+    # create project with pods
+    Given I have a project
+    And evaluation of `project.name` is stored in the :proj1 clipboard
+
+    # add the egress ip to the hostsubnet
+    Given I store the schedulable workers in the :nodes clipboard
+    And the valid egress IP is added to the "<%= cb.nodes[0].name %>" node
+
+
```
liangxia, line 492: Extra empty lines ?

huiran0826: Thanks, remove the extra empty line.

PR #1411, opened 2020-08-03
Fix ocp29857

miyadav:
[New story ](https://issues.redhat.com/browse/OCPCLOUD-911) have made updates to how invalid values were checked , hence this needs to be updated 

logs--
.
...
      [09:09:55] INFO> Exit Status: 1
      | n | openshift-machine-api |
    Then the output should contain:                                                                                                           # features/step_definitions/common.rb:33
      | The MachineHealthCheck "mhc-malformed" is invalid: spec.maxUnhealthy: Invalid value: "": spec.maxUnhealthy in body should match '^((100 | [0-9]{1,2})% | [0-9]+)$' |
      [09:09:55] INFO> === After Scenario: [MHC] MaxUnhealthy should not allow malformed values ===
      [09:09:55] INFO> Shell Commands: rm -r -f -- /root/workdir/26351978eef8-
      
      [09:09:55] INFO> Exit Status: 0
      [09:09:55] INFO> === End After Scenario: [MHC] MaxUnhealthy should not allow malformed values ===

1 scenario (1 passed)
7 steps (7 passed)
0m9.445s
[09:09:55] INFO> === At Exit ===

### DIFF
--- a/features/machine/machine_healthcheck.feature
+++ b/features/machine/machine_healthcheck.feature
@@ -211,19 +211,9 @@ Feature: MachineHealthCheck Test Scenarios
     Given I obtain test data file "cloud/mhc/mhc_malformed.yaml"
     When I run oc create over "mhc_malformed.yaml" replacing paths:
       | n  | openshift-machine-api |
-    Then the step should succeed
-    And I ensure "mhc-malformed" machinehealthcheck is deleted after scenario
-
-    Then a pod becomes ready with labels:
-      | api=clusterapi, k8s-app=controller |
-
-    When I run the :logs admin command with:
-      | resource_name | <%= pod.name %>                |
-      | c             | machine-healthcheck-controller |
     Then the output should contain:
-      | remediation won't be allowed: invalid value for IntOrString  |
-      | total targets: <%= cb.machines.count %>                      | #This covers OCP-29062 - empty selectors watches all machines in cluster
-
+      | The MachineHealthCheck "mhc-malformed" is invalid: spec.maxUnhealthy: Invalid value: "": spec.maxUnhealthy in body should match '^((100|[0-9]{1,2})%|[0-9]+)$'|
+      
   # @author miyadav@redhat.com
   # @case_id OCP-28859
   @admin


miyadav:
/unhold

@sunzhaohua2  @jhou1  PTAL

features/machine/machine_healthcheck.feature
```diff
@@ -211,19 +211,9 @@ Feature: MachineHealthCheck Test Scenarios
...
-      | api=clusterapi, k8s-app=controller |
-
-    When I run the :logs admin command with:
-      | resource_name | <%= pod.name %>                |
-      | c             | machine-healthcheck-controller |
     Then the output should contain:
-      | remediation won't be allowed: invalid value for IntOrString  |
-      | total targets: <%= cb.machines.count %>                      | #This covers OCP-29062 - empty selectors watches all machines in cluster
-
+      | The MachineHealthCheck "mhc-malformed" is invalid: spec.maxUnhealthy: Invalid value: "": spec.maxUnhealthy in body should match '^((100|[0-9]{1,2})%|[0-9]+)$'|
```
jhou1, line 215: Is this test backward compatible for 4.4/4.5? Can you please make the text shorter to catch minimum keywords possible? 

The reason I'm asking this is because I hope future changes to the logs do not suddenly fail our tests.

miyadav: for us we need  it compatible with 4.5 , I have let the dev know about it , it might be made backward compatible , but not confirmed , will update the string as you suggested , will keep it on hold till we get confirmation.
Thanks

jhou1: Thanks, what I meant to say was, the longer the text we want to match, the greater chance the test fails due to some changes in the product. I think that having `"mhc-malformed" is invalid` and `Invalid value: ""` captured in our test can be suffficient, and it's more likely to work for 4.5 as well.

miyadav:
this should be good for now - https://github.com/openshift/verification-tests/pull/1417

PR #1419, opened 2020-08-05
[OCPQE-1641] Placeholder directory for essential Dockerfiles

quarterpin:
There some essential Dockerfiles that the Edge team uses for testing and modifying the docker test container images. Most of these files are not maintained or locally used. For safekeeping and maintaining these files, this push basically creates a central repo for them. 

@lihongan , request for a peer review.


testdata/routing/images/caddy-docker/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+RUN chmod 0755 /usr/bin/caddy
+
+EXPOSE 8080 8443
+WORKDIR /srv
+
+ADD Caddyfile /etc/caddy/Caddyfile
+ADD public/ /srv/public/
+ADD publics/ /srv/publics/
+ADD certs/ /srv/certs
```
lihongan, line 16: For completeness, those files/folders that `ADD` to the image should be upload to the same folder as well and ensure using relative path.

quarterpin: Good catch!  I have added all those files to the relative directory in the same tree structure as the original ones. Let me know if there is anything further to be done. 

Thanks.

lihongan:
/lgtm
Thanks @quarterpin

liangxia:
BIN +14.8 MB testdata/routing/images/caddy-docker/caddy
How about add this file into the image ?

quarterpin:
Hi @liangxia ,

Thanks for the review, as suggested, I have removed the bin file which I think can be introduced during the image build process. 

Let me know in case if there is anything else to be changed.

PR #1422, opened 2020-08-05
xt_u32 test coverage for SDN

anuragthehatter:
Tested on 4.4,4.5 and 4.6 - OCP 33413
Logs: https://privatebin-it-iso.int.open.paas.redhat.com/?98628e852ce477a3#jnpIS4NC6qLXpVt9ZtQoZJ6jkxLBhnDL6RAzc+IPyVM=

@rbbratta @zhaozhanqi @huiran0826 @weliang1

### DIFF
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -303,8 +303,8 @@ Feature: Pod related networking scenarios
     And evaluation of `service(cb.host_pod1.name).node_port(port: 8080)` is stored in the :nodeport clipboard
     #Creating a simple client pod to generate traffic from it towards the exposed node IP address
     Given I obtain test data file "networking/aosqe-pod-for-ping.json"
-    When I run the :create client command with:
-      | f | aosqe-pod-for-ping.json |
+    When I run oc create over "aosqe-pod-for-ping.json" replacing paths:
+      | ["spec"]["nodeName"] | <%= cb.nodes[0].name %> |
     Then the step should succeed
     Given a pod becomes ready with labels:
       | name=hello-pod |
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -452,3 +452,56 @@ Feature: Pod related networking scenarios
     """
     OVN is functional on the cluster
     """
+  # @author anusaxen@redhat.com
+  # @case_id OCP-33413
+  @admin
+  @destructive
+  Scenario: xt_u32 kernel module functionality check from NET_ADMIN pods
+    Given I have a project
+    And I have a pod-for-ping in the project
+    Then evaluation of `pod.name` is stored in the :client_pod clipboard
+    Given I select a random node's host
+    #Making sure the iptables on latest centos are actually nf_tables as the xt_u32 is only supported with that
+    When I run commands on the host:
+      | iptables --version |
+    Then the step should succeed
+    Then the output should contain "nf_tables"
+
+    Given I obtain test data file "networking/centos_latest_admin.yaml"
+    When I run oc create as admin over "centos_latest_admin.yaml" replacing paths:
+      | ["metadata"]["name"]      | server-pod          |
+      | ["metadata"]["namespace"] | <%= project.name %> |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | name=centos-pod |
+    Then evaluation of `pod` is stored in the :server_pod clipboard
+    Given I use the "<%= cb.server_pod.node_name %>" node
+    #Enabling XT_u32 module on server pod's node
+    And I run commands on the host:
+      | modprobe xt_u32 |
+    Then the step should succeed
+
+    #Checking whether server pod also has nf_tables version installed
+    When admin executes on the "<%= cb.server_pod.name %>" pod:
+      | bash | -c | iptables --version |
+    Then the step should succeed
+    And the output should contain "nf_tables"
+
+    #Creating iptable rule on server pod with module u32 and Successful pings will represent a successful xt_u32 match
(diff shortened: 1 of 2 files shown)


features/networking/pod.feature
```diff
@@ -452,3 +452,56 @@ Feature: Pod related networking scenarios
...
+      | modprobe xt_u32 |
+    Then the step should succeed
+
+    #Checking whether server pod also has nf_tables version installed
+    When admin executes on the "<%= cb.server_pod.name %>" pod:
+      | bash | -c | iptables --version |
+    Then the step should succeed
+    And the output should contain "nf_tables"
+
+    #Creating iptable rule on server pod with module u32 and Successful pings will represent a successful xt_u32 match
```
weliang1, line 490: @anuragthehatter Could you describe what's the detail rule the line 492 try to match? I mean you may do the same thing as the line 498 mentioning "less than 256 bytes".

features/networking/pod.feature
```diff
@@ -452,3 +452,56 @@ Feature: Pod related networking scenarios
...
+    #Successful pings will represent a successful xt_u32 match. Here we allow ping for less than 256 bytes packet else Reject
+    Given admin executes on the "<%= cb.server_pod.name %>" pod:
+      | bash | -c | iptables -t filter -A INPUT -i eth0 -m u32 --u32 '0 & 0xFFFF = 0x100:0xFFFF' -j REJECT |
+    Then the step should succeed
+    When I execute on the "<%= cb.client_pod %>" pod:
+      | ping | -s 64 |-c1 | <%= cb.server_pod.ip %> |
+    Then the step should succeed 
+    Given I execute on the "<%= cb.client_pod %>" pod:
+      | ping | -s 256 |-c1 | <%= cb.server_pod.ip %> |
+    Then the step should fail
```
weliang1, line 507: Your testing covered two xt_u32 match rules created in the testing pods, not sure if we need a testing to define the xt_u32 match rules in the node level, WDYT?

anuragthehatter: @weliang1 The e2e test should be to run iptables in a pod as per talk with Yuval and customers are going to use this feature on pods only. Unless there is requirement for node as well, we can invest time on that in future.

features/networking/pod.feature
anuragthehatter: Sure its bit complicated actually. We are using xt_u32 module for creating bit level rules in an IP header. We are using ICMP/ping here. Maximum payload a ping packet can take is 56 bytes(excluding 8 header) so a 2000 bytes source will trigger some fragmentation. Detailed explanation for this rule is [here](http://www.stearns.org/doc/iptables-u32.current.html) and search ```Moving on to the TCP header``` section . 256 bytes test was taken from here as well and explains how iptables entry was designed for that.

In short we are extracting 6&0xFF=0x6 from the packet which will tell whether its a TCP packet. This is stored in byte 9 of the IP header, so we'll pull 4 bytes starting at byte 6, drop 6-8, and check to see if it's 6 and so on.

The first rule is actually just to check if -m 32 (module xt_u32) works/matches with iptables and 2nd rule is to check extended functionality like byte level ACCEPT/REJECT. Combining both rules may not be a good idea for debugging. At least if test fails at first rule, we know something is wrong with module and if fails at second, it might depict packet bit matching level issues.

features/networking/pod.feature
weliang1: /LGTM

testdata/networking/centos_latest_admin.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: centos-pod
+  labels:
+    name: centos-pod
+spec:
+  containers:
+  - name: centos-pod
+    command: ["/bin/bash", "-c"]
+    args:
+        ["yum install -y iptables && sleep inf"]
+    image: centos:latest
```
zhaozhanqi, line 13: image need to push to qe registry

features/networking/pod.feature
```diff
@@ -452,3 +452,56 @@ Feature: Pod related networking scenarios
...
+    Then evaluation of `pod.name` is stored in the :client_pod clipboard
+    Given I select a random node's host
+    #Making sure the iptables on latest centos are actually nf_tables as the xt_u32 is only supported with that
+    When I run commands on the host:
+      | iptables --version |
+    Then the step should succeed
+    Then the output should contain "nf_tables"
+
+    Given I obtain test data file "networking/centos_latest_admin.yaml"
+    When I run oc create as admin over "centos_latest_admin.yaml" replacing paths:
```
zhaozhanqi, line 471: here need `admin` user to create the server pod?

testdata/networking/centos_latest_admin.yaml
anuragthehatter: Its fine if we keep this with ```Disonnected:No```. Okay to keep regression on connected clusters. WDYT

features/networking/pod.feature
anuragthehatter: This is a NET_ADMIN pod so need to be deal administratively

testdata/networking/centos_latest_admin.yaml
zhaozhanqi: team require our test image need to from QE registry, maybe consider in case the image is gone for public in future. :)

features/networking/pod.feature
zhaozhanqi: it's ok just curious

testdata/networking/centos_latest_admin.yaml
anuragthehatter: Yea we should do that. Great suggestion @zhaozhanqi.

PR #1427, opened 2020-08-10
Fix ip route show default parsing

rbbratta:
On OVN the default interface is `br-ex`.
If we use the `\W` regex to split the output,
the `\W` will match the `-` in `br-ex` since the
dash is not a word character `([A-z_])`

```
sh-4.2# ip route show default
default via 10.0.0.1 dev br-ex proto dhcp metric 800
```
switch to `\s` and split on whitespace to match `br-ex`
instead of `br`.

Idealy we would use `-j` to output JSON but the iproute2
version in the container doesn't support it yet.

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -712,13 +712,15 @@ Given /^the default interface on nodes is stored in the#{OPT_SYM} clipboard$/ do
   end
   case networkType
   when "OVNKubernetes"
-    step %Q/I run command on the node's ovnkube pod:/, table("| bash | -c | ip route show default |")
+    # use -4 to limit output to just `default` interface, fixed in later iproute2 versions
+    step %Q/I run command on the node's ovnkube pod:/, table("| ip | -4 | route | show | default |")
   when "OpenShiftSDN"
-    step %Q/I run command on the node's sdn pod:/, table("| bash | -c | ip route show default |")
+    step %Q/I run command on the node's sdn pod:/, table("| ip | -4 | route | show | default |")
   else
     raise "unknown networkType"
   end
-  cb[cb_name] = @result[:response].split("\n").first.split(/\W+/)[7]
+  # OVN uses `br-ex` and `-` is not a word char, so we have to split on whitespace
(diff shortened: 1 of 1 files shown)


anuragthehatter:
Thanks @rbbratta . Could you also run a test using such steps on both SDN and OVN cluster? We should merge it soon to avoid premature test exits in CI.

rbbratta:
OVN http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-08-10-11-13-45.log
SDN http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-08-10-11-19-40.log

rbbratta:
```
Feature: Quick

  @admin
  Scenario: Quick test
    Given I store the masters in the :masters clipboard
    And the Internal IP of node "<%= cb.masters[0].name %>" is stored in the :master_ip clipboard
    Given the node's default gateway is stored in the :def_gw clipboard
    Given the default interface on nodes is stored in the :def_intf clipboard
```

anuragthehatter:
Thanks!

/LGTM

features/step_definitions/networking.rb
```diff
@@ -915,15 +917,17 @@
...
+    # use -4 to limit output to just `default` interface, fixed in later iproute2 versions
+    step %Q/I run command on the node's ovnkube pod:/, table("| ip | -4 | route | show | default |")
   when "OpenShiftSDN"
-    step %Q/I run command on the node's sdn pod:/, table("| bash | -c | ip route show default |")
+    step %Q/I run command on the node's sdn pod:/, table("| ip | -4 | route | show | default |")
   else
     raise "unknown networkType"
   end
-  def_inf = @result[:response].split("\n").first.split(/\W+/)[7]
+  # OVN uses `br-ex` and `-` is not a word char, so we have to split on whitespace
```
akostadinov, line 927: I think these comment lines are not needed.

rbbratta: not having suitable extrema documented is partly the reason the issue arose, but this could be re-worded, will try to address in #1252

PR #1428, opened 2020-08-11
automated 2 new test cases

lihongyan1:
Depend on PR:
https://github.com/openshift/cucushift/pull/8037
V3 Job:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/130925/console
Automated 2 cases
OCP-33446 - Node CPU stats should be accurate
OCP-33244 - [BZ 1846805] kubelet_running_pod_count shouldn't take into account completed pods

### DIFF
--- a/lib/rules/cli/4.1.yaml
+++ b/lib/rules/cli/4.1.yaml
@@ -1405,8 +1405,8 @@
     :etcd_version: !ruby/regexp '/etcd (.+)$/'
 :oadm_top_node:
   :cmd: oc adm top node
-:oadm_top_specific_node:
-  :cmd: oc adm top node <node_name>
+  :options:
+    :node_name: <value>
 :registry_info:
   :cmd: oc registry info
   :options:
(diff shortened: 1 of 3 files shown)


lib/rules/cli/4.1.yaml
```diff
@@ -1405,6 +1405,8 @@
     :etcd_version: !ruby/regexp '/etcd (.+)$/'
 :oadm_top_node:
   :cmd: oc adm top node
+:oadm_top_specific_node:
```
juzhao, line 1408: see from
```
# oc adm top node --help
Examples:
  # Show metrics for all nodes
  oc adm top node
  
  # Show metrics for a given node
  oc adm top node NODE_NAME
```
I think we can combine oadm_top_node and oadm_top_specific_node as 
```
:oadm_top_node:
  :cmd: oc adm top node
  :options:
    :node_name: <value>
```

PR #1434, opened 2020-08-12
Fix OCP-23501

duanwei33:
In OCP 4.6, we cannot get the default storage class from co storage, before a better solution, I'd like to update it to unblock all the failure upgrade to 4.6, now it still work for default sc check.

[pass log](https://privatebin-it-iso.int.open.paas.redhat.com/?bb2482b18913c241#3G6JxYKji5H1DV/WTnyUP3sPTU5qQ6CR7HzQ2c4PR2w=)

@qinpingli @chao007 Please help review.

### DIFF
--- a/features/upgrade/storage/upgrade.feature
+++ b/features/upgrade/storage/upgrade.feature
@@ -13,33 +13,34 @@ Feature: Storage upgrade tests
     Given the expression should be true> cluster_operator('storage').condition(type: 'Upgradeable')['status'] == "True"
 
     # There should be one and only one default storage class
-    When I run the :get client command with:
-      | resource      | clusteroperator                                                         |
-      | resource_name | storage                                                                 |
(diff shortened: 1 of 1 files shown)


liangxia:
> In OCP 4.6, we cannot get the default storage class from co storage

@duanwei33  Is there a bug or issue tracking that ?

duanwei33:
> > In OCP 4.6, we cannot get the default storage class from co storage
> 
> @duanwei33 Is there a bug or issue tracking that ?

Good catch, I just confirmed with our dev and we think it is a regression, but not sure when they will fix it with the new cluster-storage-operator...  
https://bugzilla.redhat.com/show_bug.cgi?id=1868424

liangxia:
@duanwei33    How about comment out those lines ?  And with some explanation and bug reference in comments ?  Then we can easily recover them back when the bug is fixed.

liangxia:
/lgtm
Merging to unblock upgrade testing.

PR #1435, opened 2020-08-12
Allow operations without teardown to support upgrade

liangxia:
/cc @akostadinov @pruan-rht 

We need some operations to persistent in pre-upgrade, E.g, adding scc to user/service account.

### DIFF
--- a/features/step_definitions/scc.rb
+++ b/features/step_definitions/scc.rb
@@ -62,7 +62,7 @@ Given /^scc policy #{QUOTED} is restored after scenario$/ do |policy|
   }
 end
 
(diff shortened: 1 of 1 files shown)


features/step_definitions/scc.rb
```diff
@@ -105,7 +105,7 @@
     @result = _admin.cli_exec(_command, **_opts)
     @result[:success]
   }
-  if @result[:success]
+  if @result[:success] && !no_teardown
```
akostadinov, line 108: @liangxia , could you make sure scenario is tagged with `@ugrade_prepare` in case no teardown is selected?

liangxia: @akostadinov  https://github.com/openshift/verification-tests/pull/1444

PR #1436, opened 2020-08-13
add pod method "reviseVersion"

lyman9966:
fix: add pod method "reviseVersion" which can return reviseVersion attribute to check if a pod has been redeployed.
This fix will support case OCP-33816 and OCP-33818


lyman9966:
@jhou1 @weinliu @sunilcio Can you help to review? Thanks

PR #1438, opened 2020-08-13
fix upgrade failure for metric name change

lihongyan1:
Fix upgrade failure by updating metric name which is applicable to 4.4, 4.5 and 4.6
https://issues.redhat.com/browse/OCPQE-1847


lihongyan1:
@juzhao @pruan-rht Please help reivew.

PR #1444, opened 2020-08-17
No teardown should apply to upgrade-prepare only

liangxia:
/cc @akostadinov @pruan-rht 

Address https://github.com/openshift/verification-tests/pull/1435/files#r470642284

### DIFF
--- a/features/step_definitions/scc.rb
+++ b/features/step_definitions/scc.rb
@@ -64,6 +64,9 @@ end
 
 Given /^SCC #{QUOTED} is (added to|removed from) the #{QUOTED} (user|group|service account)( without teardown)?$/ do |scc, op, which, type, no_teardown|
   ensure_admin_tagged
+  if no_teardown
+    ensure_upgrade_prepare_tagged
+  end
   _admin = admin
 
   case type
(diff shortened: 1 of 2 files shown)


features/step_definitions/scc.rb
```diff
@@ -64,6 +64,9 @@
 
 Given /^SCC #{QUOTED} is (added to|removed from) the #{QUOTED} (user|group|service account)( without teardown)?$/ do |scc, op, which, type, no_teardown|
   ensure_admin_tagged
+  if no_teardown
+    ensure_upgrade_prepare_tagged
```
akostadinov, line 68: I think method should be called `ensure_upgrade_tagged` and also allow the upgrade check tag. In both - upgrade check and upgrade prepare we do not want clean-up. Otherwise after one attempt of check, the data from upgrade prepare will be gone.

In this case though things may make sense to be left as you did them. Because within upgrade check we shouldn't change SCC.

WDYT? Perhaps lets leave as you have it and wi will see if upgrade-check needs to be added in the future.

liangxia: @akostadinov   Add in https://github.com/openshift/verification-tests/pull/1448

PR #1446, opened 2020-08-17
pod label changed to 'deployment' from 'deploymentconfig'

pruan-rht:


xiuwang:
@pruan-rht  the label 'deploymentconfig' changeing to 'deployment'  is since 4.4+ , then your change won't work below 4.4.
Could you change to use other pod label as selector?

features/cli/pod.feature
```diff
@@ -201,15 +201,15 @@ Feature: pods related scenarios
   Scenario: 4.0 Oauth provider info should be consumed in a pod
     Given I have a project
     When I run the :new_app client command with:
-      | docker_image     | aosqe/ruby-ex        |
+      | docker_image | aosqe/ruby-ex |
```
liangxia, line 204: We can use `-l/--labels` when new-app.
```
-l, --labels='': Label to set in all resources for this application.
```

pruan-rht:
@xiuwang @liangxia thanks for the suggestion...PR updated.  PTAL again. thx!

PR #1450, opened 2020-08-18
some fix and new rules

yapei:
@xiaocwan  Ready for your review


yapei:
@liangxia  @pruan-rht  Can you help review and merge if it looks good?

PR #1460, opened 2020-08-24
Properly wait for mcp to roll out

jhou1:
This fixes a frequently failed test. The test fails because we were only fetching the mcp's updating condition, but there could be a chance that the mcp has not started to update and returns 'Updating: false' to the client. With this fix we strictly verifies mcp has successfully rolled up new machineconfig before having the next verification step.

@sunilcio @weinliu @lyman9966 PTAL.

```
1 scenario (1 passed)
15 steps (15 passed)
9m2.542s
```

### DIFF
--- a/features/node/seccomp.feature
+++ b/features/node/seccomp.feature
@@ -29,11 +29,11 @@ Feature: Secure Computing Test Scenarios
     Then the step should succeed
     And admin ensures "custom-seccomp" machineconfig is deleted after scenario
 
-    # Wait for machineconfigpool to process the new config
-    Given I wait up to 600 seconds for the steps to pass:
+    # Wait for machineconfigpool to roll out the new machineconfig
+    Given I wait up to 900 seconds for the steps to pass:
     """
-    Given evaluation of `machine_config_pool('worker').condition(type: 'Updating', cached: false)` is stored in the :mcp clipboard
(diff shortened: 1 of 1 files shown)


akostadinov:
raw_resource usage is an anti-pattern. But implementing in a better way needs some good work. If you think this construct will be needed in other places, I can take a look. Otherwise we can leave as is. 

/lgtm

features/node/seccomp.feature
```diff
@@ -29,11 +29,11 @@ Feature: Secure Computing Test Scenarios
...
     And admin ensures "custom-seccomp" machineconfig is deleted after scenario
 
-    # Wait for machineconfigpool to process the new config
-    Given I wait up to 600 seconds for the steps to pass:
+    # Wait for machineconfigpool to roll out the new machineconfig
+    Given I wait up to 900 seconds for the steps to pass:
     """
-    Given evaluation of `machine_config_pool('worker').condition(type: 'Updating', cached: false)` is stored in the :mcp clipboard
-    Then the expression should be true> cb.mcp["status"] == "False"
+    Then the expression should be true> machine_config_pool('worker').raw_resource(cached: false).dig('status', 'configuration', 'source').select { |c| c['name'] == 'custom-seccomp' }.empty? == false
```
pruan-rht, line 35: @jhou1 perhaps put this into a method within machine_config_pool?

PR #1462, opened 2020-08-24
Auto-SCTP

weliang1:
Update one old case and automate three new cases.

Fixed the intermittent failure by using  "--sctp -k -m 5 -w 60 " which will let sctpserver pod to accept
5 connection at same time and keep port open until 60 seconds. 

Test log: http://file.rdu.redhat.com/~weliang/sctp.txt

### DIFF
--- a/features/networking/sctp.feature
+++ b/features/networking/sctp.feature
@@ -44,12 +44,12 @@ Feature: SCTP related scenarios
     And I wait up to 60 seconds for the steps to pass:
     """"
     When I run the :exec client command with:
-      | pod              | sctpclient                                                               |
-      | namespace        | <%= project.name %>                                                      |
-      | oc_opts_end      |                                                                          |
-      | exec_command     | sh                                                                       |
-      | exec_command_arg | -c                                                                       |
-      | exec_command_arg | /usr/bin/nc -v <%= cb.serverpod_ip %> 30102 --sctp  <<< 'test-openshift' |
+      | pod              | sctpclient                                                                |
+      | namespace        | <%= project.name %>                                                       |
+      | oc_opts_end      |                                                                           |
+      | exec_command     | sh                                                                        |
+      | exec_command_arg | -c                                                                        |
+      | exec_command_arg | echo test-openshift \| /usr/bin/nc -v <%= cb.serverpod_ip %> 30102 --sctp |
     Then the step should succeed
     And the output should contain:
       | Connected to <%= cb.serverpod_ip %>:30102 |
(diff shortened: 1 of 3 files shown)


features/networking/sctp.feature
```diff
@@ -33,23 +32,257 @@ Feature: SCTP related scenarios
...
-      | oc_opts_end      |                                                                  |
-      | exec_command     | sh                                                               |
-      | exec_command_arg | -c                                                               |
-      | exec_command_arg | echo test-openshift \| nc -v <%= cb.serverpod_ip %> 30102 --sctp | 
+      | pod              | sctpclient                                                               |
+      | namespace        | <%= project.name %>                                                      |
+      | oc_opts_end      |                                                                          |
+      | exec_command     | sh                                                                       |
+      | exec_command_arg | -c                                                                       |
+      | exec_command_arg | /usr/bin/nc -v <%= cb.serverpod_ip %> 30102 --sctp  <<< 'test-openshift' |
```
rbbratta, line 52: `<<<` is only valid in `bash` not `sh`.   The fact that `sh` is symlinked to `bash` should be not always be assumed as on Debian-based systems `sh` is symlinked to `dash`.   

If we are going to use bash-isms we should be clear and use `bash -c` specifically.

We should probably always use `bash -c`, if we need to use a shell, to ensure we have all the cool bash features.

rbbratta: also it looks like `<<<` here-strings are implemented with tmp files in case that matters vs. pipes.

```shell
$ ls -l /proc/self/fd/ <<< "TEST"
total 0
lr-x------. 1 rbrattai rbrattai 64 Aug 25 00:01 0 -> '/tmp/sh-thd.H2fCpL (deleted)'
lrwx------. 1 rbrattai rbrattai 64 Aug 25 00:01 1 -> /dev/pts/14
lrwx------. 1 rbrattai rbrattai 64 Aug 25 00:01 2 -> /dev/pts/14
```
https://askubuntu.com/a/678919

features/networking/sctp.feature
```diff
@@ -33,23 +32,257 @@ Feature: SCTP related scenarios
...
+    Then the step should succeed
+    And the output should contain:
+      | Connected to <%= cb.serverpod_ip %> |
+      | 15 bytes sent                       |
+    """"
```
rbbratta, line 288: do we prefer ending newlines?

features/networking/sctp.feature
```diff
@@ -33,23 +32,257 @@ Feature: SCTP related scenarios
...
+    And evaluation of `service.ip(user: user)` is stored in the :service_ip clipboard
+
+    # sctpserver pod start to wait for sctp traffic
+    When I run the :exec background client command with:
+      | pod              | sctpserver                                |
+      | namespace        | <%= project.name %>                       |
+      | oc_opts_end      |                                           |
+      | exec_command     | sh                                        |
+      | exec_command_arg | -c                                        |
+      | exec_command_arg | /usr/bin/nc -l 30102 --sctp -k -m 5 -w 60 |
```
rbbratta, line 102: can we use `service.port(user: user)` so we don't have to hard-code the port?

features/networking/sctp.feature
weliang1: Fixed by using "echo test-openshift \| /usr/bin/nc -v <%= cb.serverpod_ip %> 30102 --sctp", thanks!

features/networking/sctp.feature
weliang1: Fixed, thanks!

features/networking/sctp.feature
weliang1: @rbbratta  30102 is SCTP port not service port, and it is statically defined in the pod yaml file, so it's OK by using the hard-code sctp port number. Thanks!

anuragthehatter:
@weliang1 Do you think we should check their repeatability to make sure? Like running group of all tests 5 times in a row? Thats might give some idea if they tend to stuck somewhere. Juts a thought. If you have already done that. Pls ignore.

weliang1:
> @weliang1 Do you think we should check their repeatability to make sure? Like running group of all tests 5 times in a row? Thats might give some idea if they tend to stuck somewhere. Juts a thought. If you have already done that. Pls ignore.

@anuragthehatter Yes, did that already. Thanks!

PR #1464, opened 2020-08-25
some fix and new rules

yapei:
@pruan-rht  Can you help merge?


yapei:
@pruan-rht  Could you help merge?

PR #1465, opened 2020-08-25
obtain test data: refactor and add missing dest dir option

rbbratta:
In some tests we may need to copy and modify the same file twice.
To support tnis add a dest_dir option

```
    Given I create the "policy2" directory
    When I obtain test data file "networking/egressnetworkpolicy/limit_policy.json" into the "policy2" dir
```

cp_r can handle directories if we append `/.` to copy the contents

prevent relative path escapes usind `../../`

pre-create the dest_dir using `mkdir_p` to prevent errors whem EEXIST

### DIFF
--- a/features/step_definitions/file.rb
+++ b/features/step_definitions/file.rb
@@ -1,20 +1,29 @@
-Given /^I obtain test data (file|dir) #{QUOTED}$/ do |type, path|
-  accepted_roots = [
-    "#{BushSlicer::HOME}/testdata",
-    "#{BushSlicer::HOME}/features/tierN/testdata",
-  ]
+Given /^I obtain test data (file|dir) #{QUOTED}(?: into the #{QUOTED} dir)?$/ do |type, path, dest_dir|
+  if path.include?("../") || (dest_dir && dest_dir.include?("../"))
+    # prevent escape, e.g. "../../../../../../../../../etc/passwd"
+    raise "invalid relative path access"
+  end
+  accepted_roots = %W[#{BushSlicer::HOME}/testdata #{BushSlicer::HOME}/features/tierN/testdata]
   tried_paths = []
   found = accepted_roots.any? do |root|
-    test_file = "#{root}/#{path}"
-    tried_paths << test_file
-    if File.exist? test_file
-      if type == "dir"
-        FileUtils.cp_r(test_file, File.basename(path))
-        cb.test_file = File.absolute_path(File.basename(path))
-      else 
-        FileUtils.cp(test_file, File.basename(path))
-        cb.test_file = File.absolute_path(File.basename(path))
+    src = "#{root}/#{path}"
+    tried_paths << src
+    if File.exist? src
+      dest = File.basename(path)
+      if dest_dir
+        # cp_r does not do a mkdir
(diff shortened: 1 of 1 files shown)


rbbratta:
tests

```
Feature: Obtain tests

  @admin
  @destructive
  Scenario: test obtain
    Given I obtain test data file "networking/egressnetworkpolicy/limit_policy.json"
    Given I obtain test data file "networking/egressnetworkpolicy/limit_policy.json" into the "policy2" dir
    Given I obtain test data file "networking/egressnetworkpolicy/limit_policy.json" into the "policy2" dir
    # this works because dest egressnetworkpolicy doesn't exist yet so it isn't a dir
    Given I obtain test data file "networking/egressnetworkpolicy"
    Given I obtain test data file "networking/egressnetworkpolicy"
    Given I obtain test data file "networking/egressnetworkpolicy" into the "policy3" dir
    Given I obtain test data dir "networking/egressnetworkpolicy"
    Given I obtain test data dir "networking/egressnetworkpolicy" into the "policy3" dir
    # empty string doesn't match the regexp
#    Given I obtain test data file ""
#    Given I obtain test data file "" into the "policy4" dir
    # could not find test file in
#    Given I obtain test data file "  "
#    Given I obtain test data file "  " into the "policy4" dir
#    Given I obtain test data dir "/tmp" into the "policy3" dir
    # invalid relatve path access
#    Given I obtain test data file "../../../../../../../../../etc/passwd"
    Given I obtain test data file "networking/egressnetworkpolicy" into the "../../../../../../tmp" dir
```

features/step_definitions/file.rb
```diff
@@ -1,20 +1,29 @@
...
+    tried_paths << src
+    if File.exist? src
+      dest = File.basename(path)
+      if dest_dir
+        # cp_r does not do a mkdir
+        # mkdir_p to prevent errors when EEXIST
+        FileUtils.mkdir_p(dest_dir)
+        dest = File.join(dest_dir, dest)
+      end
+      if type == "dir" or File.directory?(src)
```
akostadinov, line 19: I would like people to be careful and properly use `dir` vs `file` in step name. That way if we change location of test data files, we can more easily handle these usages.

So please remove  `or File.directory?(src)`

features/step_definitions/file.rb
```diff
@@ -1,20 +1,29 @@
...
+        # copy the contents not the dir by appending '/.'
+        # without this if the dest dir already exists (e.g. if we run the step twice)
+        # cp_r('src', 'dest') will copy src/x -> dest/src/x when instead we want src/x -> dest/x
+        src += "/."
       end
+      FileUtils.cp_r(src, dest)
```
akostadinov, line 25: please use `cp` when doing a file and `cp_r` when doing a dir. Otherwise people may be flaky with `file` vs `dir`  usage in step name.

akostadinov:
I totally missed that. Looks good except for two comments. Thanks!

akostadinov:
I ported and addressed my concerns in #1786
I left out the `..` check because there are many ways if somebody wants to access lower paths. And giving false sense of security is not good IMO. But if I missed your point with the check, please let me know. Thanks!

PR #1468, opened 2020-08-25
Add machineconfig method to machine_config_pool

jhou1:
As a follow up of https://github.com/openshift/verification-tests/pull/1460#discussion_r475857294, remove the direct usage of `raw_resource`. 

@akostadinov @pruan-rht PTAL

```
      [05:50:46] INFO> === End After Scenario: Using Secure Computing Profiles with Pod Annotations ===

1 scenario (1 passed)
14 steps (14 passed)
9m51.645s
```

### DIFF
--- a/lib/openshift/machine_config_pool.rb
+++ b/lib/openshift/machine_config_pool.rb
@@ -8,5 +8,10 @@ module BushSlicer
       rr = raw_resource(user: user, cached: cached, quiet: quiet).dig("spec")
       MachineConfigPoolSpec.new rr
     end
+
+    def machineconfig(mc_name, user: nil, cached: true, quiet: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'configuration', 'source')
+      rr.select { |c| c["name"] == mc_name }
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/openshift/machine_config_pool.rb
```diff
@@ -8,5 +8,10 @@ def spec(user: nil, cached: true, quiet: false)
       rr = raw_resource(user: user, cached: cached, quiet: quiet).dig("spec")
       MachineConfigPoolSpec.new rr
     end
+
+    def machineconfig(mc_name, user: nil, cached: true, quiet: false)
```
pruan-rht, line 12: @jhou1 since this is a `status` property, I'm wondering if we should just default to `cached: false`...that way you don't need to specify that when calling it.  WDYT @akostadinov ?  LGTM otherwise.

akostadinov: We should always default to `cached: true` because most common use cases benefit with performance from this. When we loop-wait people should get used to doing `cached: false`. It is simpler convention than trying to think what would be more common for each individual method and then never know what is the default.

akostadinov, line 12: I didn't want to request such approach in the other PR because machine config source is a complex object and we better have a special class for it. Let me have a look tomorrow.

jhou1:
This has become stale, closing

PR #1469, opened 2020-08-25
fix issue: fluentd is not ready

QiaolingTang:
Fix the below issue and remove collector type `rsyslog` since we never support it.
```
fluentd did not become ready within 300 seconds
RuntimeError
/home/jenkins/workspace/Runner-v3/lib/openshift/cluster_logging.rb:128:in `wait_until_fluentd_is_ready'
/home/jenkins/workspace/Runner-v3/features/step_definitions/logging.rb:160:in `/^I wait for clusterlogging(?: named "(.+)")? with "(.+?)" log collector to be functional in the(?: "(.+?)")? project$/'
features/logging/elasticsearch.feature:47:in `Given I wait for clusterlogging with "fluentd" log collector to be functional in the project'
```

Log: https://url.corp.redhat.com/3153b7e

@anpingli @pruan-rht PTAL, thanks.

### DIFF
--- a/features/logging/elasticsearch.feature
+++ b/features/logging/elasticsearch.feature
@@ -44,7 +44,10 @@ Feature: Elasticsearch related tests
     And the expression should be true> cluster_logging('instance').logstore_storage_class_name == cb.default_sc.name
     Given I wait for the "elasticsearch" elasticsearches to appear
     And the expression should be true> elasticsearch('elasticsearch').nodes[0]['storage']['storageClassName'] == cb.default_sc.name
-    Given I wait for clusterlogging with "fluentd" log collector to be functional in the project
+    #Given I wait for clusterlogging with "fluentd" log collector to be functional in the project
+    Given I wait until ES cluster is ready
+    And I wait until "fluentd" log collector is ready
+    And I wait until kibana is ready
     Given I wait up to 300 seconds for the steps to pass:
     """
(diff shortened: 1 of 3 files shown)


features/logging/elasticsearch.feature
```diff
@@ -44,7 +44,10 @@ Feature: Elasticsearch related tests
     And the expression should be true> cluster_logging('instance').logstore_storage_class_name == cb.default_sc.name
     Given I wait for the "elasticsearch" elasticsearches to appear
     And the expression should be true> elasticsearch('elasticsearch').nodes[0]['storage']['storageClassName'] == cb.default_sc.name
-    Given I wait for clusterlogging with "fluentd" log collector to be functional in the project
+    #Given I wait for clusterlogging with "fluentd" log collector to be functional in the project
```
QiaolingTang, line 47: EO is unable to update the Kibana status(https://bugzilla.redhat.com/show_bug.cgi?id=1849564), so I use the below 3 steps to replace the step `I wait for clusterlogging with "fluentd" log collector to be functional in the project`.

pruan-rht: @QiaolingTang once the bug is fixed then we can use the `old` approach, right?  If it works, then I don't recommend removing the various functions.

QiaolingTang: Yes, this step should work once the bug is fixed,  but I'm not removing the functions related to this step, I just want to remove some useless functions, we never used them.

PR #1473, opened 2020-08-25
Quick fix for conntrack usecase

anuragthehatter:
```Yes``` utility doesn't seem to be smooth while sending traffic to udp sockets every time and this use case suffered lot of failures. Moving back to old school simple logic:) Killing background process is also not required now as that also stuck a lot. Making that logic smooth. Not a big change.

This should be merged soon. Thanks

Logs: https://privatebin-it-iso.int.open.paas.redhat.com/?7c2c573bdb1dbe7f#l5/KX92PIvwetbvBvPUAqMHF0eHnHDAgkAbT3mo+1Nk=

@rbbratta @weliang1 @zhaozhanqi @huiran0826

### CODE REVIEW

File: features/networking/pod.feature [hunk 1]
```diff
@@ -309,16 +309,14 @@ Feature: Pod related networking scenarios
     Given a pod becomes ready with labels:
       | name=hello-pod |
     And evaluation of `pod` is stored in the :client_pod clipboard
-    # 'yes' command will send a character "h" continously for 3 seconds to /dev/udp on listener where the node is listening for udp traffic on exposed nodeport. The 3 seconds mechanism will create an Assured
```
weliang1, line 312: @anuragthehatter I thought your new version should remove line 312 about 'yes' command, but your testing log still has this comment, it's not a big deal, just want to make sure the testing log you post is correct one.

anuragthehatter: Yeah that was taken care before your comment. Thanks for concern though.

PR #1476, opened 2020-08-27
rules fixed for OCP-27586

yapei:


yapei:
@liangxia @pruan-rht  Can you help merge at your preference? Thanks

PR #1477, opened 2020-08-28
update for OCP-11326

zhouying7780:
@kasturinarra Could you help review ? thanks.

### DIFF
--- a/testdata/deployment/extensions/deployment.yaml
+++ b/testdata/deployment/extensions/deployment.yaml
@@ -1,8 +1,11 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: hello-openshift
 spec:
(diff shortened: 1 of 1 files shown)


testdata/deployment/extensions/deployment.yaml
```diff
@@ -1,8 +1,11 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
```
kasturinarra, line 1: @zhouying7780 could you  please help provide a description on why is this change? Thanks !!

zhouying7780: @kasturinarra  The older api version has discarded , when use it will hit error: no matches for kind "Deployment" in version "extensions/v1beta1"

pruan-rht (commented):
/lgtm

PR #1478, opened 2020-08-28
use Ruby system libray instead of pure linux commands

pruan-rht:
@akostadinov PTAL, the reason for this PR is that currently it's checking for the string `No such`, @lwan-wanglin ran into a situation in which her system language is Chinese, so the output string was not matched and the `delete` function returned the wrong status.  Using the ruby system library should solve this kind of corner case.

### DIFF
--- a/lib/host.rb
+++ b/lib/host.rb
@@ -511,10 +511,9 @@ module BushSlicer
         if bad_files.include? file
           raise "should not remove file named '#{file}'"
         end
-
-        r = "-r"
+        meth = :rm
       else
-        r = ""
+        meth = :remove
       end
       file = shell_escape file
       if opts[:home] && ! file.start_with?('/','\\')
@@ -522,20 +521,14 @@ module BushSlicer
         # relies strongly on bad_files checking above
         file = '~/' + file
       end
-      if opts[:raw]
-        exec_method = :exec_raw
-      elsif opts[:admin]
-        exec_method = :exec_admin
+      if File.file? file
+        FileUtils.public_send(method, file, force: true)
+        check_cmd = File.file? file
       else
-        exec_method = :exec
+        FileUtils.public_send(:remove_dir, file, force: true)
+        check_cmd = File.directory? file
       end
-      public_send(exec_method, "rm #{r} -f -- #{file}", **opts)
-      opts[:quiet] = true
-      res = public_send(exec_method, "ls -d -- #{file}", **opts)
-
-      # OCDebugAccessibleHost does not return exit status of executed command
-      # return ! res[:success]
-      return res[:response].include? "No such"
+      return !check_cmd
     end
 
     # wait until one file in the list is found and returns its name


lib/host.rb
```diff
@@ -511,31 +511,24 @@ def delete(file, opts={})
...
       else
-        exec_method = :exec
+        FileUtils.public_send(:remove_dir, file, force: true)
+        check_cmd = File.directory? file
       end
-      public_send(exec_method, "rm #{r} -f -- #{file}", **opts)
-      opts[:quiet] = true
-      res = public_send(exec_method, "ls -d -- #{file}", **opts)
-
-      # OCDebugAccessibleHost does not return exit status of executed command
```
rbbratta, line 536: we still need something in OCDebugAccessibleHost to handle cleaning up in the chroot.

```
self = {BushSlicer::OCDebugAccessibleHost} root@ip-10-0-187-176
bad_files = Array (8 elements)
exec_method = :exec_raw
file = "/tmp/workdir/rbrattai-rbrattai"
opts = Hash (2 elements)
 :r => true
 :raw => true
r = "-r"
res = nil
```

It looks like we can use Ruby methods for `LocalLinuxLikeHost` but we have to use shell for `SSHAccessibleHost` and  `OCDebugAccessibleHost`.

Maybe we can move methods around between the classes.

rbbratta: OCDebugAccessibleHost test case

```
  @admin
  @destructive
  Scenario: net.ipv4.ip_forward should be always enabled on node service startup
    Given I select a random node's host

      | sysctl -w net.ipv4.ip_forward=0 |
    Then the step should succeed
```

pruan-rht: @rbbratta does that testcase call the `delete` methods?  I just ran it and it passed.

lib/host.rb
```diff
@@ -511,31 +511,24 @@ def delete(file, opts={})
...
         # relative to host local user home directory
         # relies strongly on bad_files checking above
         file = '~/' + file
       end
-      if opts[:raw]
-        exec_method = :exec_raw
-      elsif opts[:admin]
-        exec_method = :exec_admin
+      if File.file? file
+        FileUtils.public_send(method, file, force: true)
```
rbbratta, line 525: `s/method/meth/`

lib/host.rb
```diff
@@ -511,31 +511,24 @@ def delete(file, opts={})
...
           raise "should not remove file named '#{file}'"
         end
-
-        r = "-r"
+        meth = :rm
       else
-        r = ""
+        meth = :remove
       end
       file = shell_escape file
```
rbbratta, line 518: we shouldn't need to shell escape when using `Fileutils.remove_dir`

lib/host.rb
```diff
@@ -511,31 +511,24 @@ def delete(file, opts={})
...
-      if opts[:raw]
-        exec_method = :exec_raw
-      elsif opts[:admin]
-        exec_method = :exec_admin
+      if File.file? file
+        FileUtils.public_send(method, file, force: true)
+        check_cmd = File.file? file
       else
-        exec_method = :exec
+        FileUtils.public_send(:remove_dir, file, force: true)
```
rbbratta, line 528: need to check `File.dir?` here otherwise if the path doesn't exist at all the `force: true` will eat the error and we won't know that the path is neither file nor dir because it doesn't exist.

lib/host.rb
```diff
@@ -511,31 +511,24 @@ def delete(file, opts={})
         if bad_files.include? file
           raise "should not remove file named '#{file}'"
         end
-
-        r = "-r"
+        meth = :rm
       else
-        r = ""
+        meth = :remove
```
rbbratta, line 516: `:rm` and `:remove` are aliases in `Fileutils`  

```ruby
  alias remove rm
  module_function :remove
```

lib/host.rb
rbbratta: yes, the original call trace when using `When I run commands on the host:` is

```
delete [host.rb:532] (BushSlicer::LinuxLikeHost)
user_opts [cli_executor.rb:287] (BushSlicer::SharedLocalCliExecutor)
exec [cli_executor.rb:303] (BushSlicer::SharedLocalCliExecutor)
cli_exec [api_accessor.rb:71] (BushSlicer::APIAccessor)
get_matching [cluster_resource.rb:162] (BushSlicer::ClusterResource)
nodes [environment.rb:412] (BushSlicer::OpenShiftEnvironment)
node.rb:29

```

rbbratta: it passes because `if File.file? file`  fails because the file is on the remote `oc debug` host so it calls `FileUtils.public_send(:remove_dir, file, force: true)` instead, and the `force: true` eats the error

lib/host.rb
pruan-rht: @rbbratta the point of `delete` is to remove the target, if the target doesn't exists already, then we don't really care if there's error or not... the ultimate goal is `target` resource is not there.

lib/host.rb
akostadinov: you should add method under `module LocalHost`. Then any local host class should pick it up by including the module. Presently you are changing the generic linux like host and it can be a remote host.

lib/host.rb
akostadinov: we need to follow current convention. Honestly I didn't check what it is. But we just keep existing behavior.

lib/host.rb
rbbratta: if we strip out the Linux `~/` stuff, otherwise under LocalLinuxHost

lib/host.rb
```diff
@@ -511,31 +511,24 @@ def delete(file, opts={})
         if bad_files.include? file
           raise "should not remove file named '#{file}'"
         end
-
-        r = "-r"
+        meth = :rm
```
rbbratta, line 514: It looks like `rm_r` is recursive

```ruby
        meth = :rm_r
```

rbbratta: or it looks like there is a method already

https://rubydoc.info/stdlib/fileutils/FileUtils#remove_entry_secure-class_method

This method removes a file system entry path. path shall be a regular file, a directory, or something. If path is a directory, remove it recursively. This method is required to avoid TOCTTOU (time-of-check-to-time-of-use) local security vulnerability of rm_r. #rm_r causes security hole when:

Parent directory is world writable (including /tmp).

Removing directory tree includes world writable directory.

The system has symbolic link.

lib/host.rb
rbbratta: the system library has it all  `:remove_entry` or `:remove_entry_secure` will remove anything. :smiley: 

https://rubydoc.info/stdlib/fileutils/FileUtils#remove_entry_secure-class_method

PR #1481, opened 2020-09-01
Wait for CRD to be running first before createing meteringconfig and do proper cleanup

pruan-rht:
@akostadinov @liangxia @yapei  PTAL.

### DIFF
--- a/features/step_definitions/crd.rb
+++ b/features/step_definitions/crd.rb
@@ -29,18 +29,3 @@ Given /^the #{QUOTED} #{QUOTED} CR is restored after scenario$/ do |name, crd|
     raise "Cannot restore #{crd}: #{name}" unless @result[:success]
   }
 end
-
-Given /^I remove all CRDs with labels:$/ do | table |
-  ensure_admin_tagged
-  ensure_destructive_tagged
-  labels = table.raw.flatten
-  timeout = 60
-  res = BushSlicer::CustomResourceDefinition.wait_for_labeled(*labels, user: user, seconds: timeout)
-  if res[:success]
-    crd_names = res[:parsed]['items'].map { |r| r['metadata']['name'] }
-    crd_names.each do | crd |
-      step %Q/I ensure "#{crd}" custom_resource_definition is deleted/
-    end
-  end
-end
-
--- a/features/step_definitions/crd.rb
+++ b/features/step_definitions/crd.rb
@@ -29,3 +29,18 @@ Given /^the #{QUOTED} #{QUOTED} CR is restored after scenario$/ do |name, crd|
     raise "Cannot restore #{crd}: #{name}" unless @result[:success]
   }
 end
+
+Given /^I remove all CRDs with labels:$/ do | table |
+  ensure_admin_tagged
+  ensure_destructive_tagged
+  labels = table.raw.flatten
+  timeout = 60
+  res = BushSlicer::CustomResourceDefinition.wait_for_labeled(*labels, user: user, seconds: timeout)
+  if res[:success]
+    crd_names = res[:parsed]['items'].map { |r| r['metadata']['name'] }
+    crd_names.each do | crd |
+      step %Q/I ensure "#{crd}" custom_resource_definition is deleted/
+    end
+  end
+end
+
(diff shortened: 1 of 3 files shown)


features/step_definitions/crd.rb
```diff
@@ -29,3 +29,18 @@
...
+Given /^I remove all CRDs with labels:$/ do | table |
+  ensure_admin_tagged
+  ensure_destructive_tagged
+  labels = table.raw.flatten
+  timeout = 60
+  res = BushSlicer::CustomResourceDefinition.wait_for_labeled(*labels, user: user, seconds: timeout)
+  if res[:success]
+    crd_names = res[:parsed]['items'].map { |r| r['metadata']['name'] }
+    crd_names.each do | crd |
+      step %Q/I ensure "#{crd}" custom_resource_definition is deleted/
```
akostadinov, line 42: shouldn't this be `admin ensures ...`?

pruan-rht: since this step is tagged with `ensure_admin_tagged`, wouldn't that already implies the `I` in this case is `admin`?

pruan-rht: @akostadinov I've resolved the conflicts...anything else that needs to be modified before it can merged?

features/step_definitions/metering.rb
```diff
@@ -294,12 +296,21 @@
...
+    # need to remove CRDs as well.
+    step %Q(I remove all metering CRDs with label "operators.coreos.com/metering-ocp.openshift-metering=")
+    step %Q(I ensure "openshift-metering" meteringconfig is deleted)
     step %Q(I ensure "metering-ocp-sub" subscription is deleted)
     step %Q(I ensure "metering-ocp-og" subscription is deleted)
     step %Q/I ensure "#{metering_ns}" project is deleted/
   end
 end
 
+Given /^I remove all metering CRDs with label #{QUOTED}$/ do | label |
```
akostadinov, line 308: This seems to copy the other step. It implies it selects only metering CRDs. But in fact it has no special extra logic. IMO we should remove this step.

features/step_definitions/crd.rb
akostadinov: > since this step is tagged with ensure_admin_tagged, wouldn't that already implies the I in this case is admin?

no, this is not the case

features/step_definitions/crd.rb
```diff
@@ -29,3 +29,18 @@
     raise "Cannot restore #{crd}: #{name}" unless @result[:success]
   }
 end
+
+Given /^I remove all CRDs with labels:$/ do | table |
```
akostadinov, line 33:
```
Given /^I remove all #{WORD} with labels:$/ do | resource, table |
```

I would like to make this step more generic so it works with any resource type.

features/step_definitions/crd.rb
akostadinov:
```
list = clazz = resource_class(resource_type)
if BushSlicer::ProjectResource > clazz
  list = clazz.list(*labels, user: user, project: project(namespace))
else
  ensure_destructive_tagged
  list = clazz.list(*labels, user: user)
end
list.each { |r| r.ensure_deleted }
```

akostadinov, line 42:
```
list = clazz = resource_class(resource_type)
if BushSlicer::ProjectResource > clazz
  list = clazz.list(*labels, user: user, project: project(namespace))
else
  ensure_destructive_tagged
  list = clazz.list(*labels, user: user)
end
list.each { |r| r.ensure_deleted }
```

akostadinov: but making step universal makes this moot

features/step_definitions/crd.rb
akostadinov: also needs to go into features/step_definitions/resources.rb

pruan-rht:
@akostadinov PR updated, PTAL.  thx!

features/step_definitions/resources.rb
```diff
@@ -221,3 +221,18 @@
     cb[cb_name] = clazz.list(user: _user)
   end
 end
+
+Given /^I remove all #{WORD}(?: in the#{OPT_QUOTED} project) with labels:$/ do | resource_type, namespace, table |
+  ensure_admin_tagged
+  ensure_destructive_tagged
```
akostadinov, line 227: 2 lines to remove, only for cluster resources are needed and I see you have there already

PR #1483, opened 2020-09-01
Hongyli new case aug17 veri

lihongyan1:
Related to PR:
https://github.com/openshift/cucushift/pull/8043

### DIFF
--- a/lib/openshift/flakes/container_spec.rb
+++ b/lib/openshift/flakes/container_spec.rb
@@ -22,6 +22,10 @@ module BushSlicer
         return struct['image']
       end
 
+      def args
+        return struct['args']
+      end
+
       def image_pull_policy
         return struct['imagePullPolicy']
       end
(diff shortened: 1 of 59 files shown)


lihongyan1:
Please help review and merge

lib/openshift/flakes/container_spec.rb
```diff
@@ -22,6 +22,10 @@ def image
         return struct['image']
       end
 
+      def args
```
liangxia, line 25: The functions under the module are sorted.

pruan-rht: @liangxia that convention is not really followed in practice at all...we do encourage the sorting to be done alphabetically in the rules file.  I'll merge this to unblock @lihongyan1

liangxia: @pruan-rht  OK. Got it.

PR #1485, opened 2020-09-01
fix deduplication of more than 256 messages where UTF is used

akostadinov:
/assign @pruan-rht

### DIFF
--- a/lib/log.rb
+++ b/lib/log.rb
@@ -189,6 +189,7 @@ module BushSlicer
         @messages = []
       end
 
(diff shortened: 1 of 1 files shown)


pruan-rht:
@akostadinov thanks!  let's just merge it and I'll test it out with my scenario.  BTW, do we really need to do a `raise` if there's a mismatch?  Since this is only for log output, I think a warning should be sufficient.

akostadinov:
@pruan-rht , in your case things raise because of missing method and this raise has no effect.

PR #1495, opened 2020-09-03
rules updated

yapei:


yapei:
@liangxia  Can you help add lgtm label if it looks good?

PR #1496, opened 2020-09-03
Add config file for cco feature

lwan-wanglin:
Added yaml file for cloud credential operator feature
@pruan-rht @wjiangjay those yaml files are used for cloudcredential.feature, could you please take a look?


pruan-rht (commented):
/lgtm

PR #1498, opened 2020-09-04
Delete new format oauth token

barleyer:
@xingxingxia @akostadinov PTAL!

This is for `OCPQE-2082`


xingxingxia:
Thanks!
/lgtm
CC'ing @rhpmali

akostadinov:
Hmm, new token deleting doesn't seem user friendly.

/lgtm

PR #1505, opened 2020-09-08
add type for credentialsrequest

lwan-wanglin:
I need to add type for credentialsrequest if I want to use the step (admin ensures "my-cred-request" credentialsrequest is deleted).


pruan-rht (commented):
/lgtm

PR #1510, opened 2020-09-09
No operatorsource from 4.6

yuhui-12:
Modify the case OCP-22618 since there is no operatorsource from OCP 4.6.

### DIFF
--- a/features/step_definitions/operators.rb
+++ b/features/step_definitions/operators.rb
@@ -40,6 +40,22 @@ Given /^I create a new CatalogSourceConfig$/ do
   end
 end
 
+Given /^I create a new OperatorSource$/ do
+  ensure_admin_tagged
+  # Create OperatorSource in 4.6-
+  if env.version_lt("4.6", user: user)
+    os_yaml ||= "#{BushSlicer::HOME}/testdata/olm/operatorsource-template.yaml"
+    step %Q/I process and create:/, table(%{
+      | f | #{os_yaml}                 |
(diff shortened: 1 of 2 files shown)


yuhui-12:
The log is pasted [here](http://pastebin.test.redhat.com/900508).

features/step_definitions/operators.rb
```diff
@@ -40,6 +40,22 @@
   end
 end
 
+Given /^I create a new OperatorSource$/ do
+  ensure_admin_tagged
+  # Create OperatorSource in 4.6-
+  if env.version_lt("4.6", user: user)
```
akostadinov, line 46: I think it makes more sense to check whether this already exist and create if not. What happens if two tests try to create?
It appears to me as a destructive test because it creates a cluster global resource. Then what cleans this resource?

features/step_definitions/operators.rb
```diff
@@ -40,6 +40,22 @@
...
+    os_yaml ||= "#{BushSlicer::HOME}/testdata/olm/operatorsource-template.yaml"
+    step %Q/I process and create:/, table(%{
+      | f | #{os_yaml}                 |
+      | p | NAME=test-operators          |
+      | p | SECRET=                      |
+      | p | DISPLAYNAME=Test Operators   |
+      | p | REGISTRY=jiazha              |
```
akostadinov, line 53: we usually don't commit resources with a personal name

PR #1514, opened 2020-09-09
auto OCP-34718

sunzhaohua2:
Auto OCP-34718: Node labels and Affinity definition in PV should match. This is for bug https://bugzilla.redhat.com/show_bug.cgi?id=1860128. @jhou1 @miyadav please help to review, thanks!
```
      [07:08:21] INFO> === End After Scenario: Node labels and Affinity definition in PV should match ===

1 scenario (1 passed)
18 steps (18 passed)
1m2.072s
```

### DIFF
--- a/features/machine/machine.feature
+++ b/features/machine/machine.feature
@@ -328,3 +328,32 @@ Feature: Machine features testing
       | attempting to acquire leader lease  openshift-machine-api/cluster-api-provider |
     """
 
+  # @author zhsun@redhat.com
+  # @case_id OCP-34718
+  @admin
+  Scenario: Node labels and Affinity definition in PV should match	
+    Given I have a project
+
+    # Create a pvc
+    Given I obtain test data file "cloud/pvc-34718.yml"
+    When I run the :create client command with:
+      | f | pvc-34718.yml |
+    Then the step should succeed
+
+    # Create a pod
+    Given I obtain test data file "cloud/pod-34718.yml"
+    When I run the :create client command with:
+      | f | pod-34718.yml |
+    Then the step should succeed
+
+    #Check node labels and affinity definition in PV are match
+    Given the pod named "task-pv-pod" becomes ready
+    And I use the "<%= pod.node_name %>" node
+    And evaluation of `node.region` is stored in the :default_region clipboard
+    And evaluation of `node.zone` is stored in the :default_zone clipboard
+    When I run the :describe admin command with:
+      | resource | pv |
+    Then the step should succeed
+    And the output should contain:
+      | failure-domain.beta.kubernetes.io/zone in [<%= cb.default_zone %>]     |
+      | failure-domain.beta.kubernetes.io/region in [<%= cb.default_region %>] |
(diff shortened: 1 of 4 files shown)


miyadav:
@sunzhaohua2  , we will have issues in private clusters while pulling nginx image ..
other than that LGTM

sunzhaohua2:
> @sunzhaohua2 , we will have issues in private clusters while pulling nginx image ..
@miyadav thanks for reminder, I have changed it to sha256 format. please help to review again. This is the test result on a disconnection env. https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/153221/console

features/machine/machine.feature
```diff
@@ -328,3 +328,37 @@ Feature: Machine features testing
...
+      | f | pvc-34718.yml |
+    Then the step should succeed
+    And admin ensures "pvc-cloud" pvc is deleted after scenario
+
+    # Create a pod
+    Given I obtain test data file "cloud/pod-34718.yml"
+    When I run the :create admin command with:
+      | f | pod-34718.yml |
+    Then the step should succeed
+    And admin ensures "task-pv-pod" pod is deleted after scenario
```
liangxia, line 352: Create PVC/pod do not need admin privilege. Also we should not use project `openshift-machine-api` here as any project seems to be working here.

testdata/cloud/pod-34718.yml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: task-pv-pod
+spec:
+  volumes:
+    - name: task-pv-storage
+      persistentVolumeClaim:
+        claimName: pvc-cloud
+  containers:
+    - name: task-pv-container
+      image: quay.io/openshifttest/nginx@sha256:3936fb3946790d711a68c58be93628e43cbca72439079e16d154b5db216b58da
```
liangxia, line 12: hello-openshift image can also work, which does not require admin privilege.

features/machine/machine.feature
```diff
@@ -328,3 +328,37 @@ Feature: Machine features testing
       | attempting to acquire leader lease  openshift-machine-api/cluster-api-provider |
     """
 
+  # @author zhsun@redhat.com
+  # @case_id OCP-34718
+  @admin
+  @destructive
```
liangxia, line 334: Is the scenario destructive ?

sunzhaohua2:
@lxia Thank you, I have updated according to your comments, please help to review again.

features/machine/machine.feature
liangxia: We do not need line 336, 343 and 350. You can test that. If you met errors, replace client with admin in line 357, **or** switch to pseudo admin before line 354

liangxia: PVC/pod are project resources, they will be removed once the project is deleted, so no need explicitly clean-up (line 343 and 350). Also creating pvc/pod does not require storage-admin privilege (line 336).

sunzhaohua2: Thank you. Updated again, help to take a look.

PR #1520, opened 2020-09-10
fix OCP-23044

yapei:
@xiaocwan  Please help review, thanks.
Jenkins pass log can be got from #148881/console


yapei:
@liangxia  @pruan-rht  Could you please help review?

PR #1521, opened 2020-09-10
Add WA for OCP-28630

chao007:
Test case OCP-29741 with cmd `oc patch scc privileged -p {} --type merged` will cause existed serviceaccount lost priviled scc.
This will impact OCP-28630.
So as a WA, adding scc back to serviceaccount csi-xxx in post-checking.

@qinpingli @duanwei33 Please help to reivew.

### DIFF
--- a/features/upgrade/storage/upgrade.feature
+++ b/features/upgrade/storage/upgrade.feature
@@ -293,6 +293,28 @@ Feature: Storage upgrade tests
 
     #Restore works
     When I use the "csihostpath" project
+    
+    #Test case OCP-29741 with cmd oc patch scc privileged -p {} --type merged will cause existed serviceaccount lost priviled scc in pre-action, so add scc back in post-action as a WA.
+
+    Given SCC "privileged" is added to the "csi-provisioner" service account 
+    Given SCC "privileged" is added to the "csi-attacher" service account
+    Given SCC "privileged" is added to the "csi-snapshotter" service account 
+    Given SCC "privileged" is added to the "csi-plugin" service account 
+
+    Given I ensures "csi-hostpath-attacher-0" pod is deleted
+    Given I ensures "csi-hostpath-provisioner-0" pod is deleted
+    Given I ensures "csi-hostpath-snapshotter-0" pod is deleted
+    Given I ensures "csi-hostpathplugin-0" pod is deleted
+
+    And I wait up to 360 seconds for the steps to pass:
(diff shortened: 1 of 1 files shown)


qinpingli:
Cause this case is used to test csi-snapshot-controller cluster operator,  the WA doesn't affect the CO,  so I'm ok with this WA, once OCP-29741 issue is fixed, let's remove the WA.

liangxia:
> Test case OCP-29741 with cmd `oc patch scc privileged -p {} --type merged` will cause existed serviceaccount lost priviled scc.

Any bug or issue tracking this ?

chao007:
> > Test case OCP-29741 with cmd `oc patch scc privileged -p {} --type merged` will cause existed serviceaccount lost priviled scc.
> 
> Any bug or issue tracking this ?

If needed, we can create a jira issue here. 
And can we update OCP-29741 that if want to patch scc , we need first to copy the existed user/accountservice.
oc patch scc privileged -p \{\"users\":\ \[\"system:admin\",\"system:serviceaccount:openshift-infra:build-controller\",\"system:serviceaccount:test-scc:test-scc\"\]\} --type=merge
Above cmd only copy the system used serviceaccount like system:admin and system:serviceaccount:openshift-infra:build-controller

@liangxia WDYT?

liangxia:
Where is the bug ?  Will customer met the same issue ?

liangxia:
At first, i thought it might be an product bug. So i would like a bug/issue to track that.
After reading the prepare scenario of OCP-29741, it's more likely to be a script issue. No issue required from me now.

We can add a get command to get existing content of the SCCs and save to clipboard, then use merge patch command with both existing content and the content we want to add to solve the issue.

barleyer:
First:
add users or groups to SCCs wont change the SCCs itself, it just creates a  clusterrolebinding.get existing content of the SCCs and `save to clipboard`  makes no effect,I can do nothing on scripts.

Second:

I just test this `oc patch` didnt remove the `clusterrolebinding` that oc adm policy add-scc-to-user/group created.So I dont think the steps under OCP-29741 will affect ur cases.The `clusterrolebinding` may lost by upgrade action or anything else.

barleyer:
In one word, I dont think OCP-29741 break this test case.

I think `post-check` contains this PR steps guarantee those **clusterolebindings** still exsit after upgrade well.

/lgtm

PR #1522, opened 2020-09-10
OCP-25832 - [web console] Check upgrade channels automate the test case.

schituku:
@yapei  PTAL.

### DIFF
--- a/lib/rules/web/admin_console/4.5/clustersettings.xyaml
+++ b/lib/rules/web/admin_console/4.5/clustersettings.xyaml
@@ -1,0 +1,26 @@
+click_update_channel_button:
+  elements:
+  - selector:
+      xpath: //button[@data-test-id='current-channel-update-link']
+    op: click
+click_channels_drop_down:
+  elements:
+  - selector:
+      xpath: //button[@data-test-id='dropdown-button']
+    op: click
+check_drop_down_items:
+  elements:
+  - selector:
+      xpath: //button[@data-test-id='dropdown-menu' and contains(., '<menu_item_1>')]
+  - selector:
+      xpath: //button[@data-test-id='dropdown-menu' and contains(., '<menu_item_2>')]
+  - selector:
+      xpath: //button[@data-test-id='dropdown-menu' and contains(., '<menu_item_3>')]
+check_list_of_channels_in_channel_modal:
+  params:
+    menu_item_1: stable-<product_version>
+    menu_item_2: fast-<product_version>
+    menu_item_3: candidate-<product_version>
+  action: click_update_channel_button
+  action: click_channels_drop_down
+  action: check_drop_down_items
--- a/lib/rules/web/admin_console/4.6/clustersettings.xyaml
+++ b/lib/rules/web/admin_console/4.6/clustersettings.xyaml
@@ -30,3 +30,17 @@ click_update_channel_button:
   - selector:
       xpath: //button[@data-test-id='current-channel-update-link']
     op: click
+click_channels_drop_down:
+  elements:
+  - selector:
+      xpath: //button[@data-test-id='dropdown-button']
+    op: click
+check_drop_down_items: {}
+check_list_of_channels_in_channel_modal:
+  params:
+    menu_item_1: stable-<product_version>
+    menu_item_2: fast-<product_version>
+    menu_item_3: candidate-<product_version>
+  action: click_update_channel_button
+  action: click_channels_drop_down
+  action: check_drop_down_items
--- a/lib/rules/web/admin_console/4.6/clustersettings.xyaml
+++ b/lib/rules/web/admin_console/4.6/clustersettings.xyaml
@@ -30,17 +30,15 @@ click_update_channel_button:
   - selector:
       xpath: //button[@data-test-id='current-channel-update-link']
     op: click
-click_channels_drop_down:
+check_drop_down_items:
   elements:
   - selector:
-      xpath: //button[@data-test-id='dropdown-button']
-    op: click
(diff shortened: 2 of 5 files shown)


lib/rules/web/admin_console/4.5/clustersettings.xyaml
```diff
@@ -0,0 +1,26 @@
...
+    op: click
+click_channels_drop_down:
+  elements:
+  - selector:
+      xpath: //button[@data-test-id='dropdown-button']
+    op: click
+check_drop_down_items:
+  elements:
+  - selector:
+      xpath: //button[@data-test-id='dropdown-menu' and contains(., '<menu_item_1>')]
```
yapei, line 14: It looks like we don't need `params`
```
//button[@data-test-id='dropdown-menu' and contains(., 'stable-<product_version>')]
```

lib/rules/web/admin_console/4.5/clustersettings.xyaml
```diff
@@ -0,0 +1,26 @@
+click_update_channel_button:
+  elements:
+  - selector:
+      xpath: //button[@data-test-id='current-channel-update-link']
+    op: click
+click_channels_drop_down:
+  elements:
+  - selector:
+      xpath: //button[@data-test-id='dropdown-button']
+    op: click
```
yapei, line 10: We have a same rule with same definition but a different name
```
click_dropdown_button:
  element:
    selector:
      xpath: //button[contains(@data-test-id, 'dropdown-button')]
    op: click
```

lib/rules/web/admin_console/4.5/clustersettings.xyaml
```diff
@@ -0,0 +1,26 @@
...
+      xpath: //button[@data-test-id='dropdown-menu' and contains(., '<menu_item_2>')]
+  - selector:
+      xpath: //button[@data-test-id='dropdown-menu' and contains(., '<menu_item_3>')]
+check_list_of_channels_in_channel_modal:
+  params:
+    menu_item_1: stable-<product_version>
+    menu_item_2: fast-<product_version>
+    menu_item_3: candidate-<product_version>
+  action: click_update_channel_button
+  action: click_channels_drop_down
```
yapei, line 25: as [above comment](https://github.com/openshift/verification-tests/pull/1522/files#r487615194), here we can use `action: click_dropdown_button` instead.

lib/rules/web/admin_console/4.5/clustersettings.xyaml
yapei: Also on OCP 4.5, we also need to check there are `stable-4.6, fast-4.6, candidate-4.6` channels as well

lib/rules/web/admin_console/4.6/clustersettings.xyaml
```diff
@@ -30,3 +30,17 @@ click_update_channel_button:
   - selector:
       xpath: //button[@data-test-id='current-channel-update-link']
     op: click
+click_channels_drop_down:
+  elements:
+  - selector:
+      xpath: //button[@data-test-id='dropdown-button']
+    op: click
```
yapei, line 37: same as above, we can use `click_dropdown_button` directly, no need to add a new rule

lib/rules/web/admin_console/4.6/clustersettings.xyaml
```diff
@@ -30,3 +30,17 @@ click_update_channel_button:
...
+check_drop_down_items: {}
```
yapei, line 38: Why `check_drop_down_items` is empty?

lib/rules/web/admin_console/4.5/clustersettings.xyaml
schituku: @yapei Updates:
- removed the params
- updated the xpath to have the params value instead.
- updated the corresponding version xyaml files 
     note the drop down values have additional version in 4.4 only. Both 4.5 and 4.6 have just one set of values.

lib/rules/web/admin_console/4.5/clustersettings.xyaml
schituku: @yapei update:
- removed the redundant rule and used the existing web rule as suggested.

lib/rules/web/admin_console/4.5/clustersettings.xyaml
schituku: @yapei update:
- removed the redundant rule and used the existing web rule as suggested.

lib/rules/web/admin_console/4.6/clustersettings.xyaml
schituku: @yapei update:
- removed the redundant rule and used the existing web rule as suggested.

lib/rules/web/admin_console/4.6/clustersettings.xyaml
schituku: @yapei update:
- corrected it. I was thinking for released versions this would be empty but now I see it will not.

lib/rules/web/admin_console/4.5/clustersettings.xyaml
yapei: @schituku 

>  note the drop down values have additional version in 4.4 only. Both 4.5 and 4.6 have just one set of values.

Yes, after 4.6 is released we will add `stable-4.6`, `fast-4.6` and `candidate-4.6` channels to 4.5 cluster. And when 4.7 is released we will add `stable-4.7`, `fast-4.7` and `candidate-4.7` channels to 4.6 cluster.

PR #1531, opened 2020-09-14
no need for deep change into symkeys

akostadinov:
/assign @jianlinliu

In the past we changed all keys in the variables files to symbols. This is only
needed on root level. A quick look doesn't show me existing usage where we need
keys to by `Symbol` type deeper in the maps.

Do you think we can merge this. There is potential for an issue if we expect symbols as keys deeper in some template.

WDYT?

### DIFF
--- a/tools/launch_instance.rb
+++ b/tools/launch_instance.rb
@@ -689,7 +689,7 @@ module BushSlicer
           raise "LAUNCHER_VARS not a mapping but #{launcher_vars.inspect}"
         end
       end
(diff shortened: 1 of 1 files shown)


jianlinliu:
Sorry, actually I am not sure what part of installation will be affected by this change in details. If you know, I suggest to run a install using your own branch via flexy for a validation.

akostadinov:
I found incompatibilities with 3.11 `openshift_ansible_vars`. I will not introduce such change. Not high priority.

PR #1533, opened 2020-09-15
host-ovs: add missing network_type and fix pgrep

rbbratta:


rbbratta:
fixes OCP-25707

features/step_definitions/networking.rb
```diff
@@ -744,6 +742,8 @@
     end
   end
   unless host_ovs
+    network_operator = BushSlicer::NetworkOperator.new(name: "cluster", env: env)
```
pruan-rht, line 745: @rbbratta probably can combine these two lines into one?

rbbratta: I'd rather replace all the instances at once, but this current PR fixes current test failures, can we merge while I refactor `network_type`?

pruan-rht (commented):
/lgtm

PR #1539, opened 2020-09-17
rules added for OCP-33568

yapei:
@xiaocwan  Please help review


yapei:
@pruan-rht  @liangxia  Can you help review and merge if no objections?

PR #1544, opened 2020-09-18
Auto OCP-35068

xiuwang:
Passed locally.

### DIFF
--- a/lib/rules/web/images/jenkins_2/jenkins_2.xyaml
+++ b/lib/rules/web/images/jenkins_2/jenkins_2.xyaml
@@ -593,7 +593,7 @@ jenkins_oauth_login:
   action: jenkins_verify_logged_in
 
 jenkins_multi_oauth_login:
-  url: /securityRealm/commenceLogin?from=%2F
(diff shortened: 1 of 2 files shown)


lib/rules/web/images/jenkins_2/jenkins_2.xyaml
```diff
@@ -593,7 +593,7 @@ jenkins_oauth_login:
   action: jenkins_verify_logged_in
 
 jenkins_multi_oauth_login:
-  url: /securityRealm/commenceLogin?from=%2F
+  url: /
```
liangxia, line 596: @yapei  Could you help check ?  I'm not quite sure it will affect your team or not.

liangxia: I'll merge later today if still no response.

xiuwang:
@liangxia  Thanks Liang, this change will not affect UI, the rules/web/images/jenkins_2/jenkins_2.xyaml is only for jenkins feature.

PR #1547, opened 2020-09-21
Bugfix for PV is deleted before PVC

qinpingli:
Patch for v3 branch.

@duanwei33 @chao007 Please help review.

### DIFF
--- a/features/storage/nfs.feature
+++ b/features/storage/nfs.feature
@@ -33,12 +33,16 @@ Feature: NFS Persistent Volume
       | touch | /mnt/nfs/testfile_2 |
     Then the step should succeed
 
+    Given I ensure "hellopod" replicationcontroller is deleted
+
     # Finally verify both files created by each pod are under the same export dir in the nfs-server pod
     When I execute on the "nfs-server" pod:
       | ls | /mnt/data |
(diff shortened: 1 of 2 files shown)


features/storage/nfs.feature
```diff
@@ -135,3 +139,5 @@ Feature: NFS Persistent Volume
     And the output should contain:
       | Permission denied |
 
+   Given I ensure "nfspd-<%= project.name %>" pod is deleted
+   And I ensure "nfsc-<%= project.name %>" pvc is deleted
```
akostadinov, line 143: I don't mind these. But if test fails, we never reach these steps. Shouldn't we register deletion `after scenario` earlier as clean-up steps?

akostadinov: @jhou1 ?

akostadinov:
How do things fail before this patch? Shall we fix standard clean-up logic instead?

akostadinov:
/lgtm

to unblock but I think a better fix is preferred.

features/storage/nfs.feature
liangxia: Added that in https://github.com/openshift/verification-tests/pull/1562

PR #1552, opened 2020-09-21
we need to set `-i` when stdin is provided

akostadinov:
@rhpmali , please check if this works for you.


rhpmali:
Provided patch worked for me, Thanks @akostadinov

rhpmali:
Here is the successful testing file
[groups_sync.txt](https://github.com/openshift/verification-tests/files/5256764/groups_sync.txt)

pruan-rht (commented):
/lgtm

PR #1556, opened 2020-09-22
add rules for OCP-33748

yapei:
Ready for review


yapei:
@pruan-rht  @liangxia  Could you help review and merge if no objections? Thanks

PR #1562, opened 2020-09-23
Ensure pvc clean-up, regardless of test step failures

liangxia:
Even some of the steps in the middle failed, e.g, pod failed to be ready for any reason, there will be clean-up for the PVC to unblock the clean-up of PV.

As a follow up of https://github.com/openshift/verification-tests/pull/1547

/cc @akostadinov @qinpingli @chao007 @duanwei33

### DIFF
--- a/features/storage/nfs.feature
+++ b/features/storage/nfs.feature
@@ -12,6 +12,7 @@ Feature: NFS Persistent Volume
     And admin creates a PV from "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/storage/nfs/auto/pv-retain.json" where:
       | ["metadata"]["name"]      | nfs-<%= project.name %>          |
       | ["spec"]["nfs"]["server"] | <%= service("nfs-service").ip %> |
+    Given I ensure "nfsc" pvc is deleted after scenario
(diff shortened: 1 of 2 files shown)


akostadinov:
/lgtm

this is good. But why don't we change `delete_graceful` to `ensure_deleted` [here](https://github.com/openshift/verification-tests/blob/master/features/step_definitions/pv.rb#L178)? This should solve all issues AFAICT.

features/storage/nfs.feature
```diff
@@ -12,6 +12,7 @@ Feature: NFS Persistent Volume
     And admin creates a PV from "https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/storage/nfs/auto/pv-retain.json" where:
       | ["metadata"]["name"]      | nfs-<%= project.name %>          |
       | ["spec"]["nfs"]["server"] | <%= service("nfs-service").ip %> |
+    Given I ensure "nfsc" pvc is deleted after scenario
```
qinpingli, line 15: IIUC, this teardown step will be run after the PV delete teardown step, then this case will be blocked, @liangxia Could you paste the test result here?

qinpingli: Ah, sorry, it's added after pv is created, then lgtm.

PR #1380, opened 2020-07-22
auto sriov cases

zhaozhanqi:
automated more cases and test pass in local

PR #1566, opened 2020-09-25
Fix resource type of NetworkAttachmentDefinition

liangxia:
Fix https://github.com/openshift/verification-tests/pull/1380#discussion_r494296702

```
$ oc api-resources | grep -E 'NAME|attachment'
NAME                                  SHORTNAMES       APIGROUP                              NAMESPACED   KIND
network-attachment-definitions        net-attach-def   k8s.cni.cncf.io                       true         NetworkAttachmentDefinition

```


pruan-rht (commented):
/lgtm

PR #1569, opened 2020-09-27
Fix test failure of case OCP-27841 and automate new case OCP-21874

lihongyan1:
Related to PR https://github.com/openshift/cucushift/pull/8167

### DIFF
--- a/lib/rules/cli/4.4.yaml
+++ b/lib/rules/cli/4.4.yaml
@@ -447,6 +447,8 @@
     :keys: --keys=<value>
     :resource: <value>
     :to: --to=<value>
+:get_raw:
+  :cmd: oc get --raw <resource>
 :get:
   :cmd: oc get <resource>
   :options:
--- a/lib/rules/cli/4.4.yaml
+++ b/lib/rules/cli/4.4.yaml
@@ -447,8 +447,6 @@
     :keys: --keys=<value>
     :resource: <value>
     :to: --to=<value>
-:get_raw:
-  :cmd: oc get --raw <resource>
 :get:
   :cmd: oc get <resource>
   :options:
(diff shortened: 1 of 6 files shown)


lib/rules/cli/4.4.yaml
```diff
@@ -447,6 +447,8 @@
     :keys: --keys=<value>
     :resource: <value>
     :to: --to=<value>
+:get_raw:
+  :cmd: oc get --raw <resource>
```
liangxia, line 451: This is against CLI rules. 
Given we already have a lot of real command, we do not add specific options as another command, or else we will have dozens of specific commands.

lihongyan1: The reason I add a rule, the --raw can't be at the end for my command as the following
[hongyli@hongyli-fed Downloads]$ oc get /apis/custom.metrics.k8s.io/v1beta1 --raw
Error: flag needs an argument: --raw
See 'oc get --help' for usage.
[hongyli@hongyli-fed Downloads]$ oc get --raw /apis/custom.metrics.k8s.io/v1beta1

lihongyan1: so --raw can't be added as an option and the following way can't work either
  | resource | --raw path |

liangxia: It works fine as below,
```
    When I run the :get client command with: # features/step_definitions/cli.rb:13
      | resource | --raw=/api/v1/namespaces/<%= project.name %> |
      [03:51:20] INFO> Shell Commands: oc get --raw\=/api/v1/namespaces/t2sah --kubeconfig=/home/lxia/workdir/ocp-lxia/ocp4_testuser-43.kubeconfig
      [03:51:22] INFO> Exit Status: 0
    Then the step should succeed             # features/step_definitions/common.rb:4

```

PR #1570, opened 2020-09-27
update OCP-12625

yapei:


yapei:
@liangxia  @pruan-rht  Could you help review and add lgtm label if no objections? Thanks

PR #1573, opened 2020-09-28
add specify for container

zhouying7780:
@kasturinarra  The case will failed with ocp4.4. need to specify the container for `oc logs` .  And the output for ocp4.4 is also different from ocp4.6 . So short the output check.


zhouying7780:
Failed log for ocp4.4: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/167431/console

kasturinarra:
@zhouying7780  thanks for fixing !!

PR #1574, opened 2020-09-29
add automation scenarios to test forward logs with tags

QiaolingTang:
Cases: https://url.corp.redhat.com/1a12958
Runner-v3 logs: https://url.corp.redhat.com/d7afc62

@anpingli PTAL, thanks.

### DIFF
--- a/lib/openshift/cluster_log_forwarder.rb
+++ b/lib/openshift/cluster_log_forwarder.rb
@@ -60,6 +60,17 @@ module BushSlicer
         return raw_resource(user: user, cached: cached, quiet: quiet).dig('status')
     end
 
+    def outputs_status(user: nil, quiet: false, cached: true)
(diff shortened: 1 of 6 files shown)


anpingli (commented):
/lgtm

lib/openshift/cluster_log_forwarder.rb
```diff
@@ -60,6 +60,17 @@ def status_raw(user: nil, quiet: false, cached: true)
         return raw_resource(user: user, cached: cached, quiet: quiet).dig('status')
     end
 
+    def outputs_status(user: nil, quiet: false, cached: false)
```
liangxia, line 63: We should default cached to true. Set cached to false in your scenario when you do not want to use cached value.

QiaolingTang: Done, https://github.com/openshift/cucushift/pull/8173 is updated too.

PR #1575, opened 2020-09-29
Fix the manifest issue

duanwei33:
The previous secret yaml file has bad indentation, which makes the secret has no real data.


duanwei33:
@qinpingli @chao007 Could you help review it?

pruan-rht (commented):
/lgtm

PR #1576, opened 2020-09-30
add lib support for argocds

yapei:


yapei:
@liangxia  Please help review, thanks

PR #1584, opened 2020-10-06
Machine API co should go degraded if machine-controller pods crashes

miyadav:
Getting below exception while running the test: 
.
.

Expand ERROR with argument: Given a pod becomes ready with labels:
  | api=clusterapi, k8s-app=controller |
When admin ensure "<%= pod.name %>" pod is deleted from the "openshift-machine-api" project
And the step should succeed
#<RuntimeError: what Pod are you talking about?>
    /home/jenkins/workspace/Runner-v3/lib/world.rb:313:in `project_resource'
.
.
Although for test purpose it works fine , the functionality that I want to test gets validated , but I need help in reviewing what is causing above and the overall code .
Job url - https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/175580/console
This is the related bug - https://bugzilla.redhat.com/show_bug.cgi?id=1859221

### DIFF
--- a/features/machine/machine.feature
+++ b/features/machine/machine.feature
@@ -429,3 +429,27 @@ Feature: Machine features testing
     And the output should contain:
       | Forbidden: publicIP is not allowed in Azure disconnected installation |
 
+  # @author miyadav@redhat.com
+  # @case_id OCP-33047
+  @admin
+  @destructive
+  Scenario: Machine API Operator should go degraded if any pod controller is crash looping
+    Given I have an IPI deployment
+    When I switch to cluster admin pseudo user
+    Then I use the "openshift-machine-api" project
+
+    And I run the steps 20 times or exit when co machine-api is degraded:
+    """
+    Given a pod becomes ready with labels:
+      | api=clusterapi     |
+      | k8s-app=controller |
+    When admin ensure "#{pod.name}" pod is deleted from the "openshift-machine-api" project
+    And the step should succeed
+    """
+
+    When I wait up to 180 seconds for the steps to pass:
+    """
+    And evaluation of `cluster_operator('machine-api').condition(type: 'Degraded')` is stored in the :co_degraded clipboard
+    Then the expression should be true> cb.co_degraded["status"]=="False"
+    """
+
--- a/features/step_definitions/meta_steps.rb
+++ b/features/step_definitions/meta_steps.rb
@@ -114,3 +114,22 @@ Given /^I repeat the following steps for each #{SYM} in cb\.([\w]+):$/ do |varna
   end
 end
 
+# To exit successfully if output is obtained - mapi
+Given /^I run the steps #{NUMBER} times or exit when co machine-api is degraded:$/ do |num, steps_string|
+  eval_regex = /\#\{(.+?)\}/
+  eval_found = steps_string =~ eval_regex
+  step %Q{evaluation of `cluster_operator('machine-api').condition(type: 'Degraded')` is stored in the :co_degraded clipboard}
+  begin
+    logger.dedup_start
+    (1..Integer(num)).each { |i|
+      cb.i = i
+      if eval_found && cb.co_degraded["status"]=="False"
+        steps steps_string.gsub(eval_regex) { |s| "<%= #{$1} %>"}
+      else
+        steps steps_string
+      end
+    }
+  ensure
+    logger.dedup_flush
+  end
+end


jhou1:
I think the error is because you are trying to delete a pod repeatedly in a loop, where a second delete attempt finds the pod does not exit because it's deleted by the first delete attempt.

jhou1:
If you want to delete the pod, you can scale down the machine-api-operator as well:

```
oc scale -n openshift-cluster-version deployment/cluster-version-operator --replicas=0
oc scale -n openshift-machine-api deployment/machine-api-operator --replicas=0
```

miyadav:
@jhou1 , we don't want to scale deplopyments of cvo or machine-api-operator , this is to create a condition when everything is as is and just the pod gets crashed due to some reason repeatedly  , then we want to make sure , it shows degraded at some point of time and not miss it ..

features/machine/machine.feature
```diff
@@ -357,3 +357,24 @@ Feature: Machine features testing
...
+    Then I use the "openshift-machine-api" project
+
+    And I run the steps 20 times or exit when co machine-api is degraded:
+    """
+    Given a pod becomes ready with labels:
+      | api=clusterapi, k8s-app=controller |
+    When admin ensure "<%= pod.name %>" pod is deleted from the "openshift-machine-api" project
+    And the step should succeed
+    """
+    When 180 seconds have passed
```
sunzhaohua2, line 377: use `And I wait up to 180 seconds for the steps to pass:` to avoid hard code.

features/machine/machine.feature
```diff
@@ -429,3 +429,23 @@ Feature: Machine features testing
...
+    And evaluation of `cluster_operator('machine-api').condition(type: 'Degraded')` is stored in the :co_degraded clipboard
```
akostadinov, line 450: It seems to me that you don't need a new step. You can just put the above step in the loop. So you delete pod, then check if API is degraded. 

Then wait 180 seconds and check it is still degrated.

akostadinov: I mean you put these steps inside `I wait(?: up to #{NUMBER} seconds)? for the steps to pass:`

features/machine/machine.feature
miyadav: done

miyadav:
Made new changes as suggested by Alexander .. 
Here is the Validation results 
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/216774/console

/unhold

miyadav:
@sunzhaohua2  @jhou1  PTAL

features/step_definitions/meta_steps.rb
```diff
@@ -114,3 +114,22 @@
   end
 end
 
+# To exit successfully if output is obtained - mapi
+Given /^I run the steps #{NUMBER} times or exit when co machine-api is degraded:$/ do |num, steps_string|
```
akostadinov, line 118: my request was to remove this new step and use `I wait(?: up to #{NUMBER} seconds)? for the steps to pass:` instead. See our slack conversation.

miyadav: Hey @akostadinov but , it wont give the same result if I make that change , we will not know if the clusteroperator machine-api , actually degraded ..
Here is what I changed -  And I wait up to 300 seconds for the steps to pass:
    """
The output , is not the intended one , it just deletes pod and waits the co doesnt get degraded by deleting ones ..
.
.
      [04:35:47] INFO> Shell Commands: rm -r -f -- /root/workdir/3ec7b6b1d17c-
      
      [04:35:47] INFO> Exit Status: 0
      [04:35:47] INFO> === End After Scenario: Machine API Operator should go degraded if any pod controller is crash looping ===

1 scenario (1 passed)
5 steps (5 passed)
0m12.816s
[04:35:47] INFO> === At Exit ===

akostadinov: I can't decipher your output. Could you more carefully report what steps you use and what output you get and use code blocks instead of normal text?

miyadav: Hi @akostadinov , here is the complete output - http://pastebin.test.redhat.com/923625 , the scenario executes in less than 13 sec (when I remove the step definition) , whereas it takes around 4 mins once the actual functional test that we want gets executed ..

features/machine/machine.feature
```diff
@@ -429,3 +429,27 @@ Feature: Machine features testing
...
+    When I switch to cluster admin pseudo user
+    Then I use the "openshift-machine-api" project
+
+    And I run the steps 20 times or exit when co machine-api is degraded:
+    """
+    Given a pod becomes ready with labels:
+      | api=clusterapi     |
+      | k8s-app=controller |
+    When admin ensure "#{pod.name}" pod is deleted from the "openshift-machine-api" project
+    And the step should succeed
```
akostadinov, line 447: FYI this step is not needed. Previous step will fail in case result is failure anyway.

features/step_definitions/meta_steps.rb
akostadinov: In your custom wait step, you are waiting for some steps to complete and insert a step after them 
```
  step %Q{evaluation of `cluster_operator('machine-api').condition(type: 'Degraded')` is stored in the :co_degraded clipboard}
```

So basically you can remove new step and insert the extra step behind the others like:
```
And I wait up to 180 seconds for the steps to pass:
"""
  | api=clusterapi     |
  | k8s-app=controller |
When admin ensures "#{pod.name}" pod is deleted from the "openshift-machine-api" project
Then the expression should be true> cluster_operator('machine-api').condition(type: 'Degraded')["status"]=="False"
"""
```
What I see in your log is that you have split the steps into separate wait steps which breaks the semantics.

miyadav: Hi @akostadinov  , I might not be able to express clearly , but this is ok for test to have machine-api  degraded as true at first and then false , but we are not aware when it will go degraded for few seconds , so I think it is not breaking semantics .. here is what we are trying to validate - https://bugzilla.redhat.com/show_bug.cgi?id=1859221#c5

akostadinov: I pasted wrong step. What I mean is to do
```
And I wait up to 180 seconds for the steps to pass:
"""
  | api=clusterapi     |
  | k8s-app=controller |
When admin ensures "#{pod.name}" pod is deleted from the "openshift-machine-api" project
Then the expression should be true> cluster_operator('machine-api').condition(type: 'Degraded')["status"]=="True"
"""
```

And then, once you know status became degraded, you can continue with the rest of the steps. This is basically what your new step is doing AFAIU. No?

miyadav: you are correct logically , but  , when I use it the way you suggested above it is unable to find pod.name and never kill machine-controller pod 
Checked it in debugging shell :
"[1] pry(#<BushSlicer::DefaultWorld>)> cluster_operator('machine-api').condition(type: 'Degraded')
=> {"lastTransitionTime"=>"2020-12-15T08:57:05Z", "reason"=>"AsExpected", "status"=>"False", "type"=>"Degraded"}
[2] pry(#<BushSlicer::DefaultWorld>)> pod.name
=> "\#{pod.name}""
When I used the way I have : it picks up the podname and keep destroying it until we get degraded status as "True"
Debugging shell when used with meta_step.rb : 
"[1] pry(#<BushSlicer::DefaultWorld>)> pod.name
=> "machine-api-controllers-7f67dc495-mgwt6"
"
Might be some issue with the way I coded it .. will check , but functionally works as expected when I used the way I have used ..

akostadinov: @miyadav , please start using code blocks when pasting something. It takes much more effort to understand what you want to say otherwise. I'm not sure why you never do it but it is really important to do.

The issue is that the step `And I wait up to 180 seconds for the steps to pass:` does not process steps individually and doesn't handle resolution of expressions. It is not very easy to fix but is much better to improve the existing step instead of creating a new step with limited application.

Until somebody has time to improve the step, I'd suggest you to use step
```
/^I remove all #{WORD}(?: in the#{OPT_QUOTED} project) with labels:$/
```
i.e.
```
Then I remove all pods with labels:
  | api=clusterapi     |
  | k8s-app=controller |
```
Will this do?

miyadav: thanks @akostadinov  for pointing  out , I will use code blocks . 
I am not sure , above will help , as it will still not monitor the cluster operator status  ?

akostadinov: @miyadav , what you are presently doing is delete the pods and exit when status becomes degraded. Instead of writing a new step, you can use the existing wait step and do a last check
`Then the expression should be true> cluster_operator('machine-api').condition(type: 'Degraded')["status"]=="True"` as I suggested above. What other monitoring is needed? I can't understand what is the issue.

```
And I wait up to 180 seconds for the steps to pass:
"""
Then I remove all pods with labels:
  | api=clusterapi     |
  | k8s-app=controller |
Then the expression should be true> cluster_operator('machine-api').condition(type: 'Degraded')["status"]=="True"
```

Or something similar. It is almost the same of what you do, except that you are waiting for pod to become available. Accoirding to your description it doesn't sound to me that pod status checking is important. Or what else is missing?

PR #1585, opened 2020-10-06
wait_till: fix failure case, return false instead of nil

rbbratta:
In the `wait_till_*` functions if the timeout has already
expired before the first status check then we set
`res` to `nil` which causes the access `res[:response]`
to raise

```ruby
      undefined method `[]' for nil:NilClass (NoMethodError)
      verification-tests/lib/openshift/resource.rb:351:in `wait_till_ready'
```

This is not the correct error message.

The steps that call `wait_till_ready` have a more detailed error
message that indicates which exact resource failed.

```ruby
  @result = _resource.wait_till_ready(_user, timeout)
  unless @result[:success]
    raise %Q{#{type} "#{_resource.name}" did not become ready within } \
       "#{timeout} seconds"
  end
```

If we set `res = BushSlicer::ResultHash.new(:success => false)` then
`@result[:success]` will fail and we get a good error message

```ruby
RuntimeError: BushSlicer::Pod ovnkube-master-vw8v6 did not become ready within timeout
verification-tests/features/step_definitions/resources.rb:107:in `block (2 levels) in <top (required)>'
```

Also use `wait_for` `stats` to calculate duration

### DIFF
--- a/lib/base_helper.rb
+++ b/lib/base_helper.rb
@@ -138,20 +138,33 @@ module BushSlicer
       # @param seconds [Numeric] the max number of seconds to try operation to
       #   succeed
       # @param interval [Numeric] the interval to wait between attempts
+      # @param stats [Hash] collect stats
+      # @param jitter_multiplier [Numeric] (value >= 1) used to generate
+      #   random jitter in the wait interval using the decorrelated jitter
+      #   algorithm from https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+      #   Jitter helps reduce contention between simultaneous clients
       # @yield block the block will be yielded until it returns true or timeout
       #   is reached
-      def wait_for(seconds, interval: 1, stats: nil)
-        if seconds > 60
-          Kernel.puts("waiting for operation up to #{seconds} seconds..")
-        end
+      def wait_for(seconds, interval: 1, stats: nil, jitter_multiplier: 3)
         iterations = 0
-
         start = monotonic_seconds
+        # pre-compute deadline
+        deadline = start + seconds
         success = false
-        until monotonic_seconds - start > seconds
-          iterations += 1
-          success = yield and break
+        base_interval = interval
+        until monotonic_seconds > deadline
+          (success = yield) and break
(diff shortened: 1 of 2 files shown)


rbbratta:
```ruby

[21:29:39] INFO> After 0 iterations and 0 seconds:

RuntimeError: BushSlicer::Pod ovnkube-master-vw8v6 did not become ready within timeout
verification-tests/features/step_definitions/resources.rb:107:in `block (2 levels) in <top (required)>'
````

**Comment by akostadinov** [comment]
I'm not sure this is correct fix here. Within the `wait_for` block we have as first operation `res = ready?(user: user, quiet: true)`.

This operation will not be interrupted no matter if timeout ended or not. Timeout will not cause operation to be canceled. Only at the end of the block time is checked. If it has run out, then no further iterations are performed. Otherwise another iteration is performed and only when whole iteration completes, time is checked again.

As a last step of the `wait_for` block on line 346 we have `res[:success]`. So in case `res` is `nil`, then the `nil:NilClass` error should happen at that line, not 351.

My guess is that something called the `wait_till_*` methods with a negative timeout value so no iterations have been performed. I'm not sure what should be a correct behavior of a wait with timeout when negative timeout value is given. 

- Shall method fail immediately?
- return some kind of `false` value`?
- do one check and return its result?

**Comment by rbbratta** [comment]
The `if` in `wait_for` was not showing negative `seconds`

```ruby
        if seconds > 60
          Kernel.puts("waiting for operation up to #{seconds} seconds..")
        end
```

Not sure why we don't display `seconds < 60`, so we should remove that.

It `seconds < 0` then the block won't be yielded so `res` will not be set.

`wait_for` returns false if `seconds` is negative, so `wait_till*` should also return false.

What does a false `res` value look like.  Based on `ready?` it should be `{:success=>false}` and that's it.

rbbratta:
rebased on top of the jitter work, since I was testing it all.

rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-11-12-015619.log

PR #1588, opened 2020-10-10
rules update

yapei:
fixed
OCP-24281 RBAC check to normal operations for users with view access - rule update
OCP-20688 Check log streaming - rule update
OCP-28297 Cluster Inventory block has a full list of required items - remove `s` to reduce instability caused by single and plural nouns 
OCP-19722 Check nodes list columns and terminal tab - feature changed and there is no need to choose namespace, rule updated


yapei:
pass log in Runner-v3/179725/console

yapei:
@liangxia  Please help merge if no objections, thank you

PR #1596, opened 2020-10-14
Fix OVN tests: namespace, expect grep fail, pod(0) references

rbbratta:
Use explicit namespace in scale commands with `-n`

Move the CNO scale to 1 register cleanup steps to before
we scale down.

It seemed like the `oc exec` commands were timing
out, so we switched to using `grep` on the server
to filter the output, but for the negative test
the `grep` will not match `"hello-pod"` so the
`oc exec` returns exit status `1`, and the step
will fail.  Adjust the steps to expect the failure.

Fix the pod references when finding the leader pods.
When we find the OVN leader pod we cache the resource
so the leader pod appears as `pod(0)`, this broke
the pod references as the `test-pods` are now
`pod(1) and pod(2)` instead of `pod(0)` and `pod(1)`.

Instead grab the pods by label and reference them that way.

### DIFF
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -48,22 +48,22 @@ Feature: OVN related networking scenarios
   @destructive
   Scenario: OVN DB should be updated correctly if a resource only exist in Kube API but not in OVN NB db
     Given the env is using "OVNKubernetes" networkType
-    # Now scale down CNO pod to 0
-    Given admin uses the "openshift-network-operator" project
-    And I run the :scale admin command with:
-      | resource | deployment       |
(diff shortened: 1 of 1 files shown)


rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-10-13-22-37-02.log

/cc anuragthehatter
/cc zhaozhanqi
/cc huiran0826
/cc weliang1

anuragthehatter:
Approach looks good and logs as well. Thanks Ross.

/LGTM

pruan-rht (commented):
/lgtm

PR #1597, opened 2020-10-15
OCP-33848: wait for nodeport svc

rbbratta:
We are testing a nodeport out of the standard range
so the create always returns a warning and exit status 1
but the actual service is actually created

The loop won't work because the `hello-pod` is already created
and the command will still continue to show the warning and
return exit status 1.

Instead of retrying, instead use the wait for service ready step
and then wait for the pod.

Failures when trying the loop, we can see that `hello-pod` is created,
but when retrying we get `AlreadyExists`

```
    And I wait up to 300 seconds for the steps to pass:                                             # features/step_definitions/meta_steps.rb:33
      [03:04:14] INFO> {"apiVersion":"v1","kind":"List","items":[{"kind":"Pod","apiVersion":"v1","metadata":{"name":"hello-pod","labels":{"name":"hello-pod"}},"spec":{"containers":[{"name":"hello-pod","image":"aosqe/hello-openshift","ports":[{"containerPort":8081}]}]}},{"kind":"Service","apiVersion":"v1","metadata":{"name":"hello-pod","labels":{"name":"hello-pod"}},"spec":{"ports":[{"name":"http","protocol":"TCP","port":27017,"nodePort":32929,"targetPort":8081}],"type":"NodePort","selector":{"name":"hello-pod"}}}]}
      [03:04:14] INFO> Shell Commands: oc create -f - --kubeconfig
      pod/hello-pod created

      STDERR:
      The Service "hello-pod" is invalid: spec.ports[0].nodePort: Invalid value: 32929: provided port is not in the valid range. The range of valid ports is 30000-32767

      [03:04:15] INFO> Exit Status: 1
      [03:04:16] INFO> {"apiVersion":"v1","kind":"List","items":[{"kind":"Pod","apiVersion":"v1","metadata":{"name":"hello-pod","labels":{"name":"hello-pod"}},"spec":{"containers":[{"name":"hello-pod","image":"aosqe/hello-openshift","ports":[{"containerPort":8081}]}]}},{"kind":"Service","apiVersion":"v1","metadata":{"name":"hello-pod","labels":{"name":"hello-pod"}},"spec":{"ports":[{"name":"http","protocol":"TCP","port":27017,"nodePort":32929,"targetPort":8081}],"type":"NodePort","selector":{"name":"hello-pod"}}}]}
      [03:04:16] INFO> Shell Commands: oc create -f - --kubeconfig

      STDERR:
      Error from server (AlreadyExists): pods "hello-pod" already exists
      Error from server (Invalid): Service "hello-pod" is invalid: spec.ports[0].nodePort: Invalid value: 32929: provided port is not in the valid range. The range of valid ports is 30000-32767

      INFO> last 4 messages repeated 81 times
      [03:06:53] INFO> Exit Status: 1
      [03:06:54] INFO> {"apiVersion":"v1","kind":"List","items":[{"kind":"Pod","apiVersion":"v1","metadata":{"name":"hello-pod","labels":{"name":"hello-pod"}},"spec":{"containers":[{"name":"hello-pod","image":"aosqe/hello-openshift","ports":[{"containerPort":8081}]}]}},{"kind":"Service","apiVersion":"v1","metadata":{"name":"hello-pod","labels":{"name":"hello-pod"}},"spec":{"ports":[{"name":"http","protocol":"TCP","port":27017,"nodePort":32929,"targetPort":8081}],"type":"NodePort","selector":{"name":"hello-pod"}}}]}
      [03:06:54] INFO> Shell Commands: oc create -f - --kubeconfig
      service/hello-pod created

      STDERR:
      Error from server (AlreadyExists): pods "hello-pod" already exists

      [03:06:55] INFO> Exit Status: 1

### DIFF
--- a/features/networking/service.feature
+++ b/features/networking/service.feature
@@ -490,12 +490,9 @@ Feature: Service related networking scenarios
     Given as admin I successfully merge patch resource "networks.config.openshift.io/cluster" with:
       | {"spec":{"serviceNodePortRange": "30000-33000"}} |
     When I obtain test data file "networking/nodeport_service.json"
-    And I wait up to 300 seconds for the steps to pass:
-    """    
     When I run oc create over "nodeport_service.json" replacing paths:
       | ["items"][1]["spec"]["ports"][0]["nodePort"] | <%= cb.port %> |
-    Then the step should succeed
-    """
+    Given I wait for the "hello-pod" service to be created
     Given the pod named "hello-pod" becomes ready
     Given I select a random node's host
     When I run commands on the host:
--- a/features/networking/service.feature
+++ b/features/networking/service.feature
@@ -82,13 +82,13 @@ Feature: Service related networking scenarios
     Given I have a project
     And SCC "privileged" is added to the "system:serviceaccounts:<%= project.name %>" group
     And I have a pod-for-ping in the project
-    
+
     #Creating laodbalancer service that points to MCS IP
-    When I run the :create_service_loadbalancer client command with: 
+    When I run the :create_service_loadbalancer client command with:
       | name | <%= cb.ping_pod.name %>  |
-      | tcp  | 22623:8080               | 
+      | tcp  | 22623:8080               |
     Then the step should succeed
-    
+
     # Editing endpoint to point to master ip
     When I run the :patch client command with:
       | resource      | ep                         				      						   |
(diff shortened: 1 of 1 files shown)


rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2020-10-14-21-30-56.log

/cc anuragthehatter
/cc zhaozhanqi
/cc huiran0826
/cc weliang1

rbbratta:
/hold

this might be busted, since we don't revert the `serviceNodePortRange`

zhaozhanqi:
> revert

@rbbratta serviceNodePortRange cannot be reverted once expand. So no need this step

rbbratta:
ah, well then I can't really test this fix, because my cluster has already expanded the port range and so the services always succeed.

rbbratta:
well I can keep expanding the range I guess.

rbbratta:
I have a different fix.

zhaozhanqi:
yes, hold on, seems the hello-pod service cannot be ready using current fix. 
/lgtm cancel

rbbratta:
we need to split the features/tierN/testdata/networking/nodeport_service.json  into pod and service, and just retry the service maybe

zhaozhanqi:
@rbbratta the original issue should be already fixed when I updated this from 200 second to 300 second due to the api has not take effect. 
or maybe you have a better way

rbbratta:
it failed with 300 seconds here http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/10/14/02:00:17/User_can_expand_the_nodePort_range_by_patch_the_serviceNodePortRange_in_network/console.html

zhaozhanqi:
hmm, 300 is not enough, update it to 600 
https://github.com/openshift/verification-tests/pull/1601

rbbratta:
better fix in #1611 
closing

PR #1598, opened 2020-10-15
fix ocp-24490

huiran0826:
The original case failed with error : 
STDERR:
      arping: socket: Operation not permitted
      command terminated with exit code 2

Updated the script to use ping.
Logs:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/182390/console

@anuragthehatter @rbbratta @zhaozhanqi @weliang1

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -699,15 +699,15 @@ Feature: Multus-CNI related scenarios
     
     #making sure the pods on same node can ping while pods on diff nodes can't
     When I execute on the "<%= cb.pod1 %>" pod:
-      | arping | -I | net1 |-c1 | -w2 | <%= cb.pod2_net1_ip %> |
+      | ping | -c1 | -w2 | <%= cb.pod2_net1_ip %> |
     Then the step should succeed
 
     When I execute on the "<%= cb.pod3 %>" pod:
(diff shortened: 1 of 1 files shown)


anuragthehatter:
Thanks Huiran. It was ping initially but we had to change to arping to counter same net1 ip address issues on different nodes. Means we are creating multus pods on different nodes but with same net-attach-def so two pods on diff nodes might get same net1 ip as ```bridge-host-local-vlan``` net-attach-def has a particular address range and that range will work independently for each node. So every node will see that range and assign net1 IPs from that but won't compare whether 2nd node has gotten Ip from that and vice versa.

The reason being is,  arping is not getting set with correct capatabilities on ubi8 kernel/images i believe. We need to look into that. I am going to spend some time on this figure this out.

sh-4.2$ getcap /usr/sbin/arping
/usr/sbin/arping = cap_net_raw+p     <<<<<<<<<<<<<<  cap_net_admin is missing
sh-4.2$ getcap /bin/ping
/bin/ping = cap_net_admin,cap_net_raw+p  <<<<<<<<<<<<cap_net_admin is there hence ping is working fine

We need to keep arping for such cases.

huiran0826:
Thanks Anurag, close this PR for now.

PR #1600, opened 2020-10-15
WIP: add exponential backoff with jitter wait step

rbbratta:
The `I wait up to 30 seconds for the steps to pass`
step calls `wait_for` but does not allow for changing
the interval.

The proper way to do retry loops in a distributed system is
to use exponetial backoff with jitter as described in the Amazon paper.

https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

And a prototype manual wait step for now instead of trying to alter all the loops.

### DIFF
--- a/features/step_definitions/common.rb
+++ b/features/step_definitions/common.rb
@@ -121,6 +121,28 @@ Then /^the output should equal #{QUOTED}$/ do |value|
   end
 end
 
+Given /^we wait using exponential backoff with jitter(?: using #{SYM} clipboard)?$/ do |cb_name|
+  # https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+  max_wait = 120
+  base_wait_seconds = 1
+  cb_name = 'backoff' unless cb_name
+  cb[cb_name] ||= 1
+  tmp = [max_wait, (base_wait_seconds*(2**cb[cb_name]))].min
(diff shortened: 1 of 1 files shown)


rbbratta:
might use this for OCP-33848
 https://github.com/openshift/verification-tests/blob/master/features/networking/service.feature#L493

rbbratta:
The main point is that hammering the API every second for 300 seconds isn't the best way to get a call to succeed or to wait for something to happen.  Better to add some randomness.

akostadinov:
It's a good point that many places could benefit from jitter and/or exponential back-off.

My suggestion is to alter method `wait_for` [1] which is now:
```
def wait_for(seconds, interval: 1, stats: nil)
```
It could become something like
```
def wait_for(seconds, interval: 1, stats: nil, jitter: 0, exp_back: false)
```

Jitter would be in seconds. Then we can update existing `wait_for` users whenever it is needed.

WDYT?

[1] https://github.com/openshift/verification-tests/blob/master/lib/base_helper.rb#L143

rbbratta:
`wait_for` implementation in #1649

akostadinov:
Not sure how are new steps supposed to be used. Could you show example usage?

rbbratta:
Sorry meant to close this in favor of #1649 which just updates main wait_for loop itself.

PR #1607, opened 2020-10-21
Only check sctp module in worker nodes

weliang1:
Update sctp script to:
1. Only check sctp module in worker nodes, no need check sctp modules in all the nodes.

2. Add step wait up to 60 seconds for sctpserver pod too avoid randomly see below failure
 # sctpserver pod start to wait for sctp traffic
    When I run the :exec background client command with:                              # features/step_definitions/cli.rb:13
      [17:56:14] INFO> Shell Commands: oc exec sctpserver  --kubeconfig=/home/weliang/workdir/weliang-weliang/ocp4_testuser-0.kubeconfig --namespace=y2fh9  -- sh -c /usr/bin/nc\ -l\ 30102\ --sctp\ -k\ -m\ 5\ -w\ 60
      
      STDERR:
      sh: /usr/bin/nc: No such file or directory
      command terminated with exit code 127

Test logs: http://file.rdu.redhat.com/~weliang/sctp.log

### DIFF
--- a/features/networking/sctp.feature
+++ b/features/networking/sctp.feature
@@ -10,7 +10,7 @@ Feature: SCTP related scenarios
     Given I have a project
     And I wait up to 800 seconds for the steps to pass:
     """
-    Given I check load-sctp-module in all nodes
+    Given I check load-sctp-module in all workers
     """
     Given I obtain test data file "networking/sctp/sctpserver.yaml"
     When I run oc create as admin over "sctpserver.yaml" replacing paths:
@@ -63,7 +63,7 @@ Feature: SCTP related scenarios
     Given I have a project
     And I wait up to 800 seconds for the steps to pass:
     """
-    Given I check load-sctp-module in all nodes
+    Given I check load-sctp-module in all workers
     """
     Given I obtain test data file "networking/sctp/sctpserver.yaml"
     When I run oc create as admin over "sctpserver.yaml" replacing paths:
@@ -123,7 +123,7 @@ Feature: SCTP related scenarios
     Given I have a project
     And I wait up to 800 seconds for the steps to pass:
     """
-    Given I check load-sctp-module in all nodes
+    Given I check load-sctp-module in all workers
     """
     Given I obtain test data file "networking/sctp/sctpserver.yaml"
     When I run oc create as admin over "sctpserver.yaml" replacing paths:
@@ -182,7 +182,7 @@ Feature: SCTP related scenarios
     Given I have a project
     And I wait up to 800 seconds for the steps to pass:
     """
-    Given I check load-sctp-module in all nodes
+    Given I check load-sctp-module in all workers
     """
     Given I obtain test data file "networking/sctp/sctpserver.yaml"
     When I run oc create as admin over "sctpserver.yaml" replacing paths:
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1182,13 +1182,13 @@ Given /^I install machineconfigs load-sctp-module$/ do
   end
 end
 
-Given /^I check load-sctp-module in all nodes$/ do
+Given /^I check load-sctp-module in all workers$/ do
   ensure_admin_tagged
   _admin = admin
-  env.nodes.each do |node|
-    @result = node.host.exec_admin("lsmod \| grep sctp")
-    unless @result[:response].include? "sctp"
-      raise "no sctp module"
+  cb.workers.each do |workers|
+    @result = workers.host.exec_admin("cat /sys/module/sctp/initstate")
+    unless @result[:response].include? "live"
+      raise "No sctp module installed"
     end
   end
 end


features/step_definitions/networking.rb
```diff
@@ -1157,17 +1157,21 @@
...
-Given /^I check load-sctp-module in all nodes$/ do 
+Given /^I check load-sctp-module in #{NUMBER} workers$/ do | workers_num |
   ensure_admin_tagged
   _admin = admin
-  env.nodes.each do |node|
-    @result = node.host.exec_admin("lsmod \| grep sctp")
+  $i = 0
+  $num = Integer(workers_num)
+  while $i < $num  do
+    @result = cb.workers[$i].host.exec_admin("lsmod \| grep sctp")
```
rbbratta, line 1166: we can check modules directly in `/sys/modules`

`/sys/module/sctp/initstate`  shows `live`

```bash
sh-4.4# cat /sys/module/sctp/initstate
live
```

this way we don't have to dump all the modules and grep, because the grep can match substrings as well unless we anchor it, but `lsmod` is just messy in general.  It matches the `libcrc32c` line for example.

```bash
      'lsmod\ \|\ grep\ sctp
      sctp                  401408  32
      libcrc32c              16384  5 nf_conntrack,nf_nat,openvswitch,xfs,sctp
```

`/sys/module/sctp/initstate` is cleaner

```ruby
@result = cb.workers[$i].host.exec_admin("cat /sys/module/sctp/initstate")
unless @result[:response].include? "live"
```

features/step_definitions/networking.rb
```diff
@@ -1157,17 +1157,21 @@
   end
 end
 
-Given /^I check load-sctp-module in all nodes$/ do 
+Given /^I check load-sctp-module in #{NUMBER} workers$/ do | workers_num |
```
rbbratta, line 1160: is there a use case where we wouldn't want to check all the workers in `cb.workers`?   Can we just loop over the `cb.workers` array?

features/step_definitions/networking.rb
rbbratta: Ruby is fancy, but I think the loop could be simplified, using `.to_i` and `.times`

```ruby
  workers_num.to_i.times do |i|
    @result = cb.workers[i].host.exec_admin("cat /sys/module/sctp/initstate")
    unless @result[:stdout].to_s.include? "live"
      raise "no sctp module"
    end
  end
```

features/step_definitions/networking.rb
akostadinov: We better have the step either:
```
I check load-sctp-module in all workers
```
or
```
I check load-sctp-module in nodes from the #{SYM} clipboard
```

If we just specify number, then there is no sane way for the scenario to specify which exactly workers we care about. Thus making application very limited. While providing a clipboard with nodes, would allow you to choose what you want.

And if such flexibility is not needed, then just go with the `all workers` one.

weliang1: > is there a use case where we wouldn't want to check all the workers in `cb.workers`? Can we just loop over the `cb.workers` array?

@rbbratta we can loop over 'cb.nodes' , below code works:
env.nodes.each do |node|
    @result = node.host.exec_admin("lsmod \| grep sctp")

But loop over 'cb.workers' not support, below code will not work
env.workers.each do |workers|
    @result = workers.host.exec_admin("cat /sys/module/sctp/initstate")

For sctp testing, we need check sctp module in all worker nodes which not include master nodes, so old code env.nodes.each do |node| will cause test fail.

weliang1: @rbbratta Just got this works:
cb.workers.each do |workers|

features/networking/sctp.feature
```diff
@@ -5,13 +5,15 @@ Feature: SCTP related scenarios
...
   Scenario: Establish pod to pod SCTP connections
-    Given I store the workers in the :workers clipboard    
+    Given I store the workers in the :workers clipboard
+    And I store the number of worker nodes to the :num_workers clipboard
     Given I install machineconfigs load-sctp-module
     Given I have a project
     And I wait up to 800 seconds for the steps to pass:
     """
-    Given I check load-sctp-module in all nodes
+    Given I check load-sctp-module in <%= cb.num_workers %> workers
```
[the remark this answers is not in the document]
weliang1: Fixed by using "Given I check load-sctp-module in all workers"

features/networking/sctp.feature
```diff
@@ -59,11 +64,12 @@ Feature: SCTP related scenarios
   @destructive
   Scenario: Expose SCTP ClusterIP Services
     Given I store the workers in the :workers clipboard    
+    And I store the number of worker nodes to the :num_workers clipboard
     Given I install machineconfigs load-sctp-module
     Given I have a project
     And I wait up to 800 seconds for the steps to pass:
     """
-    Given I check load-sctp-module in all nodes
+    Given I check load-sctp-module in <%= cb.num_workers %> workers
```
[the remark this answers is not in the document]
weliang1: Fixed by using "Given I check load-sctp-module in all workers"

features/step_definitions/networking.rb
weliang1: Fixed by using "cat /sys/module/sctp/initstate", thanks!

features/step_definitions/networking.rb
weliang1: > We better have the step either:
> 
> ```
> I check load-sctp-module in all workers
> ```
> 
> or
> 
> ```
> I check load-sctp-module in nodes from the #{SYM} clipboard
> ```
> 
> If we just specify number, then there is no sane way for the scenario to specify which exactly workers we care about. Thus making application very limited. While providing a clipboard with nodes, would allow you to choose what you want.
> 
> And if such flexibility is not needed, then just go with the `all workers` one.

Fixed by using "Given I check load-sctp-module in all workers"

weliang1:
New testing log: http://file.rdu.redhat.com/~weliang/sctp_test.txt

PR #1610, opened 2020-10-23
update test ingresscontroller to use NodePort instead of HostNetwork

lihongan:
`HostNetwork` might causes port conflict (rarely but possibly) when running parallel CI jobs on some platforms, so replace it with `NodePortService`, and `nodePlacement` part is not required any more.

@quarterpin please help review.


quarterpin:
/lgtm

Thanks, @lihongan !

lihongan:
log: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/191299/console

PR #1613, opened 2020-10-27
rules added for OCP-28832

yapei:
/assign @yanpzhan


yapei:
@liangxia  Could you help review and add lgtm label if no objections? Thank you

pruan-rht (commented):
/lgtm

PR #1615, opened 2020-10-27
Quickfix

anuragthehatter:
Fixes OCP-25787 glitches

@weliang1 @rbbratta @zhaozhanqi @huiran0826

PR #1618, opened 2020-10-28
OCP-25787 quick fix

anuragthehatter:
bash c to enable treating string with pmtrs as a whole single command
This will fix glitches in OCP-25787
@pruan-rht (same as #1615 which was closed to to squashing issue)


pruan-rht (commented):
/lgtm

PR #1619, opened 2020-10-29
Add retry for oauth_bearer_token_challenge

liangxia:
As described in https://bugzilla.redhat.com/show_bug.cgi?id=1890724, there are API server flappings, which can cause automation test execution job quit.

The server can recover itself after some time, we should continue the test as well, so add retry for oauth_bearer_token_challenge, or else we will fail with "Error getting bearer token"

/cc @akostadinov @pruan-rht @jhou1 @kasturinarra

### DIFF
--- a/lib/login.rb
+++ b/lib/login.rb
@@ -21,12 +21,17 @@ module BushSlicer
       debug_res = Http.request(**opts)
 
       # try challenging client auth
-      res = oauth_bearer_token_challenge(
-        server_url: env.api_endpoint_url,
-        user: user,
-        password: password,
-        proxy: env.client_proxy
-      )
+      res = {}
+      (1..3).each { |seconds|
+        res = oauth_bearer_token_challenge(
+          server_url: env.api_endpoint_url,
+          user: user,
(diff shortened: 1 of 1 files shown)


lib/login.rb
```diff
@@ -21,12 +21,17 @@ def new_token_by_password(user:, password:, env:)
...
+      res = {}
+      (1..3).each { |seconds|
+        res = oauth_bearer_token_challenge(
+          server_url: env.api_endpoint_url,
+          user: user,
+          password: password,
+          proxy: env.client_proxy
+        )
+        break if res[:success]
+        sleep (seconds * 30)
```
jhou1, line 33: Thanks for having this! Why do we increase the sleep seconds by each iteration?

kasturinarra: +1 for this question, rest looks good to me, thanks !!

liangxia: @jhou1 @kasturinarra  The idea is that we wait longer for another retry, as previous retry failed means the cluster haven't recovered itself.

PR #1620, opened 2020-10-29
Fix resource type MachineHealthCheck

liangxia:
@jhou1 @akostadinov


jhou1:
@liangxia Can you please provide the context as to what the problem is that this PR fixes?

liangxia:
> @liangxia Can you please provide the context as to what the problem is that this PR fixes?

Using the same name style for all the resources.

liangxia:
Also it's a project resource, so no need to require cluster resource.

jhou1:
Thanks.
/lgtm

PR #1630, opened 2020-11-03
Route53 support

akostadinov:


lib/launchers/amz.rb
```diff
@@ -471,6 +480,63 @@ def get_regions
...
+    #     :action=>"UPSERT",
+    #     :resource_record_set=> {
+    #       :name=>"myhost.qe.devcluster.openshift.com",
+    #       :resource_records=> [{:value=>"192.0.2.44"}],
+    #       :ttl=>60,
+    #       :type=>"A"
+    #     }
+    #   }]
+    # @see https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Aws/Route53/Client.html#change_resource_record_sets-instance_method
+    def change_resource_record_sets(zone_id: nil, changes:)
```
pjanouse, line 511: Is this method waiting for DNS propagation before returns? I'm asking because normal calling of change-resource-record-set is ASYNC - you get an ID of change and then you have to call 'aws wait resource-record-sets-changed --id ...' to hold execution until the change is propagated (usually it takes couple of seconds (10-30) in common situation).

Probably we should extend this impl. with the attribute whether we want to wait until the propagation finishes or not.

lib/launchers/amz.rb
```diff
@@ -471,6 +480,63 @@ def get_regions
...
+      client_r53.list_resource_record_sets(hosted_zone_id: zone_id).each_page { |p|
+        res.concat p.resource_record_sets.map(&:to_h)
+      }
+      return res
+    end
+
+    # @param [Regexp] re
+    # @note be very careful to avoid interfering regular expressions
+    # example: delete_resource_records_re(/my-hostname/)
+    def delete_resource_records_re(re, zone_id: nil)
```
pjanouse, line 531: SYNC vs ASYNC as above

pjanouse: Tried today and based on the speed it is definitely an ASYNC operation right now which isn't the best in this case from my PoV.

lib/launchers/amz.rb
```diff
@@ -471,6 +480,89 @@ def get_regions
...
+      logger.warn("Removing Route53 records matching #{re.inspect}")
+      records = list_resource_record_sets(zone_id: zone_id)
+      to_delete = records.select { |r| r[:name] =~ re }
+      logger.debug("Removing records: #{to_delete.map {|r| r[:name]}}")
+      list = to_delete.map { |r| { action: "DELETE", resource_record_set: r } }
+      change_resource_record_sets(zone_id: zone_id, changes: list)
+    end
+
+    def create_a_records(name, ips, zone_id: nil, ttl: 180)
```
pjanouse, line 546: Can we change the default TTL to 60(s)? This value is used by other team already and it is more legible for users I guess.

lib/launchers/amz.rb
akostadinov: It does not presently do so. It can be extended. Is it needed right now for installation?

If it works now, I'd suggest to merge as is and implement some param like `wait: nil`. When `nil` or negative number, then we do not wait and return ID of request. If it is a positive integerm then we wait for up to that number of seconds.

Please tell me.

lib/launchers/amz.rb
akostadinov: For deleting records, I think it doesn't matter. It is for clean-up only purposes anyway so we shouldn't care.

akostadinov: btw knowing that operation completed, doesn't mean that item will be unresolvable immediately, it will depend on TTL.

akostadinov: In case it is needed for something, we can also implement waiting until TTL finished to make sure nothing will resolve the name after method returned. Still I wonder whether that is needed in our use case.

lib/launchers/amz.rb
akostadinov: TTL is time that caching DNS servers may keep record it received. It's not something user visible. I didn't look at OCP installation details, do you think that after creating the record, installer will need it immediately resolvable?

My assumption was that there will be several minutes before these DNS records matter at all. But if I'm mistaken, I'll be happy to change.

akostadinov: Just to clarify, TTL only matters when removing or changing an existing record. Then if that record has been cached on some DNS server, the server will try to resolve it again (and see the new value) at most after TTL finished since name was resolved.

lib/launchers/amz.rb
pjanouse: Just only one concern with a negative number - I was thinking of "-1" as a value indicating - I'm willing to wait until it is propagated whatever time it needs (including indefinitely). In your case this needs to be handled on user side however I prefer to leave it on API side...

So what about:
* nil - don't like to wait, provide me an ID aka ASYNC mode - like now (in that case we need another `amz` helper - wait-for ID finished)
* positive number - wait at most specified amount of time (how to indicate we finished/don't finish in time limit? - exception OutOfTime for ex.)
* "-1" - wait until propagated (including counting with unlimited amount of time)

lib/launchers/amz.rb
pjanouse: I'm not fully for the statement we shouldn't care - I'm not sure if requests are somehow serialized on AWS side, but if we spin-down and spin-up a cluster fast enough, this operations can be interleaved and so executed in a wrong ordering... - we create a new records (they overwrite the older values because we use `UPSERT` operation) but later we delete them just because the previous delete operation takes so much time. Is it just a hypothetical case? Don't know...

TTL value as of 60s is more legible for me than 100s general time period, it doesn't matter for installer more I guess.

lib/launchers/amz.rb
pjanouse: Yes, it has to be present because installer depends on DNS resolving too - based on my measuring, it calls DNS for that ~5minutes after installer called but it depends on the power so it might be a bit sooner or later.

5 mins isn't a large timeframe, that's the reason I'd prefer to by in SYNC mode at least during DNS records creation.

lib/launchers/amz.rb
akostadinov: agreed

lib/launchers/amz.rb
akostadinov: Ok, lets add sync mode in another PR. My understanding was that installer would retry for some time until things start working. But I don't mind having SYNC mode as we discussed above.

lib/launchers/amz.rb
akostadinov: Wrong ordering can't happen. Records are deleted when full record is provided. The upserted record will have different IPs thus it shouldn't be deleted.

Actually DNS providers give a one hour or so default TTL IIRC. To me 60 seconds seems too aggressive leading to many more DNS queries. But on the other hand we overwrite records so 1 hour is too much. 3 minutes was my gut feeling not to be too bad to the DNS. Lets keep it 3 minutes unless we observe some issue with that value.

lib/launchers/amz.rb
pjanouse: Yes installer tries it until timeout (hard-coded 40minutes now if I recall correctly).

lib/launchers/amz.rb
pjanouse: Fine, 3min -> 180s (not 100), was a typo originally?

pjanouse: Just a note - UPSERTed records don't have a different IPs every time (as I observed), at least some of the records... but you are right, we are forced to provide an exact record to be deleted, it is now hidden from user's perspective only (to a RegExp).

akostadinov:
Ok, please let me know if things work for you with the v3 gem. And I will try to coordinate merging this change with images rebuild. Then we can make further improvements.

pjanouse:
@akostadinov I've tested your changes with my local setup and also with our established CI. I don't see any other issue.

The outstanding issue (will be processed by other PR, it is out of scope now) is only SYNC vs. ASYNC changes and handling of them.

It'd be nice to have merged this ASAP to be able to process the dependent MR.

pjanouse:
I've spot a new issue (see [here](https://flexy-one-ocp-deploy.cloud.paas.psi.redhat.com/job/Flexy-destroy/92/console)):
```
18:30:57  [17:30:52] WARN> Removing Route53 records matching /.+[.]pjanouse[.]intentionally-so-long-name-of-the-subdomain-of-third-level[.]pjanouse-rht[.]com[.]/
18:30:57  [17:30:53] ERROR> Error during deleting AWS Route 53 DNS records: #<Aws::Route53::Errors::MalformedInput: Unexpected list element termination>
[2020-12-01T17:30:57.186Z]     /home/jenkins/.gem/ruby/gems/aws-sdk-core-3.109.3/lib/seahorse/client/plugins/raise_response_errors.rb:17:in `call'
[2020-12-01T17:30:57.186Z]     /home/jenkins/.gem/ruby/gems/aws-sdk-core-3.109.3/lib/aws-sdk-core/plugins/jsonvalue_converter.rb:22:in `call'
[2020-12-01T17:30:57.186Z]     /home/jenkins/.gem/ruby/gems/aws-sdk-core-3.109.3/lib/aws-sdk-core/plugins/idempotency_token.rb:19:in `call'
[2020-12-01T17:30:57.186Z]     /home/jenkins/.gem/ruby/gems/aws-sdk-core-3.109.3/lib/aws-sdk-core/plugins/param_converter.rb:26:in `call'
[2020-12-01T17:30:57.186Z]     /home/jenkins/.gem/ruby/gems/aws-sdk-core-3.109.3/lib/seahorse/client/plugins/request_callback.rb:71:in `call'
[2020-12-01T17:30:57.186Z]     /home/jenkins/.gem/ruby/gems/aws-sdk-core-3.109.3/lib/aws-sdk-core/plugins/response_paging.rb:12:in `call'
[2020-12-01T17:30:57.186Z]     /home/jenkins/.gem/ruby/gems/aws-sdk-core-3.109.3/lib/seahorse/client/plugins/response_target.rb:24:in `call'
[2020-12-01T17:30:57.186Z]     /home/jenkins/.gem/ruby/gems/aws-sdk-core-3.109.3/lib/seahorse/client/request.rb:72:in `send_request'
[2020-12-01T17:30:57.186Z]     /home/jenkins/.gem/ruby/gems/aws-sdk-route53-1.44.0/lib/aws-sdk-route53/client.rb:1164:in `change_resource_record_sets'
[2020-12-01T17:30:57.186Z]     /home/jenkins/agent/workspace/Flexy-destroy/flexy/lib/launchers/amz.rb:519:in `change_resource_record_sets'
[2020-12-01T17:30:57.186Z]     /home/jenkins/agent/workspace/Flexy-destroy/flexy/lib/launchers/amz.rb:543:in `delete_resource_records_re'
```

akostadinov:
@pjanouse I wonder if this is caused by no record with matching name exist and we create an empty request. Just a wild guess. Is it reproducible?

pjanouse:
Damm, GitHub swallowed one of my msg from yesterday where I'd explained later what went wrong, so again...

Later I've realized it was basically fault on my side... (however it would be handled better inside helpers) I though config.yaml is copied from install job when I've executing destroy, but it isn't the case obviously.

So my case what I did when I've spotted issue above is:
- specified `install_base_domain` as A
- executed install
- changed `install_base_domain` to B (zone already exists too however it is empty - only SOA and NS record are here as usual defaults)
- execute destroy with data from above install (obviously install_base_domain=B was taken)

Hope provided details helps you.

akostadinov:
Got it, we should be more graceful with deletions it sounds going forward.

akostadinov:
Internal testing turned good.

PR #1634, opened 2020-11-04
update ca to another file since there is same name exist

zhaozhanqi:
since there is file named ca.pem in the /tmp of the image and cannot be overwrited by user. So change another name
@lihongan  PTAL,thanks.


lihongan:
/lgtm
Thank you @zhaozhanqi

pruan-rht (commented):
/lgtm

PR #1635, opened 2020-11-04
update rules for case OCP-36102

QiaolingTang:
Update the kibana rules.

### DIFF
--- a/lib/rules/web/admin_console/4.5/logging_kibana.xyaml
+++ b/lib/rules/web/admin_console/4.5/logging_kibana.xyaml
@@ -178,3 +178,9 @@ check_log_count:
       }
       return doc_count
     expect_result: true
+
+click_refresh_button_in_kibana:
+  element:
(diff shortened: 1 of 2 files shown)


lib/rules/web/admin_console/4.5/logging_kibana.xyaml
```diff
@@ -178,3 +178,9 @@ check_log_count:
       }
       return doc_count
     expect_result: true
+
+click_refresh_button_in_kibana:
+  element:
+    selector:
+      xpath: //span[@class="euiButton__text euiSuperUpdateButton__text" and contains(., "Refresh")]
```
yapei, line 185: better use `contains` instead of `=`
```
//span[contains(@class,'euiButton__text euiSuperUpdateButton__text') and contains(., 'Refresh')]
```

pruan-rht (commented):
/lgtm

PR #1637, opened 2020-11-04
Update for OCP-28630

chao007:
Add selector to make the plugin pod scheduled to same node with other csi pods.
@qinpingli @duanwei33  Please help to view.

### DIFF
--- a/features/upgrade/storage/upgrade.feature
+++ b/features/upgrade/storage/upgrade.feature
@@ -290,7 +290,6 @@ Feature: Storage upgrade tests
   # @author chaoyang@redhat.com
   @upgrade-check
   @admin
-  @destructive
   Scenario: Snapshot operator should be in available status after upgrade and can created pod with snapshot
     Given I switch to cluster admin pseudo user
 
--- a/features/upgrade/storage/upgrade.feature
+++ b/features/upgrade/storage/upgrade.feature
@@ -182,10 +182,20 @@ Feature: Storage upgrade tests
   Scenario: Snapshot operator should be in available status after upgrade and can created pod with snapshot - prepare
     Given the master version >= "4.4"
  
-    #Deploy csi hostpath driver 
-    When I run the :new_project client command with:
-      | project_name | csihostpath |
-    
+    Given I store the ready and schedulable workers in the :worker clipboard
+
+    When I run the :get admin command with:
+      | resource      | nodes                     |
+      | resource_name | <%= cb.worker[0].name %>  |
+      | show_label    | true                      |
+      
+    Given evaluation of `@result[:stdout].split(/\n/)[-1].split(/\s/)[-1].split(",").select{|n| n=~ /hostname/}.join("")` is stored in the :label clipboard
+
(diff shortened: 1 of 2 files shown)


qinpingli:
where the label "kubernetes.io/hostname=app=csi-hostpathplugin" is added to the node?

chao007:
1. Adding Pod affinity to make sure all the pod scheduled to one node. 
2. Adding node label and namespace node-selector to make sure test-pod also scheduled to the same node.

features/upgrade/storage/upgrade.feature
```diff
@@ -179,13 +179,19 @@ Feature: Storage upgrade tests
...
-    When I run the :new_project client command with:
-      | project_name | csihostpath |
-    
+    When I run the :oadm_new_project admin command with:
+      | project_name  | csihostpath            |
+      | node_selector | app=csi-hostpathplugin |
+      | admin         | <%= user.name %>       |
+ 
+    Given I store the ready and schedulable nodes in the :nodes clipboard
+    And label "app=csi-hostpathplugin" is added to the "<%= cb.nodes[0].name %>" node
```
qinpingli, line 193: 1. Looks there is no step to save some nodes in the `cb`
2. I don't think we need to use a different label, just create the `csihostpath` project with the node label, so we don't need to mark the test case is a destructive case?

testdata/storage/csi/csi-hostpath-plugin.yaml
```diff
@@ -34,6 +34,16 @@ spec:
...
+      affinity:
+        podAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - csi-hostpathplugin
+            topologyKey: kubernetes.io/hostname
```
qinpingli, line 46: If the project has nodeselectors, we don't need to update this manifest.

features/upgrade/storage/upgrade.feature
```diff
@@ -280,6 +290,7 @@ Feature: Storage upgrade tests
   # @author chaoyang@redhat.com
   @upgrade-check
   @admin
+  @destructive
```
qinpingli, line 293: destructive tag is no needed.

qinpingli:
@chao007 Thanks for the PR, except the destructive tag, others lgtm.

testdata/storage/csi/csi-hostpath-plugin.yaml
liangxia: @qinpingli  @chao007  Is the concern addressed here ?

qinpingli: I think it's unnecessary, but from @chao007 test, looks like this will not affect the csi host plugin pod schedule.
Anyway, I suggest we can revert this update.

liangxia: @chao007  Could you revert the change as @qinpingli's comments above.

PR #1638, opened 2020-11-04
Fix scenarioname

kasturinarra:
Fixing scenario name as it is same for both ocp43 and ocp44 cases due to which script is failing on 4.4


kasturinarra:

kasturinarra:

PR #1639, opened 2020-11-04
Non ping binaries use case to fix with NET_RAW and supported yamls/json files

anuragthehatter:
Adding two general pod yamls as well to be used in future for supporting use cases using bin utilities other than ping
- pod-for-ping-net-raw.json 
- generic_multus_pod_net_raw.yaml

The idea to keep above generalize pods for testing and debugging where utilities requires NET_RAW cap. Would be easy to debug in case. Not a good idea to integrate this logic with existing standard pod ping and generic multus pods we use, IMO.

Logs: https://privatebin-it-iso.int.open.paas.redhat.com/?246f314821eefd88#G542ULH9jHcF6VANUaw6f7RXAT1N3Quqa1EBhCRiXXBc

@rbbratta @zhaozhanqi @weliang1 @huiran0826


testdata/networking/multus-cni/Pods/generic_multus_pod_net_raw.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-pod
+  annotations:
+    k8s.v1.cni.cncf.io/networks: bridgevlan100
+spec:
+  containers:
+  - name: test-pod
+    command: ["/bin/bash", "-c", "sleep 2000000000000"]
```
rbbratta, line 10: I think technically we need to trap SIGTERM due to the PID 1 issues where whatever process is run as PID 1 won't receive SIGTERM and won't shutdown quickly (< 1min)

https://github.com/openshift/cluster-network-operator/pull/859/files#diff-6f3fb59cc9404c296e2f7aa6b442137bdb2d4559ced7bfc51b7a188c2b64233fR65

https://github.com/openshift/cluster-dns-operator/issues/65

> Listen for SIGTERM explicitly
> Always `sleep X & wait` instead of just `sleep X`

https://vagga.readthedocs.io/en/latest/pid1mode.html
https://hynek.me/articles/docker-signals/

The current `["/bin/bash", "-c", "sleep 2000000000000"]` is causing us to sleep ~42 seconds each time we delete the pod.

So we need to trap SIGTERM and then exit
```shell
    command: ["/bin/bash", "-c", "trap 'kill $(jobs -p); exit 0' TERM ; sleep 2000000000000 & wait"]
```

Time to delete for  `["/bin/bash", "-c", "sleep 2000000000000"]`

```
pod "test-pod" deleted

real    0m41.459s
user    0m0.175s
sys     0m0.050s
```

Time to delete for `["/bin/bash", "-c", "trap 'kill $(jobs -p); exit 0' TERM ; sleep 2000000000000 & wait"]`

```
pod "test-pod" deleted

real    0m2.738s
user    0m0.186s
sys     0m0.057s
```

averages over 20 runs with
```
for f in {1..20} ; do oc create -f generic_multus_pod.yaml ; oc wait --for=condition=Ready pod/test-pod ; time oc delete pod test-pod ; done
grep -P -o '(?<=real    0m)[^s]+'  | awk '{ sum += $1; n++ } END { if (n > 0) print sum / n; }'
```

`["/bin/bash", "-c", "sleep 2000000000000"]`
36.317s

`["/bin/bash", "-c", "trap 'kill $(jobs -p); exit 0' TERM ; sleep 2000000000000 & wait"]`
6.9327


I guess I can submit a diff PR to fix this in all the sleeping pods.

rbbratta: I imagine this applies to all cleanups so this is adding ~30 seconds to all our project deletes in each test case?

rbbratta: It looks like Pod termination doesn't delay project deletion, so this might not be that big of a deal unless we delete Pods in tests.

But it is a hidden gotcha that we should be aware of.

akostadinov: That's a good point. One question, can we quickly fix this by using `["/bin/sleep", "2000000000000"]? In this way PID 1 will be `sleep` so a SIGTERM should kill it and pod quickly be removed. WDYT?

rbbratta: `/bin/sleep` doesn't handle SIGTERM, so it will still take 30 seconds.  Almost nothing explicitly handles SIGTERM, so whatever binary is PID 1 needs to explicitly handle SIGTERM to die quickly.  Bash can trap SIGTERM, so it is the simplest wrapper.

`    command: ["sleep", "2000000000000"]`
real    0m34.111s


`    command: ["/bin/sleep", "2000000000000"]`
real    0m34.077s

rbbratta: https://man7.org/linux/man-pages/man7/pid_namespaces.7.html

> Likewise, a process in an ancestor namespace can—subject to the usual permission checks described in kill(2)—send signals to the "init" process of a child PID namespace only if the "init" process has established a handler for that signal.

I think this is it, since our `sleep` process does not have a handler for `SIGTERM`, as PID 1 in it's own namespace it does not receive `SIGTERM` from the ancestor namespace unless it has a handler.

akostadinov: Apparently one can learn something every day. Thanks. For better UX we do not ship any simple init in our products so we have to do ugly workarounds.

So I agree with your bash trap being the simplest approach here. Why did you comment with lgtm? Shouldn't that be fixed in the PR?

anuragthehatter: Sorry for the delay guys. @rbbratta  Your analysis seems right with the bash trap which will be a point to consider in future pod yamls but I also figured that we don't need this whole command ```command: ["/bin/bash", "-c", "sleep 2000000000000"]```  in multus pods anymore. Its redundant. As we know this is usually needed so that the pod just stays up "forever" and then we can kubectl exec foo -- /bin/bash or oc exec. But this was a legacy yaml we got from multus dev which initially pointed to image ```dougbtv/centos-network``` and but now pointing to our standard ```hello-sdn``` which already has an hello-openshift entrypoint. So we don't need this command at all. Its just redundant.  Test runs fine without that https://privatebin-it-iso.int.open.paas.redhat.com/?88769ce95bb48dbb#8UGZH2iZVp3TajcEEaFtymwY5CP1kKXfof5FYyxWsaEC

So I guess I can go ahead and remove this redundancy from other pods as well. Let me know

anuragthehatter: cc @weliang1

anuragthehatter: umm...anyways seems like #1640 is merged. Better not to marry go round and just fix this PR with suggested bash trap. Its okay to keep redundancy for future references sometime :) Thanks

akostadinov: @anuragthehatter , I'm fine to leave the default entrypoints when this doesn't cause immediate troubles as long as @rbbratta is ok with that.

anuragthehatter: @akostadinov yea its completely fine in whatever state its now after #1640 merge. this one is good to merge now.

weliang1: @anuragthehatter Good catch, agree that without using dougbtv/centos-network we can remove that redundancy

akostadinov:
/lgtm

@rbbratta , given your earlier comment I'll merge. If you find issues with current removed entrypoint, lets discuss in another PR.

testdata/networking/multus-cni/Pods/generic_multus_pod_net_raw.yaml
rbbratta: > But this was a legacy yaml we got from multus dev which initially pointed to image `dougbtv/centos-network` and but now pointing to our standard `hello-sdn` which already has an hello-openshift entrypoint. 

Whatever the entrypoint is, if it doesn't handle SIGTERM it will take >30s to die.  I'll try to track down the `hello-openshift` source, hopefully it already handles SIGTERM.

PR #1641, opened 2020-11-05
add rule for kibana

QiaolingTang:
For logging 4.7, same rule as https://github.com/openshift/verification-tests/pull/1635.


pruan-rht (commented):
/lgtm

PR #1642, opened 2020-11-05
fix cirucular refrence in arg and indentation issue

pruan-rht:
indentation was wrong
```
    def ready?(user: nil, quiet: false, cached: false)
    	res = get(user: user, quiet: quiet)
    	res[:success] = res[:parsed]["status"]["phase"] == "Succeeded" && res[:parsed]["status"]["reason"] == "InstallSucceeded" if res[:success]
    	return res
    end
```

### DIFF
--- a/lib/openshift/cluster_service_version.rb
+++ b/lib/openshift/cluster_service_version.rb
@@ -8,7 +8,9 @@ module BushSlicer
     # status phase=Succeeded and reason=InstallSucceeded;
     def ready?(user: nil, quiet: false, cached: false)
       res = get(user: user, quiet: quiet)
-      res[:success] = res[:parsed]["status"]["phase"] == "Succeeded" && res[:parsed]["status"]["reason"] == "InstallSucceeded" if res[:success]
+      if res[:success]
+        res[:success] = res[:parsed]["status"]["phase"] == "Succeeded" && res[:parsed]["status"]["reason"] == "InstallSucceeded"
+      end
       return res
     end
   end
--- a/lib/openshift/cluster_service_version.rb
+++ b/lib/openshift/cluster_service_version.rb
@@ -4,15 +4,12 @@ module BushSlicer
   class ClusterServiceVersion < ProjectResource
     RESOURCE = "clusterserviceversions"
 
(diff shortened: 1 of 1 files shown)


lib/openshift/cluster_service_version.rb
```diff
@@ -4,15 +4,12 @@ module BushSlicer
   class ClusterServiceVersion < ProjectResource
     RESOURCE = "clusterserviceversions"
 
-    # @return [BushSlicer::ResultHash] with :success depending on 
+    # @return [BushSlicer::ResultHash] with :success depending on
     # status phase=Succeeded and reason=InstallSucceeded;
-    def ready?(user: user, quiet: false, cached: false)
-    	res = get(user: user, quiet: quiet)
-    	if res[:success]
```
akostadinov, line 11: this has a purpose. If the `get` call failed then there is no `:parsed` element. In the new version, if `get`  failed, then no method for Nil class will be thrown.

lib/openshift/cluster_service_version.rb
```diff
@@ -4,15 +4,12 @@ module BushSlicer
...
-    		res[:success] = 
-    			res[:parsed]["status"]["phase"] == "Succeeded" && res[:parsed]["status"]["reason"] == "InstallSucceeded"
-    	end
-    	return res
+    def ready?(user: nil, quiet: false, cached: false)
+      res = get(user: user, quiet: quiet)
+      res[:success] = res[:parsed]["status"]["phase"] == "Succeeded" && res[:parsed]["status"]["reason"] == "InstallSucceeded" if res[:success]
```
akostadinov, line 11: Is there any change in logic? When `if` is at the end, it's not very visible. Can we keep using `if` on separate lines?

pruan-rht: no change in the logic, it was originally to fix the circular reference warning, but then I noticed the indentation issue (which only shows up in `cat`)

akostadinov:
Thank you.

/lgtm

lib/openshift/cluster_service_version.rb
pruan-rht: @akostadinov anyway, I've put back the original formatting with the indentation fix

PR #1643, opened 2020-11-05
update the ip bin path

zhaozhanqi:
1 update `/usr/sbin/ip` to `/sbin/ip`
2. update the net_admin_cap_pod.yaml file with origin image to use `conntrack`

@weliang1 @anuragthehatter @huiran0826 @rbbratta 
PTAL, thanks

### DIFF
--- a/features/networking/multus-ipv6.feature
+++ b/features/networking/multus-ipv6.feature
@@ -28,11 +28,11 @@ Feature: Multus-CNI ipv6 related scenarios
 
     # Check that the macvlan with mode bridge is added to the pod
     When I execute on the pod:
-      | /usr/sbin/ip | -d | link |
+      | ip | -d | link |
     Then the output should contain "net1"
     And the output should contain "macvlan mode bridge"
     When I execute on the pod:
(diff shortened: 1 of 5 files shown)


weliang1:
/LGTM
Better to have testing log link

pruan-rht:
@zhaozhanqi LG, can you perhaps add the reason for the update and whether or not it affects backward compatibilty?

zhaozhanqi:
> @zhaozhanqi LG, can you perhaps add the reason for the update and whether or not it affects backward compatibilty?
we changed the image and the ping path is in '/sbin/' of new image

anuragthehatter:
@zhaozhanqi That looks good to me unless if you think its okay to just use ```ip``` instead of full path. This will shield us in making such changes again in future as binary will point to default path always IMO.

liangxia:
> just use `ip` instead of full path

+1 for this

zhaozhanqi:
> @zhaozhanqi That looks good to me unless if you think its okay to just use `ip` instead of full path. This will shield us in making such changes again in future as binary will point to default path always IMO.

yes, have no idea the background why use the full path before.  use 'ip' should more flex 
updated all. thanks.

zhaozhanqi:
> > just use `ip` instead of full path
> 
> +1 for this
updated, thanks

PR #1645, opened 2020-11-06
auto case OCP-36287

zhaozhanqi:
@huiran0826 @anuragthehatter @weliang1 @rbbratta  PTAL, thanks
logs: 
http://pastebin.test.redhat.com/916187

### DIFF
--- a/features/networking/sdn.feature
+++ b/features/networking/sdn.feature
@@ -412,3 +412,18 @@ Feature: SDN related networking scenarios
       | since         | 30s                            |
     And the output should contain:
       | Not applying unsafe change: invalid configuration |
+
+  # @author zzhao@redhat.com
+  # @case_id OCP-36287
+  @admin
+  @destructive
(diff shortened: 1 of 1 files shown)


features/networking/sdn.feature
```diff
@@ -412,3 +412,18 @@ Feature: SDN related networking scenarios
...
       | Not applying unsafe change: invalid configuration |
+
+  # @author zzhao@redhat.com
+  # @case_id OCP-36287
+  @admin
+  @destructive
+  Scenario: Netnamespace should be recreated after deleting it before the project is deleted
+    Given the env is using "OpenShiftSDN" networkType
+    Given I have a project
+    And admin checks that the "<%= project.name %>" net_namespace exists
```
liangxia, line 423: Is admin here (and below) a must ? or we are OK to use `And I check that ...` ?

zhaozhanqi: yes, admin is a must since the netnamespace is cluster scope.

PR #1647, opened 2020-11-09
Auto OCP-35341

huiran0826:
Logs: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2493/console


features/networking/egress-ingress.feature
```diff
@@ -464,3 +464,24 @@ Feature: Egress-ingress related networking scenarios
...
+  Scenario: EgressNetworkPolicy maxItems is 1000
+    Given I have a project
+    Given I obtain test data file "networking/egressnetworkpolicy/egressnetworkpolicy_1000.yaml"
+    When I run the :create admin command with:
+      | f | egressnetworkpolicy_1000.yaml |
+      | n | <%= project.name %>           |
+    Then the step should succeed
+    And the output should contain:
+      | egressnetworkpolicy.network.openshift.io/default created |
+    And admin ensures "default" egress_network_policy is deleted
```
weliang1, line 480: @huiran0826 Not sure when creating the EgressNetworkPolicy the new iptable rule will also be created. If this is the case, besides checking EgressNetworkPolicy maxItems is 1000, should we also check 1000 new iptable rules are also created?

liangxia: Any updates/follow up ?

huiran0826: I did a check and there is no iptable rule created when creating the EgressNetworkPolicy. This case is from bug https://bugzilla.redhat.com/show_bug.cgi?id=1879910. And the fix is just bump the maxItems from 50 to 1000. Thanks for checking!

PR #1650, opened 2020-11-11
some rules changes

yapei:
Some element definition changes


yapei:
pass log Runner-v3#204644

PR #1652, opened 2020-11-11
ONEOCPDEPL-43: Add support for OSD4

apodhrad:

### DIFF
--- a/lib/bushslicer.rb
+++ b/lib/bushslicer.rb
@@ -16,6 +16,7 @@ module BushSlicer
   autoload :OpenStack, "launchers/openstack"
   autoload :VSphere, "launchers/v_sphere"
   autoload :Packet, "launchers/packet"
+  autoload :OCM, "launchers/ocm"
   autoload :EnvironmentLauncher, "launchers/environment_launcher"
   autoload :PolarShift, "polarshift/autoload"
 
--- a/lib/bushslicer.rb
+++ b/lib/bushslicer.rb
@@ -16,7 +16,7 @@ module BushSlicer
   autoload :OpenStack, "launchers/openstack"
   autoload :VSphere, "launchers/v_sphere"
   autoload :Packet, "launchers/packet"
-  autoload :OCM, "launchers/ocm"
+  autoload :OCMCluster, "launchers/o_c_m_cluster"
   autoload :EnvironmentLauncher, "launchers/environment_launcher"
   autoload :PolarShift, "polarshift/autoload"
 
--- a/lib/launchers/amz.rb
+++ b/lib/launchers/amz.rb
@@ -622,6 +622,11 @@ module BushSlicer
       ec2.client.config.credentials.secret_access_key
     end
 
+    # @return [String]
+    def account_id
+      client_sts.get_caller_identity.to_h[:account]
+    end
+
     # @return [Object] undefined
     def terminate_instance(instance)
       # we don't really have root permission to terminate, we'll just label it
--- a/lib/launchers/o_c_m_cluster.rb
+++ b/lib/launchers/o_c_m_cluster.rb
@@ -168,15 +168,14 @@ module BushSlicer
           raise "Unsupported OS"
         end
       end
-      #doesn't work
-      #File.open('/tmp/ocm', 'wb') do |file|
-      #  @result = Http.get(url: url) do |chunk|
-      #    file.write chunk
-      #  end
-      #end
-      shell("curl -L #{url} -o /tmp/ocm")
-      File.chmod(0775, '/tmp/ocm')
-      return '/tmp/ocm'
+      ocm_path = File.join(Host.localhost.workdir, 'ocm')
+      File.open(ocm_path, 'wb') do |file|
+        @result = Http.get(url: url, raise_on_error: true) do |chunk|
+          file.write chunk
+        end
+      end
+      File.chmod(0775, ocm_path)
+      return ocm_path
     end
 
     def shell(cmd, output = nil)
--- a/lib/launchers/o_c_m_cluster.rb
+++ b/lib/launchers/o_c_m_cluster.rb
@@ -14,13 +14,13 @@ require 'os'
 require 'http'
 
 module BushSlicer
-  class OCM
+  class OCMCluster
     include Common::Helper
 
     attr_reader :config, :ocm_cli
     attr_reader :token, :token_file, :url, :region, :version, :num_nodes, :lifespan, :cloud, :cloud_opts, :multi_az, :aws_account_id, :aws_access_key, :aws_secret_key
 
-    def initialize(**options) 
+    def initialize(**options)
       service_name = ENV['OCM_SERVICE_NAME'] || options[:service_name] || 'ocm'
       @opts = default_opts(service_name)&.merge options
       unless @opts
@@ -82,7 +82,7 @@ module BushSlicer
           end
           case @cloud_opts[:cloud_type]
           when "aws"
-            aws = Amz_EC2.new(service_name: @cloud)            
+            aws = Amz_EC2.new(service_name: @cloud)
             @aws_account_id = aws.account_id
             @aws_access_key = aws.access_key
             @aws_secret_key = aws.secret_key
--- a/lib/launchers/ocm.rb
+++ b/lib/launchers/ocm.rb
@@ -1,0 +1,296 @@
+#!/usr/bin/env ruby
+
+lib_path = File.expand_path(File.dirname(File.dirname(__FILE__)))
+unless $LOAD_PATH.any? {|p| File.expand_path(p) == lib_path}
+  $LOAD_PATH.unshift(lib_path)
+end
+
+require 'common'
+require 'json'
+require 'yaml'
+require 'tmpdir'
+require 'git'
+require 'os'
+require 'http'
+
+module BushSlicer
+  class OCM
+    include Common::Helper
+
+    attr_reader :config, :ocm_cli
+    attr_reader :token, :token_file, :url, :region, :version, :num_nodes, :lifespan, :cloud, :cloud_opts, :multi_az, :aws_account_id, :aws_access_key, :aws_secret_key
+
+    def initialize(**options) 
+      service_name = ENV['OCM_SERVICE_NAME'] || options[:service_name] || 'ocm'
+      @opts = default_opts(service_name)&.merge options
+      unless @opts
+        @opts = options
+      end
+
+      # OCM token is mandatory
+      # it can be defined by token or by token_file
+      @token = ENV['OCM_TOKEN'] || @opts[:token]
+      @token_file = @opts[:token_file]
+      unless @token
+        if @token_file
+          token_file_path = expand_private_path(@token_file)
+          @token = File.read(token_file_path)
+        else
+          raise "You need to specify OCM token by 'token' or by 'token_file'"
+        end
+      end
+
+      # region is mandatory
+      # in the future we can extend support for other clouds, e.g. GCP and ARO
+      @region = ENV['OCM_REGION'] || ENV['AWS_REGION'] || @opts[:region]
+
+      # url defines the OCM environment (prod, integration or stage)
+      # currently, the url is ignored as many teams use the stage environment
+      @url = ENV['OCM_URL'] || @opts[:url] || 'https://api.stage.openshift.com'
+
+      # openshift version is optional
+      @version = ENV['OCM_VERSION'] || ENV['OCP_VERSION'] || @opts[:version]
+
+      # number of worker nodes
+      # minimum is 2
+      # default value is 4
+      @num_nodes = ENV['OCM_NUM_NODES'] || @opts[:num_nodes]
+
+      # lifespan in hours
+      # default value is 24 hours
+      @lifespan = ENV['OCM_LIFESPAN'] || @opts[:lifespan]
+
+      # multi_az is optional
+      # default value is false
+      @multi_az = ENV['OCM_MULTI_AZ'] || @opts[:multi_az]
+
+      # BYOC (Bring Your Own Cloud)
+      # you can refer to already defined cloud in config.yaml
+      # currently, only AWS is supported
+      if ENV['AWS_ACCOUNT_ID'] && ENV['AWS_ACCESS_KEY'] && (ENV['AWS_SECRET_ACCESS_KEY'] || ENV['AWS_SECRET_KEY'])
+        # account_id will not be required once the following issue is fixed
+        # https://github.com/openshift-online/ocm-cli/issues/216
+        @aws_account_id = ENV['AWS_ACCOUNT_ID']
+        @aws_access_key = ENV['AWS_ACCESS_KEY']
+        @aws_secret_key = ENV['AWS_SECRET_ACCESS_KEY'] || ENV['AWS_SECRET_KEY']
+      else
+        @cloud = ENV['OCM_CLOUD'] || @opts[:cloud]
+        if @cloud
+          @cloud_opts = default_opts(@cloud)
+          unless @cloud_opts
+            raise "Cannot find cloud '#{cloud}' defined in '#{service_name}'"
+          end
+          case @cloud_opts[:cloud_type]
+          when "aws"
+            aws = Amz_EC2.new(service_name: @cloud)            
+            @aws_account_id = aws.account_id
+            @aws_access_key = aws.access_key
+            @aws_secret_key = aws.secret_key
+          end
+        end
+      end
+    end
+
+    # @param service_name [String] the service name of this openstack instance
+    #   to lookup in configuration
+    def default_opts(service_name)
+      return  conf[:services, service_name.to_sym]
+    end
+
+    private :default_opts
+
+    def to_seconds(string)
+      regex_m = /^(\d+)\s*(m|min|minutes|mins)+$/
+      regex_h = /^(\d+)\s*(h|hour|hours|hrs)+$/
+      regex_d = /^(\d+)\s*(d|day|days)+$/
+      regex_w = /^(\d+)\s*(w|week|weeks|wks)+$/
+      case string
+      when regex_m
+        return string.match(regex_m)[1].to_i * 60
+      when regex_h
+        return string.match(regex_h)[1].to_i * 60 * 60
+      when regex_d
+        return string.match(regex_d)[1].to_i * 24 * 60 * 60
+      when regex_w
+        return string.match(regex_w)[1].to_i * 7 * 24 * 60 * 60
+      else
+        raise "Cannot convert '#{string}' to seconds!"
+      end
+    end
+
+    # Generate a cluster data used for creating OSD cluster
+    def generate_cluster_data(name)
+      json_data = {
+        "name" => name,
+        "managed" => true,
+        "multi_az" => false,
+        "byoc" => false
+      }
+
+      if @multi_az
+        json_data.merge!({"multi_az" => @multi_az})
+      end
+
+      if @region
+        json_data.merge!({"region" => {"id" => @region}})
+      end
+
+      if @version
+        json_data.merge!({"version" => {"id" => "openshift-v#{@version}"}})
+      end
+
+      if @num_nodes
+        json_data.merge!({"nodes" => {"compute" => @num_nodes.to_i}})
+      end
+
+      if @lifespan
+        expiration = Time.now + to_seconds(@lifespan)
+        json_data.merge!({"expiration_timestamp" => expiration.strftime("%Y-%m-%dT%H:%M:%SZ")})
+      end
+
+      if @aws_account_id && @aws_access_key && @aws_secret_key
+        json_data.merge!({"aws" => {"account_id":@aws_account_id, "access_key_id":@aws_access_key, "secret_access_key":@aws_secret_key}})
+        json_data.merge!({"byoc" => true})
+      end
+
+      return json_data
+    end
+
+    def download_ocm_cli
+      url = ENV['OCM_CLI_URL']
+      unless url
+        url_prefix = ENV['OCM_CLI_URL_PREFIX'] || 'https://github.com/openshift-online/ocm-cli/releases/download/v0.1.46'
+        if OS.mac?
+          url = "#{url_prefix}/ocm-darwin-amd64"
+        elsif OS.linux?
+          url = "#{url_prefix}/ocm-linux-amd64"
+        else
+          raise "Unsupported OS"
+        end
+      end
+      #doesn't work
+      #File.open('/tmp/ocm', 'wb') do |file|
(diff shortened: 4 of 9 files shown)


pruan-rht:
@apodhrad what does OCM stands for?

akostadinov:
OpenShift Cluster Manager I guess. I not sure it belongs to IaaSes but having a class for it should be fine. Need to look at it thoroughly. Feel welcome to review as well.

apodhrad:
Yes, Openshift Cluster Manager. The implementation uses a bash script which is built on top of ocm-cli
https://github.com/openshift-online/ocm-cli

apodhrad:
Here is an example of config.yaml
```
  AWS-INT-QE:
    awscred: config/.awscred-int-qe
    config_opts:
      region: us-east-2
    cloud_type: aws
  ocm:
    token_file: config/.ocm-token-int-qe
    url: https://api.stage.openshift.com
    cloud: AWS-INT-QE
    region: eu-central-1
    flavor: m5.xlarge
    nodes: 4
    lifespan: 36
    cloud_type: ocm
```

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,197 @@
...
+      # region is mandatory
+      # in the future we can extend support for other clouds, e.g. GCP and ARO
+      @region = ENV['OCM_REGION'] || ENV['AWS_REGION'] || @opts[:region]
+
+      # url defines the OCM environment (prod, integration or stage)
+      # currently, the url is ignored as many teams use the stage environment
+      @url = ENV['OCM_URL'] || @opts[:url] || 'https://api.stage.openshift.com'
+
+      # openshift version is optional
+      @version = ENV['OCM_VERSION'] || @opts[:version]
```
akostadinov, line 48: is this version of installed OpenShift of OCM?

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,197 @@
...
+
+      # number of worker nodes
+      # minimum is 2
+      # default value is 4
+      @nodes = ENV['OCM_NODES'] || @opts[:nodes]
```
akostadinov, line 53: Seems like these variables select what cluster will be installed. I think variable names should be more like `OCM_INSTALL_NODES` or something like this.

My bigger concern though is that and OCM object is bound to an installed OCP cluster. I would expect to see an `OpenshiftClusterManager` class that contains interface to interact with OCM. And then an `OCMInstalledCluster` class which will represent an OCP cluster that has been installed via OCM.

In that way you will have separation of concerns about interfacing with OCM and operating on the cluster installed by it. e.g. first `OCMInstalledCluster` object will be created with the cluster definition where `ocm_installed_cluster.version` will return version you want,  `ocm_installed_cluster.lifespan` will return configured lifepan, etc. And the `OpenshiftClusterManager` class will be able to install, remove and manage clusters based on `OCMInstalledCluster` objects.

Makes sense or am I missing something?

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,197 @@
...
+        return string.match(regex_m)[1].to_i * 60
+      elsif string.match(regex_h)
+        return string.match(regex_h)[1].to_i * 60 * 60
+      elsif string.match(regex_d)
+        return string.match(regex_d)[1].to_i * 24 * 60 * 60
+      elsif string.match(regex_w)
+        return string.match(regex_w)[1].to_i * 7 * 24 * 60 * 60
+      else
+        raise "Cannot convert '#{string}' to seconds!"
+      end
```
akostadinov, line 96:
```
case string
when regex_m
  return string.match(regex_m)[1].to_i * 60
when regex_h
  ...
else
  ...
end
```

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,197 @@
...
+      File.open(ocm_token_file, "w") do |f|
+        f.write(@token)
+      end
+      # create a temp file with cluster specification
+      puts File.read(ocm_token_file.path)
+      ocm_json_file = Tempfile.new("ocm-json-file", Host.localhost.workdir)
+      File.open(ocm_json_file, "w") do |f|
+        data = generate_json(name)
+        f.write(data)
+      end
```
akostadinov, line 175:
```
File.write(ocm_json_file, generate_json(name))
```

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,197 @@
...
+
+    # download the script 'osd-provision.sh' which takes care of the OSD installation/uninstallation
+    def download_osd_script
+      default_osd_script_url = "https://gitlab.cee.redhat.com/mk-bin-packing/mk-performance-tests/-/raw/master/scripts/osd-provision.sh?inline=false"
+      osd_script_url = ENV['OSD_SCRIPT_URL'] || default_osd_script_url
+      # we need another parent dir as the script downloads ocm tool to its parent
+      osd_scripts = "#{Dir.tmpdir}/osd-scripts"
+      %x(
+        rm -rf #{osd_scripts} && mkdir #{osd_scripts} && \
+        curl -s #{osd_script_url} --output #{osd_scripts}/osd-provision.sh && \
```
akostadinov, line 150: Please never use `curl` and other native commands unless absolutely necessary and after discussion. Please see how to download file example in step `I download a file from` inside `features/step_definitions/net.rb` using method `BushSlicer::Http.get`. You can chmod file like `File.chmod(0600, ssh_key)`.

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,197 @@
...
+      ocm_json_file = Tempfile.new("ocm-json-file", Host.localhost.workdir)
+      File.open(ocm_json_file, "w") do |f|
+        data = generate_json(name)
+        f.write(data)
+      end
+      # now, download the script which will take care of the OSD cluster installation
+      osd_script = download_osd_script
+      shell("#{osd_script} --create --cloud-token-file #{ocm_token_file.path} -f #{ocm_json_file.path} --wait")
+      shell("#{osd_script} --get api_url -f #{ocm_json_file.path}")
+      shell("#{osd_script} --get credentials -f #{ocm_json_file.path}")
```
akostadinov, line 180: I see we can't avoid using shell commands given we depend on a script actually. But lets limit local command execution to this script only. Just a note, not request to change.

lib/launchers/ocm.rb
apodhrad: It is Openshift version in OCM. Probably we should think of better naming, e.g. `OCM_OCP_VERSION`

lib/launchers/ocm.rb
apodhrad: It makes sense and I'm ok to refactor current implementation in the way you have described.
I did not notice anything similar for `amz`, `azure`, `openstack`, ... that's why I did it in such a straightforward way.
But before refactoring it, I'd like to be sure that having this among launchers and having `cloud_type=ocm` is a good approach.

lib/launchers/ocm.rb
apodhrad: nice catch ;-) btw, wdyt about keeping the json file in the workspace?
Currently it is a temporary file which is deleted at the end of ruby execution.
Note that the json file contains AWS credentials (no other credentials).
Sometimes I found it useful to take a look in that file when I hit some errors.

lib/launchers/ocm.rb
apodhrad: I wasn't aware of the `BushSlicer::Http.get` so I will use it.

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,190 @@
...
+      git.clone
+      osd_script = File.join(osd_repo_dir, 'scripts', 'osd-provision.sh')
+      if !File.exists?(osd_script)
+        rais "Cannot find #{osd_script}"
+      end
+      return osd_script
+    end
+
+    def shell(cmd)
+      if !system(cmd)
```
akostadinov, line 158: This might be fine but to allow catching signals properly in Jenkins Job abort, it might be needed to use `Host.localhost.exec_raw()` method. I'm not sure yet if it makes any difference.

lib/launchers/ocm.rb
akostadinov: The other cloud classes deal with VMs, not clusters and one cloud management object creates potentially many VMs. In some classes we have separate VM classes, in some refer to them only with string names. You can keep OCP class in same file as the OCM class. 

I'm still not sure whether OCM should be among the IaaS providers. It is a quet different beast. On the other hand `cloud_type=ocm` sounds reasonable. Ideally I'd like to move IaaS providers to `/lib/iaas` but always keep on the back burner. 

I think you can keep file where it is now and it can be moved later if needed.

akostadinov: Would you include a sample configuration into `config/config.yaml`?

lib/launchers/ocm.rb
akostadinov: You better save it in `workdir`. Access by `Host.localhost.workdir`.

lib/launchers/ocm.rb
akostadinov: Yes, please use `Host.localhost.exec()` otherwise there are delays with with signal handling.
```
res = Host.localhost.exec(command, arg1, arg2, ..., env: env, single: true, stderr: :stdout, stdout: STDOUT, timeout: 3600)
if res[:success]
  ...
else
  ...
end
```

You can skip `env` if you don't want to change env. You can pass a single argument with full bash command I think as you are passing to `system` atm.

akostadinov: This will run command in workdir.

lib/launchers/ocm.rb
apodhrad: I have changed it to cloning full git repo.

lib/launchers/ocm.rb
akostadinov: At the end I figured out that th esignaling flakiness I have been observing were caused by the fact that Jenkins sends SIGTERM to all processes instead of the parent process and rely on signal forwarding. That's why using `system()` is alright. But it will run in whatever is current dir while I prefer to run all processes in the localhost workdir for consistency.

lib/launchers/ocm.rb
apodhrad: now, it is also possible to use `OCP_VERSION`

lib/launchers/ocm.rb
apodhrad:
```
  # AWS account dedicated to OSD testing
  aws-osd:
    awscred: config/.awscred-osd
    host_opts:
      user: root
      class: SSHAccessibleHost
      ssh_private_key: config/keys/integration-qe.pem
    config_opts:
      region: us-east-2
    cloud_type: aws
  # Openshift Cluster Management for Openshift Dedicated (OSD)
  ocm:
    token_file: config/.ocm-token
    url: https://api.stage.openshift.com
    cloud: aws-osd
    region: us-east-2
    flavor: m5.xlarge
    nodes: 4
    lifespan: 8h
    cloud_type: ocm
 ```

Comment (apodhrad) [reply lib/launchers/ocm.rb:54]: @akostadinov I was thinking of refactoring this OCM class as you have described and now I think that it is not needed. I do not implement anything like `OpenshiftClusterManager` as this is implemented by `ocm-cli` tool which I execute.
The class `ocm.rb` just takes a cluster configuration, transforms it to a json and posts this json via the `ocm-cli` tool.

File: lib/launchers/ocm.rb (hunk 3 above)
Comment (apodhrad) [reply lib/launchers/ocm.rb:98]: I have applied your suggestion

File: lib/launchers/ocm.rb (hunk 4 above)
Comment (apodhrad) [reply lib/launchers/ocm.rb]: I have applied your suggestion

File: lib/launchers/ocm.rb (hunk 6 above)
Comment (apodhrad) [reply lib/launchers/ocm.rb]: ok

File: lib/launchers/ocm.rb (hunk 7 above)
Comment (apodhrad) [reply lib/launchers/ocm.rb]: I use `Host.localhost.exec(` as you have suggested

File: lib/launchers/amz.rb [hunk 8]
```diff
@@ -52,6 +52,8 @@ def initialize(access_key: nil, secret_key: nil, service_name: nil, region: nil)
         )
       }) )
       Aws.config.update( config[:config_opts].merge({region: region})) if region
+
+      @account_id = awscred["aws_account_id"]
```
akostadinov, line 56: please remove this line, the ID should be obtained within the method, not at initialization time

akostadinov: Also it is not correct to put account id inside the awscred file. It is not standard part AFAIK.

lib/launchers/amz.rb
```diff
@@ -615,6 +617,11 @@ def secret_key
       ec2.client.config.credentials.secret_access_key
     end
 
+    # @return [String]
+    def account_id
+      return @account_id
```
akostadinov, line 622: Please change to `client.sts.get_caller_identity.to_h[:user_id]` , see what it returns in [1]. Or if by account id you mean `arn`, then use the `#arn` method directly. I'm not sure what exactly is needed because I can't see example of account ID you are using in the two PRs.

[1] https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/STS/Client.html#get_caller_identity-instance_method

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,238 @@
...
+
+      # multi_az is optional
+      # default value is false
+      @multi_az = ENV['OCM_MULTI_AZ'] || @opts[:multi_az]
+
+      # BYOC (Bring Your Own Cloud)
+      # you can refer to already defined cloud in config.yaml
+      # currently, only AWS is supported
+      if ENV['AWS_ACCOUNT_ID'] && ENV['AWS_ACCESS_KEY'] && (ENV['AWS_SECRET_ACCESS_KEY'] || ENV['AWS_SECRET_KEY'])
+        @aws_account_id = ENV['AWS_ACCOUNT_ID']
```
akostadinov, line 69: Are you sure OCM requires account_id? This is the first tool I see to request that. I commented about it above but still.

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,238 @@
...
+            @aws_access_key = aws.access_key
+            @aws_secret_key = aws.secret_key
+          end
+        end
+      end
+    end
+
+    # @param service_name [String] the service name of this openstack instance
+    #   to lookup in configuration
+    def default_opts(service_name)
```
akostadinov, line 92: make this method private

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,238 @@
...
+        json_data.merge!({"aws" => {"account_id":@aws_account_id, "access_key_id":@aws_access_key, "secret_access_key":@aws_secret_key}})
+        json_data.merge!({"byoc" => true})
+      end
+
+      return json_data.to_json
+    end
+
+    # download the script 'osd-provision.sh' which takes care of the OSD installation/uninstallation
+    def download_osd_script
+      osd_repo_uri = ENV['GIT_OSD_URI'] || 'https://gitlab.cee.redhat.com/mk-bin-packing/mk-performance-tests.git'
```
akostadinov, line 156: If this class relies on non-public resources, it is not really useful in a public open source project. If this class is needed for internal use cases, then it will need to be stored somewhere internally. Lets discuss this option in chat.

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,238 @@
...
+        res = Host.localhost.exec(cmd, single: true, timeout: 3600)
+      end
+      if res[:success]
+        return res[:response]
+      else
+        raise "Error when executing '#{cmd}'. Response: #{res[:response]}"
+      end
+    end
+
+    def get_last_line(text)
```
akostadinov, line 181: for small texts you can just do `str.lines.last` and not need a special method

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,238 @@
...
+        "ocp_api_url" => "https://api.#{domain}:6443",
+        "ocp_console_url" => "https://console-openshift-console.apps.#{domain}",
+        "user" => credentials["user"],
+        "password" => credentials["password"]
+      }
+      return ocp_info
+    end
+
+    # create OSD cluster
+    def create_osd(name)
```
akostadinov, line 207: probably makes sense to check if cluster exists already and/or keep status of that within the object
I mean before attempting an installation.

lib/launchers/ocm.rb
akostadinov: If this represents a asingle OCM cluster, then please rename class to `OCMCluster` and do not add it as an IaaS in `cloud_helper`. Given it relies on non-public script and probably moved to internal repo (e.g. the template repo), then it will not make sense anyway.

lib/launchers/ocm.rb
```diff
@@ -0,0 +1,238 @@
...
+      # currently, the url is ignored as many teams use the stage environment
+      @url = ENV['OCM_URL'] || @opts[:url] || 'https://api.stage.openshift.com'
+
+      # openshift version is optional
+      @version = ENV['OCM_VERSION'] || ENV['OCP_VERSION'] || @opts[:version]
+
+      # number of worker nodes
+      # minimum is 2
+      # default value is 4
+      @nodes = ENV['OCM_NODES'] || @opts[:nodes]
```
akostadinov, line 55: It will be more readable to rename `@nodes` to `@num_nodes`.

lib/launchers/ocm.rb
apodhrad: If an error occurs then it is returned by the `ocm-cli` tool. So, no need to extract some checks from that tool.

lib/launchers/ocm.rb
apodhrad: Yes, it is. This is the error when I omit it
```
Attribute 'aws.account_id' is mandatory on BYOC clusters
```

lib/launchers/ocm.rb
apodhrad: ok, I did not realized this is a public repo :(

lib/launchers/amz.rb
apodhrad: It is not standard but it doesn't break anything. Moving the id attribute somewhere else would make more mess IMHO.

apodhrad: btw, where should I move it?

lib/launchers/amz.rb
apodhrad: user_id is different from account_id... I don't know what is arn... OCM requires AWS account id (I'm not sure why) which is not included in any official AWS API library.

lib/launchers/ocm.rb
apodhrad: fixed

lib/launchers/ocm.rb
apodhrad: done

lib/launchers/ocm.rb
apodhrad: ok

lib/launchers/amz.rb
akostadinov: The awscred file is standard file used by aws official cli. Modifying standard files is more mess. You can't reuse existing AWS accounts in config.yaml.

lib/launchers/amz.rb
akostadinov: So give an example of it. It would be surprising (or impossible) not to be in any API because if it is in no API, then how could OCM know that such account exists or not?

I expect it to be of the format of the method I linked above. But if it is not, we can further dig.

lib/launchers/amz.rb
akostadinov: Also most likely we can figure that value out from API calls. Do we know any OCM dev that can tell use why is this value needed and how exactly do they use it?

lib/launchers/ocm.rb
akostadinov: So what do we do? We move this class to templates repo? Or any other ideas?

lib/launchers/amz.rb
akostadinov: After the chat, it seems like the body of this method can simply be 
```
    def account_id
      client.sts.get_caller_identity.to_h[:account]
    end
```
There is no need to ask users to provide this value. key and secret key are enough. Also this is a bug in OCM to require this value from the user. Should be reported IMO.

akostadinov: FYI I also created an upstream issue https://github.com/openshift-online/ocm-cli/issues/216 to have it resolved for everybody in the future.

apodhrad: @akostadinov that's awesome! I have just tried it and it works! Thanks!

lib/launchers/ocm.rb
apodhrad: I have refactored it so that it now directly uses https://github.com/openshift-online/ocm-cli

lib/launchers/amz.rb
apodhrad: I have removed it

akostadinov:
@apodhrad , I have moved stuff a little and fixed downloading by chunks. Let me know if you find any issues with the changes.

lib/launchers/o_c_m_cluster.rb
```diff
@@ -0,0 +1,295 @@
...
+      if res[:success]
+        return res[:response]
+      else
+        raise "Error when executing '#{cmd}'. Response: #{res[:response]}"
+      end
+    end
+
+    def exec(cmd)
+      unless @ocm_cli
+        @ocm_cli = download_ocm_cli
```
akostadinov, line 196: btw might be a good idea to first check `ocm` binary in local PATH and only then download

lib/bushslicer.rb
```diff
@@ -16,7 +16,7 @@ module BushSlicer
   autoload :OpenStack, "launchers/openstack"
   autoload :VSphere, "launchers/v_sphere"
   autoload :Packet, "launchers/packet"
-  autoload :OCM, "launchers/ocm"
+  autoload :OCMCluster, "launchers/o_c_m_cluster"
```
apodhrad, line 19: Hi @akostadinov is there any reason for using `o_c_m_cluster` instead of `ocm_cluster`?

lib/launchers/o_c_m_cluster.rb
apodhrad: Yes, I was thinking also about that as well. There are more such enhancements which could be done but they can be implemented later.

lib/bushslicer.rb
akostadinov: @apodhrad , it can work both ways but this is the convention from converting snake case to camel case in the projects, that's why I prefer to use it.

lib/launchers/ocm.rb
apodhrad: Done

apodhrad:
Hi @akostadinov the PR is ready for merge.

PR #1653, opened 2020-11-12
support kata operator installation and verification.

pruan-rht:


features/step_definitions/kata.rb
```diff
@@ -0,0 +1,88 @@
...
+    @result = user.cli_exec(:create_namespace, name: kata_ns)
+    raise "Failed to create namespace #{kata_ns}" unless @result[:success]
+  end
+  project(kata_ns)
+  iaas_type = env.iaas[:type] rescue nil
+  raise "Kata installation only supports GCE platform currently." if iaas_type != 'gcp'
+  # check to see if kata already exists
+  unless kata_config('example-kataconfig').exists?
+    # setup service account
+    role_yaml = "https://raw.githubusercontent.com/openshift/kata-operator/release-4.6/deploy/role.yaml"
```
liangxia, line 14: Hard code to 4.6 ? Will it supported in 4.7 ?

pruan-rht: @liangxia how 4.7 works is still not cleared, that's the only working model.

features/step_definitions/kata.rb
```diff
@@ -1,16 +1,88 @@
...
     @result = user.cli_exec(:create_namespace, name: kata_ns)
     raise "Failed to create namespace #{kata_ns}" unless @result[:success]
   end
   project(kata_ns)
-  # setup service account
-  role_yaml = "https://raw.githubusercontent.com/openshift/kata-operator/release-4.6/deploy/role.yaml"
-  role_binding_yaml = "https://raw.githubusercontent.com/openshift/kata-operator/release-4.6/deploy/role_binding.yaml"
-  sa_yaml = "https://raw.githubusercontent.com/openshift/kata-operator/release-4.6/deploy/service_account.yaml"
+  iaas_type = env.iaas[:type] rescue nil
+  raise "Kata installation only supports GCE platform currently." if iaas_type != 'gcp'
```
jensfr, line 10: Actually I think Azure works even better, it has nested virt enabled and you can actually run a kata container. Maybe not necessary for testing the installation only.

features/test/kata.feature
```diff
@@ -1,5 +1,16 @@
...
     Given kata container has been installed successfully in the "kata-operator" project
+    And I verify kata container runtime is installed into the a worker node
+
+  @admin
+  @destructive
+  @upgrade-prepare
+  Scenario: test delete kata installation
+    Given the master version >= "4.6"
+    Given I switch to cluster admin pseudo user
+    And I remove kata operator from "kata-operator" namespace
```
jensfr, line 16: Not sure if I read this correctly. But a successful deletion of the kataconfig should go like this: 1. delete kataconfig example-kataconfig CR (oc delete kataconfig example-kataconfig), this should block until the example-kataconfig is removed and it should also block until the worker machine-config-pool is updated  2. delete the kata-operator namespace (oc delete ns kata-operator).

features/step_definitions/kata.rb
pruan-rht: @jensfr ok, I have not tested with Azure yet.  I'll give it a try...if it works, we can filter out more laterr

features/test/kata.feature
pruan-rht: @jensfr that's what I'm doing within the step.

pruan-rht: @jensfr https://github.com/openshift/verification-tests/pull/1653/files/c0735ba128eb7726582a3e66227c0f1d8538095a#diff-119f8ac5278d90b899ce7a06bcdb073301fab26275a2d44b24964ade734a8eb1R51-R58

features/step_definitions/kata.rb
pruan-rht: @liangxia I've updated the file path to be keyed off by the master version, so that should make it more flexible.

features/step_definitions/kata.rb
pruan-rht: @jensfr I've verified that Azure works as well.  Updated the code to filter from a list of acceptable platforms.

features/step_definitions/kata.rb
```diff
@@ -0,0 +1,91 @@
...
+  })
+  # 2. check there's a process with the pod name `qemu`
+  step %Q/I switch to cluster admin pseudo user/
+  # node_cmd = "ps aux"
+  step %Q/I run the :debug client command with:/, table(%{
+    | resource     | node/<%= pod.node_name %> |
+    | oc_opts_end  |                           |
+    | exec_command | ps                        |
+    | exec_command | aux                       |
+  })
```
rbbratta, line 88: do we want `node(pod.node_name).host.exec_admin("ps aux")`  ?

akostadinov: Maybe it's time to write a `pod#node` method as `node(pod.node_name)` is too verbose. But need to be careful to select nodes from `env.nodes`. Maybe for another PR. Your suggestion is nice given current methods available.

pruan-rht: @akostadinov what is the difference between calling `oc debug node/xxx` vs `node(pod.node_name).host.exec_admin`?

rbbratta: `host.exec_admin` will always use separate `service_project` for debug pods and will use local pre-loaded multus image instead of `registry.redhat.io/rhel7/support-tools` to work around disconnected cluster issues.  https://bugzilla.redhat.com/show_bug.cgi?id=1728135

not sure if kata has other requirements.

cpmeadors (approved):
Great first set of tests.

pruan-rht:
@akostadinov @liangxia I've addressed the comments.  PTAL.  thx!

lib/openshift/kata_config.rb
```diff
@@ -0,0 +1,34 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  # represents Metering CRD
+  class KataConfig < ClusterResource
+    RESOURCE = 'kataconfigs'
```
akostadinov, line 6: @pruan-rht , I planned to review but procrastinated :/
you need to add api name to it.

pruan-rht: @akostadinov how can I find the api name from the `oc get kataconfig -o yaml` [output](http://pastebin.test.redhat.com/921121)?

akostadinov: @pruan-rht , see `selfLink`. In this case it is `kataconfiguration.openshift.io`. So you get:
```
RESOURCE = 'kataconfigs.kataconfiguration.openshift.io'
```

PR #1665, opened 2020-11-17
replace click_tab with new actions

yapei:
@yanpzhan  Could you help review? Thanks


yapei:
@pruan-rht  @liangxia  Could you help add lgtm label if it looks good to you?

PR #1666, opened 2020-11-17
Fix OCP-11695

kasturinarra:
Sometimes script proceeds when the latestVersion is 0 due to which the test script is failing with run time error, so fixing the issue. One example is as below
http://virt-openshift-05.lab.eng.nay.redhat.com/buildcorp/Runner-v3/198456/console


kasturinarra:
@zhouying7780 could you please help review this ? Thanks !!

https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2524/console

PR #1671, opened 2020-11-18
added support for testing report exipration parameter

pruan-rht:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2527/console


pruan-rht:
@akostadinov @liangxia PTAL...it's needed by https://github.com/openshift/cucushift/pull/8235

PR #1672, opened 2020-11-18
add step I wait for the last background process to finish

xingxingxia:
For Aleks's https://github.com/openshift/cucushift/pull/8229#discussion_r525081949
Pass log: Runner-v3/209171/console
@wangke19 @akostadinov help review/merge, thx.

### DIFF
--- a/features/step_definitions/cli.rb
+++ b/features/step_definitions/cli.rb
@@ -185,11 +185,7 @@ end
 
 Given /^I wait(?: up to #{NUMBER} seconds)? for the last background process to finish$/ do |seconds|
   seconds = Integer(seconds) rescue 60
-  success = wait_for(seconds) {
-    if @bg_processes.last.finished?
-      true
-    end
-  }
+  success = wait_for(seconds) { @bg_processes.last.finished? }
   raise "last process did not finish within #{seconds} seconds" unless success
 end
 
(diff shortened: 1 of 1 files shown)


features/step_definitions/cli.rb
```diff
@@ -183,6 +183,16 @@
     })
 end
 
+Given /^I wait(?: up to #{NUMBER} seconds)? for the last background process to finish$/ do |seconds|
+  seconds = Integer(seconds) rescue 60
```
pruan-rht, line 187: @xingxingxia if you are using the `#{NUMBER}` already `seconds` should be an `Integer` already.  I think you can use ` seconds ||= 60` instead of using the `rescue` clause.

akostadinov: It will be a number but still will be a String.

features/step_definitions/cli.rb
```diff
@@ -183,6 +183,16 @@
...
+  success = wait_for(seconds) {
+    if @bg_processes.last.finished?
+      true
+    end
+  }
```
akostadinov, line 192: you can simplify by replacing the 3 lines with `@bg_processes.last.finished?`. The `if` is redundant here.

wangke19:
@xingxingxia  Is there a maximum timeout limit here? I mean waiting time of 'up to #{NUMBER} seconds' with a limitation, such as 300 seconds.

akostadinov:
@wangke19 , what you choose is up to your responsibility only. And code reviewer :)

wangke19:
absolutely!

PR #1673, opened 2020-11-18
try to handle TERM/INT gracefully and generate vminfo.yml

akostadinov:
/assign @pruan-rht

### DIFF
--- a/tools/launch_instance.rb
+++ b/tools/launch_instance.rb
@@ -744,6 +744,8 @@ module BushSlicer
       end
 
       ## perform provisioning steps
+      org_term = Signal.trap('TERM') { raise "Received SIGTERM during installation." }
+      org_int = Signal.trap('INT') { raise "Received SIGINT during installation." }
       template[:install_sequence].each do |task|
         installation_task(
           task,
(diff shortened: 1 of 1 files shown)


pruan-rht (commented):
/lgtm

akostadinov:
@pruan-rht , it happened that I made some unintended changes that I reverted so your LGTM label was removed. But did you test this actually works before adding `/lgtm`?

akostadinov:
@pruan-rht , you can spin a build, abort the build in the middle and see if vminfo is generated. Make sure that abort happens after installation steps have begun.

akostadinov:
@pruan-rht , @jianlinliu , do you want to test this? Honestly I don't want to do myself because I need to then figure out how to remove the cluster resources in case it doesn't work :)

pruan-rht:
@akostadinov I'll take a look at it now using your branch.

pruan-rht:
@akostadinov hmm...it doesn't look like it worked.   https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Launch%20Environment%20Flexy/123941/.

akostadinov:
In fact it is not necessary to trap these because `ensure` block should execute anyway. It is the way that we start our ruby process and Jenkins sending the correct signal to allow vminfo to be generated on abort.

akostadinov:
Actually it seems to depend on `bash` mode we run installer as. So in some modes INT is ignored, in some modes TERM is ignored. To be sure proper signal handling is used instead of whatever we inherited from Bash, we need to `trap` explicitly.

PR #1685, opened 2020-11-20
A step to delete the buffer files in nodes for usage

anpingli:
@QiaolingTang  PTAL

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -705,3 +705,37 @@ Given /^I check the cronjob status$/ do
     end
   end
 end
+
+# Delete the fluentd buffer files in nodes. the fluentd must be deleted before running this step.
+Given /^logging collector bufferfiles are cleared on all nodes$/ do
+  ensure_admin_tagged
+  serviceaccount="collector-clean"
+  ds_name="collector-clean"
+  user(0)
+  step %Q/I have a project/
+
+  fluentd_pods = BushSlicer::Pod.get_labeled("logging-infra=fluentd", project: project("openshift-logging", switch: false), user: admin)
+  if(fluentd_pods.length()>0)
+    puts("The fluetnd pods is still running")
+    @result[":success"] = false
+  else
+    sa = service_account("#{serviceaccount}")
+    @result =  sa.create(by: admin)
+    step %Q/the step should succeed/
+
+    _opts = {scc: "privileged", user_name: "system:serviceaccount:#{project.name}:#{serviceaccount}"}
+    @result = admin.cli_exec(:oadm_policy_add_scc_to_user, **_opts)
+    step %Q/the step should succeed/
+    @result = user.cli_exec(:create, {f: "#{BushSlicer::HOME}/testdata/logging/collecter/collector_clear.json"})
+    step %Q/the step should succeed/
+    
+    pod_num=daemon_set(ds_name).replica_counters[:desired]
+    step %Q/#{pod_num} pods become ready with labels:/, table(%{
+      | logging-infra=#{ds_name} |
+    })
+
+    admin.cli_exec(:delete, object_type: 'daemonset', object_name_or_id: ds_name, n: project.name)
+    admin.cli_exec(:delete, object_type: 'sa', object_name_or_id: serviceaccount, n: project.name)
+    @result[":success"] = true
+  end
+end
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -707,7 +707,7 @@ Given /^I check the cronjob status$/ do
 end
 
 # Delete the fluentd buffer files in nodes. the fluentd must be deleted before running this step.
-Given /^logging collector bufferfiles are cleared on all nodes$/ do
+Given /^I clear all logging collector buffer files all nodes$/ do
   ensure_admin_tagged
   serviceaccount="collector-clean"
   ds_name="collector-clean"
@@ -736,6 +736,5 @@ Given /^logging collector bufferfiles are cleared on all nodes$/ do
 
     admin.cli_exec(:delete, object_type: 'daemonset', object_name_or_id: ds_name, n: project.name)
     admin.cli_exec(:delete, object_type: 'sa', object_name_or_id: serviceaccount, n: project.name)
-    @result[":success"] = true
   end
 end
--- a/testdata/logging/collecter/collector_clear.json
+++ b/testdata/logging/collecter/collector_clear.json
@@ -1,0 +1,83 @@
+{
+    "apiVersion": "apps/v1",
+    "kind": "DaemonSet",
+    "metadata": {
+        "labels": {
+            "component": "collector-clean",
+            "logging-infra": "collector-clean",
+            "provider": "aosqe"
+        },
+        "name": "collector-clean"
+    },
+    "spec": {
+        "selector": {
+            "matchLabels": {
+                "component": "collector-clean",
+                "logging-infra": "collector-clean",
+                "provider": "aosqe"
+            }
+        },
+        "template": {
+            "metadata": {
+                "labels": {
+                    "component": "collector-clean",
+                    "logging-infra": "collector-clean",
+                    "provider": "aosqe"
+                },
+                "name": "collector-clean"
+            },
+            "spec": {
+                "containers": [
+                    {
+                        "image": "quay.io/openshifttest/busybox@sha256:afe605d272837ce1732f390966166c2afff5391208ddd57de10942748694049d",
+                        "imagePullPolicy": "IfNotPresent",
+                        "name": "collector-clean",
+                        "securityContext": {
+                            "privileged": true
+                        },
+                        "command": ["/bin/sh"],
+                        "args": ["-c","rm -rf /var/lib/fluentd/*; sleep 2m"],
+                        "terminationMessagePath": "/dev/termination-log",
+                        "terminationMessagePolicy": "File",
+                        "volumeMounts": [
+                            {
+                                "mountPath": "/var/log",
+                                "name": "varlog"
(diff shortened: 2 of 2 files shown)


anpingli:
A result when  no fluent pods: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/210068/console
A result when fluent pods are running. https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/210069/console
Note: the test scenarios had been removed.

features/step_definitions/logging.rb
```diff
@@ -705,3 +705,37 @@
     end
   end
 end
+
+# Delete the fluentd buffer files in nodes. the fluentd must be deleted before running this step.
+Given /^logging collector bufferfiles are cleared on all nodes$/ do
```
pruan-rht, line 710: @anpingli perhaps change `bufferfiles` to `buffer files`

features/step_definitions/logging.rb
```diff
@@ -705,3 +705,37 @@
...
+  ensure_admin_tagged
+  serviceaccount="collector-clean"
+  ds_name="collector-clean"
+  user(0)
+  step %Q/I have a project/
+
+  fluentd_pods = BushSlicer::Pod.get_labeled("logging-infra=fluentd", project: project("openshift-logging", switch: false), user: admin)
+  if(fluentd_pods.length()>0)
+    puts("The fluetnd pods is still running")
+    @result[":success"] = false
```
QiaolingTang, line 720: I think here can use `raise "The fluentd pods is still running"`

features/step_definitions/logging.rb
```diff
@@ -705,3 +705,37 @@
...
+    step %Q/the step should succeed/
+    
+    pod_num=daemon_set(ds_name).replica_counters[:desired]
+    step %Q/#{pod_num} pods become ready with labels:/, table(%{
+      | logging-infra=#{ds_name} |
+    })
+
+    admin.cli_exec(:delete, object_type: 'daemonset', object_name_or_id: ds_name, n: project.name)
+    admin.cli_exec(:delete, object_type: 'sa', object_name_or_id: serviceaccount, n: project.name)
+    @result[":success"] = true
```
QiaolingTang, line 739: Out of curiosity, why do you add this line here?

features/step_definitions/logging.rb
```diff
@@ -705,3 +705,37 @@
...
   end
 end
+
+# Delete the fluentd buffer files in nodes. the fluentd must be deleted before running this step.
+Given /^logging collector bufferfiles are cleared on all nodes$/ do
+  ensure_admin_tagged
+  serviceaccount="collector-clean"
+  ds_name="collector-clean"
+  user(0)
+  step %Q/I have a project/
```
QiaolingTang, line 715: I don't see any steps to delete the project, will the project be deleted when the step/scenario finished?

features/step_definitions/logging.rb
```diff
@@ -705,3 +705,37 @@
...
+    step %Q/the step should succeed/
+    @result = user.cli_exec(:create, {f: "#{BushSlicer::HOME}/testdata/logging/collecter/collector_clear.json"})
+    step %Q/the step should succeed/
+    
+    pod_num=daemon_set(ds_name).replica_counters[:desired]
+    step %Q/#{pod_num} pods become ready with labels:/, table(%{
+      | logging-infra=#{ds_name} |
+    })
+
+    admin.cli_exec(:delete, object_type: 'daemonset', object_name_or_id: ds_name, n: project.name)
```
QiaolingTang, line 737: These 2 objects couldn't be removed if any one of the above steps is failed.
I think we'd better use `teardown_add`:
```
  teardown_add {
    admin.cli_exec(:delete, object_type: 'daemonset', object_name_or_id: ds_name, n: project.name)
    admin.cli_exec(:delete, object_type: 'sa', object_name_or_id: serviceaccount, n: project.name)
  }
```
And I think it's better to put this block above line 729, so these objects could be removed when the scenario finished.

features/step_definitions/logging.rb
```diff
@@ -705,3 +705,37 @@
...
+    puts("The fluetnd pods is still running")
+    @result[":success"] = false
+  else
+    sa = service_account("#{serviceaccount}")
+    @result =  sa.create(by: admin)
+    step %Q/the step should succeed/
+
+    _opts = {scc: "privileged", user_name: "system:serviceaccount:#{project.name}:#{serviceaccount}"}
+    @result = admin.cli_exec(:oadm_policy_add_scc_to_user, **_opts)
+    step %Q/the step should succeed/
```
QiaolingTang, line 728: Here should use something like:
```raise "some error messages" unless @result[:success]```

File: features/step_definitions/logging.rb [hunk 7]
```diff
@@ -705,3 +705,37 @@
...
+    @result = user.cli_exec(:create, {f: "#{BushSlicer::HOME}/testdata/logging/collecter/collector_clear.json"})
+    step %Q/the step should succeed/
```
QiaolingTang, line 730: Here should use something like:
```raise "some error messages" unless @result[:success]```

File: features/step_definitions/logging.rb [hunk 8]
```diff
@@ -705,3 +705,37 @@
...
+  step %Q/I have a project/
+
+  fluentd_pods = BushSlicer::Pod.get_labeled("logging-infra=fluentd", project: project("openshift-logging", switch: false), user: admin)
+  if(fluentd_pods.length()>0)
+    puts("The fluetnd pods is still running")
+    @result[":success"] = false
+  else
+    sa = service_account("#{serviceaccount}")
+    @result =  sa.create(by: admin)
+    step %Q/the step should succeed/
```
QiaolingTang, line 724: Here should use something like:
```raise "some error messages" unless @result[:success]```

Comment (QiaolingTang) [reply features/step_definitions/logging.rb:724]: Besides, I don't find any steps to remove this sa

File: features/step_definitions/logging.rb [hunk 9]
```diff
@@ -705,3 +705,37 @@
...
+    @result = user.cli_exec(:create, {f: "#{BushSlicer::HOME}/testdata/logging/collecter/collector_clear.json"})
+    step %Q/the step should succeed/
+    
+    pod_num=daemon_set(ds_name).replica_counters[:desired]
+    step %Q/#{pod_num} pods become ready with labels:/, table(%{
+      | logging-infra=#{ds_name} |
+    })
+
+    admin.cli_exec(:delete, object_type: 'daemonset', object_name_or_id: ds_name, n: project.name)
+    admin.cli_exec(:delete, object_type: 'sa', object_name_or_id: serviceaccount, n: project.name)
```
QiaolingTang, line 738: I don't think it's a good idea to use this step here. When we need to create some objects, it's better to add clean-up steps before creating the objects.

So here I suggest to add the step above line 723, you can use a step like:
```
admin ensures "eventrouter" service_account is deleted from the "openshift-logging" project after scenario
```

features/step_definitions/logging.rb
anpingli: i don't want to exit the test, so I don't use raise here.

features/step_definitions/logging.rb
anpingli: It is useful to run this step several times in on scenarios.

features/step_definitions/logging.rb
anpingli: Yes.  the cronjob will be running under one existing project if there is.

features/step_definitions/logging.rb
anpingli: The whole project will be deleted after the scenarios. I think we needn't put it in teardown block. If it is in tear down block. the delete steps will not be executed in tail

features/step_definitions/logging.rb
QiaolingTang: Then if the fluentd pods are running, the step won't do anything.

features/step_definitions/logging.rb
QiaolingTang: I suggest you to remove the step to create the project and suggest you to create the project before you call the step `I clear all logging collector buffer files all nodes`  in the scenario.

features/step_definitions/logging.rb
QiaolingTang: What's the usage of `@result[":success"] = false` ?

features/step_definitions/logging.rb
QiaolingTang: Then how to deal with the failures? If the step failed, but the scenario could go on, how can we ensure the test result is valid?

features/step_definitions/logging.rb
anpingli: @result[":success"] should be bool value.   If we set false here.  We can use the following step to check the status as usual.

Given /^the step (succeeded|failed)$/ do |outcome|
  if (outcome == "succeeded") != @result[:success]
    raise "the step #{@result[:success] ? "succeeded" : "failed"}"
  end
end

    raise "the step #{@result[:success] ? "succeeded" : "failed"}"
  end
end

features/step_definitions/logging.rb
QiaolingTang: No, the steps in teardown block will be executed after the scenario.

QiaolingTang: but better to use 
```
admin ensures "xxxx" daemon_set is deleted from the "xxxx" project after scenario
```

features/step_definitions/logging.rb
anpingli: I think ' step %Q/the step should succeed/' should be the similar with 'raise "some error messages" unless @result[:success]'. Of course, Raise can provide more clear message.

PR #1687, opened 2020-11-23
Map test case IDs from TCMS to Polarion, and clean-up test files

liangxia:
@dongboyan77 @pruan-rht @wewang58 

Also remove some unused test files.


liangxia:
The PR is ready for review now.

PR #1692, opened 2020-11-24
remove outdated code

QiaolingTang:
These 2 scenarios are outdated, I open this PR to remove them. 

@anpingli PTAL, thanks.


anpingli:
@QiaolingTang  Any challenge to fix it instead to drop these code?

QiaolingTang:
@anpingli , I think it's easy to update the code, but other logging cases can be easily affected by these 2 cases, what we need do in these 2 cases are:
1. delete existing logging operators
2. subscribe logging operators from console
3. delete the subscriptions from console
4. teardown step:  re-subscribe logging operators via CLI to ensure the cluster has logging operators installed
If the step 4 failed, then any logging cases executed after these 2 cases will failed.

anpingli:
@QiaolingTang  Got it.  Let's merge these two cases into the all-info-acceptance case. @gkarager

anpingli (commented):
/lgtm

pruan-rht (commented):
/lgtm

PR #1693, opened 2020-11-24
Add pdb

kasturinarra:
Adding poddisruptionbudget

### DIFF
--- a/lib/openshift/pod_disruption_budget.rb
+++ b/lib/openshift/pod_disruption_budget.rb
@@ -1,0 +1,6 @@
+module BushSlicer
+  class PodDisruptionBudget < ProjectResource
+    RESOURCE = "poddisruptionbudget.policy"
+
+  end
+end


kasturinarra:

akostadinov:
`RESOURCE` is missing api name. Could you post full output of `oc get -o yaml` of an example object and I will tell you how it should look.

akostadinov:
Should be
```
RESOURCE = "poddisruptionbudget.policy"
```

kasturinarra:
@akostadinov have updated, could you please help check ? Thanks !!

lib/openshift/pod_disruption_budget.rb
```diff
@@ -0,0 +1,6 @@
+module BushSlicer
+  class PodDisruptionBudget < ClusterResource
```
akostadinov, line 2: from output
```
namespace: openshift-kube-descheduler-operator
```

So this object should be `ProjectResource`, not `ClusterResource`.

kasturinarra: Thought so, but me missed to update as well, thanks !!

PR #1699, opened 2020-11-25
Fix case OCP-11762

kasturinarra:
I see that the file name used to run oc create -f is incorrect, so adding the right file name.


kasturinarra:

https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/212709/console

pruan-rht (commented):
/lgtm

PR #1701, opened 2020-11-26
Add update_caseruns_smart

liangxia:
/cc @akostadinov @pruan-rht 

Add retry for update_caseruns

### DIFF
--- a/lib/polarshift/request.rb
+++ b/lib/polarshift/request.rb
@@ -209,7 +209,6 @@ module BushSlicer
           end
         }
 
-
         Http.request(
           method: :post,
           url: "#{base_url}project/#{project_id}/test-cases/query",
@@ -218,7 +217,6 @@ module BushSlicer
           **common_opts
         )
       end
-
 
       def query_test_cases_smart(timeout: 360, **opts)
         res = query_test_cases(**opts)
@@ -305,6 +303,22 @@ module BushSlicer
           raise_on_error: false,
           **common_opts
         )
+      end
+
+      def update_caseruns_smart(project_id, run_id, updates, timeout: 360)
+        res = nil
+        success = wait_for(timeout, interval: 30) {
+          res = update_caseruns(project_id, run_id, updates)
+          if res[:success]
+            return res
+          else
+            logger.info %Q{got status "#{res[:exitstatus]}" updating test run "#{run_id}" in project "#{project_id}":\n#{res[:response]}}
+            return false
+          end
+        }
+        unless success
+          raise %Q{timeout ("#{timeout}" seconds) updating run "#{run_id}" in project "#{project_id}":\n#{res[:response]}}
+        end
       end
 
       # @param project_id [String]
--- a/lib/polarshift/request.rb
+++ b/lib/polarshift/request.rb
@@ -313,7 +313,7 @@ module BushSlicer
             return res
           else
             logger.info %Q{got status "#{res[:exitstatus]}" updating test run "#{run_id}" in project "#{project_id}":\n#{res[:response]}}
-            return false
+            false
           end
         }
         unless success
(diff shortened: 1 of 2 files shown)


lib/polarshift/request.rb
```diff
@@ -307,6 +305,16 @@ def update_caseruns(project_id, run_id, updates)
         )
       end
 
+      def update_caseruns_smart(project_id, run_id, updates, timeout: 360)
+        res = update_caseruns(project_id, run_id, updates)
+        if res[:exitstatus] == 202
+          wait_op(url: JSON.load(res[:response])["operation_result_url"],
+                timeout: timeout)
```
akostadinov, line 312: I think we shouldn't wait here. 

We rather need to repeat the operation a few times in case we get a server error 5xx. Otherwise we will not solve the issue reported in JIRA.

liangxia: @akostadinov   Updated.  Repeat the operation via wait_for, rather than wait the operation to finish.

liangxia:
@akostadinov  PR ready for review.

lib/polarshift/request.rb
```diff
@@ -307,6 +305,22 @@ def update_caseruns(project_id, run_id, updates)
...
 
+      def update_caseruns_smart(project_id, run_id, updates, timeout: 360)
+        res = nil
+        success = wait_for(timeout, interval: 30) {
+          res = update_caseruns(project_id, run_id, updates)
+          if res[:success]
+            return res
+          else
+            logger.warn %Q{got status "#{res[:exitstatus]}" updating test run "#{run_id}" in project "#{project_id}":\n#{res[:response]}}
+          end
```
akostadinov, line 316: Last statement needs to be `false` in case a repeat is needed. I'm not sure what `logger.warn` returns but we shouldn't rely on it.

liangxia: @akostadinov  Updated.

akostadinov: I'm not sure what changed. I don't see a `false` added anywhere. Did you push your change?

lib/polarshift/request.rb
```diff
@@ -307,6 +305,22 @@ def update_caseruns(project_id, run_id, updates)
         )
       end
 
+      def update_caseruns_smart(project_id, run_id, updates, timeout: 360)
+        res = nil
+        success = wait_for(timeout, interval: 30) {
+          res = update_caseruns(project_id, run_id, updates)
+          if res[:success]
+            return res
+          elsif res[:error]
```
akostadinov, line 314: I think a straight `else` should work here. Do you have any specific reason to use `elsif res[:error]`?

lib/polarshift/request.rb
```diff
@@ -307,6 +305,22 @@ def update_caseruns(project_id, run_id, updates)
...
+            logger.warn %Q{got status "#{res[:exitstatus]}" updating test run "#{run_id}" in project "#{project_id}":\n#{res[:response]}}
+          end
+        }
+        unless success
+          raise %Q{timeout ("#{timeout}" seconds) updating run "#{run_id}" in project "#{project_id}":\n#{res[:response]}}
+        end
+
```
akostadinov, line 321: this empty line not needed I think

lib/polarshift/request.rb
```diff
@@ -307,6 +305,22 @@ def update_caseruns(project_id, run_id, updates)
...
 
+      def update_caseruns_smart(project_id, run_id, updates, timeout: 360)
+        res = nil
+        success = wait_for(timeout, interval: 30) {
+          res = update_caseruns(project_id, run_id, updates)
+          if res[:success]
+            return res
+          else
+            logger.info %Q{got status "#{res[:exitstatus]}" updating test run "#{run_id}" in project "#{project_id}":\n#{res[:response]}}
+            false
```
akostadinov, line 316: In this case we should just leave `false` as the last statement. Return will return from the whole method. But this is Ruby madness so no worries. I fixed it. 

Thanks for implementing this feature!

PR #1706, opened 2020-11-27
handle Unable to locate element collection

yapei:
WIP

### DIFF
--- a/lib/webauto/web4cucumber.rb
+++ b/lib/webauto/web4cucumber.rb
@@ -556,10 +556,11 @@ require_relative 'chrome_extension'
         element = elements.first.first.last
         opres = handle_operation(element, op, **user_opts)
 
-        if opres[:response].include? "#<Watir::Exception::UnknownObjectException: unable to locate element"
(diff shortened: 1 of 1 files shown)


lib/webauto/web4cucumber.rb
```diff
@@ -556,10 +556,11 @@ def handle_element(element_rule, **user_opts)
         element = elements.first.first.last
         opres = handle_operation(element, op, **user_opts)
 
-        if opres[:response].include? "#<Watir::Exception::UnknownObjectException: unable to locate element"
+        if opres[:response].include? "Watir::Exception::LocatorException"
```
akostadinov, line 559: you think we never hit `UnknownObjectException` ?
Maybe we get it with Firefox?

It would be safer to keep check for it. But it is also fine if we remove it as you did above and we can revert back in case we hit that error in the future.

yapei: @akostadinov  good point, it would be safer to keep it, let me revert the changes

akostadinov:
LGTM, let me know if you want to merge immediately or what to test first. Or merge yourself if you can.

yapei:
@akostadinov this error is rarely seen in CI, I would like to keep it and watch

PR #1707, opened 2020-11-27
update case OCP-22202 to use admin user

zhaozhanqi:
this cases failed due to normal user cannot get machines in `clean up` steps
```
^[[36mError from server (Forbidden): machines.machine.openshift.io "qeci-11104-qkwcq-worker-us-east-2c-mclf6" is forbidden: User "testuser-40" cannot get resource "machines" in API group "machine.openshift.io" in the namespace "aos-qe-ci"^[[0m
```
Add one step to make use 'admin' user

logs: http://pastebin.test.redhat.com/921515
@anuragthehatter @huiran0826 @rbbratta @weliang1 ^^
@liangxia @pruan-rht  No big changes

### DIFF
--- a/features/networking/operator.feature
+++ b/features/networking/operator.feature
@@ -127,8 +127,12 @@ Feature: Operator related networking scenarios
   @destructive
   Scenario: The clusteroperator should be able to reflect the realtime status of the network when a new node added
     Given I have an IPI deployment
-    # Check that the operator is not progressing
-    Given the expression should be true> cluster_operator('network').condition(type: 'Progressing')['status'] == "False"
+    And I switch to cluster admin pseudo user
+    # Check that the operator is not progressing at the beginning to make sure the network operator is normal
+    Given I wait up to 160 seconds for the steps to pass:
+    """
+    Given the status of condition "Progressing" for network operator is :False
+    """
 
     # Record the original machine replica and scale it up to number +1
     Given I pick a random machineset to scale


features/networking/operator.feature
```diff
@@ -127,8 +127,12 @@ Feature: Operator related networking scenarios
...
   Scenario: The clusteroperator should be able to reflect the realtime status of the network when a new node added
     Given I have an IPI deployment
-    # Check that the operator is not progressing
-    Given the expression should be true> cluster_operator('network').condition(type: 'Progressing')['status'] == "False"
+    And I switch to cluster admin pseudo user
+    # Check that the operator is not progressing at the beginning to make sure the network operator is normal
+    Given I wait up to 160 seconds for the steps to pass:
+    """
+    Given the status of condition "Progressing" for network operator is :False
+    """
```
liangxia, line 135: Is it necessary to wrap a wait here ? I think we should fail immediately if the condition is not meet at the begging.

zhaozhanqi: hmm, sometimes last case may reboot the node or restart the network caused the multus or network pods are still in recreating. So here just double make sure the cluster is a normal. if the cluster has a problem, I guess the `progressing` still will be 'true` after 160 seconds.

liangxia: If every test case do wait like this, then the total execution time will be much longer.

zhaozhanqi: if the cluster is normal, then no need to wait up to 160 seconds since the processing is 'false'. 
if the cluster is not normal, then 160 seconds is needed for testing this case

liangxia: We should always assume the cluster is functional healthy. That's all the test cases (except the disaster recovery ones) should start with.

If everyone do assume the cluster is partly functional, and add wait logic in code to wait it to recover, the code will become really redundancy and un-efficiency. 
Let just imaging the worst situation, the cluster is broken, and all the cluster operator are keep restarting. We also need to wait 160 seconds. If we keep doing like this, then we will have more test cases do wait like this. Sooner or later, the execution time will be not bearable.

zhaozhanqi: > We should always assume the cluster is functional healthy. That's all the test cases (except the disaster recovery ones) should start with.

I definitely agree with you we should assume the cluster is well.   but the real world is we have destructive cases in our testing.  it may affect the result of operator 'processing' eg.  this scenario `Scaling a machineset with providerSpec.publicIp set to true`.   when this scenario is done and all node will be restored and be ready when it return pass, but even though the cluster of network operator 'processing' may still 'true' since the network operator need to update this status. it may need some seconds. I cannot sure how long it will cost. but I sometimes met this issue( the 'processing' has not updated to 'false' when running this case)


> 
> If everyone do assume the cluster is partly functional, and add wait logic in code to wait it to recover, the code will become really redundancy and un-efficiency.

> Let just imaging the worst situation, the cluster is broken, and all the cluster operator are keep restarting. We also need to wait 160 seconds. If we keep doing like this, then we will have more test cases do wait like this. Sooner or later, the execution time will be not bearable.

if you think 160 is too long.  hmm, I think 10 seconds is needed at least for this case unless we can make sure the operator status is normal before executing this case.

liangxia: I think i got your concern now.

zhaozhanqi: thanks Liang

PR #1710, opened 2020-11-30
fix ocp-33704 ocp-34938

huiran0826:
ocp-33704 sometimes failed as the hello-pod may located on same node as reboot node which will cause the pod killed after reboot, fix it by explicitly specifying  the node for the pod .
ocp-34938 failed as it get the event from project aos-qe-ci namespace which is incorrect,  fix it by explicitly using the default project to get event.

Logs:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2622/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2621/console

### DIFF
--- a/features/networking/egress-ip.feature
+++ b/features/networking/egress-ip.feature
@@ -121,12 +121,11 @@ Feature: Egress IP related features
   @destructive
   Scenario: Should remove the egressIP from the array if it was not being used
     Given I store a random unused IP address from the reserved range to the clipboard
-    And evaluation of `IPAddr.new("<%= cb.valid_ip %>").to_i + 1 ` is stored in the :newipint clipboard
-    And evaluation of `IPAddr.new(<%= cb.newipint %>, Socket::AF_INET).to_s` is stored in the :newip clipboard
+    And evaluation of `IPAddr.new("<%= cb.subnet_range %>").to_s+"/"+IPAddr.new("<%= cb.subnet_range %>").prefix.to_s` is stored in the :valid_subnet clipboard
 
     #Patch egress cidr to the node
     Given as admin I successfully merge patch resource "hostsubnet/<%= node.name %>" with:
-      | {"egressCIDRs": ["<%= cb.subnet_range %>"] }   |
+      | {"egressCIDRs": ["<%= cb.valid_subnet %>"] }   |
     And I register clean-up steps:
     """
     as admin I successfully merge patch resource "hostsubnet/<%= node.name %>" with:
@@ -136,18 +135,30 @@ Feature: Egress IP related features
     # Patch egress IP to the project twice
     Given I have a project
     And as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
-      | {"egressIPs": ["<%= cb.newip %>"]} |
-    And as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
-      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+      | {"egressIPs": ["<%= cb.valid_ips[0]%>"]} |
+
+    And I wait up to 30 seconds for the steps to pass:
+    """
+    When I run command on the "<%= node.name%>" node's sdn pod:
+      | bash | -c | ip address show label <%= cb.interface %>:eip |
+    Then the step should succeed
+    And evaluation of `@result[:response].chomp.match(/inet.*eip/)[0].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :egress_ip clipboard
+    Then the expression should be true> cb.egress_ip == cb.valid_ips[0]
+    Then the expression should be true> cb.egress_ip != cb.valid_ips[1]
+    """
+
+    Given as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
+      | {"egressIPs": ["<%= cb.valid_ips[1] %>"]} |
 
     # Check the egress ip is the last one applied
     And I wait up to 30 seconds for the steps to pass:
     """
     When I run command on the "<%= node.name%>" node's sdn pod:
-      | bash | -c | ip address show <%= cb.interface %> |
+      | bash | -c | ip address show label <%= cb.interface %>:eip |
     Then the step should succeed
-    And the output should contain "<%= cb.valid_ip %>"
-    And the output should not contain "<%= cb.newip %>"
+    And evaluation of `@result[:response].chomp.match(/inet.*eip/)[0].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :egress_new_ip clipboard
+    Then the expression should be true> cb.egress_new_ip == cb.valid_ips[1]
+    Then the expression should be true> cb.egress_new_ip != cb.valid_ips[0]
     """
 
   # @author huirwang@redhat.com
--- a/features/networking/ovn-egress-ip.feature
+++ b/features/networking/ovn-egress-ip.feature
@@ -356,7 +356,7 @@ Feature: OVN Egress IP related features
   @destructive
   Scenario: After reboot node or reboot OVN services EgressIP still work
     Given I save ipecho url to the clipboard
-    Given I store the schedulable nodes in the :nodes clipboard
+    Given I store the schedulable workers in the :nodes clipboard
     Then label "k8s.ovn.org/egress-assignable=true" is added to the "<%= cb.nodes[0].name %>" node
 
     #Get unused IP as egress ip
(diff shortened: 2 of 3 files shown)


huiran0826:
Fixed one more case OCP-21812.  As fix of the bug https://bugzilla.redhat.com/show_bug.cgi?id=1878624, CIDR like 136.144.52.247/24  is considered as not in canonical form, cannot be applied to egressCIDR, need use form like 136.144.52.0/24.

Logs: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/215338/console

features/networking/egress-ip.feature
```diff
@@ -136,18 +135,18 @@ Feature: Egress IP related features
...
     And I wait up to 30 seconds for the steps to pass:
     """
     When I run command on the "<%= node.name%>" node's sdn pod:
       | bash | -c | ip address show <%= cb.interface %> |
     Then the step should succeed
-    And the output should contain "<%= cb.valid_ip %>"
-    And the output should not contain "<%= cb.newip %>"
+    And the output should contain "<%= cb.valid_ips[1] %>"
+    And the output should not contain "<%= cb.valid_ips[0] %>"
     """
```
weliang1, line 150: @huiran0826 To make sure the last one is applied, I would suggest to check both valid_ip[0] and valid_ip[1]:

First step apply {"egressIPs": ["<%= cb.valid_ips[0]%>"]} , then cb.valid_ips[0] will be shown in pod,

Second step apply {"egressIPs": ["<%= cb.valid_ip[1] %>"]}, then cb.valid_ips[1] will be shown in pod

huiran0826: @weliang1 Thanks, updated it.

features/networking/egress-ip.feature
```diff
@@ -136,18 +135,28 @@ Feature: Egress IP related features
...
-      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+      | {"egressIPs": ["<%= cb.valid_ips[0]%>"]} |
+
+    And I wait up to 30 seconds for the steps to pass:
+    """
+    When I run command on the "<%= node.name%>" node's sdn pod:
+      | bash | -c | ip address show <%= cb.interface %> |
+    Then the step should succeed
+    And the output should contain "<%= cb.valid_ips[0] %>"
+    And the output should not contain "<%= cb.valid_ips[1] %>"
```
rbbratta, line 146: do we ever have collisions with substrings?  Maybe we should add `/` to the end of the string match to delimit it.

for example
```
valid_ips[0] = 192.168.1.212
valid_ips[1] = 192.168.1.2  
```

will `192.168.1.2` substring match `192.168.1.212`?  

maybe this never happens for some other reason?

Only risk is when the last octet is `.1` and `.2` I guess.

huiran0826: Thanks @rbbratta ,  hmmm, this is a kind of corner case but may actually happen.  Updated code use equal to exactly match.

features/networking/egress-ip.feature
```diff
@@ -121,12 +121,11 @@ Feature: Egress IP related features
   @destructive
   Scenario: Should remove the egressIP from the array if it was not being used
     Given I store a random unused IP address from the reserved range to the clipboard
```
rbbratta, line 123: This fails for me when there is an error message on stderr, because we chomp on `@result[:response]` which combines stdout and stderr.


```
      [17:48:06] INFO> Shell Commands: oc exec sdn-546bz  --kubeconfig=/home/rbrattai/Documents/deployments/127342/workdir/ocp4_admin.kubeconfig -n openshift-sdn  --container=sdn -i -- bash -c ip\ -4\ -brief\ a\ show\ \"ens5\"\ \|\ awk\ \'\{print\ \$3\}\'
      10.0.136.130/19
      
      STDERR:
      E1215 12:48:06.767680  392724 v2.go:105] write tcp 172.16.16.104:57442->3.97.80.33:6443: use of closed network connection
      
      [17:48:06] INFO> Exit Status: 0
      [17:48:06] INFO> Subnet range for the primary interface on the node is stored in the subnet_range clipboard.
      invalid address: 19
    
```
This fails because we are trying `IPAddr.new("10.0.136.130/19\n\nSTDERR:\nE1215 18:25:29.802564  488056 v2.go:105] read /dev/stdin: resource temporarily unavailable").to_range.to_a.shuffle.each`




We need to fix the step to use `cb[cb_name] = @result[:stdout].chomp` and we also don't need the extra string interpolation for `cb.subnet_range`


```diff
diff --git a/features/step_definitions/networking.rb b/features/step_definitions/networking.rb
index f42843ac..2222c60e 100644
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -528,7 +528,7 @@ Given /^I store a random unused IP address from the reserved range to the#{OPT_S
   cb_name = "valid_ip" unless cb_name
   step "the subnet for primary interface on node is stored in the clipboard"

-  reserved_range = "#{cb.subnet_range}"
+  reserved_range = cb.subnet_range

   unused_ips=[]
   #Save four unused ip in the clipboard
@@ -827,7 +827,7 @@ Given /^the subnet for primary interface on node is stored in the#{OPT_SYM} clip
     "| bash | -c | ip -4 -brief a show \"<%= cb.interface %>\" \\| awk '{print $3}' |"
   )
   raise "Failed to get the subnet range for the primary interface on the node" unless @result[:success]
-  cb[cb_name] = @result[:response].chomp
+  cb[cb_name] = @result[:stdout].chomp
   logger.info "Subnet range for the primary interface on the node is stored in the #{cb_name} clipboard."
 end
```

features/networking/egress-ip.feature
```diff
@@ -136,18 +135,30 @@ Feature: Egress IP related features
...
     And as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
-      | {"egressIPs": ["<%= cb.newip %>"]} |
-    And as admin I successfully merge patch resource "netnamespace/<%= project.name %>" with:
-      | {"egressIPs": ["<%= cb.valid_ip %>"]} |
+      | {"egressIPs": ["<%= cb.valid_ips[0]%>"]} |
+
+    And I wait up to 30 seconds for the steps to pass:
+    """
+    When I run command on the "<%= node.name%>" node's sdn pod:
+      | bash | -c | ip address show <%= cb.interface %> |
```
rbbratta, line 143: all egress IPs have the `:eip` suffix label on the device?  And they all have `secondary` flag?

```
      2: ens5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 qdisc mq state UP group default qlen 1000
          link/ether 0e:82:96:67:cf:6c brd ff:ff:ff:ff:ff:ff
          inet 10.0.222.222/19 brd 10.0.223.255 scope global dynamic noprefixroute ens5
             valid_lft 3428sec preferred_lft 3428sec
          inet 10.0.214.169/19 brd 10.0.223.255 scope global secondary ens5:eip
             valid_lft forever preferred_lft forever
```

It looks like we can lookup by label
```
ip a show label ens5:eip
```

It is also possible to lookup all addrs with the `secondary` flag.
```
ip a show secondary
```
or
```
ip a show ens5 secondary
```
It looks like we use the `:eip` label on purpose
https://github.com/openshift/sdn/commit/88d43b6d0e8b02a6ece88076d3e1eb28418bb74d#diff-74ce0259328d8a7c7c720e38bd60d2d97df35c3e9cf9acd7fede641fdf4a2b9aR95

```diff
diff --git a/features/networking/egress-ip.feature b/features/networking/egress-ip.feature
index 310c24e5..6449de17 100644
--- a/features/networking/egress-ip.feature
+++ b/features/networking/egress-ip.feature
@@ -140,7 +140,7 @@ Feature: Egress IP related features
     And I wait up to 30 seconds for the steps to pass:
     """
     When I run command on the "<%= node.name%>" node's sdn pod:
-      | bash | -c | ip address show <%= cb.interface %> |
+      | bash | -c | ip address show label <%= cb.interface %>:eip |
     Then the step should succeed
     And evaluation of `@result[:response].chomp.match(/inet.*eip/)[0].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :egress_ip clipboard
     Then the expression should be true> cb.egress_ip == cb.valid_ips[0]
@@ -154,7 +154,7 @@ Feature: Egress IP related features
     And I wait up to 30 seconds for the steps to pass:
     """
     When I run command on the "<%= node.name%>" node's sdn pod:
-      | bash | -c | ip address show <%= cb.interface %> |
+      | bash | -c | ip address show label <%= cb.interface %>:eip |
     Then the step should succeed
     And evaluation of `@result[:response].chomp.match(/inet.*eip/)[0].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :egress_new_ip clipboard
     Then the expression should be true> cb.egress_new_ip == cb.valid_ips[1]
```


```
      [23:40:18] INFO> Exit Status: 0
      [23:40:18] INFO> Shell Commands: oc exec sdn-546bz  --kubeconfig=/home/rbrattai/Documents/deployments/127342/workdir/ocp4_admin.kubeconfig -n openshift-sdn  --container=sdn -i -- bash -c ip\ address\ show\ label\ ens5:eip
          inet 10.0.148.29/19 brd 10.0.159.255 scope global secondary ens5:eip
             valid_lft forever preferred_lft forever
      
```

features/networking/egress-ip.feature
huiran0826: @rbbratta Thanks a lot!  Learned a lot from you!

features/networking/egress-ip.feature
huiran0826: Thanks Ross, updated it!

features/networking/ovn-egress-ip.feature
```diff
@@ -419,6 +427,8 @@ Feature: OVN Egress IP related features
     #Check events
     Then I wait up to 30 seconds for the steps to pass:
     """
+    Given I switch to cluster admin pseudo user
+    And I use the "default" project
```
liangxia, line 431: The added steps can be put outside of 'wait up to', as no need to duplicate those steps.

huiran0826: Thanks @liangxia ,  you are right, updated it!

PR #1712, opened 2020-11-30
removing unwanted runlevel block

anuragthehatter:
Setting run-level is not needed not in 4.7. Small change

@zhaozhanqi @huiran0826 @rbbratta @weliang1


pruan-rht (commented):
/lgtm

PR #1713, opened 2020-12-01
Fix ocm install page case and add new cases

tzhou5:


tzhou5:
New PR 1744 created, so close this one.

PR #1648, opened 2020-11-10
Added wait_for_tcp for ldap

rhpmali:
@xingxingxia @barleyer please review. Added wait_for_tcp for ldap

Job link: Runner-v3-smoke/2497/console

PR #1719, opened 2020-12-03
Make sure some conditions of the operator are ready

wangke19:
@xingxingxia, Please help to review the PR, here is test results: Runner-v3-smoke/2641/console

### DIFF
--- a/features/step_definitions/operators.rb
+++ b/features/step_definitions/operators.rb
@@ -160,3 +160,49 @@ Given /^the major.minor version of the cluster is stored in the#{OPT_SYM} clipbo
   cb_name = 'operator_channel_name' unless cb_name
   cb[cb_name] = cluster_version('version').channel.split('-')[1]
 end
+
+Given /^operator #{QUOTED} becomes "([^"]*)" within(?: (\d+) seconds)?$/ do | operator_name, conditions, timeout |
+  ensure_admin_tagged
+
+  expected = {}
+  interval_time = 5
+  timeout = Integer(timeout) rescue 60
+  interval_time = 20 if timeout > 100
+  actual_results = {}
+  stats = {}
+ 
+  # Parse the conditions to Hash table {Available"=>"True", "Progressing"=>"False", "Degraded"=>"False"}
+  arr_conditions = conditions.split("/")
+  arr_conditions.each do |v|
+    case v
+    when /^(available|non-available)$/
+      (v.include? "non-") ? expected["Available"] = "False" : expected["Available"] = "True"
+    when /^(progressing|non-progressing)$/
+      (v.include? "non-") ? expected["Progressing"] = "False" : expected["Progressing"] = "True"
+    when /^(degraded|non-degraded)$/
+      (v.include? "non-") ? expected["Degraded"] = "False" : expected["Degraded"] = "True"
+    else
+      raise "#### Invalid condition: #{v}, please input one or more condition(s) of (available|non-available)/(progressing|non-progressing)/(degraded|non-degraded) !"
+    end                                 
+  end 
+  raise "#### Without any conditions, please input at least one condition!!!" unless expected.length >0
+
+  begin
+    # Log to StdOut dedup
+    logger.dedup_start
+    success = wait_for(timeout, interval: interval_time, stats: stats) {
+      # Does not get conditions data from the cache
+      current_conditions = cluster_operator(operator_name).conditions(cached: false)
+      expected.keys.each do |c|
+        actual_results[c] = current_conditions.select{ |t| t['type'] == c }.first['status']
+      end
+      expected == actual_results
+    }
+    logger.info("#### Operator #{operator_name} Expected conditions: #{expected}")
+    logger.info "#### After #{stats[:seconds]} seconds and #{stats[:iterations]} iterations " <<
+      "operator #{operator_name} becomes: #{actual_results}" if success
+    raise "The #{operator_name} operator still didn't become #{expected} after #{timeout} seconds" unless success
+  ensure
+    logger.dedup_flush
+  end
+end
--- a/features/step_definitions/operators.rb
+++ b/features/step_definitions/operators.rb
@@ -161,7 +161,7 @@ Given /^the major.minor version of the cluster is stored in the#{OPT_SYM} clipbo
   cb[cb_name] = cluster_version('version').channel.split('-')[1]
 end
 
-Given /^operator #{QUOTED} becomes "([^"]*)" within(?: (\d+) seconds)?$/ do | operator_name, conditions, timeout |
+Given /^operator #{QUOTED} becomes #{NO_SPACE_STR}(?: within #{NUMBER} seconds)?$/ do | operator_name, conditions, timeout |
   ensure_admin_tagged
 
   expected = {}
@@ -170,7 +170,7 @@ Given /^operator #{QUOTED} becomes "([^"]*)" within(?: (\d+) seconds)?$/ do | op
   interval_time = 20 if timeout > 100
   actual_results = {}
   stats = {}
- 
+
   # Parse the conditions to Hash table {Available"=>"True", "Progressing"=>"False", "Degraded"=>"False"}
   arr_conditions = conditions.split("/")
   arr_conditions.each do |v|
@@ -183,8 +183,8 @@ Given /^operator #{QUOTED} becomes "([^"]*)" within(?: (\d+) seconds)?$/ do | op
       (v.include? "non-") ? expected["Degraded"] = "False" : expected["Degraded"] = "True"
     else
       raise "#### Invalid condition: #{v}, please input one or more condition(s) of (available|non-available)/(progressing|non-progressing)/(degraded|non-degraded) !"
-    end                                 
-  end 
+    end
+  end
   raise "#### Without any conditions, please input at least one condition!!!" unless expected.length >0
 
   begin
--- a/features/test/apiserver.feature
+++ b/features/test/apiserver.feature
@@ -1,0 +1,27 @@
+Feature: API server Test features
+  # @author kewang@redhat.com
+  @admin
+  @destructive
+  Scenario: Quick test for step definitions
+    Given the master version >= "4.6"
+    Given I switch to cluster admin pseudo user      
+
+    # Set WriteRequestBodies profile to audit log
+    Given as admin I successfully merge patch resource "apiserver/cluster" with:
+      | {"spec": {"audit": {"profile": "WriteRequestBodies"}}} |
+    And I register clean-up steps:
+    # Set original Default profile to audti log
+    """
+    Given as admin I successfully merge patch resource "apiserver/cluster" with:
+      | {"spec": {"audit": {"profile": "Default"}}} |
+    Given operator "kube-apiserver" becomes "progressing" within 100 seconds
+    Given operator "kube-apiserver" becomes "available/non-progressing/non-degraded" within 1200 seconds
+    """
+    Given operator "kube-apiserver" becomes "progressing" within 100 seconds
+    Given operator "kube-apiserver" becomes "available/non-progressing/non-degraded" within 1200 seconds
+    Given operator "kube-apiserver" becomes "available/non-progressing" within 100 seconds
+    Given operator "kube-apiserver" becomes "available/non-degraded" within 100 seconds
+    Given operator "kube-apiserver" becomes "available" within 100 seconds
+    Given operator "kube-apiserver" becomes "non-progressing" within 100 seconds
+    Given operator "kube-apiserver" becomes "non-degraded" within 100 seconds
+    Given operator "kube-apiserver" becomes "" within 100 seconds
--- a/features/test/apiserver.feature
+++ b/features/test/apiserver.feature
@@ -4,7 +4,7 @@ Feature: API server Test features
   @destructive
   Scenario: Quick test for step definitions
     Given the master version >= "4.6"
-    Given I switch to cluster admin pseudo user      
+    Given I switch to cluster admin pseudo user
 
     # Set WriteRequestBodies profile to audit log
     Given as admin I successfully merge patch resource "apiserver/cluster" with:
@@ -17,11 +17,10 @@ Feature: API server Test features
     Given operator "kube-apiserver" becomes "progressing" within 100 seconds
     Given operator "kube-apiserver" becomes "available/non-progressing/non-degraded" within 1200 seconds
     """
-    Given operator "kube-apiserver" becomes "progressing" within 100 seconds
-    Given operator "kube-apiserver" becomes "available/non-progressing/non-degraded" within 1200 seconds
-    Given operator "kube-apiserver" becomes "available/non-progressing" within 100 seconds
-    Given operator "kube-apiserver" becomes "available/non-degraded" within 100 seconds
-    Given operator "kube-apiserver" becomes "available" within 100 seconds
-    Given operator "kube-apiserver" becomes "non-progressing" within 100 seconds
-    Given operator "kube-apiserver" becomes "non-degraded" within 100 seconds
-    Given operator "kube-apiserver" becomes "" within 100 seconds
+    Given operator "kube-apiserver" becomes progressing
+    Given operator "kube-apiserver" becomes available/non-progressing/non-degraded within 1200 seconds
+    Given operator "kube-apiserver" becomes available/non-progressing within 100 seconds
+    Given operator "kube-apiserver" becomes available/non-degraded within 100 seconds
+    Given operator "kube-apiserver" becomes available within 100 seconds
+    Given operator "kube-apiserver" becomes non-progressing within 100 seconds
+    Given operator "kube-apiserver" becomes non-degraded within 100 seconds
--- a/features/step_definitions/_constants.rb
+++ b/features/step_definitions/_constants.rb
@@ -7,5 +7,6 @@ USER = 'the( \\S+)? user'
 NUMBER = '([0-9]+|<%=.+?%>)'
 WORD = '(\w+|<%=.+?%>)'
 OPT_WORD = "(?: #{WORD})?"
+NO_SPACE_STR = '([\S]+|<%=.+?%>)'
 # use for regular expression
 RE='/(.+?)/'


wangke19:
@xingxingxia I have updated the code, supported more combinations of input conditions. Here is test results: Runner-v3-smoke/2646/consoleFull, please help to review, thanks.

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,35 @@
   cb_name = 'operator_channel_name' unless cb_name
   cb[cb_name] = cluster_version('version').channel.split('-')[1]
 end
+
+Given /^I wait(?: up to (\d+) seconds)? for operator "([^"]*)" to check conditions: "([^"]*)"$/ do | time, operator, conditions |
```
xingxingxia, line 164: `I wait ... operator "..." to check conditions: "Conditionname=Value [...]"`
I think `check conditions:` is hard to understand what it means. Let's use `become`, as many step definitions used.

wangke19: what about below?
```
I wait ... operator "..." to become ready with conditions: "Conditionname=Value [...]"
```

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,35 @@
...
+  interval_time = 15
+  expected = {}
+  # Parse the conditions to Hash value {Available"=>"True", "Progressing"=>"False", "Degraded"=>"False", "Upgradeable"=>"True"}
+  if conditions.include? "="
+    conditions.split(/\s+/).map{|ar| ar.split('=')}.each {|k, v| expected[k]=v}
+  else
+    arr_conditions = conditions.split(/\s+/)
+    ["Available", "Progressing", "Degraded", "Upgradeable"][0,arr_conditions.length].zip(arr_conditions).map {|k, v| expected[k]=v}
+  end
+  raise "### Without valid conditions, unable to continue to check !!!" unless expected.length >0
```
xingxingxia, line 176: If you like, you could validate all values must be True or False.

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,35 @@
...
+
+  time = Integer(time) rescue 60
```
xingxingxia, line 178: This seems not needed, because `(\d+)` ensures it.

features/step_definitions/operators.rb
xingxingxia: Many step definitions used `ready` too. `ready` means normal. So don't use `ready`.
`I wait ... operator "..." to become "Conditionname=Value [...]"` is simple and straightforward.

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,35 @@
...
+  interval_time = 30 if time > 100
+  actual_results = {}
+  success = wait_for(time, interval: interval_time) {
+    runtime_conditions = cluster_operator(operator).conditions
+    runtime_conditions.each do |c|
+      # only care about the `expected`, ignore others
+      if expected.keys.include? c['type']
+        actual_results[c['type']] = cluster_operator(operator).condition(type: c['type'], cached: false)['status']
+      end
+    end
```
xingxingxia, line 188: You should iterate `expected` instead of `runtime_conditions`:
`expected.each do | c |`

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,35 @@
...
+    logger.info("### Operator #{operator} Expected conditions: #{expected}")
+    logger.info("### Operator #{operator} Actual conditions: #{actual_results}")
```
xingxingxia, line 190: Move the logs outside of wait_for, otherwise too many logs when waiting.

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,35 @@
...
+    expected == actual_results
```
xingxingxia, line 191: I think expected should not be changed after you parsed.

xingxingxia:
@wangke19 Great PR! Done review, please check/update.

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,35 @@
...
+  if conditions.include? "="
+    conditions.split(/\s+/).map{|ar| ar.split('=')}.each {|k, v| expected[k]=v}
+  else
+    arr_conditions = conditions.split(/\s+/)
+    ["Available", "Progressing", "Degraded", "Upgradeable"][0,arr_conditions.length].zip(arr_conditions).map {|k, v| expected[k]=v}
+  end
+  raise "### Without valid conditions, unable to continue to check !!!" unless expected.length >0
+
+  time = Integer(time) rescue 60
+  interval_time = 30 if time > 100
```
xingxingxia, line 179: 30 is too coarse. If using 30, then 100s only tries at 30s, 60s, 90s. If success at 40s, then it still needs wait 20s to exit the loop.
I think need not use interval_time

features/step_definitions/operators.rb
wangke19: Line 170~173 I add validations as below,
```
if conditions.split(/\s+/).any? { |i| /.*=True|False/ } 
    arr_conditions = conditions.split(/\s+/)
    raise "### Invalid number of conditions!!! (Valid number:1~4)" unless arr_conditions.length <=4
    arr_conditions.split(/\s+/).map{|ar| ar.split('=')}.each {|k, v| expected[k]=v}
  else
    arr_conditions = conditions.split(/\s+/)
    raise "### Invalid conditions #{arr_conditions} !!! " unless arr_conditions.any? { |i| ["True","Flase"].include? i }
```

features/step_definitions/operators.rb
wangke19: Without interval_time,  waiting time will be the whole timeout value, the more time to wait is uncertain. This is not an optimal solution. With interval_time, the more time to wait is interval_time in the worst case, will quit the process early.

features/step_definitions/operators.rb
wangke19: Because` wait_for` with interval time, `runtime_conditions` is current conditions of operator every attempt , `expected` is the  comparison value that we got from input parameter `conditions`.

features/step_definitions/operators.rb
wangke19: Yes, just look at the final result at the end.

features/step_definitions/operators.rb
wangke19: This looks like better.

features/step_definitions/operators.rb
wangke19: For example,
```
timeout = 100 
with interval = 10
10,20,30,40,50,60,70,80,90,100 <- attempts
best case: more time to wait -> 0
worst case: more time to wait -> 10

without interval,
one attempt
we assume the program quit in 10s, 50s, 100s
more time to wait -> 90s, 50s, 0s.

So with internal, the result is better.
```

xingxingxia: No. Check lib/base_helper.rb:143:      def wait_for(seconds, interval: 1, stats: nil)
Default interval is 1 if not set.

features/step_definitions/operators.rb
xingxingxia: No. If `oc get clusteroperator the_operator -o yaml` has 20 condition names, runtime_conditions will have 20 conditions, runtime_conditions.each will iterate 20 times.

features/step_definitions/operators.rb
wangke19: yes, I noticed this, interval:1 is inadvisable, more attempts. An appropriate increase in interval time is desirable for a long wait.

features/step_definitions/operators.rb
wangke19: `runtime_conditions` got  all the conditions of operator every attempt, then iterate the every condition only matches included in `expected` ({Available"=>"True", "Progressing"=>"False", "Degraded"=>"False", "Upgradeable"=>"True"}), see following line,
`if expected.keys.include? c['type']`

features/step_definitions/operators.rb
xingxingxia: `any?` means if any one satisfies. But we want all must satisfy. So `any?` isn't OK.
You can use
expected.each {|k, v| raise ... unless ["True","Flase"].include? v}

features/step_definitions/operators.rb
xingxingxia: OK, you can change `> 100` to larger:
  interval_time = 30 if time > 600

features/step_definitions/operators.rb
wangke19: sure, need to check each value

features/step_definitions/operators.rb
wangke19: Attempt once a second is too frequent, it is necessary to extend the interval time for detecting the status of operator. I adjust the interval time as below, 
```
  interval_time = 5
  ...
  interval_time = 20 if time > 100
```

xingxingxia: > OK, you can change > 100 to larger:
> interval_time = 30 if time > 600

This comment already agreed to your previous comment. OK, you can feel free to adjust the values.

wangke19: There is something wrong with page refreshing, sometime I couldn't get the new content.

wangke19:
@xingxingxia Updated the PR, here is test results: Runner-v3-smoke/2655/console, please help to review, thanks.

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,34 @@
...
+  interval_time = 20 if time > 100
+  actual_results = {}
+  success = wait_for(time, interval: interval_time) {
+    expected.keys.each do |c|
+        actual_results[c] = cluster_operator(operator).condition(type: c, cached: false)['status']
+    end
+    expected == actual_results
+  }
+  logger.info("#### Operator #{operator} Expected conditions: #{expected}")
+  logger.info("#### Operator #{operator} Actual conditions: #{actual_results}")
```
xingxingxia, line 191: Not only use logger.info from Pravin's PR https://github.com/openshift/verification-tests/pull/1648 , but also use stats[:seconds] and stats[:iterations], so that we easily know how much time the wait_for actually takes when success is true.

xingxingxia: Line 184 ~ 192, change to below:
```
  result = {
    props: {stats: {}}
  }         
  success = wait_for(time, interval: interval_time, stats: result[:props][:stats]) {
    ...
  }     
  stats = result[:props][:stats]
  logger.info "#### after #{stats[:seconds]} seconds and #{stats[:iterations]} iterations " <<
    "operator #{operator} is: #{actual_results}"
    raise "The #{operator} operator still didn't become #{expected} after #{time} seconds" unless success
```

xingxingxia: My test result:
```
$ cat features/myquicktest.feature
Feature: Test feature
  @admin
  Scenario: test kewang verification-tests/pull/1719
    Given I wait up to 20 seconds for operator "kube-apiserver" to become conditions: "Progressing=False"
    Given I wait up to 10 seconds for operator "kube-apiserver" to become conditions: "Progressing=True"
```
Got:
```
[08:25:50] INFO> #### after 17.005530637223274 seconds and 1 iterations operator kube-apiserver is: {"Progressing"=>"False"}
The kube-apiserver operator still didn't become {"Progressing"=>"True"} after 10 seconds (RuntimeError)
/data/src/github.com/openshift/verification-tests/features/step_definitions/mytest.rb:64:in `/^I wait(?: up to (\d+) seconds)? for operator "([^"]*)" to become conditions: "([^"]*)"$/' 
features/myquicktest.feature:5:in `Given I wait up to 10 seconds for operator "kube-apiserver" to become conditions: "Progressing=True"'
```

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,34 @@
...
+  raise "The #{operator} operator still didn't become ready after #{time} seconds" unless success
```
xingxingxia, line 192: Not "`become ready`". Change to `become #{expected}`

xingxingxia (changes requested):

xingxingxia:
In addition to the inline comments, one more comment is, as we synced before, "not only care PASSED, but also remember to browse your Jenkins job automation execution logs: avoid printing many useless logs".

Check your job's logs, there are many logs of `oc get clusteroperators kube-apiserver --output=yaml`
First, in below code, if `expected` has 3 keys, your below code will run 3 times of `oc get clusteroperators kube-apiserver --output=yaml`, you can see this in the job logs.
```
    expected.keys.each do |c|
        actual_results[c] = cluster_operator(operator).condition(type: c, cached: false)['status']
    end
```
This snippet should change to 
```
    conditions = cluster_operator(operator).conditions # this runs one time of `oc get clusteroperators kube-apiserver --output=yaml`
    expected.keys.each do |c|
      actual_results[c] = conditions.select{ |t| t['type'] == c }.first['status']
    end
```

xingxingxia:
Second, if the wait_for loop tries N attempts, it will print N times of `oc get clusteroperators kube-apiserver --output=yaml`. We can study your Runner-v3-smoke/2616/consoleFull , `Given I wait up to 1200 seconds for the steps to pass:` reduces the logs by "`[05:37:16] INFO> last 3 messages repeated 153 times`". But I don't know how to use that in your PR, maybe @akostadinov @liangxia have idea.

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,34 @@
...
+  # Parse the conditions to Hash value {Available"=>"True", "Progressing"=>"False", "Degraded"=>"False", "Upgradeable"=>"True"}
+  arr_conditions = conditions.split(/\s+/)    
+  raise "#### Invalid number of conditions!!! (Valid number:1~4)" unless arr_conditions.length <=4
+  if arr_conditions.any? { |x| x.include? "=" }
+    arr_conditions.map{|ar| ar.split('=')}.each {|k, v| expected[k]=v}
+  else
+    # Parse the conditions with "True False ..." input 
+    ["Available", "Progressing", "Degraded", "Upgradeable"][0,arr_conditions.length].zip(arr_conditions).map {|k, v| expected[k]=v}
+  end
+  expected.each {|k,v| raise "#### Please make sure the input status of condition is True or False!" unless ["True","False"].include? v }
```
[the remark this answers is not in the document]
wangke19: Sounds good.

features/step_definitions/operators.rb
wangke19: agree.

xingxingxia:
@wangke19 for above "Second", I studied `Given /^I wait(?: up to #{NUMBER} seconds)? for the steps to pass:$/`, it has logger.dedup_start and logger.dedup_flush, so I copied and tried:
```
  actual_results = {}
  stats = {}
  begin
    logger.dedup_start
    success = wait_for(time, interval: interval_time, stats: stats) {
      ...
    }
    ...
    raise "The #{operator} operator still didn't become #{expected} after #{time} seconds" unless success
  ensure
    logger.dedup_flush
  end
```
$ cucumber features/myquicktest.feature:3
It reduces the printing times of `oc get clusteroperators kube-apiserver --output=yaml`.

features/step_definitions/operators.rb
wangke19: Useful output.

wangke19:
> @wangke19 for above "Second" one, I studied `Given /^I wait(?: up to #{NUMBER} seconds)? for the steps to pass:$/`, it has logger.dedup_start and logger.dedup_flush, so I copied and tried:
> 
> ```
>   ...
>   actual_results = {}
>   stats = {}
>   begin
>     logger.dedup_start
>     success = wait_for(time, interval: interval_time, stats: stats) {
>       ...
>     }
>     ...
>     raise "The #{operator} operator still didn't become #{expected} after #{time} seconds" unless success
>   ensure
>     logger.dedup_flush
>   end
> end
> ```
> 
> Then I got:
> $ cucumber features/myquicktest.feature:3
> ...
> [11:58:11] INFO> last 4 messages repeated 10 times
> ...
> It reduces the printing times of `oc get clusteroperators kube-apiserver --output=yaml`, you can copy to use.

I noticed this way before when I start this PR, but I didn't figure out how it works. I applied them, it works as expected, reduced the output of getting operator.

wangke19:
@xingxingxia Updated above suggestions, here is test results: Runner-v3-smoke/2667/console, the log looks good now. Please help to review when you are free.

xingxingxia:
/lgtm
Thanks for the great PR.
@liangxia , @akostadinov , help review / merge, thanks.

features/test/apiserver.feature
```diff
@@ -0,0 +1,27 @@
+Feature: API server Test features
+  # @author kewang@redhat.com
+  @admin
+  Scenario: Quick test for step definitions
+    Given the master version >= "4.6"
+    Given I switch to cluster admin pseudo user      
+
+    # Set WriteRequestBodies profile to audit log
+    Given as admin I successfully merge patch resource "apiserver/cluster" with:
+      | {"spec": {"audit": {"profile": "WriteRequestBodies"}}} |
```
liangxia, line 10: Change api server is a destructive action in my understanding.

wangke19: Yes, this part is for testing the step definition in operator.rb, not the real OCP test case.

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,44 @@
   cb_name = 'operator_channel_name' unless cb_name
   cb[cb_name] = cluster_version('version').channel.split('-')[1]
 end
+
+Given /^I wait(?: up to (\d+) seconds)? for operator "([^"]*)" to become conditions: "([^"]*)"$/ do | time, operator, conditions |
```
akostadinov, line 164: Lets make the conditions a table for brevity.

```
Given /^I wait(?: up to #{NUMBER} seconds)? for operator #{QUOTED} to meet conditions:$/ do | time, operator, conditions |
  arr_conditions = conditions.raw.flatten
  ...
```

xingxingxia: Hi, Aleks, I prefer current step def, one single line is more of brevity than additional table lines.
If you think the single line is a little lengthy, I suggest Ke remove characters `conditions: ` and revise it to be:
Given /^I wait(?: up to (\d+) seconds)? for operator "([^"]*)" to become "([^"]*)"$/ do | time, operator, conditions |

The frequent uses we actually possibly need would be AFAIK:
Given I wait up to XXX seconds for operator "XXX" to become "True False False"
Given I wait up to XXX seconds for operator "XXX" to become "Progressing=True"
Sufficient brevity.

liangxia: `"True False False"`, this is hard to read/review/maintain.
We should force to use format `Progressing=True`

features/test/apiserver.feature
liangxia: Without tag destructive, the scenarios will run in parallel.
We never know how the test cases will interfere with each other if they change the cluster at the same time.

features/step_definitions/operators.rb
xingxingxia: oc get co shows Available Progressing Degraded order "True False False", this is a special but very common meaning to everyone, therefor Ke specially implemented for this. So not hard to read. Let's not make the special implementation effort wasted.

wangke19: For those people who are familiar with status of operator, uses 'True False ..." is simple and convenient, for those who are unfamiliar, there is another option: "Progressing=True".

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,44 @@
...
+
+  expected = {}
+  # Parse the conditions to Hash value {Available"=>"True", "Progressing"=>"False", "Degraded"=>"False", "Upgradeable"=>"True"}
+  arr_conditions = conditions.split(/\s+/)    
+  raise "#### Invalid number of conditions!!! (Valid number:1~4)" unless arr_conditions.length <=4
+  if arr_conditions.any? { |x| x.include? "=" }
+    arr_conditions.map{|ar| ar.split('=')}.each {|k, v| expected[k]=v}
+  else
+    # Parse the conditions with "True False ..." input 
+    ["Available", "Progressing", "Degraded", "Upgradeable"][0,arr_conditions.length].zip(arr_conditions).map {|k, v| expected[k]=v}
```
xingxingxia, line 175: @wangke19 actually in Slack I said only wanting to support the short format for below special case:
the condition number is EXACTLY 3 and the conditions EXACTLY are Available, Progressing, Degraded.
So "True False False" is widely understood.
I don't want to support for other cases:
the condition number is 1, 2 or 4, and the conditions are not EXACTLY Available, Progressing, Degraded
So "True True", or "True" are not supported.

features/step_definitions/operators.rb
akostadinov: If we implement such option, we should make sure in code that proper order of checking will be performed. We should not rely on order provided by API. Even better, we can have 
`operator "..." becomes available/non-progressing/non-degraded within ## seconds`
or
`operator "..." becomes non-available/progressing/degraded within ## seconds`

Should be readable by everybody. The reason to use Gherkin is to make tests readable by non-technical leads and managers (whether they will ever do, I don't know). They will certainly have troubles reading `true false false`.

features/step_definitions/operators.rb
akostadinov: I get your point. See my previous comment. If we use the keywords with or without `non-` prefix, then we can support in a short format any combination desired. WDYT?

features/step_definitions/operators.rb
wangke19: I changed as following:
```
Given /^I wait(?: up to (\d+) seconds)? for operator "([^"]*)" to become conditions[available/non-progressing/non-degraded]: "([^"]*)"$/ do | time, operator, conditions |
```
the conditions `[available/non-progressing/non-degraded]:` is optional.

features/step_definitions/operators.rb
wangke19: For lines 174~175 I changed as following:
```
    # Parse the conditions with "True False ..." input 
    expected_order = {"Available"=>"True", "Progressing"=>"False", "Degraded"=>"False"}
    ["Available", "Progressing", "Degraded"][0,arr_conditions.length].zip(arr_conditions).map {|k, v| expected[k]=v}
    raise "#### Only the brevity conditions order 'True False False' is supported!!!" unless expected == actual_results 
```
There is only one format for qualifying brevity input: True False False

features/test/apiserver.feature
wangke19: Added destructive tag.

wangke19:
@xingxingxia @liangxia @akostadinov I have updated the code, please help to review, here is  test result: Runner-v3-smoke/2694/console

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,46 @@
   cb_name = 'operator_channel_name' unless cb_name
   cb[cb_name] = cluster_version('version').channel.split('-')[1]
 end
+
+Given /^I wait(?: up to (\d+) seconds)? for operator "([^"]*)" to become conditions available\/non-progressing\/non-degraded: "([^"]*)"$/ do | time, operator, conditions |
```
akostadinov, line 164: Starting a new thread. Not that above is wrong. Just it is very long step name and provides redundant information. My suggestion is to make step definition:
```
Given /^operator #{QUOTED} becomes (\S+) (?: within #{NUMBER} seconds)$/ do | operator_name, conditions, timeout |
```

Then `conditions` will be split on slash `/` and when `available` is one of the words, then we should check whether condition `Available` is `"True"`. If we see `non-avalable`, then we check that condition `Available` is `"False"`. If `available` and `non-available` are both missing, then we don't care about condition `Available`.

Same thing with `progressing` and `non-progressing`, ditto `degraded`, `non-degraded`.

If another word is found, then we return an error that such condition does not exist.

It is very similar why we write in code `if my_variable { ... }` instead of `if my_variable == true { ... }`. It is about brewity and conventions.

So later you will call the step like:
```
Then operator "my-operator" becomes available/non-progressing/non-degraded within 130 seconds
```
or
```
Then operator "my-operator" becomes non-available/progressing/degraded within 130 seconds
```

wangke19: It's good idea. I used regular expression to restrict input conditions, used one placeholder "_" to indicate the unconditional input. 
```
reg = /(\_|available|non-available)\/(\_|progressing|non-progressing)\/(\_|degraded|non-degraded)/
```
And we can call the step like:
Given operator "kube-apiserver" becomes "_/progressing/_" within 100 seconds
Given operator "kube-apiserver" becomes "available/non-progressing/non-degraded" within 1200 seconds

wangke19:
@akostadinov  Already updated the code, please help to review, here is test result: Runner-v3-smoke/2699/console

features/step_definitions/operators.rb
```diff
@@ -160,3 +160,47 @@
   cb_name = 'operator_channel_name' unless cb_name
   cb[cb_name] = cluster_version('version').channel.split('-')[1]
 end
+
+Given /^operator #{QUOTED} becomes "([^"]*)" within(?: (\d+) seconds)?$/ do | operator_name, conditions, timeout |
+  ensure_admin_tagged
+
+  expected = {}
+  reg = /(\_|available|non-available)\/(\_|progressing|non-progressing)\/(\_|degraded|non-degraded)/
```
akostadinov, line 168: I don't see need to use underscores `_` and quotes. When you get the string we could `conditions.split("/")`, then check if we have `non-` prefix for each word and then check which word do we have. I didn't have time to write implementation od this today but if you don't have time for it, I can hopefully do it tomorrow.

wangke19: I have no time to work on this currently, will give it a try in another time.

wangke19: I changed the code to get conditions by `conditions.split("/")`, then check `non-` prefix to get the word we want.

To call the step, see below,
```
    Given operator "kube-apiserver" becomes "progressing" within 100 seconds
    Given operator "kube-apiserver" becomes "available/non-progressing/non-degraded" within 1200 seconds
    Given operator "kube-apiserver" becomes "available" within 100 seconds
    Given operator "kube-apiserver" becomes "" within 100 seconds

```

wangke19: @akostadinov, Here is test results: Runner-v3-smoke/2709/console, when you are available, please help to review, thanks.

liangxia:
/lgtm

/hold
for @akostadinov to review.

features/test/apiserver.feature
```diff
@@ -0,0 +1,27 @@
...
+    Given operator "kube-apiserver" becomes "available/non-progressing/non-degraded" within 1200 seconds
+    """
+    Given operator "kube-apiserver" becomes "progressing" within 100 seconds
+    Given operator "kube-apiserver" becomes "available/non-progressing/non-degraded" within 1200 seconds
+    Given operator "kube-apiserver" becomes "available/non-progressing" within 100 seconds
+    Given operator "kube-apiserver" becomes "available/non-degraded" within 100 seconds
+    Given operator "kube-apiserver" becomes "available" within 100 seconds
+    Given operator "kube-apiserver" becomes "non-progressing" within 100 seconds
+    Given operator "kube-apiserver" becomes "non-degraded" within 100 seconds
+    Given operator "kube-apiserver" becomes "" within 100 seconds
```
akostadinov, line 27: What is this step checking? `""` looks like checking for nothing. I think we better fail in such case.

akostadinov:
@liangxia , if @wangke19 accept my updates, feel welcome to merge it.

features/test/apiserver.feature
wangke19: Yes, it's a negative case and will be fail,

wangke19: I agree with your updates.

pruan-rht (commented):
/lgtm

wangke19:
> @liangxia , if @wangke19 accept my updates, feel welcome to merge it.
I accepted, please go on.

akostadinov:
@wangke19 , Bot should have merged already. Let me manually merge to avoid delay.

features/test/apiserver.feature
```diff
@@ -0,0 +1,26 @@
...
+    # Set WriteRequestBodies profile to audit log
+    Given as admin I successfully merge patch resource "apiserver/cluster" with:
+      | {"spec": {"audit": {"profile": "WriteRequestBodies"}}} |
+    And I register clean-up steps:
+    # Set original Default profile to audti log
+    """
+    Given as admin I successfully merge patch resource "apiserver/cluster" with:
+      | {"spec": {"audit": {"profile": "Default"}}} |
+    Given operator "kube-apiserver" becomes "progressing" within 100 seconds
+    Given operator "kube-apiserver" becomes "available/non-progressing/non-degraded" within 1200 seconds
```
xingxingxia, line 18: @wangke19 did these 2 lines with quotes passed test? If not, could you update them?

wangke19: Fixed in PR https://github.com/openshift/verification-tests/pull/1837

PR #1720, opened 2020-12-03
Replace image from docker to quay

liangxia:
/cc @jhou1 @xiuwang @pruan-rht 
Action item from automation bi-weekly meeting on Dec 2nd.

Why we switch from docker to quay ?
1. There are rate limit added recently on docker, which will result pulling image to fail as we reach the threshold.
2. We have mirror the image from quay to internal registry for disconnect testing.


pruan-rht (commented):
/lgtm

PR #1723, opened 2020-12-04
replace caddy to nginx for network part

zhaozhanqi:
@lihongan  ^^

### DIFF
--- a/features/networking/egress-ip.feature
+++ b/features/networking/egress-ip.feature
@@ -354,12 +354,11 @@ Feature: Egress IP related features
     """
 
     # Create a pod
-    Given I obtain test data file "routing/caddy-docker.json"
+    Given I obtain test data file "routing/web-server-1.yaml"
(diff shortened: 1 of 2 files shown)


features/networking/egress-ip.feature
```diff
@@ -354,11 +354,12 @@ Feature: Egress IP related features
     """
 
     # Create a pod
-    Given I obtain test data file "routing/caddy-docker.json"
+    Given I obtain test data file "routing/web-server-rc.yaml"
```
lihongan, line 357: `web-server-rc.yaml` creates pod and svc as well, seems here you don't need the service, if that's true then just use `web-server-1.yaml`.

zhaozhanqi: correct.  updated, thanks

lihongan:
/lgtm
thank you @zhaozhanqi

PR #1726, opened 2020-12-08
OCP-26139: increase jitter check

rbbratta:
We where checking iperf
`['end']['sum']['jitter_ms'].to_f < 1`

Increase it to 5 ms.
`['end']['sum']['jitter_ms'].to_f < 5`

Anytning less than 20ms is supposed to be good
for VOIP calls, but we want to be better.

It seems the minimum resolution is 1ms as well.


rbbratta:
http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/11/28/09:56:49/Traffic_flow_shouldn_t_be_interrupted_when_master_switches_the_leader_positions/console.html

PR #1728, opened 2020-12-08
fix xpath change for 4.7

lihongyan1:


juzhao:
https://github.com/openshift/verification-tests/blob/master/lib/rules/web/admin_console/4.7/monitoring_metrics.xyaml#L298
`content: Memory Usage`
should be
`content: Memory usage`

lihongyan1:
@pruan-rht Please help merge when available.

PR #1730, opened 2020-12-08
Fix a test flake due to 1875324

jhou1:
Due to 1875324, the test flakes with an error:
```
Scenario: The multus admission controller should be able to detect the syntax issue in the net-attach-def ==>@  features/networking/multus.feature:560
Error from server (NotFound): error when creating "STDIN": namespaces "aos-qe-ci" not found
```

This PR workarounds that and ensures the test no longer fails hereafter: https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/218605/console

@zhaozhanqi @anuragthehatter PTAL.

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -563,20 +563,23 @@ Feature: Multus-CNI related scenarios
     # Create the net-attach-def via cluster admin and simulating syntax errors
     Given I have a project
     Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml"
-    When I run oc create as admin over "macvlan-bridge.yaml" replacing paths:
+
+    Given I switch to cluster admin pseudo user
+    And I use the "<%= project.name %>" project
+    When I run oc create over "macvlan-bridge.yaml" replacing paths:
       | ["metadata"]["name"] | macvlan-bridge-21756 |
       | ["spec"]["config"]   | 'asdf'               |
     Then the step should fail
     And the output should contain:
       | admission webhook "multus-validating-config.k8s.io" denied the request |
-    And admin ensures "macvlan-bridge-21756" network_attachment_definition is deleted from the "default" project after scenario
+    And admin ensures "macvlan-bridge-21756" network_attachment_definition is deleted from the "<%= project.name %>" project after scenario
     Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml"
-    When I run oc create as admin over "macvlan-bridge.yaml" replacing paths:
+    When I run oc create over "macvlan-bridge.yaml" replacing paths:
       | ["metadata"]["name"] | macvlan-bridge@$ |
     Then the step should fail
     And the output should contain:
       | subdomain must consist of lower case alphanumeric characters |
-    And admin ensures "macvlan-bridge@$" network_attachment_definition is deleted from the "default" project after scenario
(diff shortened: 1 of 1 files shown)


jhou1:
@zhaozhanqi @anuragthehatter if you see similar error in your future automation analysis, you can follow this approach to remove the flakes.

zhaozhanqi:
Thanks @jhou1 

@anuragthehatter   the net-attach-def is created in user project. but the following is checking in 'default' namespace. please confirm this. 
`And admin ensures "macvlan-bridge-21756" network_attachment_definition is deleted from the "default" project after scenario`

jhou1:
Thanks Zhanqi, I was not aware of that. It's fixed! https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/218725/console

anuragthehatter:
> Thanks @jhou1
> 
> @anuragthehatter the net-attach-def is created in user project. but the following is checking in 'default' namespace. please confirm this.
> `And admin ensures "macvlan-bridge-21756" network_attachment_definition is deleted from the "default" project after scenario`

The logic here is right. By default net-attach-def would be created under default namespace not test-project unless you force it on a particular ```metadata.namespace```. ```Admin ensures``` was an emergency step in a way that if somehow net-attach-def is created (means test fails) it would stay there under default namespace if not deleted during cleanup. I guess its okay to remove both admin ensures statement. Its fine, we will catch the failure anyway at the step ```then the step should fail```. @weliang1

weliang1:
> > Thanks @jhou1
> > @anuragthehatter the net-attach-def is created in user project. but the following is checking in 'default' namespace. please confirm this.
> > `And admin ensures "macvlan-bridge-21756" network_attachment_definition is deleted from the "default" project after scenario`
> 
> The logic here is right. By default net-attach-def would be created under default namespace not test-project unless you force it on a particular `metadata.namespace`. `Admin ensures` was an emergency step in a way that if somehow net-attach-def is created (means test fails) it would stay there under default namespace if not deleted during cleanup. I guess its okay to remove both admin ensures statement. Its fine, we will catch the failure anyway at the step `then the step should fail`. @weliang1

@jhou1 @anuragthehatter 

The logic in the original script is right. 
For multus testing, only admin has a right to create a new net-attach-def CRD in the new project. One update in original code is needed to delete NAD from "<%= project.name %>" not default project.  
In original code, even deleting steps delete the NAD in the wrong project, but it does not matter, because net-attach-def CRD configuration is not completed due to the syntax errors written in the net-attach-def YAML file, which is the purpose of this test. But it's OK to leave deleting steps there if the code can not catch those syntax error and still create the NAD (this should be a bug).

From the script logic of view I did not see the difference between the original code and the new code excepting the deleting steps I just mentioned.

My confusion is when the step of "Given I have a project" execute, it normally create a random project name such as "dvq2e"(got from my local running log), I do not think the name of "aos-qe-ci" is created by that step.  

We may see this "aos-qe-ci" not founding again in other multus cases, do you have any ideas about where the "aos-qe-ci" namespace come from?

jhou1:
The aos-qe-ci project is from another OCP cluster... The oc client runs on the PSI openshift uses the in-cluster data to communicate with its master instead of communicating with the tested cluster from kubeconfig. A solution is to ensure a step `And I use the "<%= project.name %>" project` before creating the resource.

This is described in https://bugzilla.redhat.com/show_bug.cgi?id=1875324 and https://coreos.slack.com/archives/CH76YSYSC/p1599129179356900

zhaozhanqi:
Thanks for helping correct this @jhou1 
/lgtm

PR #1737, opened 2020-12-09
Window container case for Loadbalancer svc

anuragthehatter:
We Should have a quick LB svc check for window container as well like we do for linux 
Logs: https://privatebin-it-iso.int.open.paas.redhat.com/?4513f4efce5d3048#9nJ1G2onaUnGxqU3RjjMMGrkbtHwNy6CgKGKd4c6jocZ

@rbbratta @zhaozhanqi @huiran0826 @weliang1 @brahaney @asood-rh

### DIFF
--- a/features/networking/ovn_winc.feature
+++ b/features/networking/ovn_winc.feature
@@ -70,7 +70,7 @@ Feature: OVNKubernetes Windows Container related networking scenarios
       | name | win-webserver |
       | tcp  | 80            |
     Then the step should succeed
-    And I pry
+    
     # Get the external ip of the loadbalancer service
     And I wait up to 60 seconds for the steps to pass:
     """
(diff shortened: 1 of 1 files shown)


zhaozhanqi:
@sgaoshang met an issue yesterday and we had some debug on . found the LB service is not working well when there are two windows workers and two test pods located on each one.

anuragthehatter:
> @sgaoshang met an issue yesterday and we had some debug on . found the LB service is not working well when there are two windows workers and two test pods located on each one.
> https://bugzilla.redhat.com/show_bug.cgi?id=1905950
> 
> for now I'm not sure what's the root reason cause it. I think we may need two windows workers and two test pods here . and also try 'curl $external_ip' more times to make sure it works.

Ya we can definitely add such test case but I also tried to repro that bug and seems like it was not needed couple of nodes but i was able to repro that on single node itself https://bugzilla.redhat.com/show_bug.cgi?id=1905950#c3. Seems like its the specific window image issue as shared by Aravind in comment4. Let see what dev finds out.

PR #1738, opened 2020-12-10
Use images from quay.io

liangxia:
/cc @jhou1 @wzheng1 @pruan-rht 

Hopefully, this is the last PR to update images to quay.io in batch for verification-tests.


pruan-rht (commented):
/lgtm

PR #1740, opened 2020-12-11
add support for packet & vSphere auditing.

pruan-rht:


pruan-rht:
@akostadinov @liangxia PTAL.  I have tested it locally and on Jenkins...they have [passed](https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/cloud_usage_report/206/console).

akostadinov:
@wjiangjay , could you review this PR, at least the packet part?

lib/launchers/packet.rb
```diff
@@ -17,5 +17,65 @@ class Packet
...
+      res = Http.request(**req_opts)
+
+      if res[:success]
+        if res[:headers] && res[:headers]['content-type']
+          content_type = res[:headers]['content-type'][0]
+          case
+          when content_type.include?('json')
+            res[:parsed] = JSON.load(res[:response])
+          when content_type.include?('yaml')
+            res[:parsed] = YAML.load(res[:response])
```
akostadinov, line 59: you can just parse YAML in both cases because JSON is a subset of YAML

lib/launchers/v_sphere.rb
```diff
@@ -391,6 +391,33 @@ def object.conwrapper
       end
     end
 
+    # @return Array of VirtualMachine objects
+    private def vms(folder, machines)
+      folder.childEntity.each do |x|
+        name, junk = x.to_s.split('(')
```
akostadinov, line 397: a little bit more efficient with `x.to_s.split('(', 2)`

lib/launchers/v_sphere.rb
```diff
@@ -391,6 +391,33 @@ def object.conwrapper
       end
     end
 
+    # @return Array of VirtualMachine objects
+    private def vms(folder, machines)
```
akostadinov, line 395: It would suggest this method to take only one argument `folder`. Then have in body
```
private def vms(folder: nil)
folder ||= datacenter.vmFolder
machines = []
...
        when "Folder"
          machines.concat vms(x)
...
return machines
```

Then you can alter method `#get_running_instances` to filter out only `poweredOn` instances and return. This will make for less boilerplate
```
    def get_running_instances
      return vms.select { |i| i.runtime.powerState == "poweredOn" }
    end
```

lib/launchers/v_sphere.rb
```diff
@@ -391,6 +391,33 @@ def object.conwrapper
       end
     end
 
+    # @return Array of VirtualMachine objects
```
akostadinov, line 394: Could you see example of other methods how object type is documented?

lib/launchers/v_sphere.rb
pruan-rht: @akostadinov done...

pruan-rht:
@akostadinov all comments have been addessed.  PTAL.  thx!

pruan-rht:
@akostadinov can you PTAL when you have a moment?  thx!

lib/launchers/packet.rb
akostadinov:
```
case content_type
when /json/, /yaml/
  res[:parsed] = YAML.load(res[:response])
end
```

tools/instance_summary.rb
```diff
@@ -287,5 +287,73 @@ def get_summary
...
+      # summary = summarize_instances(rg_name, instances)
+      print_summary(summary) if summary.count > 0
+      grand_summary << {platform: 'openstack', region: summary.first[:region], inst_count: summary.count}
+      print_grand_summary(grand_summary)
+    end
+  end
+
+
+  class VSphereSummary < InstanceSummary
+    attr_accessor :vms
```
akostadinov, line 327: what does `vms` stand for here? I would expect to be `vsphere` or something. Or did I mistake the abbreviation?

akostadinov:
/assign @wjiangjay 

@wjiangjay , please take a look at packet part of this PR

tools/instance_summary.rb
pruan-rht: @akostadinov  vms==vSphere AFAIK, @jinyunma do you know why it's called vms?

lib/launchers/packet.rb
```diff
@@ -17,5 +17,65 @@ class Packet
...
+        end
+
+        yield res if block_given?
+      end
+      return res
+    end
+
+    def get_running_instances()
+      params ||= {}
+      url = self.config[:api_url] + '/projects/' + self.config[:project_id] + '/devices?include=facility'
```
wjiangjay, line 70: how about using `URI.join(iaas.config[:api_url], "projects", iaas.config[:project_id], "devices?include=facility")` to make a safety URL join?

wjiangjay, line 70: Also need  to add the `api_url` to our creds repo

wjiangjay:
For the packet functional part, it works well.
```
[1] pry(#<BushSlicer::DefaultWorld>)> iaas = iaas_by_service("packet")
=> #<BushSlicer::Packet:0x000055dcf4c667f0
 @config=
  {:api_url=>"https://api.equinix.com/metal/v1/",
   :auth_token=>"HIDDEN",
   :facility=>"ny5",
   :project_id=>"xxxxxx",
......
   :cloud_type=>"packet"}>
[2] pry(#<BushSlicer::DefaultWorld>)> iaas.get_running_instances().length()
=> 8
```

lib/launchers/packet.rb
pruan-rht: yes, that's a separate PR.

pruan-rht: ok, i can update that, but not sure we gain that much since the URL components are in valid forms already.

tools/instance_summary.rb
akostadinov: I will be surprised if in vmware community vms means vsphere but in either case that is confusing for me and probably other code contributers because it associates with virtual machines in generic context.

While `vsphere` will be much more legible. SO I suggest we switch regardless of vmware community conventions.

lib/launchers/packet.rb
pruan-rht: @wjiangjay what you have suggested didn't work because `URI.join()` does not seem to work same as `String.join()`doing more [research](url) with `URI.join`,  it seems to be not very intuitive in the way it works.  So I'm inclined to keep it as it.  WDYT @akostadinov

pruan-rht: @wjiangjay @akostadinov so I got it to work by doing something like 
```a = "https://api.equinix.com/metal/"
e = "v1/"
b = "projects/"
c = "ceae1ad7-aa8d-4ce4-9efb-cac1970cea38/"
d = "devices?include=facility"
url = URI.join(a, e, b, c, d)
1] pry(main)> url                                                                                                               
=> #<URI::HTTPS https://api.equinix.com/metal/v1/projects/ceae1ad7-aa8d-4ce4-9efb-cac1970cea38/devices?include=facility>
```
IMO, this is an overkill and not a very friend lib to use, I vote to keep the current implementation.

akostadinov: In this case direct string manipulation is ok. If we process parameters that potentially need escaping, then we need to use special methods. But I'm not sure `URI.join` does a very good job.

pruan-rht:
@akostadinov  @wjiangjay I've addressed the comments that got overlooked.  I've also changed `upshift` to `psi`.  PTAL. Thanks!

pruan-rht:
@akostadinov can this be merged now?

lib/launchers/packet.rb
```diff
@@ -17,5 +17,63 @@ class Packet
...
+      res = self.rest_run(url, "GET", params, self.config[:auth_token])
+      raise "Failed to get running packet instances..." unless res[:success]
+      res[:parsed]['devices']
+    end
+
+    # current time - creation time
+    def instance_uptime(timestamp)
+      ((Time.now  - Time.parse(timestamp)) /(60 * 60)).round(2)
+    end
```
akostadinov, line 77: Why do we accept `timestamp` as a parameter? We should know creation time and current time? Why should user change that.

lib/launchers/v_sphere.rb
akostadinov: bump

lib/launchers/v_sphere.rb
akostadinov: hmm, forgot to push? I still see two args.

lib/launchers/packet.rb
pruan-rht: @akostadinov this is to recalculate it into human readable format.  Otherwise Jenkins uses milliseconds for the unit, and not seconds.

pruan-rht:
@akostadinov PR updated again....PTAL, thx!

tools/cloud_usage.rb
```diff
@@ -77,14 +77,32 @@ def run
         end
       end
       # internal openstack
-      command :"upshift" do |c|
+      command :"psi" do |c|
```
liangxia, line 80: @pruan-rht   There are PSI OCP and PSI OSP.   `upshift = PSI OSP`

liangxia:
@pruan-rht  Would you like to merge or wait for review from @akostadinov  ?

tools/cloud_usage.rb
pruan-rht: @liangxia perhaps just openstack would be better then because that's what the backend is

liangxia: We also have several OpenStack there.  How about rhos-d ?

pruan-rht:
superseded by https://github.com/openshift/verification-tests/pull/1855

PR #1743, opened 2020-12-14
fix OCP-21482 and OCP-18998

xiuwang:
On openstack, intergrated registry is default setting to Removed.
We add post action to change registry to Managed, but not ipv6 and ipv4 in edge team. Since set registry here in case registry are setting to Removed scenarios.


xiuwang:
Add fix for OCP-18998, `oc image mirror` cmd need wait for image registry pod ready.

PR #1748, opened 2020-12-16
fix eventrouter imagepullbackoff issue in 4.7

QiaolingTang:
The eventrouter pod always in imagepullbackoff status, I checked the images in brew registry, there doesn't have image openshift4/ose-logging-eventrouter:v4.7(it only has image with image digest, it's hard to get the image digest when running automation), but it has rh-osbs/openshift-ose-logging-eventrouter:v4.7. To resolve the issue, I update the image registry, @anpingli PTAL, thanks.


anpingli (commented):
LGTM

pruan-rht (commented):
/lgtm

PR #1753, opened 2020-12-17
auto OCP-37706

sunzhaohua2:
Auto OCP-37706: baremetal clusteroperator should be disabled in any deploy that is not baremetal @jhou1 @miyadav please help to take a look.
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/226815/console

### DIFF
--- a/features/machine/machine.feature
+++ b/features/machine/machine.feature
@@ -10,6 +10,7 @@ Feature: Machine features testing
   # @author jhou@redhat.com
   # @case_id OCP-22115
   @smoke
+  @admin
   Scenario: machine-api clusteroperator should be in Available state
     Given evaluation of `cluster_operator('machine-api').condition(type: 'Available')` is stored in the :co_available clipboard
     Then the expression should be true> cb.co_available["status"]=="True"
@@ -22,6 +23,14 @@ Feature: Machine features testing
 
     Given evaluation of `cluster_operator('machine-api').condition(type: 'Progressing')` is stored in the :co_progressing clipboard
     Then the expression should be true> cb.co_progressing["status"]=="False"
+
+  # @author zhsun@redhat.com
+  # @case_id OCP-37706
+  @smoke
+  @admin
+  Scenario: Baremetal clusteroperator should be disabled in any deployment that is not baremetal
+    Given evaluation of `cluster_operator('baremetal').condition(type: 'Disabled')` is stored in the :co_disabled clipboard
+    Then the expression should be true> cb.co_disabled["status"]=="True"
 
   # @author jhou@redhat.com
   # @case_id OCP-25436


features/machine/machine.feature
```diff
@@ -23,6 +23,13 @@ Feature: Machine features testing
     Given evaluation of `cluster_operator('machine-api').condition(type: 'Progressing')` is stored in the :co_progressing clipboard
     Then the expression should be true> cb.co_progressing["status"]=="False"
 
+  # @author zhsun@redhat.com
+  # @case_id OCP-37706
+  @smoke
+  Scenario: baremetal clusteroperator should be disabled in any deploy that is not baremetal
```
jhou1, line 29: Baremetal clusteroperator should be disabled in any deployment that is not baremetal

jhou1: I'm not sure if this works on our UPI baremetal test, let's check when we have a chance.

sunzhaohua2: ok, will check

jhou1: Thanks, can you please update the scenario name?

features/machine/machine.feature
```diff
@@ -23,6 +23,13 @@ Feature: Machine features testing
     Given evaluation of `cluster_operator('machine-api').condition(type: 'Progressing')` is stored in the :co_progressing clipboard
     Then the expression should be true> cb.co_progressing["status"]=="False"
 
+  # @author zhsun@redhat.com
+  # @case_id OCP-37706
+  @smoke
+  Scenario: Baremetal clusteroperator should be disabled in any deployment that is not baremetal
+    Given evaluation of `cluster_operator('baremetal').condition(type: 'Disabled')` is stored in the :co_disabled clipboard
```
liangxia, line 30: Do we need admin privilege here ?

sunzhaohua2: I checked this need admin privilege. But it's strange why another similar case can be passed in our runner-v3 job with normal TEST_MODE.
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/231698/console

liangxia: Generally we require admin test cases tag with admin by add `ensure_admin_tagged` in the step definiation, but the step use `evaluation of xxx` which can not be judged whether it requires admin or not (and in most case it does not require admin, so it does not add `ensure_admin_tagged`).

normal test mode in Runner-v3 job only ensures it does not run test cases which has tag admin or destructive. As the test case OCP-22115 does not have the tag, it's executed. While at the same time, you have passed in admin_creds_spec, so the test case use pseudo admin and got passed (The issue here is that the test case actually needs admin, but it does not tag the test case with admin. While during execution, admin privilege is provided, so it got pass).

sunzhaohua2: Thanks for the clear clarification, have added @admin

PR #1757, opened 2020-12-21
Wait for more time when update externalIP policy

weliang1:
Intermittent externalIP automation failed due to waiting time is not enough:
http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2020/12/18/04:29:14/externalIP_defined_in_service_with_set_ExternalIP_in_allowedCIDRs/console.html
STDERR: Error from server (Forbidden): error when creating "STDIN": services "service-unsecure" is forbidden: spec.externalIPs: Forbidden: externalIPs have been disabled

Update script from waiting 300s to 500s when update externalIP policy

### DIFF
--- a/features/networking/service.feature
+++ b/features/networking/service.feature
@@ -157,7 +157,7 @@ Feature: Service related networking scenarios
     # Create a svc with externalIP
     Given I switch to the first user
     Given I obtain test data file "networking/externalip_service1.json"
-    And I wait up to 300 seconds for the steps to pass:
+    And I wait up to 500 seconds for the steps to pass:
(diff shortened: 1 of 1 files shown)


anuragthehatter:
@weliang1 Thanks for investigating it. Looks good to me. Hopefully it never reaches 500 wait time for each test else it will always create a 20ish minute delay in testruns :( In the long term, we can keep finding workaround if any or grepping through sdn/ovn logs to find out its effectiveness.

pruan-rht (commented):
/lgtm

PR #1760, opened 2020-12-22
Adjust existing function to accept kubescheduler as well

kasturinarra:
Adjust existing function so that it accepts both kubescheduler & scheduler CRD.

### DIFF
--- a/features/step_definitions/scheduler.rb
+++ b/features/step_definitions/scheduler.rb
@@ -7,11 +7,11 @@ Given /^the #{QUOTED} scheduler priorityclasses is restored after scenario$/ do 
   }
 end
 
-Given /^the #{QUOTED} scheduler CR is restored after scenario$/ do |name|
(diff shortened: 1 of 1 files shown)


features/step_definitions/scheduler.rb
```diff
@@ -7,11 +7,11 @@
   }
 end
 
-Given /^the #{QUOTED} scheduler CR is restored after scenario$/ do |name|
+Given /^the #{QUOTED} #{QUOTED} scheduler CR is restored after scenario$/ do |name, crd|
```
liangxia, line 10: How about `Given /^the CR #{QUOTED} named #{QUOTED} is restored after scenario$/ do |type, name|`

kasturinarra: should be fine i guess, i will update it accordingly

kasturinarra: Addressed

kasturinarra:
@zhouying7780 @liangxia could you please help review ? Thanks!!

PR #1761, opened 2020-12-23
Add build option

wewang58:
@openshift/devexp-qe  PTAL thanks

### DIFF
--- a/lib/rules/cli/4.5.yaml
+++ b/lib/rules/cli/4.5.yaml
@@ -1091,6 +1091,7 @@
   :options:
     :buildconfig: <value>
     :build_arg: --build-arg=<value>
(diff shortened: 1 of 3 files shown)


lib/rules/cli/4.5.yaml
```diff
@@ -1091,6 +1091,7 @@
   :options:
     :buildconfig: <value>
     :build_arg: --build-arg=<value>
+    :build_loglevel: --build-loglevel=<value>
```
xiuwang, line 1094: How about 4.4.yaml file, do we need update this too?

wewang58: Yes, missed it, Updated 4.4 yaml now.

PR #1769, opened 2020-12-26
[WIP] added kata-webhook support

pruan-rht:
kata-webhook support using admission controller, which requires admin privileges.  In an effort to run existing tests w/o modification, this PR will leverage off environment variable `USE_KATA_RUNTIME` to install the admission controller for kata when an new project is created.

### DIFF
--- a/features/step_definitions/kata.rb
+++ b/features/step_definitions/kata.rb
@@ -83,3 +83,78 @@ And /^I verify kata container runtime is installed into the a worker node$/ do
   @result = node(pod.node_name).host.exec_admin(node_cmd)
   raise "No qemu process detected inside pod node" unless @result[:response].include? 'qemu'
 end
+
+# generate a cert for the kata-webhook pod which involves
+# 1. generate a cert for the webhook
+# 2. create certs secrets for k8s
+# 3. set the CABundle on the webhook registration yaml
+Given /^I install kata-webhook for the#{OPT_QUOTED} namespace$/ do | ns |
+  # ignore the admin check if it's for kata
+  unless ENV.has_key? 'USE_KATA_RUNTIME' and ENV['USE_KATA_RUNTIME']
+    ensure_admin_tagged
+  end
+
+  org_user = user
+  @user = admin
+  ns ||= project.name
+  project(ns)
+  step %Q{I use the "#{ns}" project}
+  key = OpenSSL::PKey::RSA.new(2048)
+  public_key = key.public_key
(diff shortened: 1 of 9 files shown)


features/step_definitions/kata.rb
```diff
@@ -83,3 +83,78 @@
...
   raise "No qemu process detected inside pod node" unless @result[:response].include? 'qemu'
 end
+
+# generate a cert for the kata-webhook pod which involves
+# 1. generate a cert for the webhook
+# 2. create certs secrets for k8s
+# 3. set the CABundle on the webhook registration yaml
+Given /^I install kata-webhook for the#{OPT_QUOTED} namespace$/ do | ns |
+  # ignore the admin check if it's for kata
+  unless ENV.has_key? 'USE_KATA_RUNTIME' and ENV['USE_KATA_RUNTIME']
```
akostadinov, line 93: you probably want `to_bool(ENV['USE_KATA_RUNTIME']}`, no need to check key exists

features/step_definitions/kata.rb
```diff
@@ -83,3 +83,78 @@
...
+    ensure_admin_tagged
```
akostadinov, line 94: I think we spoke that we don't require admin for this step. It performs operations as admin but we don't put such tag on the test cases, because for non-kata, admin is not needed.

akostadinov:
Could you rename variable to `NEW_PROJECTS_RUNTIME=kata` that would potentially be extensible going forward to other runtimes.

But first would you remind me what was the benefit to do this with an environment modifier instead of creating the cluster with a project template that contains the web hook to begin with and `oc new project` will always produce a project with that inside? It will not be necessary to modify any steps, tags and whatnot.

pruan-rht:
@akostadinov `kata` is a pod spec, so I think the env `NEW_PROJECT_RUNTIME=kata` will be misleading.

WRT to your suggestion of `cluster with a project template that contains the web hook to begin`.  How can we achieve that?  There are system projects (anything with `openshift-` prefix that won't work with kata, so doing what you have suggested will have adverse effects on the cluster.  @cpmeadors, can you please chime in with your take?

cpmeadors:
I don't think we want to force all new project to use kata runtime.  This does not match the use cases I am aware of.  To future proof  the solution, we should have the ability to create projects with different runtimes: for now, runc and kata.  We will need to test that that is the case.  Not sure if the product will use webhook to do that or some other solution(s).  That is another reason to be flexible.

PR #1770, opened 2020-12-28
update nginx-alpine image digest

lihongan:
@xiuwang please help take a look. thanks


lihongan:
@liangxia PTAL, thanks

PR #1771, opened 2020-12-28
Fix parameters in step definition

kasturinarra:
Have updated the params as discussed in the patch here https://github.com/openshift/cucushift/pull/8327#discussion_r548801213


kasturinarra:
@liangxia @zhouying7780 could you please help review ? Thanks !!

PR #1774, opened 2021-01-04
update rules

yapei:


lib/rules/web/admin_console/4.4/common_ui_elements.xyaml
```diff
@@ -597,10 +597,6 @@ check_restricted_access_text:
   params:
     content: Restricted Access
   action: check_page_contains
-check_no_configmaps_found_text:
-  params:
-    content: No Config Maps Found
-  action: check_page_contains
```
yapei, line 603: `check_no_configmaps_found_text` is no longer needed after cucushift#8345 is merged so removed it

lib/rules/web/admin_console/4.5/route.xyaml
```diff
@@ -32,10 +32,6 @@ check_custom_dns_help_info:
   params:
     content: To use a custom route, you must update your DNS provider
   action: check_page_contains
-check_routes_web_page_contains:
-  element:
-    selector:
-      xpath: //*[contains(.,'<content>')]
```
yapei, line 38: this step is not needed since we updated the steps in cucushift#8346

pruan-rht (commented):
/lgtm

PR #1777, opened 2021-01-05
fix failure when finding the index patterns in kibana

QiaolingTang:
Fix failure:
```
      [05:35:09] INFO> found 0 element elements with selector: {:xpath=>"//h2[@class=\"index-pattern-label\" and @id=\"index_pattern_id\" and contains(., \"app*\")] | //span[contains(., \"app*\")]"}
      [05:35:09] INFO> embedding "2021-01-05T05:35:09+00:00-screenshot.png"
      [05:35:09] INFO> embedding "2021-01-05T05:35:09+00:00.html"
      | index_pattern_name | app* |/home/jenkins/ws/workspace/Runner-v3/features/support/cucu_formatter.rb:258: warning: URI.escape is obsolete

    Then the step should succeed                                                               # features/step_definitions/common.rb:4
      the step failed (RuntimeError)
      /home/jenkins/ws/workspace/Runner-v3/features/step_definitions/common.rb:6:in `/^the step should( not)? (succeed|fail)$/'
      features/logging/kibana.feature:89:in `Then the step should succeed'
```
When using `send_keys` option, it sends the characters one-by-one,  but the kibana will add `*` automatically after the first character is sent. 
According to https://github.com/elastic/kibana/blob/master/test/functional/page_objects/settings_page.ts#L388-L414, change the index pattern name to `*app` and `*infra`.

### DIFF
--- a/features/logging/kibana.feature
+++ b/features/logging/kibana.feature
@@ -76,7 +76,7 @@ Feature: Kibana related features
     When I login to kibana logging web console
     Then the step should succeed
     When I perform the :create_index_pattern_in_kibana web action with:
-      | index_pattern_name | app |
+      | index_pattern_name | "*app" |
     Then the step should succeed
     Given I wait up to 300 seconds for the steps to pass:
     """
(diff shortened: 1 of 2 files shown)


QiaolingTang:
Runner-v3 log:  https://url.corp.redhat.com/3afc69b
@anpingli PTAL, thanks.

anpingli (approved):
/lgtm

PR #1786, opened 2021-01-10
better testfile copy

akostadinov:
supersedes #1465

@rbbratta, do you think something important for you is missing from this PR or needs amendment?

/assign @rbbratta


xiuwang:
@akostadinov @liangxia The pr works for me.
/lgtm

PR #1791, opened 2021-01-12
Auto some egressfirewall cases for dnsname

huiran0826:
Auto EgressFirewall cases: OCP-37496,OCP-37495,OCP-37491

Logs:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2794/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2793/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2792/console

@zhaozhanqi @anuragthehatter @rbbratta @weliang1

### DIFF
--- a/features/networking/egress-ingress.feature
+++ b/features/networking/egress-ingress.feature
@@ -485,3 +485,85 @@ Feature: Egress-ingress related networking scenarios
     Then the step should fail
     And the output should contain:
       | spec.egress in body should have at most 1000 items |
+
+  # @author huirwang@redhat.com
+  # @case_id OCP-37491
+  @admin
+  Scenario: [SDN-1181] EgressFirewall allows traffic to destination dnsName
+    Given I have a project
+    Given I have a pod-for-ping in the project
+
+    When I obtain test data file "networking/ovn-egressfirewall/egressfirewall-policy4.yaml"
+    And I run the :create admin command with:
+      | f | egressfirewall-policy4.yaml  |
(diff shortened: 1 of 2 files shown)


features/networking/egress-ingress.feature
```diff
@@ -485,3 +485,85 @@ Feature: Egress-ingress related networking scenarios
...
+    # Check curl from pod
+    When I execute on the pod:
+      | curl | --connect-timeout | 5 | --head | www.test.com |
+    Then the step should succeed
+    When I execute on the pod:
+      | curl | -k | --connect-timeout | 5 | --head | https://yahoo.com:443 |
+    Then the step should fail
+    When I execute on the pod:
+      | curl | --connect-timeout | 5 | --head | http://yahoo.com:80 |
+    Then the step should succeed
```
weliang1, line 510: How about  | curl | --connect-timeout | 5 | --head | http://yahoo.com |? Should it pass or fail?

huiran0826: Actually it depends on the default port which will be used when just using domain name without port. Generally the default port is 80 and the above step will pass even without specify port in domain name (Tested it and passed).  I considerate that the egress firewall limits the access in port level, and here it only allows 80 for access, so I think   http://yahoo.com:80 is okay and more accurate.

PR #1799, opened 2021-01-19
support added for OCP-37098

yapei:
@yanpzhan  Ready for your review, PTAL thanks

### DIFF
--- a/lib/openshift/catalog_source.rb
+++ b/lib/openshift/catalog_source.rb
@@ -9,5 +9,30 @@ module BushSlicer
   		res[:success] = res[:parsed]["status"]["connectionState"]["lastObservedState"] == "READY"
   		return res
   	end
+
+  	def endpoint(user: nil, quiet: false, cached: true)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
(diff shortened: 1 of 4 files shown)


lib/openshift/catalog_source.rb
```diff
@@ -9,5 +9,30 @@ def ready?(user:, quiet: false, cached: false)
...
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
+      return rr['spec'].has_key?('publisher') ? rr.dig('spec', 'publisher') : '-'	
+  	end
+
+  	def status(user: nil, quiet: false, cached: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
+      return rr.dig('status', 'connectionState', 'lastObservedState')  		
+  	end
+
+  	def displayname(user: nil, quiet: false, cached: false)
```
liangxia, line 33: I think we can cache most of the resources, except status, by default

yapei: @liangxia  Good idea, let me update

yapei: updated, PTAL @liangxia

PR #1800, opened 2021-01-19
Modify to use registry.access.redhat.com/rhscl/ruby-22-rhel7:latst ba…

xiuwang:
…ck instead of ruby-25 image

@openshift/devexp-qe please have a look, thanks

### DIFF
--- a/testdata/build/ruby20rhel7-context-docker.json
+++ b/testdata/build/ruby20rhel7-context-docker.json
@@ -125,7 +125,7 @@
           "dockerStrategy": {
             "from": {
               "kind": "DockerImage",
-              "name": "quay.io/openshifttest/ruby-25-centos7@sha256:575194aa8be12ea066fc3f4aa9103dcb4291d43f9ee32e4afe34e0063051610b"
+              "name": "registry.access.redhat.com/rhscl/ruby-22-rhel7:latest"
             },
             "env": [
               {


testdata/build/ruby20rhel7-context-docker.json
```diff
@@ -125,7 +125,7 @@
           "dockerStrategy": {
             "from": {
               "kind": "DockerImage",
-              "name": "quay.io/openshifttest/ruby-25-centos7@sha256:575194aa8be12ea066fc3f4aa9103dcb4291d43f9ee32e4afe34e0063051610b"
+              "name": "registry.access.redhat.com/rhscl/ruby-22-rhel7:latest"
```
wewang58, line 128: Can we use  newer ruby "registry.access.redhat.com/rhscl/ruby-25-rhel7" instead of ruby-22-rhel7 in here?

xiuwang: @wewang58 we used an old repo https://github.com/openshift-qe/ruby-hello-world-context.git in the testfile to make the check point of case, so only ruby-22-* images could match the repo.

wewang58: Got it
/lgtm

liangxia: Please make sure the test cases using this test file set disconnect to false in Polarion.

xiuwang: @liangxia The test case has set Disconnect = No

PR #1649, opened 2020-11-11
add jitter to wait_for delay to reduce contention

rbbratta:
The `I wait up to 30 seconds for the steps to pass`
step calls `wait_for` but does not allow for changing
the interval.

The proper way to do retry loops in a distributed system is
to use jitter and/or exponential backoff as described in the Amazon paper.

https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

Implement the Decorrelated Jitter algorithm in the wait_for
method.

PR #905, opened 2020-04-01
Automate OCP-28968

weliang1:
Auto cases OCP-28968: IPv6 testing for OCP-21151: Create pods with multus-cni - macvlan bridge mode	

Test case in polarion:
https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-28968

Automation logs:
https://openshift-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Runner-v3-smoke/855/console

@zhaozhanqi @huiran0826 @anuragthehatter @rbbratta
Please help to review both test case and automation scripts, thanks!

PR #1804, opened 2021-01-19
Automate OCP-38521 Whereabouts should exclude IPv6 ranges

weliang1:
Automate OCP-38521 Whereabouts should exclude IPv6 ranges

Test log: https://privatebin-it-iso.int.open.paas.redhat.com/?c1403aaf0fbf6db1#Bf6R8xghZPCd42RCXEUU3bfjXJazMePe3x9S2jZbSyBf

@zhaozhanqi @huiran0826 @anuragthehatter @rbbratta @asood @dbrahane


features/networking/multus-ipv6.feature
```diff
@@ -50,3 +50,64 @@ Feature: Multus-CNI ipv6 related scenarios
...
+    Given I obtain test data file "networking/multus-cni/Pods/IPv6/ipv6-pod.yaml"
+    When I run oc create over "ipv6-pod.yaml" replacing paths:
+      | ["spec"]["nodeName"] | "<%= cb.target_node %>" |
+    Then the step should succeed
+    When I wait up to 30 seconds for the steps to pass:
+    """
+    When I run the :describe client command with:
+      | resource | pod |
+    Then the output should contain:
+      | fd00:dead:beef:1::1 |
```
pruan-rht, line 82: @weliang1 hahah I used to write the pattern `0xdeadbeef` at one of my first firmware job's memory diagnostic test. :)

anuragthehatter (commented):
cc @brahaney @kedark3

anuragthehatter:
@weliang1 May be we can keep a generic ```ipv6-pod.yaml``` and just change its parameters wherever we use it

features/networking/multus-ipv6.feature
pruan-rht: to make it more English like he should've used `f00d:dead:beef:cafe::1`

weliang1: `fd00:dead:beef::` is a valid IPv6 unique local addresses and suggested by @rbbratta in https://github.com/openshift/verification-tests/pull/905.

weliang1:
> @weliang1 May be we can keep a generic `ipv6-pod.yaml` and just change its parameters wherever we use it. The difference btw both pod yamls is net-attach-def name and pod names.

@anuragthehatter Good suggestion. Remove three files and re-use existing NAD and pod yaml files now.

weliang1:
New test log: https://privatebin-it-iso.int.open.paas.redhat.com/?15ff106e0f360063#4zsaR2yS4PzGhJyhnBPmbRHwCEJTx47a9xFhFf9XKyqS

features/networking/multus-ipv6.feature
```diff
@@ -50,3 +50,70 @@ Feature: Multus-CNI ipv6 related scenarios
...
+    And evaluation of `node.name` is stored in the :target_node clipboard
+    # Create the net-attach-def via cluster admin
+    Given I have a project
+    Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/whereabouts-excludeIP.yaml"
+    When I run oc create as admin over "whereabouts-excludeIP.yaml" replacing paths:
+      | ["metadata"]["name"]      | whereabouts-excludeipv6                                                                                                                                                                                                  |
+      | ["metadata"]["namespace"] | <%= project.name %>                                                                                                                                                                                                      |
+      | ["spec"]["config"]        | '{ "cniVersion": "0.3.0", "name": "whereabouts", "type": "macvlan", "mode": "bridge", "ipam": { "type": "whereabouts", "range": "fd00:dead:beef:1::1-fd00:dead:beef:1::4/64", "exclude": ["fd00:dead:beef:1::2/128"] } }'|
```
kedark3, line 71: Is there a line length limit at which we can break this line to new line? Or it has to be on this line? Curious, as in Python we enforced line break after 80 characters.

weliang1: @kedark3  Good question, I am not really sure about the characters limitation for our cucushif testing, just follow the existing scripts format. Other forks may know it.

kedark3: Thanks @weliang1 I would definitely like to see if we can break this line. I am new to this repo so won't be able to tell if we can do that :man_shrugging:

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
```diff
@@ -0,0 +1,25 @@
...
+  template:
+    metadata:
+      annotations:
+        k8s.v1.cni.cncf.io/networks: whereabouts-excludeipv6
+      labels:
+        name: test-pod
+    spec:
+      containers:
+      - name: test-pod
+        image: quay.io/openshifttest/hello-sdn@sha256:d5785550cf77b7932b090fcd1a2625472912fb3189d5973f177a5a2c347a1f95
```
kedark3, line 19: Is it standard and required to hardcode image sha256 tag? Or we can just leave `quay.io/openshifttest/hello-sdn` and let it fetch the latest image? Ideally image should never change in such a way that it will break any existing tests right?

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
```diff
@@ -0,0 +1,25 @@
...
+        ports:
+        - containerPort: 8080
+        - containerPort: 443
+        env:
+        - name: RESPONSE
+          value: test-pod
```
kedark3, line 25: **Request:** Yaml-Lint would want us to have a newline character at the end here. 
image this symbol tells us that there is no newline character at the end of file

features/networking/multus-ipv6.feature
```diff
@@ -50,3 +50,70 @@ Feature: Multus-CNI ipv6 related scenarios
...
+    Then the step should succeed
+
+    # Create the three pods which consumes the custom resource
+   Given I obtain test data file "networking/multus-cni/Pods/IPv6/ipv6-pod.yaml"
+    When I run oc create over "ipv6-pod.yaml" replacing paths:
+      | ["spec"]["nodeName"]                                                           | "<%= cb.target_node %>"      |
+      | ["metadata"]["name"]                                                           | test-pod-list                |
+      | ["spec"]["template"]["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | whereabouts-excludeipv6-list |
+    Then the step should succeed
+    When I wait up to 30 seconds for the steps to pass:
```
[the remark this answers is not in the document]
weliang1: @kedark3 No problem, I do the same when I learn cucushift automation. Above step is not really wait 30 seconds, it means that the script will give total 30 seconds to let below steps pass. It may wait for 20 or 25 seconds which depending on the cluster environment. Sometimes it's very difficult to give exactly time. This estimate hard code time is acceptable in some scripts.

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
weliang1: The image name is fixed.

features/networking/multus-ipv6.feature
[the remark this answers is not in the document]
kedark3: @weliang1 thanks for that info. In my previous team we used `Wait_for(TaskX, Timeout=XXSeconds)` python based function from https://pypi.org/project/wait-for/ project. Is it correct if I say, in your code, the value 30 second is same as timeout from my example?

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
weliang1: Good catch, fixed it. Thanks!

features/networking/multus-ipv6.feature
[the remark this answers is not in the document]
weliang1: @kedark3 I guess they are same things. cucushif call it "I wait up to 30 seconds"

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
kedark3: Is that standard way we do things in this repo?

weliang1: The way we did here "image: quay.io/openshifttest/hello-sdn@sha256:d5785550cf77b7932b090fcd1a2625472912fb3189d5973f177a5a2c347a1f95" is standard way. sha256:d5785550cf77b7932b090fcd1a2625472912fb3189d5973f177a5a2c347a1f95 is container image ID, different image has different ID. For this testing I use this container image

features/networking/multus-ipv6.feature
pruan-rht: @kedark3 @weliang1 this has the coding guideline https://github.com/openshift/verification-tests/blob/master/doc/coding_guide.adoc. With tables, we align the `|` and historically we don't break the line even though they go over 80 characters sometimes.

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
anuragthehatter: I agree its not a compulsion now. We use https://quay.io/organization/openshifttest as our qe testfiles inventory repo. In past I had seen it to use different tags for whatever reason so we got accustomed to this sha256 practise. But seems like its okay to point to latest now without specifying tag.

features/networking/multus-ipv6.feature
```diff
@@ -50,3 +50,70 @@ Feature: Multus-CNI ipv6 related scenarios
...
+    When I run oc create as admin over "whereabouts-excludeIP.yaml" replacing paths:
+      | ["metadata"]["name"]      | whereabouts-excludeipv6                                                                                                                                                                                                  |
+      | ["metadata"]["namespace"] | <%= project.name %>                                                                                                                                                                                                      |
+      | ["spec"]["config"]        | '{ "cniVersion": "0.3.0", "name": "whereabouts", "type": "macvlan", "mode": "bridge", "ipam": { "type": "whereabouts", "range": "fd00:dead:beef:1::1-fd00:dead:beef:1::4/64", "exclude": ["fd00:dead:beef:1::2/128"] } }'|
+    Then the step should succeed
+
+    # Create the three pods which consumes the custom resource
+    Given I obtain test data file "networking/multus-cni/Pods/IPv6/ipv6-pod.yaml"
+    When I run oc create over "ipv6-pod.yaml" replacing paths:
```
anuragthehatter, line 76: @weliang1 I suggest an optional improvement here to assist debugging in future for people not familier with multus or trying to debug this flow. 
pod ```metadata.name``` and ```metadata.annotations.k8s.v1.cni.cncf.io/networks``` are not defined here because we are fetching their default values from ipv6-pod.yaml which is logically correct but it might be good to define them here (like what you did at Line 103 and 104) considering ipv6-pod.yaml is just a template. That way a debugger can correlate further steps without going and search for ipv6-pod.yaml file to look at. Doing this will depict that we are leveraging ```whereabouts-excludeipv6``` def created before this step and with any pod name. Just a small suggestion from readability and debugging perspective

features/networking/multus-ipv6.feature
kedark3: Thanks @pruan-rht

features/networking/multus-ipv6.feature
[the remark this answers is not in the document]
kedark3: Thanks @weliang1

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
kedark3: @anuragthehatter thanks for clarifying. I'd like to see if we could stick to latest, in my limited understanding, it seems best to do so. That way all our tests running same image, not different SHAs. And that way if a test fails, we know its not due to an image. And if we see many tests fail, we know we broke the image.

features/networking/multus-ipv6.feature
weliang1: @anuragthehatter Agree, just fixed it. Thanks!

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
weliang1: @kedark3 Just remove sha256 in the new commit. Thanks!

anuragthehatter:
Thanks

/LGTM

features/networking/multus-ipv6.feature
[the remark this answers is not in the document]
pruan-rht: FYI, from the step name it is `wait up to xxx seconds`, so that the time out value.  The loop will exit once the check within the enclosing step evaluate to `true`

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
anuragthehatter: Oh wait. digest is indeed required if we are testing on disconnected cluster as per #850. Disconnected clusters (not connected to internet or via proxies) will setup registry locally on the cluster itself and need digests to access the image cc @zhaozhanqi WDYT

anuragthehatter:
Waiting to resolve digest specifying argument clarification

/LGTM cancel

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
pruan-rht: yes, keeping the digest is the recommended way to go...not very human readable though.

features/networking/multus-ipv6.feature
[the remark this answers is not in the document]
rbbratta: @kedark3 I'm trying to introduce jitter for `wait_for` here  https://github.com/openshift/verification-tests/pull/1649

testdata/networking/multus-cni/Pods/IPv6/ipv6-pod.yaml
kedark3: oh interesting. I wish there was way to just say `latest` and not worry.

zhaozhanqi: digest is indeed required for disconnected. I remembered tag is not be supported before. So better use digest

weliang1: @zhaozhanqi Thanks for the confirmation. The new commit add digest back to image name for disconnected cluster.

kedark3: @weliang1 sorry if I caused confusion here. Thanks for putting up with changes twice. I know it can be unpleasant..

weliang1: @kedark3 No worry, it's process to learn something new.

PR #1809, opened 2021-01-21
Fix upgrade failures

kasturinarra:
Removing the code which was added for 4.3 as wanted to make sure that case works fine for upgrades from 4.4 to higher. For 4.3 upgrades will deal separately or keep it manual.


kasturinarra:

kasturinarra:
@liangxia could you please help review and merge ? Thanks !!

PR #1814, opened 2021-01-25
report portal support

akostadinov:


akostadinov:
@liangxia , I did local tests due to instability of PSI. Everything appears to be working alright. If you can run an actual run so then we can merge the changes, that would be of great help.

liangxia:
Sure thing.  I'll work on it today and update results there later.

liangxia:
Tested one test run with non-admin test cases and admin test cases, both work fine.

/lgtm

PR #1815, opened 2021-01-25
Adding RVM info

kedark3:
@pruan-rht adding some info about RVM for new users of this repo. Please review.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -9,7 +9,7 @@ More info you can find in link:doc/overview.adoc[Overview document].
 . Install git if not installed already, install ruby 2.4+ if not installed
 . git clone git@github.com:openshift/verification-tests.git
 . cd verification-tests
-. tools/install_os_deps.sh # need sudo; you may need to login again to terminal if ruby was installed via rvm during this phase
+. tools/install_os_deps.sh # need sudo; you may need to login again to terminal if ruby was installed via rvm during this phase (see below for RVM usage)
 . tools/hack_bundle.rb # normal user
 . Install OKD or OpenShift Container Platform client tools
 . Install the web driver if you desire to run tests which work with the Web UI. See link:doc/configuration.adoc[Web Driver]
@@ -68,3 +68,27 @@ It is in link:.gitignore[.gitignore] so that you can keep your test scenarios un
 == Contributing
 
 Please submit issues and pull requests to the project. You can discuss and ask questions on https://lists.openshift.redhat.com/openshiftmm/listinfo/dev[OpenShift mailing list].
+
+
+=== RVM for dependency management
+
+If you are planning on installing dependencies for verification Tests on your system, it is often a good idea to use some kind of Version Manager/Package manager, like RVM.
+
+RVM is analogous to virtualenv in python and it stands for Ruby Version Manager.
+
+You can install RVM and read more abou it on https://rvm.io/
+
+Usually once RVM is installed, you would want to create/install a new `rvm` interpreter using:
+
+```rvm install "ruby-2.7.1" && rvm  --create use 2.7.1@verification-tests``` which will install Ruby 2.7.1 and create a new interpreter as well as a Gemset.
+When we install Ruby 2.7.1, it creates a local directory for your user with Ruby 2.7.1 under `~/.rvm/rubies/ruby-2.7.1`.
+Once that is done, the Gemset creation is basically an attempt to have dependencies for various apps under same interpreter `2.7.1`. So in this case,
+we are creating Gemset `verification-tests` for this repo dependencies. If you had multiple repos, you would use similar command as above to create a new Gemset under 2.7.1
+and give it a name that you can recognize. Then for each respective app you may choose to use different gemset as `rvm use 2.7.1@verification-tests` or `rvm use 2.7.1@my-other-repo-or-app`
+
+Feel free to read about interpreter and gemset in the Ruby docs or elsewhere on the internet.
+
+Once ruby is installed, interpreter is created and gemset is present and activated, you can run `which ruby` to validate if the correct ruby is used and `ruby -v` to check version.
+Following this, you may continue the steps as mentioned above, to run `bundle` commands to install gems listed in Gemfile to your Gemset.
+
+Also if you are wondering what does `bundle` command do, search and read about `ruby bundler` on the internet.


README.adoc
```diff
@@ -9,7 +9,7 @@ More info you can find in link:doc/overview.adoc[Overview document].
 . Install git if not installed already, install ruby 2.4+ if not installed
 . git clone git@github.com:openshift/verification-tests.git
 . cd verification-tests
-. tools/install_os_deps.sh # need sudo; you may need to login again to terminal if ruby was installed via rvm during this phase
+. tools/install_os_deps.sh # need sudo; you may need to login again to terminal if ruby was installed via rvm during this phase(see below for RVM usage)
```
rbbratta, line 12: space maybe before paren `phase (see`

README.adoc
```diff
@@ -68,3 +68,27 @@ It is in link:.gitignore[.gitignore] so that you can keep your test scenarios un
...
+Once that is done, the Gemset creation is basically an attempt to have dependencies for various apps under same interpreter `2.7.1`. So in this case,
+we are creating Gemset `verification-tests` for this repo dependencies. If you had multiple repos, you would use similar command as above to create a new Gemset under 2.7.1
+and give it a name that you can recognize. Then for each respective app you may choose to use different gemset as `rvm use 2.7.1@verification-tests` or `rvm use 2.7.1@my-other-repo-or-app`
+
+Feel free to read about interpreter and gemset in the Ruby docs or elsewhere on the internet.
+
+Once ruby is installed, interpreter is created and gemset is present and activated, you can run `which ruby` to validate if the correct ruby is used and `ruby -v` to check version.
+Following this, you may continue the steps as mentioned above, to run `bundle` commands to install gems listed in Gemfile to your Gemset.
+
+Also if you are wondering what does `bundle` command do, search and read about `ruby bundler` on the internet.
```
rbbratta, line 94: also make sure you `rvm use` in the shell to set the environment variables before you start VS.code or other GUI IDE, otherwise `GEM_HOME`  and other env vars might not be set.

README.adoc
```diff
@@ -68,3 +68,27 @@ It is in link:.gitignore[.gitignore] so that you can keep your test scenarios un
...
+If you are planning on installing dependencies for verification Tests on your system, it is usually a good idea to use some kind of Version Manager/Package manager, like RVM.
+
+RVM is analogous to virtualenv in python and it stands for Ruby Version Manager.
+
+You can install RVM and read more abou it on https://rvm.io/
+
+Usually once RVM is installed, you would want to create/install a new `rvm` interpreter using:
+
+```rvm install "ruby-2.7.1" && rvm  --create use 2.7.1@verification-tests``` which will install Ruby 2.7.1 and create a new interpreter as well as a Gemset.
+When we install Ruby 2.7.1, it creates a local directory for your user with Ruby 2.7.1 under `/home/<user>/.rvm/rubies/ruby-2.7.1`.
```
rbbratta, line 84: aka `$HOME` or `~/`?

kedark3: good suggestion. Didn't know if there could be other home dirs haha. Never heard of DYNIX. I will update this.

README.adoc
kedark3: I haven't yet tried with VSCode. Maybe I'll ask you for tips and we can add more about it in a separate PR to help setup VS Code to our new hires.

README.adoc
```diff
@@ -68,3 +68,27 @@ It is in link:.gitignore[.gitignore] so that you can keep your test scenarios un
 == Contributing
 
 Please submit issues and pull requests to the project. You can discuss and ask questions on https://lists.openshift.redhat.com/openshiftmm/listinfo/dev[OpenShift mailing list].
+
+
+=== RVM for dependency management
+
+If you are planning on installing dependencies for verification Tests on your system, it is usually a good idea to use some kind of Version Manager/Package manager, like RVM.
```
akostadinov, line 75: `usually a good idea`, I'm not sure about it. I just use the OS version of Ruby and rarely, when I need to try another version, then I use RVM. I would change this to `sometimes a good idea`.

kedark3: @akostadinov good point. I, coming from Python world, Ruby Gems installation and management feels very similar to Python Pip package dependencies. In my humble opinion, I still feel its a good idea to have RVM or something similar. I have created a RVM environment named `2.7.1@verification-tests` so that I can use that for this repo while I can create other envs for other repo. So unless someone is planning to have only one ruby project on their system, it feels like a good idea to have separate env. In future if you add another project needing different versions, you will eventually need RVM or something, right? 

I am willing to accept the change you are requesting, just trying to provide my rationale though. Probably you know much more Ruby than I do.

pruan-rht: I personally use RVM by default instead of messing the system Ruby.  Changing the wording is not that big deal.  Also, @kedark3, you probably put way more information with regards to how to use RVM than needed.  :)

akostadinov: @kedark3 , I'm also fine if you use the word "often" instead of "sometimes". I just don't see any reason for most people to install RVM just to mess up with verification-tests.

Another reason I don't like it is because this means installing yet new non-RH built binaries on local workstation which is a potential risk. Not that I'm too much worried.

README.adoc
akostadinov: Sounds like some UNIX variant. It is usual to see home dirs in non-standard locations where shared storage is being used.

pruan-rht: just `~/.rvm` would be sufficient.  For Mac user it's `/Users/pruan/.rvm`.  Anyway,  `~/.rvm` would work regardless.

README.adoc
pruan-rht: @akostadinov I've sometime seen issue with localhost/vm when doing it against the system ruby.  For example, I recommended @kedark3 to use RVM after his issue with `bundle`  ```bundle install
Ignoring byebug-11.1.3 because its extensions are not built. Try: gem pristine byebug --version 11.1.3
Ignoring json-2.5.1 because its extensions are not built. Try: gem pristine json --version 2.5.1
Ignoring oga-3.3 because its extensions are not built. Try: gem pristine oga --version 3.3
Ignoring psych-3.3.0 because its extensions are not built. Try: gem pristine psych --version 3.3.0
Ignoring racc-1.5.2 because its extensions are not built. Try: gem pristine racc --version 1.5.2
Ignoring ruby-ll-2.1.2 because its extensions are not built. Try: gem pristine ruby-ll --version 2.1.2
Ignoring unf_ext-0.0.7.2 because its extensions are not built. Try: gem pristine unf_ext --version 0.0.7.2
Fetching gem metadata from https://rubygems.org/......
Using concurrent-ruby 1.1.8```

It was straightforward and works when he uses rvm

pruan-rht: @akostadinov the same applies to @asood-rh I believe.

README.adoc
```diff
@@ -9,7 +9,7 @@ More info you can find in link:doc/overview.adoc[Overview document].
 . Install git if not installed already, install ruby 2.4+ if not installed
```
asood-rh, line 9: Can we update this with latest version of ruby that is being used? Alternatively mention it somewhere in the document.

pruan-rht: @asood-rh I don't think we need to change that as `2.4+` indicated that the minimum version is 2.4 or above.

README.adoc
akostadinov: This happens usually after OS upgrade and the native gem extensions need to be rebuilt (or gems removed and reinstalled). If you upgrade your Fedora OS and use RVM, I expect you to hit the same issue.

akostadinov: Or maybe it happened because system ruby got updated to a new version. I'm not sure.

README.adoc
asood-rh: @pruan-rht Would it work with ruby 3.0.0 as well? If that is supported then it does not need to change otherwise intent of comment is to clarify the exact version.

pruan-rht: @asood-rh I haven't tried it...it's better to just put the minimum instead of a specific version, otherwise we would need to keep updating the doc.

README.adoc
kedark3: For the purpose of this PR I will leave out VS Code and I will add that as I set that up on my local env in a separate PR.

README.adoc
kedark3: Updated this.

README.adoc
kedark3: @akostadinov probably something on my system was not right, but as @pruan-rht pointed out, using RVM was straight forward and easy, keeping separation from system level Ruby. I will update it to "often" as you said in above comment and resolve this thread. Please feel free to re-open if that is not sufficient.

README.adoc
kedark3: @asood-rh I agree with @pruan-rht  here to keep minimum version. Adding different versions can be contentious and would need update to README. Even if we test it now against, lets say, 3.0.0, we may have to test again and update the compatibility. Not that its not a good idea, we can definitely make an effort to keep that up-to-date, but for the purpose of this PR I'd leave the version info as is.

kedark3:
@akostadinov @pruan-rht Please let me know if we are ok to merge this PR?

pruan-rht (commented):
/lgtm

PR #1817, opened 2021-01-26
update channel name

QiaolingTang:
Since the channel name `4.7` is dropped, logging couldn't subscribe from `4.7` channel, so here update the channel name.


anpingli (commented):
/lgtm

pruan-rht (commented):
/lgtm

PR #1820, opened 2021-01-27
update windows machine template

sunzhaohua2:
Update windows template on vsphere, confirmed with sgao, we will use template `openshift-qe-template-windows-2019` and try to keep the name unchanged in the future. @jhou1 @miyadav please help to take a look.


pruan-rht (commented):
/lgtm

PR #1822, opened 2021-01-27
Adding tenant to allocated FIP addresses description

pjanouse:


pjanouse:
@akostadinov can you review please?

lib/launchers/openstack.rb
```diff
@@ -763,7 +763,8 @@ def allocate_floating_ip(network_name, reuse: true)
         floatingip: {
           floating_network_id: network_id,
           project_id: self.os_tenant_id,
-          description: "IP allocated automatically by OpenShift verification-tests"
+          description: "IP allocated automatically by OpenShift verification-tests" \
+            + ((tenant.nil? || tenant.length == 0) ? "":" (#{tenant})")
```
akostadinov, line 767: Do we need tenant as a parameter? Any drawbacks to just use `#os_tenant_name` method unconditionally and remove the `tenant` parameter?

pjanouse: Parameter name changed to `designator` to avoid such a confusion

PR #1827, opened 2021-01-28
Update rules in OCM due to new change of cluster creation and taints …

xueli181114:
Here is the Jenkins console log 
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/268068/console


xueli181114:
Here is corresponding MR https://github.com/openshift/cucushift/pull/8398

PR #1828, opened 2021-01-29
Update rule for OCP-32152

wewang58:
The case version >=4.6
Related pr: https://github.com/openshift/cucushift/pull/8400
@yapei  please review it, thanks


wewang58:
@yapei Updated rules and passed in 4.6 and 4.7 cluster, please review it again.

PR #1830, opened 2021-01-29
update for auto-ocp-37363

zhouying7780:
Add lib for `oc image info` command , and update "I have a registry in my project" with deploy. I've checked no case use the step: I have a registry in my project.  So the change will not effect other case. 

@kasturinarra  Please help review, thanks.


pruan-rht (commented):
/lgtm

PR #1832, opened 2021-01-29
change the ruby-hello-world app from docker to quay

wangke19:


wangke19:
Incorrect content in PR.

PR #1838, opened 2021-02-01
enhancce OCP-38812 to support private cluster

lihongan:
Ensure the "shards" ingresscontroller always use the same LB scope to the default ingresscontroller, so the case can be executed on private cluster as well.

log
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3/276980/console

@quarterpin Please help take a look. Thanks

cc @jechen0648


quarterpin:
/lgtm

Thanks @lihongan !

pruan-rht (commented):
/lgtm

PR #1840, opened 2021-02-01
Change the pod image to a valid image

kedark3:
Updating the image for the testcase as the previous image was missing/invalid causing `ImagePullBackOff`

Test could be: `bundle exec cucumber features/networking/pod.feature:24`

### DIFF
--- a/testdata/pods/ocp10031/pod.json
+++ b/testdata/pods/ocp10031/pod.json
@@ -10,7 +10,7 @@
   "spec": {
       "containers": [{
         "name": "hello-pod",
-        "image": "quay.io/openshifttest/hello-openshift-getent"
+        "image": "quay.io/openshifttest/hello-sdn@sha256:d5785550cf77b7932b090fcd1a2625472912fb3189d5973f177a5a2c347a1f95"
       }]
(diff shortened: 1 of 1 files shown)


pruan-rht (commented):
/lgtm

testdata/pods/ocp10031/pod.json
```diff
@@ -10,7 +10,7 @@
   "spec": {
       "containers": [{
         "name": "hello-pod",
-        "image": "quay.io/openshifttest/hello-openshift-getent"
+        "image": "quay.io/openshifttest/hello-sdn@sha256:d5785550cf77b7932b090fcd1a2625472912fb3189d5973f177a5a2c347a1f95"
```
asood-rh, line 13: Do you have to provide the sha? Would it make more sense to use 'latest' tag if needed instead?

pruan-rht: @asood-rh if you look at the existing yaml/json files they are using `sha`

kedark3:
That was a quick merge. Thanks @pruan-rht

PR #1841, opened 2021-02-02
correct wrong configurations in secure fluentd server

QiaolingTang:
Runner-v3 log:  https://url.corp.redhat.com/342ffa2
@anpingli PTAL


anpingli (commented):
/lgtm

PR #1847, opened 2021-02-05
Allow to skip waiting for OSD metrics

apodhrad:
Sometimes, the metrics are synced after 3 hrs. The bug is reported at
https://issues.redhat.com/browse/OHSS-2640

Some test cases do not require these metrics.
So, we should allow to skip the waiting for them.

### DIFF
--- a/lib/launchers/o_c_m_cluster.rb
+++ b/lib/launchers/o_c_m_cluster.rb
@@ -263,12 +263,19 @@ module BushSlicer
     # Wait until OSD cluster is ready and OCP version is available
     # NOTE: we need to wait for registering all metrics - OCP version indicates this state
     def wait_for_osd(osd_name)
+      ocp_version = "NONE"
       loop do
         osd_status = get_value(osd_name, "state")
-        ocp_version = get_value(osd_name, "openshift_version")
         logger.info("Status of cluster #{osd_name} is #{osd_status} and OCP version is #{ocp_version}")
-        if osd_status == "ready" && ocp_version != "NONE"
-          break
+        if osd_status == "ready"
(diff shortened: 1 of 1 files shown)


apodhrad:
I have rebased the PR. Could someone have a look at it, please?
cc @akostadinov @liangxia @yanpzhan

lib/launchers/o_c_m_cluster.rb
```diff
@@ -263,10 +263,17 @@ def create_ocpinfo_file(osd_name, dir, filename = 'OCPINFO.yml')
     def wait_for_osd(osd_name)
       loop do
         osd_status = get_value(osd_name, "state")
-        ocp_version = get_value(osd_name, "openshift_version")
+        ocp_version = "NONE"
```
liangxia, line 266: Shall we put this line outside of the loop ?

lib/launchers/o_c_m_cluster.rb
```diff
@@ -263,10 +263,17 @@ def create_ocpinfo_file(osd_name, dir, filename = 'OCPINFO.yml')
...
         logger.info("Status of cluster #{osd_name} is #{osd_status} and OCP version is #{ocp_version}")
```
liangxia, line 267: Also better to relocate this logger line, too.

lib/launchers/o_c_m_cluster.rb
apodhrad: Hi @liangxia I have put it outside of the loop.

lib/launchers/o_c_m_cluster.rb
apodhrad: Hi @liangxia I suggest to keep the logger line as I think it is useful to print actual status to the log.

PR #1853, opened 2021-02-10
added case OCP-38468, OCP-36508 and OCP-36509

pruan-rht:


features/node/kata/install_uninstall.feature
```diff
@@ -0,0 +1,54 @@
+Feature: kata related features
+  # @author pruan@redhat.com
+  # @case_id OCP-36508
+  @admin
+  @destructive
+  @upgrade-prepare
+  Scenario: kata container operator installation
```
liangxia, line 7: If we tag it upgrade-prepare, then we need to postfix it with "- prepare"
Also, we need a scenario for upgrade check.

testdata/kata/OCP-38468/pod_with_init_container.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: myapp-pod
+  labels:
+    app: myapp
+spec:
+  containers:
+  - name: myapp-container
+    image: busybox:1.28
```
liangxia, line 10: Better to use image from quay as there are limits from docker

pruan-rht: @liangxia updated

features/node/kata/install_uninstall.feature
pruan-rht: @liangxia took those tags out.

pruan-rht:
@liangxia PTAL again, your comments have been addressed.

features/node/kata/install_uninstall.feature
```diff
@@ -0,0 +1,41 @@
...
+
+  # @author pruan@redhat.com
+  # @case_id OCP-36509
+  @admin
+  @destructive
+  Scenario: test delete kata installation
+    Given I remove kata operator from the namespace
+
+
+  Scenario: test install kata-webhook
```
liangxia, line 19: Is it a valid scenario ?

pruan-rht: @liangxia it will be, I just need to add the testcase into Polarion to get the proper Polarion testcase ID

features/node/kata/install_uninstall.feature
pruan-rht: @liangxia actually I do need that `tag` even though I'm not doing an upgrade....it's due to I'm callling this [step](https://github.com/openshift/verification-tests/blob/master/features/step_definitions/scc.rb#L65)

pruan-rht: @akostadinov @liangxia why do we need to postfix it with `-prepare`?  just a file naming convention?

lib/openshift/flakes/container.rb
```diff
@@ -12,11 +12,24 @@ def initialize(name:, pod:)
...
     # return @Hash information of container status matching the @name variable
-    def status(user: nil, cached: true, quiet: false)
-      container_statuses = @pod.status_raw(user: user, cached: cached, quiet: quiet)['containerStatuses']
+    # the parameter `type:` can be containerStatuses or initContainerStatuses.
+    #    shorthand as 'init'
+    #    we are defaulting it to containerStatuses.  For init container we are
+    #    not checking the name since we don't have a way to find out what
+    #    that would be
+    def status(user: nil, type: nil, cached: true, quiet: false)
+      valid_types = ['containerStatuses', 'initContainerStatuses']
```
akostadinov, line 21: can't we figure out automatically what type of container this is?

PR #1855, opened 2021-02-18
Cloud usage enhanced

pruan-rht:
replaces #1740


tools/instance_summary.rb
```diff
@@ -176,13 +176,19 @@ def get_summary(target_region: nil, options: nil)
...
           region.region_name = target_region
         end
-        threads << Thread.new(Amz_EC2.new(region: region.region_name)) do |aws|
-          instances = aws.get_instances_by_status('running')
-          aws_instances[region.region_name] = instances
-          break if target_region == region.region_name
-        end
+        aws = Amz_EC2.new(region: region.region_name)
+        instances = aws.get_instances_by_status('running')
+        binding.pry if instances.count > 0
```
akostadinov, line 181: leftover?

pruan-rht: yeah, fixed now.

pruan-rht:
@liangxia @JianLi-RH @dis016 PTAL, I've been running it in my own branch and static host.  I would like to have it run with a dynamic host.  Thx!

tools/creds/slack_users_map.yaml
```diff
@@ -0,0 +1,110 @@
+---
```
liangxia, line 1: Perhaps we need to sort the file ?

pruan-rht: @liangxia not necessary, I create the file and add the user as they join at the moment.

tools/creds/slack_users_map.yaml
```diff
@@ -0,0 +1,110 @@
...
+aleskandro: U02AJU6BKT2
+Shivanthi: U029X7M6KGT
+Melvin Joseph: U029EJV606T
+Ying Zhang: U029UUDF2D8
+Abhishek Banerjee: U029EH4T6H5
+shudili: U02C0U93S9H
+liangli: U028EFFD8JU
+yingzhan: U029UUDF2D8
+jshu: U02CPDBQE3A
+jiezhao: U01UYTWLTPW
```
akostadinov, line 110: Such user lists IMO are better off in some internal repo. Even path used `creds` suggests they are not good for public exposure.

pruan-rht: @akostadinov you mean in the `cucushift-internal` repo?  These are just Slack user alias and their corresponding user ID.  There's nothing secret about it, that's why I didn't put them into the internal repo to begin with.  There's a way to look up the users dynamically, but it's a slow process, which is why I went with a static mapping approach.

akostadinov: Not only a matter of secrecy. It looks more like configuration for whatever will be scheduling this operation. Not something to belong to code repo.

pruan-rht: @akostadinov @liangxia ok, I'm changed the implementation and generate the user information dynamically instead of hardcoding users.  It's looking for everyone in the `openshift-qe` slack group.  PTAL.  Please note it's dependent on https://gitlab.cee.redhat.com/aosqe/cucushift-internal/-/merge_requests/9/

pruan-rht: @liangxia @akostadinov now that above PR is merged, this should work now.  PTAL and merge if there are no more issues.

lib/launchers/amz.rb
```diff
@@ -97,6 +97,11 @@ def arn
       current_user.arn
     end
 
+    def iam_client
+      self.arn
+      @client_iam
+    end
```
akostadinov, line 103: This method is redundant, there is already a `#client_iam` method.

lib/launchers/amz.rb
```diff
@@ -575,7 +580,11 @@ def instance_uptime(instance)
     end
 
     def instance_name(instance)
-      instance.tags.select { |i| i['value'] if i['key'] == "Name" }.first['value']
+      begin
+        instance.tags.select { |i| i['value'] if i['key'] == "Name" }.first['value']
+      rescue
+        print "******\n"
```
akostadinov, line 586: debugging leftover?

lib/launchers/azure.rb
```diff
@@ -174,7 +174,12 @@ def instance_uptime(inst_view, src: "vm_disk")
     end
 
     def running? (inst_view)
-      inst_view.statuses[1].code == 'PowerState/running'
+      # I've seen inst_view.statuses[1] is Nil
+      if inst_view.statuses[1].nil?
+        return false
+      else
+        inst_view.statuses[1].code == 'PowerState/running'
+      end
```
akostadinov, line 182: If `nil` is the issue you can have:
```
      inst_view.statuses[1]&.code == 'PowerState/running'
```

lib/launchers/openstack.rb
```diff
@@ -952,8 +952,12 @@ def get_running_instances
       instances = get_objects("servers")
       instances.each do |instance|
         threads << Thread.new(instance) do | i|
-          inst_details = get_instance_detail(i['id'])
-          running_inst[inst_details.dig('server', 'name')] = inst_details['server'] if inst_details.dig('server','status') == "ACTIVE"
+          begin
+            inst_details = get_instance_detail(i['id'])
+            running_inst[inst_details.dig('server', 'name')] = inst_details['server'] if inst_details.dig('server','status') == "ACTIVE"
+          rescue
```
akostadinov, line 958:
```suggestion
          rescue => e
            logger.warn  exception_to_string(e)
```

tools/jenkins_mongo.rb
```diff
@@ -0,0 +1,85 @@
...
+  $LOAD_PATH.unshift(lib_path)
+end
+require 'mongo'
+require 'common'
+
+# Turn off debug-mode
+Mongo::Logger.logger.level = Logger::WARN
+
+module BushSlicer
+  class JenkinsMongo
```
akostadinov, line 14: maybe a better class name or explanation what is this class used for in a comment?

tools/openshift_qe_slack.rb
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env ruby
+require_relative 'common/load_path'
+
+require 'common'
+require 'slack-ruby-client'
+
+module BushSlicer
+  class CoreosSlack
```
akostadinov, line 8: Is this a class that will only work with coreos slack or it can work with other slack orgs? Name suggests it is hardcoded to coreos.

lib/launchers/azure.rb
akostadinov, line 182:
```suggestion
  inst_view.statuses[1]&.code == 'PowerState/running'
```

tools/openshift_qe_slack.rb
pruan-rht: @akostadinov the slack token is tied to the workspace which we are under `CoreosSlack`

pruan-rht:
@akostadinov all comments have been addressed...thx for the input!

tools/openshift_qe_slack.rb
akostadinov: but token is in configuration repo, no? So with another config class can be used for other orgs.

tools/jenkins_mongo.rb
akostadinov: Maybe then call it `InstallerLogMongo` or something?

tools/openshift_qe_slack.rb
pruan-rht: @akostadinov I think so, but I have not tried it with another org.

pruan-rht: @akostadinov now that I think about it more, this is tied to CoreosSlack org only....each app needs approval from the workspace's admin, so I expect the token is useless outside of CoreOS Slack workspace.  So yeah, this should be hardcoded name-wise.

tools/jenkins_mongo.rb
pruan-rht: @akostadinov it caches the jenkins Flexy-install profile...not Installer log...

tools/openshift_qe_slack.rb
akostadinov: Well, you can create a new token and approve it and put in config to work with another org. There is nothing in code to check this is coreos slack. In other words code is agnostic to organization, it only cares to use whatever token it is provided with.

tools/jenkins_mongo.rb
akostadinov: Cache with what purpose? Cache is to store temporary data for faster search/access. Log is when you store a record of something that happened. In this case it is a record of what clusters have been installed so you can later clean-them up. That's my line of thought.

tools/openshift_qe_slack.rb
pruan-rht: @akostadinov I generally don't mess with the token much...but in order for the app to work in another Org outside of the `CoreOS Slack`, the admin of the new org must approve the app first I believe...again, I have not try to use this outside of CoreOS workspace.  Are you suggesting to change the name of class to `Slack` instead?

tools/jenkins_mongo.rb
pruan-rht: @akostadinov to me the name `JenkinsMongo` is saying it's storing something from Jenkins....the db name is `cache_jenkins`.  What you suggested is more specific, where as as it stands it's more generic.

tools/openshift_qe_slack.rb
pruan-rht: @akostadinov tokens are associated to the app which is tied to a workspace...so I think the class name should be specific.

pruan-rht:
@akostadinov I've changed the class names to move this along.  PTAL

tools/openshift_qe_slack.rb
akostadinov: If token is in config and not in code, then the line of thought that this is an org specific class is at the very least unexpected.

tools/jenkins_mongo.rb
akostadinov: Exactly, it is too generic and it is not clear what and why is it storing. Naming of classes and variables when programming is one of the most important tasks and is not easy. Organizing class hierarchy even more so.

akostadinov:
PR code looks maintainable now. I have no objections to merging, subject to whatever other merging rules team may have now.

pruan-rht:
@liangxia can you pls merge if there are no more issues?

liangxia:
In the test log, it shows there are conflicts.
```
{"command":"git merge --no-ff ab90840d63666399c514b7565e92fd80d1e2379e","component":"clonerefs","error":"exit status 1","file":"prow/pod-utils/clone/clone.go:61","func":"k8s.io/test-infra/prow/pod-utils/clone.Run.func2","level":"info","msg":"Ran command","output":"Auto-merging tools/cloud_usage.rb\nCONFLICT (rename/delete): testdata/kata/release-4.8/kataconfiguration_v1_kataconfig.yaml deleted in ab90840d63666399c514b7565e92fd80d1e2379e and renamed in HEAD. Version HEAD of testdata/kata/release-4.8/kataconfiguration_v1_kataconfig.yaml left in tree.\nAuto-merging lib/launchers/openstack.rb\nAutomatic merge failed; fix conflicts and then commit the result.\n","severity":"info","time":"2021-09-17T06:46:33Z"}
```

pruan-rht:
@liangxia github is not showing there are conflicts.

liangxia:
I do see the 

> @liangxia github is not showing there are conflicts.

I've no idea about that.  You can check the detail of the logs.
Or perhaps just do fetch/rebase and see how it acts.

pruan-rht:
@liangxia it should be good now....so the issue was there was no conflict but upstream had added a file that mine branch didn't have.  It should be good to go now.

pruan-rht:
@dis016 can you do a `/lgtm` to move it along since the Beijing team won't be back until next Wednesday.

PR #1857, opened 2021-02-19
add feature for OCP-33402

yapei:
/assign @yanpzhan

### DIFF
--- a/features/step_definitions/web.rb
+++ b/features/step_definitions/web.rb
@@ -240,3 +240,33 @@ end
 Given /^I close the current browser$/ do
   browser.finalize
 end
+
+Given /^I check all relatedObjects of clusteroperator "(.*?)" are shown/ do |clusteroperator_name|
+  ensure_admin_tagged
+  #get relatedObjects of clusteroperator
+  relatedObjectsArray = cluster_operator(clusteroperator_name).related_objects
+  relatedObjectsArray.each do |related_object|
+    if !related_object.key?('name') or related_object['name'] == ""
+      next
+    end
+
+    if !related_object.key?('group') or related_object['group'] == ""
+      related_object_group = '-'
+    else
+      related_object_group = related_object['group']
+    end
+
+    if !related_object.key?('namespace') or related_object['namespace'] == ""
+      related_object_ns = '-'
+    else
+      related_object_ns = related_object['namespace']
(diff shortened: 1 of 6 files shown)


features/step_definitions/web.rb
```diff
@@ -240,3 +240,33 @@
...
 end
+
+Given /^I check all relatedObjects of clusteroperator "(.*?)" are shown/ do |clusteroperator_name|
+  ensure_admin_tagged
+  #get relatedObjects of clusteroperator
+  relatedObjectsArray = cluster_operator(clusteroperator_name).related_objects
+  relatedObjectsArray.each do |related_object|
+    if related_object['name'] == nil or related_object['name'] == ""
+      next
+    end
```
yapei, line 251: Currently when `status.relatedObjects[N].name` is empty, we will skip to next objects since the object whose name is empty will not be shown on the page

yapei:
@liangxia  @pruan-rht  Hi, could you help review the changes? Thank u

features/step_definitions/web.rb
```diff
@@ -240,3 +240,33 @@
...
   browser.finalize
 end
+
+Given /^I check all relatedObjects of clusteroperator "(.*?)" are shown/ do |clusteroperator_name|
+  ensure_admin_tagged
+  #get relatedObjects of clusteroperator
+  relatedObjectsArray = cluster_operator(clusteroperator_name).related_objects
+  relatedObjectsArray.each do |related_object|
+    if !related_object.key?('name') or related_object['name'] == ""
+      next
```
liangxia, line 250: IMO, we always need a name for whatever the resources it is.
A resource without name or null name sounds like a bug.

yapei: Yeah that's also my assumption however not every resource the cluster operators set has name field, for example 
https://github.com/openshift/cluster-ingress-operator/blob/48dd8e15152c8c9638a09de8c489131c85854253/pkg/operator/controller/status/controller.go#L116~L131
https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/starter.go#L195~L208

liangxia: Ack. Make sense now. They want all resources of the specific kind, e.g, scc

PR #1858, opened 2021-02-19
Move this case to be a '@destructive' case

weliang1:
PR is to fix https://issues.redhat.com/browse/OCPQE-3372

Any parallel running  multus cases can cause this case fail in checking the number of network_attachment_definition_instances{networks="any"} 

Simple solution is move this case to be a '@destructive' case

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -1108,6 +1108,7 @@ Feature: Multus-CNI related scenarios
   # @author weliang@redhat.com
   # @case_id OCP-25917
   @admin
+  @destructive
   Scenario: Multus Telemetry Adds capability to track usage of network attachment definitions
     # Make sure that the multus is enabled
     Given the multus is enabled on the cluster


kedark3:
@weliang1 by using `@destructive` in your initial PR description, you mistakenly tagged another GitHub user with the handle `@destructive`. It will be good to use backticks around that so we don't tag users unintentionally.

weliang1:
> @weliang1 by using `@destructive` in your initial PR description, you mistakenly tagged another GitHub user with the handle `@destructive`. It will be good to use backticks around that so we don't tag users unintentionally.



> @weliang1 by using `@destructive` in your initial PR description, you mistakenly tagged another GitHub user with the handle `@destructive`. It will be good to use backticks around that so we don't tag users unintentionally.

@k

weliang1:
> @weliang1 by using `@destructive` in your initial PR description, you mistakenly tagged another GitHub user with the handle `@destructive`. It will be good to use backticks around that so we don't tag users unintentionally.

@kedark3 Good catch, fixed it.

PR #1862, opened 2021-02-22
debug-multus

weliang1:
This PR is to collect more cluster information to find where cause the failure in https://issues.redhat.com/browse/OCPQE-3371 and https://issues.redhat.com/browse/OCPQE-3311

In our nightly e2e testing, when running the multus cases, the first step is to check multus by calling "Given the multus is enabled on the cluster", but this checking randomly failed on different mutlus test cases in different testing cluster, because the the number of desiredNumberScheduled multus pods is not equal the number of numberAvailable pod when calling "Given the multus is enabled on the cluster".

One guess is If something happened on the node and multus pod will not be be scheduled on that node. 

Add those three debug steps to collect node and multus pod information.
oc get nodes
oc get pod -n openshift-multus -o wide
oc describe daemonset -n openshift-multus

Above debug steps will be removed in the future when find the root cause and file the fix PR

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1275,3 +1275,20 @@ Given /^I save ipecho url to the#{OPT_SYM} clipboard$/ do | cb_name |
   cb[cb_name]="172.31.249.80:9095"
   logger.info "The ipecho service url #{cb[cb_name]} is stored to the #{cb_name} clipboard."
 end
+
+Given /^I Collect multus related information/ do
+  ensure_admin_tagged
+  step %Q/I run the :get admin command with:/, table(%{
+    | resource | nodes |
+    | o        | wide  |
+  })
+  step %Q/I run the :describe admin command with:/, table(%{
+    | resource | daemonset        |
+    | n        | openshift-multus |
+  })
+  step %Q/I run the :get admin command with:/, table(%{
+    | resource | pods             |
+    | n        | openshift-multus |
+    | o        | wide             |   
+  })
+end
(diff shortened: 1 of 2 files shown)


pruan-rht:
@weliang1 there's conflict

features/step_definitions/networking.rb
```diff
@@ -1275,3 +1275,20 @@
   cb[cb_name]="172.31.249.80:9095"
   logger.info "The ipecho service url #{cb[cb_name]} is stored to the #{cb_name} clipboard."
 end
+
+Given /^I Collect multus related information/ do
```
pruan-rht, line 1279: @weliang1 probably don't need to capitalize `C`

anuragthehatter:
That looks okay or may be instead of ths, we can permanently add these debugging steps under https://github.com/openshift/verification-tests/blob/master/features/step_definitions/networking.rb#L587. Would be just 3 lines. There is no harm in getting 2-25 lines extra in auto logs everytime which would be helpful always in debugging.

kedark3:
I agree with @anuragthehatter about adding it permanently if possible. Couldn't do any harm I guess.

features/step_definitions/networking.rb
```diff
@@ -1275,3 +1275,20 @@
...
+  step %Q/I run the :describe admin command with:/, table(%{
+    | resource | daemonset        |
+    | n        | openshift-multus |
+  })
+  step %Q/I run the :get admin command with:/, table(%{
+    | resource | pods             |
+    | n        | openshift-multus |
+    | o        | wide             |   
+  })
+end
```
kedark3, line 1294: Please add a newline at the end of this file so we don't see this symbol telling us no newline at the end of file: 
image

PR #1863, opened 2021-02-22
No need to capitalize the letter

weliang1:
Technical error then close https://github.com/openshift/verification-tests/pull/1862. 
All issues comments in 1862 got resolved in this PR.

This PR is to collect more cluster information to find where cause the failure in https://issues.redhat.com/browse/OCPQE-3371 and https://issues.redhat.com/browse/OCPQE-3311

In our nightly e2e testing, when running the multus cases, the first step is to check multus by calling "Given the multus is enabled on the cluster", but this checking randomly failed on different mutlus test cases in different testing cluster, because the the number of desiredNumberScheduled multus pods is not equal the number of numberAvailable pod when calling "Given the multus is enabled on the cluster".

One guess is If something happened on the node and multus pod will not be be scheduled on that node.

Add those three debug steps to collect node and multus pod information.
oc get nodes
oc get pod -n openshift-multus -o wide
oc describe daemonset -n openshift-multus

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1284,3 +1284,22 @@ Given /^the IPsec is enabled on the cluster$/ do
   default_network = network_operator.default_network(user: admin)
   raise "env doesn't have IPSec enabled" unless default_network["ipsecConfig"]
 end
+
+# For debug purpose to collect multus related information
+Given /^I collect multus related information/ do
+  ensure_admin_tagged
+  step %Q/I run the :get admin command with:/, table(%{
+    | resource | nodes |
+    | o        | wide  |
+  })
+  step %Q/I run the :describe admin command with:/, table(%{
+    | resource | daemonset        |
+    | n        | openshift-multus |
+  })
+  step %Q/I run the :get admin command with:/, table(%{
+    | resource | pods             |
+    | n        | openshift-multus |
+    | o        | wide             |   
+  })
+end
+
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -6,6 +6,7 @@ Feature: Multus-CNI related scenarios
   Scenario: Create pods with multus-cni - macvlan bridge mode
     # Make sure that the multus is enabled
     Given the master version >= "4.1"
+    Given I collect multus related information
     And the multus is enabled on the cluster
     Given the default interface on nodes is stored in the :default_interface clipboard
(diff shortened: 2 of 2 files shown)


zhaozhanqi:
I think collecting the information if the multus is not running well other than even they are running well since if so they will print a lot of logs in our CI . 

eg . we can update these steps into `Given /^the multus is enabled on the cluster$/ do` when `unless desired_multus_replicas == available_multus_replicas && available_multus_replicas != 0`

rbbratta:
@weliang1   PTAL https://github.com/openshift/verification-tests/pull/1865   It could be enabled as a permanent debug option.

features/step_definitions/networking.rb
```diff
@@ -1284,3 +1284,22 @@
...
+    | resource | daemonset        |
+    | n        | openshift-multus |
+  })
+  step %Q/I run the :get admin command with:/, table(%{
+    | resource | pods             |
+    | n        | openshift-multus |
+    | o        | wide             |   
+  })
+end
+
```
rbbratta, line 1305: From https://github.com/openshift/verification-tests/pull/1865

```diff
diff --git a/features/step_definitions/networking.rb b/features/step_definitions/networking.rb
index 3bdae8c2..4468cf65 100644
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -591,7 +591,15 @@ Given /^the multus is enabled on the cluster$/ do
   available_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:available]
   #storing desired_multus_replicas value in desired_multus_replicas clipboard variable as well
   cb.desired_multus_replicas = desired_multus_replicas
-  raise "Multus is not running correctly!" unless desired_multus_replicas == available_multus_replicas && available_multus_replicas != 0
+  unless desired_multus_replicas == available_multus_replicas && available_multus_replicas != 0
+    daemon_set('multus', project('openshift-multus')).describe(admin, quiet:false)
+    BushSlicer::Pod.get_labeled("app=multus", user: admin, project: project("openshift-multus", switch: false)) do |pod|
+      pod.describe(admin, quiet: false)
+    end
+    env.nodes(user:admin, refresh: true, quiet: false)
+    raise "Multus is not running correctly!"
+  end
+
 end
```

weliang1:
See: https://github.com/openshift/verification-tests/pull/1865

PR #1868, opened 2021-02-23
Add func to run commands on host after scenario

kasturinarra:
Add func to run commands on host after scenario

### DIFF
--- a/features/step_definitions/node.rb
+++ b/features/step_definitions/node.rb
@@ -140,6 +140,20 @@ Given /^I run( background)? commands on the nodes in the#{OPT_SYM} clipboard:$/ 
   cb[tmpcb] = nil
 end
 
+Given /^I run( background)? commands on the nodes in the#{OPT_SYM} clipboard after scenario:$/ do |bg, cbname, table|
+  ensure_admin_tagged
+  ensure_destructive_tagged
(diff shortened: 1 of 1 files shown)


kasturinarra:
@geliu2016 could you please help review this ? This func has been used in the script here: https://github.com/openshift/cucushift/pull/8427, thanks !!

geliu2016:
LGTM，thanks

kasturinarra:

kasturinarra:
@liangxia could you please help review & merge? Thanks !!

features/step_definitions/node.rb
```diff
@@ -140,6 +140,19 @@
   cb[tmpcb] = nil
 end
 
+Given /^I run( background)? commands on the nodes in the#{OPT_SYM} clipboard after scenario:$/ do |bg, cbname, table|
+  ensure_admin_tagged
```
liangxia, line 144: Perhaps we should add `ensure_destructive_tagged` ?
We can allow to run non-destructive ones in destructive mode, but we should not allow to run destructive ones in non-destructive mode.

kasturinarra:
@liangxia have resolved all comments, could you please help review again ? Thanks !!

PR #1870, opened 2021-02-24
add env vars for logging to set catsrc and channel

QiaolingTang:
Start from logging 5.0, there have several different channels, to testing different channels, it's hard to determine the channel name in scenarios. To match logging testing requirements, adding two env vars `logging_catsrc` and `logging_channel_name`.

An example about how to set them:
```
export BUSHSLICER_CONFIG='
  global:
    browser: chrome
  environments:
    ocp4:
      admin_creds_spec: "https://example.com/auth/kubeconfig/*view*/"
      version: "4.8"
      logging_catsrc: "example-registry"
      logging_channel_name: "5.0"
      idp: "example-idp"
      admin_console_url: "https://console-openshift-console.apps.example.com"
'
```
/cc @anpingli

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -326,26 +326,34 @@ end
 
 Given /^logging channel name is stored in the#{OPT_SYM} clipboard$/ do | cb_name |
   cb_name = 'logging_channel_name' unless cb_name
-  version = cluster_version('version').version.split('-')[0].split('.').take(2).join('.')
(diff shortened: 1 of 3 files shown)


anpingli (commented):
@QiaolingTang  /lgtm. @liangxia  Could you review it?
BTW, for long perspective, we will have different catalog source name and       channel name to different operators. Let's start think about this and consider to implement this in future.

PR #1874, opened 2021-02-24
Introduced FlexyEnvVariable and FlexyEnvFile helper classes

pjanouse:
We can use it in config.yaml for Flexy.

`FlexyEnvFile` class like:

```
services:
  AWS-CI:
    host_opts:
#      ssh_private_key: keys/pjanouse.pem
      ssh_private_key: !ruby/object:BushSlicer::FlexyEnvVariable
        var_name: ENVVAR1
```
where environment variable `ENVVAR1` contains the real (base64 encoded) key.

`FlexyEnvVariable` class like:
```
services:
 openstack_upshift:
#    url: https://<host-name>:1300/v3/auth/tokens
    url: !ruby/object:BushSlicer::FlexyEnvVariable
      var_name: ENVVAR1
```
where environment variable `ENVVAR1` contains the real URL.

### DIFF
--- a/lib/configuration.rb
+++ b/lib/configuration.rb
@@ -109,4 +109,51 @@ module BushSlicer
       return Object.const_get(class_opts[:class]).new(**class_opts[:opts])
     end
   end
+
+  class FlexyEnvVariable
+    attr_reader :var_name
+
+    def initialize(var_name)
+      @var_name=var_name
+    end
+    def ==(other)
+      self.class===other &&
+        @var_name==other.var_name
+    end
+    def to_s
+      ENV[@var_name]
+    end
+    
+    alias :eql? :==
+    alias :to_str :to_s
+    alias :[]= :to_s
+  end
+
+  class FlexyEnvFile
+    @@files = Hash.new
+    attr_reader :var_name
+
+    def initialize(var_name)
+      @var_name=var_name
+    end
+    def ==(other)
+      self.class===other && 
+        @var_name==other.var_name
+    end
+    def to_s
+      path=@@files[@var_name]
+      if path.nil?
+        e = ENV[@var_name]
+        Tempfile.open("env_#{@var_name}_") do |f|
+          f.write("#{e}") if !e.nil?
+          @@files[@var_name]=f.path
+        end
+      end
+      path
+    end
+
+    alias :eql? :==
+    alias :to_str :to_s
+    alias :[]= :to_s
+  end
 end
--- a/lib/configuration.rb
+++ b/lib/configuration.rb
@@ -153,6 +153,7 @@ module BushSlicer
         Tempfile.open("env_#{@var_name}_") do |f|
           f.write("#{e}") if !e.nil?
           @@files[@var_name]=f
+          file=f
         end
       end
       file.path
--- a/lib/configuration.rb
+++ b/lib/configuration.rb
@@ -142,10 +142,13 @@ module BushSlicer
     end
     def to_s
       file=@@files[@var_name]
+
+      # file has been deleted on filesystem
       if !file.nil? && !file.file?
         @@files.except!(@var_name)
         file=nil
       end
+
       if file.nil?
         e = ENV[@var_name]
         Tempfile.open("env_#{@var_name}_") do |f|
--- a/lib/configuration.rb
+++ b/lib/configuration.rb
@@ -143,7 +143,7 @@ module BushSlicer
       file=@@files[@var_name]
 
       # file has been deleted on filesystem
-      if !file.nil? && !file.file?
+      if !file.nil? && !File.file?(file.path)
         @@files.except!(@var_name)
         file=nil
       end
--- a/lib/configuration.rb
+++ b/lib/configuration.rb
@@ -141,15 +141,19 @@ module BushSlicer
         @var_name==other.var_name
     end
     def to_s
-      path=@@files[@var_name]
-      if path.nil?
+      file=@@files[@var_name]
+      if !file.nil? && !file.file?
+        @@files.except!(@var_name)
+        file=nil
+      end
+      if file.nil?
         e = ENV[@var_name]
         Tempfile.open("env_#{@var_name}_") do |f|
           f.write("#{e}") if !e.nil?
-          @@files[@var_name]=f.path
+          @@files[@var_name]=f
         end
       end
-      path
+      file.path
     end
 
     alias :eql? :==
--- a/lib/configuration.rb
+++ b/lib/configuration.rb
@@ -126,7 +126,6 @@ module BushSlicer
     
     alias :eql? :==
     alias :to_str :to_s
-    alias :[]= :to_s
   end
 
   class FlexyEnvFile
@@ -161,6 +160,5 @@ module BushSlicer
 
     alias :eql? :==
     alias :to_str :to_s
-    alias :[]= :to_s
   end
 end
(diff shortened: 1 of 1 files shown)


pjanouse:
@akostadinov please review and a feedback is more than welcomed

pjanouse:
@akostadinov I'm curious why there are repeated (in templates) block of code like (in `.../hosts/launch_opts/common_ipi_env_v1.rb`):
```
  # keep reference of temp file until installer exit to avoid it being deleted
  proc {
    _awscreds = awscreds
    at_exit { puts "deleting #{_awscreds.path}"; _awscreds.unlink }
  }.call
```

As far as I understand documentation for [Tempfile](https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html) class:

> When a Tempfile object is garbage collected, or when the Ruby interpreter exits, its associated temporary file is automatically deleted. This means that’s it’s unnecessary to explicitly delete a Tempfile after use, though it’s good practice to do so: not explicitly deleting unused Tempfiles can potentially leave behind large amounts of tempfiles on the filesystem until they’re garbage collected.

Above code is completely unnecessary from my PoV. First we don't need to explicitly call `unlink()` at all (fine, good practice however we like to intentionally force abusing the GC here until Ruby instance/process exists). Second - if we want just to avoid removing it during performing GC (inside execution, not by exit), we just need a "staled" reference to the `File` object. Everything else is handled by "Ruby core" side, right? Is there something I'm overlooking in my impl.?

akostadinov:
@pjanouse . if those are temporary files, then there should be no need to register unlink at_exit hook. On the other hand this might be a way to avoid garbage collecting the reference until exit. Did I implement these lines? I can't remember. If it was me, then I just wanted to keep a reference to the temp file until exit without keeping that variable somewhere. If you have better ideas, I'm open to them.

pjanouse:
@akostadinov Actually based on Git blame you wrote the code around `at-exit`... The whole templates tree was moved last year, but that action could stop me..:-)

I've tried to do some research regarding `Tempfile` and GC and the result is - we really need to have/save the reference to the `File` object otherwise performed GC (anytime) will *delete* the tempfile. So based on [this great article](https://www.hilman.io/blog/2016/01/tempfile/) - all we need is to close the file (to save filehandlers as a expensive system resource)  - or put it into 'the resource block' - and keep the reference to that tmpfile. We don't need to explicitly call `unlink()` nor register `at_exit` action.

pjanouse:
@akostadinov I'm happy with the current implementation and think it is done. Can you review please?

lib/configuration.rb
```diff
@@ -109,4 +109,55 @@ def get_optional_class_instance(keyword)
...
+      self.class===other &&
+        @var_name==other.var_name
+    end
+    def to_s
+      ENV[@var_name]
+    end
+    
+    alias :eql? :==
+    alias :to_str :to_s
+    alias :[]= :to_s
```
akostadinov, line 129: do we need this? 

If we need it, shall it be something like (untested):
```
def [](*args, &block)
  self.to_s[*args, &block]
end
```

akostadinov: I mean, I can't understand the use case. Maybe what you have is correct, it is just not clear to me what use case it covers.

lib/configuration.rb
```diff
@@ -109,4 +109,55 @@ def get_optional_class_instance(keyword)
...
+    def ==(other)
+      self.class===other && 
+        @var_name==other.var_name
+    end
+    def to_s
+      file=@@files[@var_name]
+      if !file.nil? && !file.file?
+        @@files.except!(@var_name)
+        file=nil
+      end
```
akostadinov, line 148: Could you give an example where `if` condition above is needed. How can file be not `nil` and not a file? btw if `file` is not a file, then it probably doesn't have the method `file?` in which case an exception will be raised. Again, you might be right, I might be missing the point.

lib/configuration.rb
pjanouse: As far as I remember during my work I've got an error like 'can't find a method "[]" for class...', but sure, I'll verify again if the issue is then still here.

lib/configuration.rb
```diff
@@ -109,4 +109,55 @@ def get_optional_class_instance(keyword)
...
+          f.write("#{e}") if !e.nil?
+          @@files[@var_name]=f
+        end
+      end
+      file.path
+    end
+
+    alias :eql? :==
+    alias :to_str :to_s
+    alias :[]= :to_s
```
akostadinov, line 161: same question as above

lib/configuration.rb
pjanouse: My intention was to check where I retrieve here a stale `File` handle or not. I know the POSIX behavior allows to have a file opened and call unlink - filesystem doesn't report that file existence onward but handle owner process can still call any operation over that file... 

On the other hand we have an opposite situation - we have a `Tempfile`, it was closed (after exiting the resource block) however GC can't delete it because there isn't a reference counter lowered to zero (intentionally, we put it into the map just for this intentional case). On the other hand, we have a handle for closed file stored then - this file can be deleted externally (directly on FS - it was closed, so I don't expect any claims here) and in that case I believe we cat get into the situation when we have stale  file handler (file.nil -> false) but the file doesn't exists (file.file? -> false). In this case I drop the reference and going forward like the first get call.

pjanouse: If I'm not wrong (see https://ruby-doc.org/core-2.5.0/File.html#method-c-file-3F) - "Returns true if the named file exists and is a regular file." Which perfectly fits to our case or am I missing sth here?

akostadinov: Got it. I wouldn't bother to test for such an edge case but it is a possible situation. If you like to keep this check it may help adding a comment line like `# file has been deleted on filesystem`.

lib/configuration.rb
akostadinov: yes, please send me a backtrace in case you can reproduce

lib/configuration.rb
pjanouse: :+1:

pjanouse: Verified, these aliases aren't necessary now thus removing them. It had to be an issue on my side when coding then...

lib/configuration.rb
pjanouse: Verified, these aliases aren't necessary now thus removing them. It had to be an issue on my side when coding then...

pjanouse:
I've realized I'd tested some stale branch this week, need to test more carefully recent changes.

lib/configuration.rb
```diff
@@ -109,4 +109,57 @@ def get_optional_class_instance(keyword)
...
+    end
+    def ==(other)
+      self.class===other && 
+        @var_name==other.var_name
+    end
+    def to_s
+      file=@@files[@var_name]
+
+      # file has been deleted on filesystem
+      if !file.nil? && !File.file?(file.path)
```
akostadinov, line 146: just a nitpick, I think it is better readable to use `file` instead of `!file.nil?`

Things look good to me so I hope we can merge soon.

pjanouse:
Well, working again as expected.

lib/configuration.rb
pjanouse: :+1: 
Just a new information about Ruby received...:)

akostadinov: In ruby only `nil` and `false` count as a falsy value and everything else as truish... including empty strings/arrays, `0`, etc.
Just for completeness.

PR #1859, opened 2021-02-22
auto some ovn egress ip cases

huiran0826:
Auto some OVN EgressIP cases: OCP-33718, OCP-33710, OCP-33617, OCP-33719

Logs:
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2893/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2894/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2895/console
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2896/console

@zhaozhanqi @anuragthehatter @weliang1  @rbbratta @asood @brahaney @kedark3

PR #1873, opened 2021-02-24
Update og to org

huiran0826:
Updated "og" to "org"

Rerun cases passed and Log is here: 
https://mastern-jenkins-csb-openshift-qe.cloud.paas.psi.redhat.com/job/Runner-v3-smoke/2901/console

@kedark3 @pruan-rht

PR #1875, opened 2021-02-24
fix pod label from og=qe to org=qe

pruan-rht:
as promised from yesterday's https://github.com/openshift/verification-tests/pull/1859


pruan-rht:
closing since https://github.com/openshift/verification-tests/pull/1873 is there already...should've checked

PR #1878, opened 2021-02-25
Fix OCP-22596

wzheng1:
@openshift/devexp-qe Please review, thanks!


wzheng1:
@openshift/devexp-qe Please review.

PR #1881, opened 2021-03-02
Update iscsi prepare step

qinpingli:
In our daily test, we found the worker node might be in "NotReady" status for the OOM, if the iscsi target pod is scheduled to the woker node and the node status is "NotReady", then the test cases for the iscsi will be blocked.

@duanwei33 @chao007 Please help review.

### DIFF
--- a/features/step_definitions/helper_services.rb
+++ b/features/step_definitions/helper_services.rb
@@ -492,32 +492,17 @@ Given /^I have a iSCSI setup in the environment$/ do
 
   if _pod.ready?(user: admin, quiet: true)[:success]
     logger.info "found existing iSCSI pod, skipping config"
-    cb.iscsi_ip = _service.ip(user: admin)
   elsif _pod.exists?(user: admin, quiet: true)
     logger.warn "broken iSCSI pod, will try to recreate keeping other config"
     @result = admin.cli_exec(:delete, n: _project.name, object_type: "pod", object_name_or_id: _pod.name)
     raise "could not delete broken iSCSI pod" unless @result[:success]
   else
-    env.node_hosts.each do |host|
-      setup_commands = [
-        "sed -i '/^node.session.auth./'d  /etc/iscsi/iscsid.conf",
-        "cat >> /etc/iscsi/iscsid.conf << EOF\n" +
-          "node.session.auth.authmethod = CHAP\n" +
-          "node.session.auth.username = 5f84cec2\n" +
-          "node.session.auth.password = b0d324e9\n" +
-          "EOF\n",
-        "systemctl enable iscsid",
-        "systemctl restart iscsid"
-      ]
-      res = host.exec_admin(*setup_commands)
-      raise "iSCSI initiator setup commands error" unless res[:success]
-    end
     @result = admin.cli_exec(:create, n: _project.name, f: "#{BushSlicer::HOME}/testdata/storage/iscsi/iscsi-target.json")
     raise "could not create iSCSI pod" unless @result[:success]
   end
 
   if !_service.exists?(user:admin, quiet: true)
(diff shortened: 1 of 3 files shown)


qinpingli:
Ran all the cases in feature/tierN/storage/iscsi.feature, passed.

qinpingli:
/hold

There is still some other issue in the iscsi automation.

features/step_definitions/helper_services.rb
```diff
@@ -492,32 +492,17 @@
...
-      res = host.exec_admin(*setup_commands)
-      raise "iSCSI initiator setup commands error" unless res[:success]
-    end
     @result = admin.cli_exec(:create, n: _project.name, f: "#{BushSlicer::HOME}/testdata/storage/iscsi/iscsi-target.json")
     raise "could not create iSCSI pod" unless @result[:success]
   end
 
   if !_service.exists?(user:admin, quiet: true)
-    @result = admin.cli_exec(:create, n: _project.name, f: 'https://raw.githubusercontent.com/openshift-qe/docker-iscsi/master/service.json')
+    @result = admin.cli_exec(:create, n: _project.name, f: "#{BushSlicer::HOME}/testdata/storage/iscsi/service.json")
```
pruan-rht, line 505: @qinpingli perhaps call the `I obtain test data...` step first and then just use `service.json` to elminate the need to use the older style of `#{BushSlicer::HOME}...`?

testdata/storage/iscsi/service.json
```diff
@@ -0,0 +1,20 @@
+{
+    "kind": "Service",
+    "apiVersion": "v1",
+    "metadata": {
+        "name": "iscsi-target"
+    },
+    "spec": {
+	"selector": {
```
liangxia, line 8: Would you mind to fix the indentions ? It's hard to read.

features/step_definitions/helper_services.rb
qinpingli: Thanks! Updated.

testdata/storage/iscsi/service.json
qinpingli: Good catch! Updated.

qinpingli:
For iscsi test, there are still some issues, such as:
1. If the node which iscsi target pod is scheduled has the old config in /etc/iscsi, then the iscsi target pod may get the wrong config .
2.  From the old step, seems it wants to enable the chap auth test, but the configuration on the target side is not correct, so the chap auth is not enabled successfully.
3. Base one of deleted testcase(OCP-17467, actually I have no idea why this case is deleted), seems we need to test more auth than the definition in the old step. Detailed auth info: https://github.com/openshift/cucushift/blob/master/testdata/storage/iscsi/chap-secret-auto.yml. I think we need more time to research again about the iscsi in openshift.
4. For OCP4.x, the nodes are based on RHEL8, our iscsi target pod based on Fedora 22, there might be some issue too.

The fixing plan for the above issue:
1. For issues 1, 2, 4, will fix in this gitrepo: https://github.com/openshift-qe/docker-iscsi/tree/master/docker and create a new iscsi target test image.
2. For issue 3, I'd like to add auth test in each iscsi testcases with chap secret not configuring it in the /etc/iscsi/iscsid.conf. The existence of this file is tested by OCP-23400.

Based on the fixing plan, this step will no need to be updated.
@duanwei33 @chao007 @jhou1 Could you help review the issues and fixing plan, if it's ok for you, I'll cancel the hold and merge this pr first.

jhou1:
Looks good to me, thanks for pushing the fixes!

duanwei33:
@qinpingli  I agree with the fix plan.

qinpingli:
Thanks, @jhou1 @duanwei33 !

/hold cancel

PR #1883, opened 2021-03-03
add rules for OCP-36912

yapei:


yanpzhan:
No issue from my side,btw, we could add 4.8 rules now.

PR #1890, opened 2021-03-05
Replace deprecated call URI::decode with URI::decode_www_from

pruan-rht:
to get rid of warnings like 

`12:15:19  tools/launch_instance.rb:202: warning: URI.unescape is obsolete
`


akostadinov:
@pruan-rht , please next time test such changes. I caught it by pure luck.

pruan-rht:
@akostadinov thanks for covering my mistake...

PR #1893, opened 2021-03-06
Prepare for Cucumber 5.x upgrade

akostadinov:
In Cucumber version 3 and later there is no support for transformation of
all parameters including tables.

This is a workaround where a separate method call performs the transformation.

We need to upgrade Cucumber to support ReportPortal as well avoid being stuck
at an unsupported library version.


akostadinov:
Tested locally and with
Runner-v3-smoke/2958

I'm gonna merge it now to avoid added and modified steps to miss the new approach.

akostadinov:
/hold

maybe I find a hack to avoid this

akostadinov:
I decided to implement a hack that will avoid the need to add a line to each scenario. See #1814

PR #1897, opened 2021-03-08
another take at URI::decode modernization

akostadinov:


akostadinov:
/assign @pruan-rht 

could you check if this works for you and merge? I also tested.

cc @XinRedhat @jianlinliu , I think this should get us working. Issue is that escaping rules can be various and it is very hard to replace old decode in exactly the same way. But I think it should be fine now.

pruan-rht:
@akostadinov do you know if this document is related to your URI call?  https://ruby-doc.org/stdlib-2.7.1/libdoc/uri/rdoc/URI/Escape.html

akostadinov:
@pruan-rht , the point is to decode as if value is in URL PATH, not query. Also to be backward compatible especially with `+` sign that people decided to use and what was the issue with first attempt.

akostadinov:
Tested with 3.11 where issue actually was Flexy-install/9517

@jiajliu @pruan-rht , if you see any strange issues, feel free to revert. I don't expect though.

PR #1901, opened 2021-03-10
fix OCP-25762

yapei:
pass log Runner-v3#310482


yapei:
@yanpzhan  Could you help review? Thanks

pruan-rht (commented):
/lgtm

PR #1902, opened 2021-03-11
files are base64 encoded and better string emulation

akostadinov:
@pjanouse, could you review and tell me what do you think about such changes?

2 issues fixed is that file content is base 64 encoded so we need to decode it.
Also added more methods from string so that usage of these values is more
transparent.

/assign pjanouse

### DIFF
--- a/lib/configuration.rb
+++ b/lib/configuration.rb
@@ -1,3 +1,4 @@
+require "base64"
 require 'yaml'
 
 require 'bushslicer'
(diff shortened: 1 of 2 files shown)


lib/configuration.rb
```diff
@@ -110,56 +111,62 @@ def get_optional_class_instance(keyword)
...
         Tempfile.open("env_#{@var_name}_") do |f|
-          f.write("#{e}") if !e.nil?
-          @@files[@var_name]=f
-          file=f
+          f.write(Base64.decode64 e)
+          @file=f
         end
       end
-      file.path
+      @file.path
```
pjanouse, line 169: Can we do this in execution time when the whole configuration is frozen?

lib/configuration.rb
```diff
@@ -110,56 +111,62 @@ def get_optional_class_instance(keyword)
...
+          raise "Environemnt variable #{@var_name} was not set."
+        end
 
-      if file.nil?
-        e = ENV[@var_name]
         Tempfile.open("env_#{@var_name}_") do |f|
-          f.write("#{e}") if !e.nil?
-          @@files[@var_name]=f
-          file=f
+          f.write(Base64.decode64 e)
```
pjanouse, line 165: What is `e` in this context? Should be `content`, shouldn't be?

lib/configuration.rb
```diff
@@ -110,56 +111,62 @@ def get_optional_class_instance(keyword)
...
+          @file=f
```
pjanouse, line 166: Can we do this in execution time when the whole configuration is frozen?

akostadinov: No, you are right. I thought about it. I have overridden the `freeze` method to make sure object is properly initialized before freezing happened. Still we create the file dynamically to avoid needless temporary files creation. But a `@cache` Hash is created that is still mutable after object is frozen.

Still feels like a hack but I thought it is a little better.

lib/configuration.rb
akostadinov: Good catch, fixed.

lib/configuration.rb
```diff
@@ -110,56 +111,82 @@ def get_optional_class_instance(keyword)
...
       end
-      file.path
+      cache[:file].path
     end
 
-    alias :eql? :==
-    alias :to_str :to_s
+    # we override this because YAML unmarshals the file without
+    # calling a constructor. Then configuration freezes all values.
+    # Thus we can't initialize anything before object is frozen.
```
pjanouse, line 182: before -> after ?

lib/configuration.rb
```diff
@@ -110,56 +111,82 @@ def get_optional_class_instance(keyword)
...
         Tempfile.open("env_#{@var_name}_") do |f|
-          f.write("#{e}") if !e.nil?
-          @@files[@var_name]=f
-          file=f
+          f.write(Base64.decode64 content)
+          cache[:file] = f
         end
       end
-      file.path
+      cache[:file].path
```
pjanouse, line 177: Are you sure this can work with a `hash` trick after self.freeze?

I tried like:
```
aa = {}
aa[:aa] = "aaa"
aa.freeze
aa[:aa] = "aa"
puts aa.inspect
```
but got "can't modify frozen Hash (FrozenError)" on line `aa[:aa] = "aa"`

pjanouse: Doesn't work even I'd removed `aa[:aa] = "aaa"`.

akostadinov: @pjanouse , the hash itself is not frozen after object is frozen. The `freeze` call is not going in depth of all object instance variables. I have tested it to work also.

With `fiddle` command you can enter a pry shell and experiment with `conf[...]` calls to see. `pry` is an essential tool to develop and debug ruby code for me.

lib/configuration.rb
akostadinov: @pjanouse, What I'm trying to say is that we can't initialize anything **before** object is frozen. Thus we need to do it at `freeze` time. Makes sense?

akostadinov: I added a line to clarify the meaning. Let me know if it is better now.

lib/configuration.rb
```diff
@@ -110,56 +111,83 @@ def get_optional_class_instance(keyword)
...
     end
 
-    alias :eql? :==
-    alias :to_str :to_s
+    # we override this because YAML unmarshals the file without
+    # calling a constructor. Then configuration freezes all values.
+    # Thus we can't initialize anything before object is frozen.
+    # That't why we initialize the cache Hash within here.
```
pjanouse, line 183: Typo -> That's

lib/configuration.rb
pjanouse: Got it. Makes sense and it's great trick how we can overcome this limit.

Regarding `fiddle`, I need probably a deeper instruction, we can continue by other channel if you want.

akostadinov:
Thank you very much! Will merge it then and wait for actual users.

PR #1903, opened 2021-03-11
update idling cases

lihongan:
After idling the service, the pod is still in `Terminating` state for a while (might be stuck on abnormal cluster), but the replicas of ReplicationController has been changed to "0" and cause the following wake up step unstable,

So replacing the checkpoint with `wait for the resource "pod" named "<%= cb.pod_name %>" to disappear within 120 seconds` to ensure the old pod is deleted before going to the next step. And we also can get correct information from the log if get stuck in Terminating.

Removed the useless step `wait until number of replicas match "1" for replicationController` since the following step `a pod becomes ready` can do the same thing.

@quarterpin @jechen0648  please help take a look. Thanks.
   
log:
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3/311840/console

### DIFF
--- a/features/routing/OWNERS
+++ b/features/routing/OWNERS
@@ -2,3 +2,5 @@ approvers:
 
 reviewers:
 - lihongan
+- quarterpin
+- jechen0648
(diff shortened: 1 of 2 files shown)


lihongan:
and updated the OWNERS and added @quarterpin @jechen0648 to reviewers

jechen0648 (commented):
looks good

quarterpin:
/lgtm 

Thanks,  @lihongan.

PR #1910, opened 2021-03-17
fix rules for OCP-19817,OCP-25761

yapei:


yapei:
@liangxia  @pruan-rht  Could you help merge? Thank you

PR #1912, opened 2021-03-17
update apiVersion for the routes

lihongan:
see below warning message if using old apiVersion, so update related routes in testdata to use `route.openshift.io/v1`.

```
W0317 11:12:03.889137    6752 shim_kubectl.go:55] Using non-groupfied API resources is deprecated and will be removed in a future release, update apiVersion to "route.openshift.io/v1" for your resource
route.route.openshift.io/secured-edge-route created
```

@quarterpin @jechen0648 please help take a look. thanks.

### DIFF
--- a/testdata/routing/edge/route_edge.json
+++ b/testdata/routing/edge/route_edge.json
@@ -1,5 +1,5 @@
 {
-	"apiVersion": "v1",
+	"apiVersion": "route.openshift.io/v1",
 		"kind": "Route",
 		"metadata": {
 			"name": "secured-edge-route"
--- a/testdata/routing/reencrypt/reencrypt-without-all-cert.yaml
+++ b/testdata/routing/reencrypt/reencrypt-without-all-cert.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 items:
 - kind: Route
-  apiVersion: v1
+  apiVersion: route.openshift.io/v1
   metadata:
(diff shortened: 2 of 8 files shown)


quarterpin:
/lgtm 

Thanks! @lihongan .

testdata/routing/edge/route_edge.json
```diff
@@ -1,5 +1,5 @@
 {
-	"apiVersion": "v1",
+	"apiVersion": "route.openshift.io/v1",
```
liangxia, line 2: Will it works with 4.2/4.5/4.8 ?

lihongan: yes, it works with 4.2 and 4.8.
```
$ oc get clusterversion
NAME      VERSION                             AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.2.0-0.nightly-2021-02-22-141219   True        False         10m     Cluster version is 4.2.0-0.nightly-2021-02-22-141219

$ oc create -f auto/verification-tests/testdata/routing/edge/route_edge.json 
route.route.openshift.io/secured-edge-route created

$ oc get route secured-edge-route -oyaml
apiVersion: route.openshift.io/v1
kind: Route
```

liangxia: @lihongan Thanks for the confirm.

PR #1919, opened 2021-03-19
[ONEOCPDEPL-10] Added configuration block of environment variables

pjanouse:
@akostadinov Please review the optimized implementation

WIP is due to missing docs (as we agreed it will be subject of ONEOCPDEPL-10 too), I'm going to create it and add as a separate commit.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -49,15 +49,23 @@ global:
   # Leave unset for default of launching local browser via geckodriver etc.
   # selenium_url: http://example.com:4444/wd/hub
 
+  # Each key:value pair from this section is exported as shell environment variable
+  # into the running context:
+  # install-envvars:
+  #   foo: "bar"
+
+  # Examples of usage YAML Helper classes
   # Any value you can set to refer to an environment variable
+  #
   # read value from ENVVAR1 environment variable:
-  # test1: !ruby/object:BushSlicer::ConfigEnvVariable
+  # input_entry: !ruby/object:BushSlicer::ConfigEnvVariable
   #   var_name: ENVVAR1
-  # `test1` key will be a string with value
-  # read value from ENVVAR2, decode as base64 and write to file, `test2` will
-  #   contain the path of this temporary file:
-  # test2: !ruby/object:BushSlicer::ConfigEnvFile
+  # `input_entry` key will be a string with value
+  #
+  # read value from ENVVAR2, decode as base64 and write to file
+  # file_entry: !ruby/object:BushSlicer::ConfigEnvFile
   #   var_name: ENVVAR2
+  # `file_entry` will contain the path of this temporary file
 
 environments:
   online_starter:
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -50,22 +50,21 @@ global:
   # selenium_url: http://example.com:4444/wd/hub
 
   # Each key:value pair from this section is exported as shell environment variable
-  # into the running context:
+  # into the running context (note: applied only when run from `tools/launch_instance.rb`)
+  #
   # install-envvars:
   #   foo: "bar"
 
   # Examples of usage YAML Helper classes
   # Any value you can set to refer to an environment variable
   #
-  # read value from ENVVAR1 environment variable:
+  # set input_entry to the content of ENVVAR1 environment variable
   # input_entry: !ruby/object:BushSlicer::ConfigEnvVariable
   #   var_name: ENVVAR1
-  # `input_entry` key will be a string with value
   #
-  # read value from ENVVAR2, decode as base64 and write to file
+  # set input_entry to the path of a temporary file with the Base64 decoded content of ENVVAR2 environment variable
   # file_entry: !ruby/object:BushSlicer::ConfigEnvFile
   #   var_name: ENVVAR2
-  # `file_entry` will contain the path of this temporary file
 
 environments:
   online_starter:
--- a/doc/configuration.adoc
+++ b/doc/configuration.adoc
@@ -97,7 +97,7 @@ Parameters:
   `var_name` - the name of shell environment variable for its content
 
 === `ConfigEnvFile`
-The aim of this helper class is a simple injection of file path. The file path is a reference to created temporary file where content from shell environment variable is stored. Temporary file is deleted at the end of execution automatically. It is useful in case when the file path is expected by the configuration.
+The aim of this helper class is an easy injection of file path. The file path is a reference to created temporary file where is stored a content from shell environment variable after its base64 decoding. Temporary file is deleted at the end of execution automatically. It is useful in case when the file path is expected by the configuration.
 
 Parameters:
   `var_name` - the name of shell environment variable for its content
--- a/doc/configuration.adoc
+++ b/doc/configuration.adoc
@@ -76,3 +76,33 @@ Whenever you are, you should usually have the convenience `#conf` method availab
 To access global options just type `conf[:option_name]`. If option is nested, you could do: `conf[:neting0, :nesting1, :options_name]`. This guarantees you that you don't get an exception looking for nested options but just get a nil when the option is not defined.
 
 To access non-global options, you need to write the root option name first, e.g. `conf[:optional_classes, :tcms_tc_manager, :class]`. What is a `root` option name you can see in `Configuration#[]` method.
+
+== Configuration YAML helper classes
+
+Instead of simple kay:value pairs in a configuration it is possible to use so called Helper classes supported by Ruby YAML processor like:
+
+`input_entry: !ruby/object:BushSlicer::FlexyEnvVariable
+  var_name: ENVVAR1`
+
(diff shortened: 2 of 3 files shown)


akostadinov:
I'm not sure whether original request was to allow multiple sets of env variables - e.g. select which set ou want. But having a hard-coded one is a reasonable start.

tools/launch_instance.rb
```diff
@@ -680,6 +680,12 @@ def launch_template(config:, launched_instances_name_prefix:)
       hosts = []
       terminate_spec = {}
       file_details = {}
+
+      # Export configuration's block of environment variables to running environment context
+      conf[:services, :"install-envvars"].each do |key, val|
```
akostadinov, line 685: We need to think whether it should be under `services` or another place. e.g. under `global`

pjanouse: Good point! Originally we agreed to do that by this way, so I'm following our agreement. 
But it make sense to extract it from *services* completely. Let touch this topic on mtg...

pjanouse: As agreed, we moved the introduced section under `global`.

Brief overview of usage in `config.yaml`:
```
services:
  cloud1:
    ..
    ..
  cloudX:
    ..
    ..
global:
  install-envvars:
    key1: value1
    key2: value2
    ..
    keyZ: valueZ
```

Variables defined by above approach are exported and available in the running context for Ruby templates and Shell scripts.

@apodhrad Andrej, can you review and provide feedback if such an implementation fulfill you requirements please?

pjanouse:
Documentation added, so hold-on until feedback received.

pjanouse:
Merge conflict resolved

apodhrad:
The MR looks ok. I have tested it and it works. The suggested improvement, mentioned in my previous comment, will be discussed separately (ONEOCPDEPL-90).

apodhrad:
Hi @akostadinov @liangxia just fyi, I have tested this MR and it is working.

pjanouse:
Thank you @apodhrad for your feedback, I guess we are ready to merge this then...

config/config.yaml
```diff
@@ -49,15 +49,23 @@ global:
...
+  # into the running context:
+  # install-envvars:
+  #   foo: "bar"
+
+  # Examples of usage YAML Helper classes
   # Any value you can set to refer to an environment variable
+  #
   # read value from ENVVAR1 environment variable:
-  # test1: !ruby/object:BushSlicer::FlexyEnvVariable
+  # input_entry: !ruby/object:BushSlicer::FlexyEnvVariable
```
akostadinov, line 61: As a last minute change, may I suggest renaming `FlexyEnvVariable` to `ConfigEnvVariable` (same with `...File`)? It should be clearer this way as well make more sense for the whole verification-tests repo.

akostadinov: If you don't want, don't take it as a blocking request.

pjanouse: This improvement should be an additional effort which is out of scope of this PR (tracked as ONEOCPDEPL-93).

tools/launch_instance.rb
```diff
@@ -681,6 +681,14 @@ def launch_template(config:, launched_instances_name_prefix:)
       hosts = []
       terminate_spec = {}
       file_details = {}
+
+      # Export configuration's block of environment variables to running environment context
+      if (conf[:global, :"install-envvars"] != nil)
+        conf[:global, :"install-envvars"].each do |key, val|
+          ENV[key.to_s] = val
```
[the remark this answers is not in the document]
pjanouse: I agree however I'd prefer to produce a warning when there are clashing sources for a `key`.

pjanouse:
Improved based on provided feedback, please check if it fits your requirements and use-cases.

apodhrad:
Verified - creating and deleting works as expected. Moreover, the warning message is a great idea.
ccing @pjanouse @akostadinov

pjanouse:
Thank you @apodhrad for verification and your feedback.

I'm going to resolve a MR conflict and then ask for merging.

pjanouse:
We are done here, customer reported it is working as he wants already.

Can you do a final check and merge, please?

tools/launch_instance.rb
```diff
@@ -681,6 +681,18 @@ def launch_template(config:, launched_instances_name_prefix:)
       hosts = []
       terminate_spec = {}
       file_details = {}
+
+      # Export configuration's block of environment variables to running environment context
+      if (conf[:global, :"install-envvars"] != nil)
```
akostadinov, line 686:
```suggestion
      if Hash === conf[:global, :"install-envvars"]
```

tools/launch_instance.rb
```diff
@@ -681,6 +681,18 @@ def launch_template(config:, launched_instances_name_prefix:)
...
+        conf[:global, :"install-envvars"].each do |key, val|
+          if ENV.key?(key.to_s)
+            say "WARNING: Clashing value coming from Environment (priority) and configuration for key: '#{key.to_s}'"
```
akostadinov, line 689: maybe better something along the lines of 

> WARNING: value from environment variable #{key.to_s} takes precedence over value in 'install-envvars' configuration

btw perhaps call it `install-env`, should be clear enough, not a hard requirement on my end

config/config.yaml
```diff
@@ -49,15 +49,23 @@ global:
   # Leave unset for default of launching local browser via geckodriver etc.
   # selenium_url: http://example.com:4444/wd/hub
 
+  # Each key:value pair from this section is exported as shell environment variable
+  # into the running context:
+  # install-envvars:
+  #   foo: "bar"
```
akostadinov, line 55: need to make clear that above only applies to `tools/launch_instance.rb`

config/config.yaml
```diff
@@ -49,15 +49,23 @@ global:
...
+
+  # Examples of usage YAML Helper classes
   # Any value you can set to refer to an environment variable
+  #
   # read value from ENVVAR1 environment variable:
```
akostadinov, line 60: > set `input_entry`  to the content of ENVVAR1 environment variable

and remove line 63

config/config.yaml
```diff
@@ -49,15 +49,23 @@ global:
...
-  # test1: !ruby/object:BushSlicer::ConfigEnvVariable
+  # input_entry: !ruby/object:BushSlicer::ConfigEnvVariable
   #   var_name: ENVVAR1
-  # `test1` key will be a string with value
-  # read value from ENVVAR2, decode as base64 and write to file, `test2` will
-  #   contain the path of this temporary file:
-  # test2: !ruby/object:BushSlicer::ConfigEnvFile
+  # `input_entry` key will be a string with value
+  #
+  # read value from ENVVAR2, decode as base64 and write to file
```
akostadinov, line 65: > set `input_entry` to the path of a temporary file with the Base64 decoded content of ENVVAR2 environment variable

akostadinov: and remove line 68

akostadinov: I think it will make it more clear but I leave you decide.

doc/configuration.adoc
```diff
@@ -76,3 +76,33 @@ Whenever you are, you should usually have the convenience `#conf` method availab
...
+* `ConfigEnvFile` - path for file with content of environment variable
+
+=== `ConfigEnvVariable`
+The aim of this helper class is a simple injecting of value from shell environment variable. It is useful in the case when the simple value is expected by the configuration.
+
+Parameters:
+  `var_name` - the name of shell environment variable for its content
+
+=== `ConfigEnvFile`
+The aim of this helper class is a simple injection of file path. The file path is a reference to created temporary file where content from shell environment variable is stored. Temporary file is deleted at the end of execution automatically. It is useful in case when the file path is expected by the configuration.
```
akostadinov, line 100: `s/simple/easy/`, also good to mention Base64 decoding

tools/launch_instance.rb
akostadinov: ok, if we don't want to break existing usage, maybe keep `envvars` string. pls check doc fixes though.

pjanouse: Agree, leaving name of section untouched, improving WARNING msg.

pjanouse:
Thank you @akostadinov for your review, I baked improvements as you'd suggested.

Is there anything else I should care?

doc/configuration.adoc
```diff
@@ -76,3 +76,33 @@ Whenever you are, you should usually have the convenience `#conf` method availab
...
+Parameters:
+  `var_name` - the name of shell environment variable for its content
+
+=== `ConfigEnvFile`
+The aim of this helper class is an easy injection of file path. The file path is a reference to created temporary file where is stored a content from shell environment variable after its base64 decoding. Temporary file is deleted at the end of execution automatically. It is useful in case when the file path is expected by the configuration.
+
+Parameters:
+  `var_name` - the name of shell environment variable for its content
+
+== Additional shell environment variables
```
akostadinov, line 105:
```suggestion
== Additional environment variables for `launch_instance.rb`
```

akostadinov:
added a minor edit

/lgtm

Thank you.

PR #1920, opened 2021-03-19
OCP-38751-[Bug 1914284] The namespace networkpolicy for an application is functional post upgrade

asood-rh:
The PR automates the network policy managing access to the application ensuring it is functional post upgrade.
The network policy allows access to application only from a specific namespace.

Testing:-
Pre upgrade:-
 bundle exec cucumber features/upgrade/sdn/policy-upgrade.feature:86
1 scenario (1 passed)
47 steps (47 passed)
1m6.249

Post upgrade:-
bundle exec cucumber features/upgrade/sdn/policy-upgrade.feature:164
1 scenario (1 passed)
20 steps (20 passed)
0m36.166s

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -305,6 +305,15 @@ Given /^the DefaultDeny policy is applied to the "(.+?)" namespace$/ do | projec
   end
 end
 
+Given /^the AllowNamespaceAndPod policy is applied to the "(.+?)" namespace$/ do | project_name |
+  ensure_admin_tagged
+
+    @result = admin.cli_exec(:create, n: project_name , f: "#{BushSlicer::HOME}/testdata/networking/networkpolicy/allow-ns-and-pod.yaml")
+    unless @result[:success]
+      raise "Failed to apply the default deny policy to specified namespace."
+    end
+end
+
 Given /^the cluster network plugin type and version and stored in the clipboard$/ do
   ensure_admin_tagged
   _host = node.host
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -308,10 +308,10 @@ end
 Given /^the AllowNamespaceAndPod policy is applied to the "(.+?)" namespace$/ do | project_name |
   ensure_admin_tagged
 
-    @result = admin.cli_exec(:create, n: project_name , f: "#{BushSlicer::HOME}/testdata/networking/networkpolicy/allow-ns-and-pod.yaml")
-    unless @result[:success]
-      raise "Failed to apply the default deny policy to specified namespace."
-    end
+  @result = admin.cli_exec(:create, n: project_name , f: "#{BushSlicer::HOME}/testdata/networking/networkpolicy/allow-ns-and-pod.yaml")
+  unless @result[:success]
+    raise "Failed to apply the default deny policy to specified namespace."
+  end
 end
 
 Given /^the cluster network plugin type and version and stored in the clipboard$/ do
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -308,10 +308,12 @@ end
 Given /^the AllowNamespaceAndPod policy is applied to the "(.+?)" namespace$/ do | project_name |
   ensure_admin_tagged
 
-  @result = admin.cli_exec(:create, n: project_name , f: "#{BushSlicer::HOME}/testdata/networking/networkpolicy/allow-ns-and-pod.yaml")
-  unless @result[:success]
-    raise "Failed to apply the default deny policy to specified namespace."
-  end
+  step %Q{I obtain test data file "networking/networkpolicy/allow-ns-and-pod.yaml"}
+  step %Q/I run the :create admin command with:/,table(%{
+    | f | allow-ns-and-pod.yaml |
+    | n | #{project_name} |
+  })
+  step %Q{the step should succeed}
 end
 
 Given /^the cluster network plugin type and version and stored in the clipboard$/ do
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -311,7 +311,7 @@ Given /^the AllowNamespaceAndPod policy is applied to the "(.+?)" namespace$/ do
   step %Q{I obtain test data file "networking/networkpolicy/allow-ns-and-pod.yaml"}
   step %Q/I run the :create admin command with:/,table(%{
     | f | allow-ns-and-pod.yaml |
-    | n | #{project_name} |
+    | n | #{project_name}       |
   })
   step %Q{the step should succeed}
 end
--- a/features/upgrade/sdn/policy-upgrade.feature
+++ b/features/upgrade/sdn/policy-upgrade.feature
@@ -109,6 +109,7 @@ Feature: SDN compoment upgrade testing
     Given a pod becomes ready with labels:
       | name=hello-idle |
     And evaluation of `pod(2).name` is stored in the :pod3 clipboard
+    And evaluation of `pod(2).ip_url` is stored in the :pod3ip clipboard
 
     When I use the "policy-upgrade2" project
     Given I obtain test data file "networking/list_for_pods.json"
@@ -155,6 +156,10 @@ Feature: SDN compoment upgrade testing
       | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
     Then the step should fail 
     And the output should not contain "Hello"
+    When I execute on the "<%= cb.pod5 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod3ip %>:8080 |
+    Then the step should succeed 
+    And the output should contain "Hello"
 
   # @author asood@redhat.com
   # @case_id OCP-38751
@@ -173,6 +178,7 @@ Feature: SDN compoment upgrade testing
     Given a pod becomes ready with labels:
       | name=hello-idle |
     And evaluation of `pod(2).name` is stored in the :pod3 clipboard
+    And evaluation of `pod(2).ip_url` is stored in the :pod3ip clipboard
     When I execute on the "<%= cb.pod3 %>" pod:
       | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
     Then the step should fail
@@ -190,3 +196,7 @@ Feature: SDN compoment upgrade testing
     When I execute on the "<%= cb.pod5 %>" pod:
       | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
     Then the step should fail 
+    When I execute on the "<%= cb.pod5 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod3ip %>:8080 |
+    Then the step should succeed 
+    And the output should contain "Hello"
--- a/features/upgrade/sdn/policy-upgrade.feature
+++ b/features/upgrade/sdn/policy-upgrade.feature
@@ -78,3 +78,116 @@ Feature: SDN compoment upgrade testing
       | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
     Then the step should fail
     And the output should not contain "Hello"
+
+
+  # @author asood@redhat.com
+  @admin
+  @upgrade-prepare
+  Scenario: Check the namespace networkpolicy for an application works well after upgrade - prepare
+    Given I switch to cluster admin pseudo user
+    When I run the :new_project client command with:
+      | project_name | test1 |
+    Then the step should succeed
+    When I run the :new_project client command with:
+      | project_name | test2 |
+    Then the step should succeed
+
+    When I use the "test1" project
+    Given I obtain test data file "networking/list_for_pods.json"
+    And I run the :create client command with:
+      | f | list_for_pods.json |
+    Then the step should succeed
+    Given 2 pods become ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(1).ip` is stored in the :pod2ip clipboard
+    And evaluation of `pod(0).name` is stored in the :pod1 clipboard
+
+    Given I obtain test data file "rc/idle-rc-1.yaml"
+    And I run the :create client command with:
+      | f | idle-rc-1.yaml |
+    Then the step should succeed
+    Given 2 pods become ready with labels:
+      | name=hello-idle |
+    And evaluation of `pod(2).name` is stored in the :pod3 clipboard
+
+    When I use the "test2" project
+    Given I obtain test data file "networking/list_for_pods.json"
+    And I run the :create client command with:
+      | f | list_for_pods.json |
+    Then the step should succeed
+    Given 2 pods become ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(4).name` is stored in the :pod5 clipboard
+
+    Given I obtain test data file "rc/idle-rc-1.yaml"
+    And I run the :create client command with:
+      | f | idle-rc-1.yaml |
+    Then the step should succeed
+    Given 2 pods become ready with labels:
+      | name=hello-idle |
+    And evaluation of `pod(6).name` is stored in the :pod7 clipboard
+
+    Given I have a project
+    When I run the :label admin command with:
+      | resource | namespace                |
+      | name     | test2                    |
+      | key_val  | team=operations          |
+    Then the step should succeed
+
+    Given the AllowNamespaceAndPod policy is applied to the "test1" namespace
+    Then the step should succeed
+
+    When I use the "test1" project
+    When I execute on the "<%= cb.pod1 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
+    Then the step should fail 
+    And the output should not contain "Hello"
+    When I execute on the "<%= cb.pod3 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
+    Then the step should fail
+    And the output should not contain "Hello"
+
+    When I use the "test2" project
+    When I execute on the "<%= cb.pod5 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
+    Then the step should succeed
+    And the output should contain "Hello"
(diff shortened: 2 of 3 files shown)


asood-rh:
Please review @zhaozhanqi @huiran0826 @rbbratta

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -78,3 +78,116 @@ Feature: SDN compoment upgrade testing
...
     And the output should not contain "Hello"
+
+
+  # @author asood@redhat.com
+  @admin
+  @upgrade-prepare
+  Scenario: Check the namespace networkpolicy for an application works well after upgrade - prepare
+    Given I switch to cluster admin pseudo user
+    When I run the :new_project client command with:
+      | project_name | test1 |
```
zhaozhanqi, line 89: hmm, How about using `policy-upgrade1` and `policy-upgrade2` here in case the cluster already have `test1` and `test2` project

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -78,3 +78,116 @@ Feature: SDN compoment upgrade testing
...
+    Then the step should succeed
+
+    When I use the "test1" project
+    Given I obtain test data file "networking/list_for_pods.json"
+    And I run the :create client command with:
+      | f | list_for_pods.json |
+    Then the step should succeed
+    Given 2 pods become ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(1).ip` is stored in the :pod2ip clipboard
```
zhaozhanqi, line 102: here better use `ip_url` instead of `ip` for compatibility of ipv6 cluster.  
/pod(1).ip/pod(1).ip_url/

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -78,3 +78,116 @@ Feature: SDN compoment upgrade testing
...
+
+    Given I obtain test data file "rc/idle-rc-1.yaml"
+    And I run the :create client command with:
+      | f | idle-rc-1.yaml |
+    Then the step should succeed
+    Given 2 pods become ready with labels:
+      | name=hello-idle |
+    And evaluation of `pod(6).name` is stored in the :pod7 clipboard
+
+    Given I have a project
```
zhaozhanqi, line 130: `Given I have a project`  should not be used.

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -78,3 +78,116 @@ Feature: SDN compoment upgrade testing
...
+    And the output should not contain "Hello"
+    When I execute on the "<%= cb.pod3 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
+    Then the step should fail
+    And the output should not contain "Hello"
+
+    When I use the "test2" project
+    When I execute on the "<%= cb.pod5 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
+    Then the step should succeed
```
zhaozhanqi, line 153: Here we need to add one step to make sure 'pod5 cannot access pod3.ip ( since pod do not have correct label)

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -78,3 +78,116 @@ Feature: SDN compoment upgrade testing
...
+    Given 2 pods become ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(1).ip` is stored in the :pod2ip clipboard
+    And evaluation of `pod(0).name` is stored in the :pod1 clipboard
+
+    Given I obtain test data file "rc/idle-rc-1.yaml"
+    And I run the :create client command with:
+      | f | idle-rc-1.yaml |
+    Then the step should succeed
+    Given 2 pods become ready with labels:
```
zhaozhanqi, line 109: we can update the replicas to 1 if only need one pod. 
```
    When I run oc create over "list_for_pods.json" replacing paths:
      | ["items"][0]["spec"]["replicas"] | 1 |
```

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -78,3 +78,116 @@ Feature: SDN compoment upgrade testing
...
+    Then the step should fail
+
+    When I use the "test2" project
+    Given status becomes :running of 2 pods labeled:
+      | name=test-pods |
+    And evaluation of `pod(4).name` is stored in the :pod5 clipboard
+    When I execute on the "<%= cb.pod5 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
+    Then the step should succeed
+    Given status becomes :running of 2 pods labeled:
```
zhaozhanqi, line 188: Here we need to add one more steps to make sure 'pod5 cannot access pod3.ip ( since pod do not have correct label)

features/upgrade/sdn/policy-upgrade.feature
asood-rh: Sure the name the folders can be changed.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: @zhaozhanqi I was still label the project without using 'Given I have a project' step so I can remove this step but what are the concerns using this step.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: pod3 has label name=hello-idle, pod5 has label name=test-pods.
So pod with label name=test-pods not being accessed by pod with label name=hello-idle is covered by the following steps:-

When I execute on the "<%= cb.pod7 %>" pod:
      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
    Then the step should fail
    And the output should not contain "Hello"
Though the pods are in different namespaces/project where pod2ip is ip of the pod with label name=test-pods and pod7 is pod with label name=hello-idle, these steps do test the flow you are referring to.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: pod3 has label name=hello-idle, pod5 has label name=test-pods.
So pod with label name=test-pods not being accessed by pod with label name=hello-idle is covered by the following steps:-

When I execute on the "<%= cb.pod7 %>" pod:
      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
    Then the step should fail
    And the output should not contain "Hello"
Though the pods are in different namespaces/project where pod2ip is ip of the pod with label name=test-pods and pod7 is pod with label name=hello-idle, these steps do test the flow you are referring to.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: Sure this can be changed for namespace policy-upgrade2.

features/upgrade/sdn/policy-upgrade.feature
zhaozhanqi: this step will create new project but not use it in the following,  means it will create one non-used project.

features/upgrade/sdn/policy-upgrade.feature
zhaozhanqi: both pod3 and pod7 do not match :
namespace with label " team=operation' and pods label 'name=test-pods'
pod3 only has pod label 'name=test-pods', however it's namespace do not have label 'team=operation'
pod7 only has namespace label 'team=operation', However without pod label 'name=test-pods'

So we need one more pod (pod5) match both label (namespace with label " team=operation' and pods label 'name=test-pods'). 

- it can access pods with label 'name=test-pods' in namespace test1  ----> the current steps already including this checkpoint

- but it cannot access pods without label 'name=test-pods' in in namespace test1  ---> **we should add this checkpoint**

testdata/networking/networkpolicy/allow-ns-and-pod.yaml
```diff
@@ -0,0 +1,19 @@
...
+      name: test-pods
+  ingress:
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            team: operations
+        podSelector:
+          matchLabels:
+            name: test-pods
+  podSelector:
```
zhaozhanqi, line 17: line '17-19' should be dup with line '6-8'

features/upgrade/sdn/policy-upgrade.feature
zhaozhanqi: you can see the policy below
```
spec:
  podSelector:
    matchLabels:
      name: test-pods       --> here is for matching namespace with this policy .  means test1 in our case
  ingress:
    - from:
      - namespaceSelector:
          matchLabels:
            team: operations
        podSelector:
          matchLabels:
            name: test-pods                 --> here is for matching test2 namespace pods in our case
```

features/upgrade/sdn/policy-upgrade.feature
asood-rh: Reduced # of pods in policy-upgrade2 namespace to 2 pods (1 test-pods and 1 hello-idle) and in policy-upgrade1 to 3 pods (2 test-pods and 1 hello-idle)

features/upgrade/sdn/policy-upgrade.feature
asood-rh: I do not mind removing this step as reduces number of steps in scenario. I do not think it tries to create a new project. If had attempted to create a new project the script would give an error - project already exits.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: Fixed

testdata/networking/networkpolicy/allow-ns-and-pod.yaml
asood-rh: Fixed

features/upgrade/sdn/policy-upgrade.feature
asood-rh: Scenario is pod labeled test-pods is only accessible by the pod test-pods in namespace policy-upgrade2
Putting in more details for the cases.
Namespace policy-upgrade1
hello-idle-5dftl   1/1     Running   0          141m   10.131.0.87   (hello-idle/ pod2)   
test-rc-jgtkr      1/1     Running   0          141m   10.131.0.86     (test-pods/ pod1)    
test-rc-rvccc      1/1     Running   0          141m   10.129.2.43  (test-pods/pod2ip)
Namespace policy-upgrade2
hello-idle-7chxv   1/1     Running   0          (hello-idle/ pod5)
test-rc-zvstp      1/1     Running   0          (test-pods/pod4)

Case 1 Curl from pod1 -> pod2ip fails, this is case where both pods have same label test-pods in namespace that is not labeled (I think you are referring to this case in your comment, this is already covered) 
Case 2 Curl from pod3->pod2ip fails, this case where hello-idle is trying to get to pod labeled test-pods in namespace that is not labeled.
Case 3 Curl from pod4->pod2ip pass, this is case where test-pods in upgrade-policy2 is accesses test-pods in upgrade-policy1
Case 4 Curl from pod5->pod2ip fails this is case where hello-idle in upgrade-policy2 is trying to access test-pods in upgrade-policy1.

zhaozhanqi: Actually I want to add is :  pod4 -> pod2(hello-idle-5dftl 1/1 Running 0 141m 10.131.0.87 (hello-idle/ pod2)),  this should be failed since pod2 do not have label 'name=test-pods'

others looks good to me

asood-rh: This will succeed as these pods are not selected by the podSelector specified in network policy or in other words network policy does not impact these pods.
Nework policy is about the test-pods being accessible pods that are labeled test-pods and are in namspace that is labeled team=operations.

zhaozhanqi: I saw you update `hello-idle-5dftl 1/1 Running 0 141m 10.131.0.87 (hello-idle/ pod3)` to pod3 from pod2

then pod4 -> pod3 should be failed. 
as you said, this pod did not be selected by podSelectror.  so it should be accessed from pod4.

asood-rh: Yes, I fixed the pod number, it was incorrect. I will add another case, though there are bunch of others that could be added.

asood-rh: Added another case.

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -78,3 +78,125 @@ Feature: SDN compoment upgrade testing
...
+    Then the step should succeed
+    Given a pod becomes ready with labels:
+      | name=hello-idle |
+    And evaluation of `pod(4).name` is stored in the :pod5 clipboard
+    When I execute on the "<%= cb.pod5 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
+    Then the step should fail 
+    When I execute on the "<%= cb.pod4 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod3ip %>:8080 |
+    Then the step should succeed
```
zhaozhanqi, line 201: I think here should be failed.  pod4 should not be able to access pod3ip since pod3 do not have label with 'name=test-pods'.
I'm not sure if run again this case after updating. Could you help paste the logs ?

asood-rh: pod3ip is of pod3 in policy-upgrade1 namespace and has label name=hello-idle. Network policy does not apply to that pod it only applies to pod with label name=test-pods. pod4 has label name=test-pods. This is the only policy in namespace policy-upgrade1. I believe there has to be deny all policy if you want completely isolate namespaces and then allow some access.

When I execute on the "<%= cb.pod4 %>" pod:                                                     # features/step_definitions/pod.rb:216
      [11:59:57] INFO> Shell Commands: oc exec test-rc-sspx2  --kubeconfig=/home/asood/workdir/asood-asood/ocp4_admin.kubeconfig -n policy-upgrade2  -i -- curl -s --connect-timeout 5 10.128.3.112:8080
      Hello Pod!
      
      STDERR:
      E0325 07:59:57.618183  280520 v2.go:105] read /dev/stdin: resource temporarily unavailable
      
      [11:59:58] INFO> Exit Status: 0
      | curl | -s | --connect-timeout | 5 | <%= cb.pod3ip %>:8080 |
    Then the step should succeed                                                                    # features/step_definitions/common.rb:4
    And the output should contain "Hello"                                                           # features/step_definitions/common.rb:113

features/step_definitions/networking.rb
```diff
@@ -305,6 +305,15 @@
   end
 end
 
+Given /^the AllowNamespaceAndPod policy is applied to the "(.+?)" namespace$/ do | project_name |
+  ensure_admin_tagged
+
+    @result = admin.cli_exec(:create, n: project_name , f: "#{BushSlicer::HOME}/testdata/networking/networkpolicy/allow-ns-and-pod.yaml")
```
pruan-rht, line 311: @asood-rh indentation issue

asood-rh: @pruan-rht fixed

features/step_definitions/networking.rb
```diff
@@ -305,6 +305,15 @@
   end
 end
 
+Given /^the AllowNamespaceAndPod policy is applied to the "(.+?)" namespace$/ do | project_name |
+  ensure_admin_tagged
+
+  @result = admin.cli_exec(:create, n: project_name , f: "#{BushSlicer::HOME}/testdata/networking/networkpolicy/allow-ns-and-pod.yaml")
```
pruan-rht, line 311: @asood-rh I think we prefer to use the generic step to download get the testdata first using ```step %Q(I obtain test data file "networking/networkpolicy/allow-ns-and-pod.yaml")``` and then you can use `f: "allow-ns-and-pod.yaml"`

asood-rh: @pruan-rht What is the advantage of using the generic step to download the data file before applying it when the current implementation does both operations in on step?
I am following the pattern used for apply another policy.

akostadinov: +1 The idea is that these files may be moved or reorganized and things can remain transparent to users.

asood-rh: @akostadinov yes, it is my first PR.

features/step_definitions/networking.rb
```diff
@@ -308,10 +308,12 @@
...
   ensure_admin_tagged
 
-  @result = admin.cli_exec(:create, n: project_name , f: "#{BushSlicer::HOME}/testdata/networking/networkpolicy/allow-ns-and-pod.yaml")
-  unless @result[:success]
-    raise "Failed to apply the default deny policy to specified namespace."
-  end
+  step %Q{I obtain test data file "networking/networkpolicy/allow-ns-and-pod.yaml"}
+  step %Q/I run the :create admin command with:/,table(%{
+    | f | allow-ns-and-pod.yaml |
+    | n | #{project_name} |
```
pruan-rht, line 314: @asood-rh table alignment.  You could've also just do` @result = admin.cli_exec(:create, n: project_name , f: "allow-ns-and-pod.yaml")`

asood-rh: Fixed alignment.

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -78,3 +78,125 @@ Feature: SDN compoment upgrade testing
...
     Then the step should fail
     And the output should not contain "Hello"
+
+
+  # @author asood@redhat.com
+  @admin
+  @upgrade-prepare
+  Scenario: Check the namespace networkpolicy for an application works well after upgrade - prepare
+    Given I switch to cluster admin pseudo user
+    When I run the :new_project client command with:
```
pruan-rht, line 88: @asood-rh do we have to use hardcoded project names? I think you can use the step `Given I have a project` and then save the project name into a clipboard.  This way, the project names are random to lessen the chance of name collision.

asood-rh: @pruan-rht this is upgrade test, so we do want a known project, policy and application so that post upgrade we can check for the specific resources.

pruan-rht: @asood-rh oh ok, didn't catch that...looks good then

pruan-rht: /lgtm

PR #1921, opened 2021-03-20
fix step to use cache: false

pruan-rht:
We need to feed the method and set `cached` to be `false` to force it to fetch new values since it's a status field


pruan-rht:
@duanwei33 this should fix the stale data issue that we discussed last night.

PR #1923, opened 2021-03-22
Update step `I pick a random machineset to scale`

sunzhaohua2:
Update step `I pick a random machineset to scale` in case windows machineset is selected in a cluster with windows nodes. @jhou1 @miyadav please help to take a look.

### DIFF
--- a/features/step_definitions/machine_set.rb
+++ b/features/step_definitions/machine_set.rb
@@ -1,7 +1,7 @@
-Given(/^I pick a random machineset to scale$/) do
+Given(/^I pick a random( windows)? machineset to scale$/) do | windows |
   ensure_admin_tagged
   machine_sets = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api")).
-    select { |ms| ms.available_replicas >= 1 }
+    select { |ms| ms.available_replicas >= 1 && (windows ? ms.is_windows_machinesets? : !ms.is_windows_machinesets?) }
   cache_resources *machine_sets.shuffle
 end
 
@@ -56,8 +56,8 @@ Then(/^the machineset should have expected number of running machines$/) do
   end
 end
 
-Given(/^I clone a machineset and name it "([^"]*)"$/) do | ms_name |
-  step %Q{I pick a random machineset to scale}
+Given(/^I clone a( windows)? machineset and name it "([^"]*)"$/) do | os_type, ms_name |
+  step %Q{I pick a random#{os_type} machineset to scale}
 
   ms_yaml = machine_set.raw_resource.to_yaml
   new_spec = YAML.load ms_yaml
--- a/lib/openshift/machine_set.rb
+++ b/lib/openshift/machine_set.rb
@@ -40,5 +40,10 @@ module BushSlicer
       raw_resource(user: user, cached: cached, quiet: quiet).
         dig('spec', 'taints')
     end
+
+    def is_windows_machinesets?(user: nil, cached: true, quiet: false)
+      rr = raw_resource(user: user, cached: cached, quiet: quiet)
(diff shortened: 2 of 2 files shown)


miyadav:
LGTM , shall we run it once and confirm the results ?

sunzhaohua2:
thanks for reminder @miyadav I tested clone a linux machineset with case OCP-28108: 
```
      [05:50:41] INFO> === End After Scenario: Cluster should automatically scale up and scale down with clusterautoscaler deployed ===

1 scenario (1 passed)
28 steps (28 passed)
12m7.297s
```
 I tested clone a windows machineset with steps:
```
  Scenario: Test create a windows machineset
    Given I have an IPI deployment
    And I switch to cluster admin pseudo user
    And I use the "openshift-machine-api" project
    And admin ensures machine number is restored after scenario

    Given I clone a windows machineset and name it "machineset-windows"
```
clone a windows machineset result:
```
      [07:25:23] INFO> === End After Scenario: Test create a windows machineset ===

1 scenario (1 passed)
5 steps (5 passed)
9m34.887s
```

features/step_definitions/machine_set.rb
```diff
@@ -1,7 +1,14 @@
 Given(/^I pick a random machineset to scale$/) do
   ensure_admin_tagged
   machine_sets = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api")).
-    select { |ms| ms.available_replicas >= 1 }
+    select { |ms| ms.available_replicas >= 1 && ms.windows_machinesets != "Windows"}
+  cache_resources *machine_sets.shuffle
+end
+
+Given(/^I pick a random windows machineset to scale$/) do
```
liangxia, line 8: `Given(/^I pick a random (windows|linux) machineset to scale$/)`

lib/openshift/machine_set.rb
```diff
@@ -40,5 +40,10 @@ def taints(user: nil, cached: true, quiet: false)
       raw_resource(user: user, cached: cached, quiet: quiet).
         dig('spec', 'taints')
     end
+
+    def windows_machinesets(user: nil, cached: true, quiet: false)
```
liangxia, line 44: Anything indicate it's a windows machineset ?

jhou1: The naming and definition of the method can be more straightforward, I suggest defining a method like `is_windows_machineset?` which returns true or false.

```
def is_windows_machineset?(user: nil, cached: true, quiet: false)
  rr.dig('spec', 'template', 'metadata', 'labels','machine.openshift.io/os-id') == "Windows"
end
```

features/step_definitions/machine_set.rb
jhou1: The original step can be updated slimilar to `I pick a random( windows)? machineset to scale` with condition to deal with windows machine scenario.

lib/openshift/machine_set.rb
akostadinov: @jhou1 , I fully agree with you. Also related to what we spoke about earlier.

sunzhaohua2:
@miyadav @jhou1 @liangxia @akostadinov have updated and tested locally, it passed, please help to review again.

features/step_definitions/machine_set.rb
```diff
@@ -1,7 +1,7 @@
-Given(/^I pick a random machineset to scale$/) do
+Given(/^I pick a random( windows)? machineset to scale$/) do | windows |
   ensure_admin_tagged
   machine_sets = BushSlicer::MachineSet.list(user: admin, project: project("openshift-machine-api")).
-    select { |ms| ms.available_replicas >= 1 }
+    select { |ms| ms.available_replicas >= 1 && (!windows || ms.is_windows_machinesets?) }
```
jhou1, line 4: What about `(windows ? ms.is_windows_machinesets? : !ms.is_windows_machinesets?)`

PR #1927, opened 2021-03-24
Disable all repos except several needed ones

liangxia:
We have met twice that a repo is enabled unexpectedly in redhat.repo. So disable all repos by default to prevent such issue from happen again.
/cc @akostadinov @jhou1 @jianlinliu

### DIFF
--- a/tools/jenkins/slaves/Dockerfile.rhel7
+++ b/tools/jenkins/slaves/Dockerfile.rhel7
@@ -9,13 +9,10 @@ ENV USER_PATHS="$HOME /etc/machine-id /etc/passwd /etc/pki"
 
 RUN set -x && \
     yum repolist all && \
-    yum-config-manager --disable rhel-7-for-power-le-rpms && \
-    yum-config-manager --disable rhel-7-server-htb-rpms && \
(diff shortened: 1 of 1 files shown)


tools/jenkins/slaves/Dockerfile.rhel7
```diff
@@ -55,7 +52,7 @@ RUN setpriv --reuid 1001 scl enable rh-ruby27 "$HOME/verification-tests/tools/ha
 
 RUN rpm -qa && \
     yum clean all -y && \
-    rm -rf $HOME/verification-tests /var/cache/yum /tmp/*
+    rm -rf $HOME/verification-tests /var/cache/yum /var/tmp/ /tmp/*
```
liangxia, line 55:
```
+ yum install -y --setopt=tsflags=nodocs https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
Examining /var/tmp/yum-root-CQKasF/epel-release-latest-7.noarch.rpm: epel-release-7-13.noarch
Marking /var/tmp/yum-root-CQKasF/epel-release-latest-7.noarch.rpm to be installed
```

PR #1930, opened 2021-03-24
add get ipv6 address for pod and service

zhaozhanqi:
for dual stack cluster .we have ipv4 and ipv6 ip for pod and service, see 

pod:
```
              f:ip: {}
    hostIP: 192.168.123.71
    podIP: 10.130.0.94
    podIPs:
    - ip: 10.130.0.94
    - ip: fd01:0:0:3::5e
```
service:
```
spec:
  clusterIP: 172.30.133.189
  clusterIPs:
   - 172.30.133.189
   - fd02::1cd8
  ipFamilies:
```
@akostadinov @pruan-rht @liangxia  could you help take a look?
 
cc @rbbratta @weliang1 @huiran0826 @asood-rh @kakkoyun @brahaney

### DIFF
--- a/features/networking/ipv6_dualstack.feature
+++ b/features/networking/ipv6_dualstack.feature
@@ -1,0 +1,78 @@
+Feature: ipv6 dual stack cluster test scenarios
+
+  # @author zzhao@redhat.com
+  # @case_id OCP-40581
+  @admin
+  Scenario: Project should be in isolation when using multitenant policy for ipv6 dual stack
+    # create project and pods
+    Given I have a project
+    And evaluation of `project.name` is stored in the :proj1 clipboard
+    Given I obtain test data file "networking/list_for_pods_ipv6.json"
+    When I run the :create client command with:
+      | f | list_for_pods_ipv6.json |
+    Then the step should succeed
+    Given a pod becomes ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(0).ip_v4` is stored in the :p1pod1ipv4 clipboard
+    And evaluation of `pod(0).ip_v6_url` is stored in the :p1pod1ipv6_url clipboard
+    Given I use the "test-service" service
+    And evaluation of `service.ip_v4_url` is stored in the :service_ipv4_url clipboard
+    And evaluation of `service.ip_v6_url` is stored in the :service_ipv6_url clipboard
+
+    # create another project and pods
+    Given I create a new project
+    And evaluation of `project.name` is stored in the :proj2 clipboard
+    Given I obtain test data file "networking/list_for_pods_ipv6.json"
+    When I run the :create client command with:
+      | f | list_for_pods_ipv6.json |
+    Then the step should succeed
+    Given a pod becomes ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(1).name` is stored in the :p2pod1 clipboard
+
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 10 | <%= cb.p1pod1ipv4 %>:<%= service.ports[0]["targetPort"] %> |
+    Then the step should succeed
+    And the output should contain "Hello"
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 10 | <%= cb.p1pod1ipv6_url %>:<%= service.ports[0]["targetPort"] %> |
+    Then the step should succeed
+    And the output should contain "Hello"
+
+    #access pod by service ipv4 and ipv6 address
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 10 | <%= cb.service_ipv4_url %> |
+    Then the step should succeed
+    And the output should contain "Hello"
+
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 10 | <%= cb.service_ipv6_url %> |
+    Then the step should succeed
+    And the output should contain "Hello"
+
+
+    #create multitetant policy in project 1
+    Given I obtain test data file "networking/networkpolicy/multitenant_policy.yaml"
+    When I run the :create admin command with:
+      | f | multitenant_policy.yaml |
+      | n | <%= cb.proj1 %>         |
+    Then the step should succeed
+
+
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 5 | <%= cb.p1pod1ipv4 %>:<%= service.ports[0]["targetPort"] %> |
+    Then the step should fail
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 5 | <%= cb.p1pod1ipv6_url %>:<%= service.ports[0]["targetPort"] %> |
+    Then the step should fail
+
+    #access pod by service ipv4 and ipv6 address
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 5 | <%= cb.service_ipv4_url %> |
+    Then the step should fail
+
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 5 | <%= cb.service_ipv6_url %> |
+    Then the step should fail
+
+
--- a/lib/openshift/pod.rb
+++ b/lib/openshift/pod.rb
@@ -50,6 +50,7 @@ module BushSlicer
       props[:volumes] = spec["volumes"]
       s = pod_hash["status"]
       props[:ip] = s["podIP"]
+      props[:ips] = s["podIPs"]
       # status should be retrieved on demand but we cache it for the brave
       props[:status] = s
 
@@ -111,6 +112,23 @@ module BushSlicer
     # @note call without parameters only when props are loaded
     def ip(user: nil, cached: true, quiet: false)
       return get_cached_prop(prop: :ip, user: user, cached: cached, quiet: quiet)
+    end
+
+    def ip_v6(user: nil, cached: true, quiet: false)
+      ips = get_cached_prop(prop: :ips, user: user, cached: cached, quiet: quiet)
+      ipv6 = ips.find { |ip| ip["ip"].include? ":" }
+      return ipv6["ip"]
+    end
+
+    # return pod ipv6 address as URL for dualstack cluster
+    def ip_v6_url(user: nil, cached: true, quiet: false)
+      return "[#{ip_v6(user: user, cached: cached, quiet: quiet)}]"
+    end
+
+    def ip_v4(user: nil, cached: true, quiet: false)
+      ips = get_cached_prop(prop: :ips, user: user, cached: cached, quiet: quiet)
+      ipv4 = ips.find { |ip| ip["ip"].include? "." }
+      return ipv4["ip"]
     end
 
     # @return [String] string IP if IPv4 or [IP] if IPv6
--- a/lib/openshift/service.rb
+++ b/lib/openshift/service.rb
@@ -80,6 +80,35 @@ module BushSlicer
       spec = raw_resource(user: user, cached: cached, quiet: quiet).dig('spec')
       return (spec.dig('portalIP') || spec.dig('clusterIP'))
     end
+    
+    # return service ipv6 address for dualstack cluster
+    def ip_v6(user: nil, cached: true, quiet: false)
+      spec = raw_resource(user: user, cached: cached, quiet: quiet).dig('spec')
+      ipv6 = spec['clusterIPs'].find { |ip| ip.include? ":" }
+      return ipv6
+    end
+
+    # return service ipv6 address as URL for dualstack cluster
+    def ip_v6_url(user: nil, cached: true, quiet: false)
+      ipv6 = self.ip_v6(user: user, cached: cached, quiet: quiet)
+      port = self.ports(user: user, cached: true, quiet: quiet)[0]["port"]
+      "[#{ipv6}]:#{port}"
+    end
+
+    # return service ipv4 address for dualstack cluster
+    def ip_v4(user: nil, cached: true, quiet: false)
+      spec = raw_resource(user: user, cached: cached, quiet: quiet).dig('spec')
+      ipv4 = spec['clusterIPs'].find { |ip| ip.include? "." }
+      return ipv4
+    end
+
+    # return service ipv4 address as URL for dualstack cluster
+    def ip_v4_url(user: nil, cached: true, quiet: false)
+      ipv4 = self.ip_v4(user: user, cached: cached, quiet: quiet)
+      port = self.ports(user: user, cached: true, quiet: quiet)[0]["port"]
+      "#{ipv4}:#{port}"
+    end
+
     # @note call without parameters only when props are loaded
     # return @Array of ports
     def ports(user: nil, cached: true, quiet: false)
--- a/testdata/networking/list_for_pods_ipv6.json
+++ b/testdata/networking/list_for_pods_ipv6.json
@@ -1,0 +1,65 @@
+{
+    "apiVersion": "v1",
+    "kind": "List",
+    "items": [
+        {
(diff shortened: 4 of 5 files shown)


lib/openshift/service.rb
```diff
@@ -80,6 +80,14 @@ def ip(user: nil, cached: true, quiet: false)
       spec = raw_resource(user: user, cached: cached, quiet: quiet).dig('spec')
       return (spec.dig('portalIP') || spec.dig('clusterIP'))
     end
+    
+    # return service ipv6 address for dualstack cluster
+    def ip_v6(user: nil, cached: true, quiet: false)
+      spec = raw_resource(user: user, cached: cached, quiet: quiet).dig('spec')
+      ipv6 = spec.dig('clusterIPs')[0].include?(":") ? spec.dig('clusterIPs')[0] : spec.dig('clusterIPs')[1]
```
akostadinov, line 87: Is there any documentation that describes IPv6 will always be the second IP? If this is not guaranteed, then both IPs need to be inspected and the correct one chosen. Same applies to Pod.

zhaozhanqi:
Add one test scenario for ipv6 dual stack cluster.  see logs: http://pastebin.test.redhat.com/950399

lib/openshift/service.rb
zhaozhanqi: I guess it should be according the order of network config:
```
  spec:
    clusterNetwork:
    - cidr: 10.128.0.0/14                --> ipv4 first
      hostPrefix: 23
    - cidr: fd01::/48                  --> then ipv6 
      hostPrefix: 64
    externalIP:
      policy: {}
    networkType: OVNKubernetes
    serviceNetwork:
    - 172.30.0.0/16
    - fd02::/112
  status:
    clusterNetwork:
    - cidr: 10.128.0.0/14
      hostPrefix: 23
    - cidr: fd01::/48
      hostPrefix: 64
    clusterNetworkMTU: 1400
    networkType: OVNKubernetes
    serviceNetwork:
    - 172.30.0.0/16             ##--> same here
    - fd02::/112

```

akostadinov: I now see that you are checking which one is the correct. I would suggest to use need something like
```
ipv6 = spec['clusterIPs'].find { |ip| ip.include? "::" }
```

In this way you will get `nil` result in case IPv6 IP is missing.

It also makes sense to add an `ip_v4` method to return an IPv4 address with same premise - that it might be missing.

The `ip` method would return the default one - v4 or v6.

zhaozhanqi: thanks @akostadinov 
add ipv4 and ipv6 method to make sure get the correct ip address

akostadinov:
Thanks,  I will let somebody from team review the added scenario.

lib/openshift/service.rb
weliang1: > I guess it should be according the order of network config:
> 
> ```
>   spec:
>     clusterNetwork:
>     - cidr: 10.128.0.0/14                --> ipv4 first
>       hostPrefix: 23
>     - cidr: fd01::/48                  --> then ipv6 
>       hostPrefix: 64
>     externalIP:
>       policy: {}
>     networkType: OVNKubernetes
>     serviceNetwork:
>     - 172.30.0.0/16
>     - fd02::/112
>   status:
>     clusterNetwork:
>     - cidr: 10.128.0.0/14
>       hostPrefix: 23
>     - cidr: fd01::/48
>       hostPrefix: 64
>     clusterNetworkMTU: 1400
>     networkType: OVNKubernetes
>     serviceNetwork:
>     - 172.30.0.0/16             ##--> same here
>     - fd02::/112
> ```

For Pod, first IP is always ipv4
Status:       Running
IP:           10.128.3.230
IPs:
  IP:           10.128.3.230
  IP:           fd01:0:0:5::1e6

For service, the order of IP is based on the the order of IP Families, ' IPv4,IPv6' or 'IPv6,IPv4', check below test logs:
$ oc create -f https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/Services/DualStack/PreferDualStack-v4v6-svc.yaml
$ oc describe svc preferdualstack-v4v6-svc
Type:              ClusterIP
IP Family Policy:  PreferDualStack
IP Families:       IPv4,IPv6
IP:                172.30.51.16
IPs:               172.30.51.16,fd02::a46f

$ oc create -f https://raw.githubusercontent.com/weliang1/Openshift_Networking/master/Features/Services/DualStack/PreferDualStack-v6v4-svc.yaml
$ oc describe svc preferdualstack-v6v4-svc
Type:              ClusterIP
IP Family Policy:  PreferDualStack
IP Families:       IPv6,IPv4
IP:                fd02::45c3
IPs:               fd02::45c3,172.30.206.223

$ oc get svc
NAME                       TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
preferdualstack-v4v6-svc   ClusterIP   172.30.51.16   <none>        8080/TCP   109s
preferdualstack-v6v4-svc   ClusterIP   fd02::45c3     <none>        8080/TCP   11s

zhaozhanqi: yes, I know service can set the order by `IP Families: IPv6,IPv4`
not sure what's the result for pod if we set ipv6 first, eg
```
    clusterNetwork:
    - cidr: fd01::/48                  -->  ipv6 
      hostPrefix: 64
    - cidr: 10.128.0.0/14                --> ipv4 
      hostPrefix: 23
```
but anyways, the current implementation can work for both

weliang1: IP Families is not for pod, pod always get both ip address in dual stack cluster.
IPs:
IP: 10.128.3.230
IP: fd01:0:0:5::1e6

zhaozhanqi: yes,  I know `IP Families: IPv6,IPv4` is only for service. 

I mean if we set ipv6 first in network config as below. 


```
clusterNetwork:
    - cidr: fd01::/48                  -->  ipv6 
      hostPrefix: 64
    - cidr: 10.128.0.0/14                --> ipv4 
      hostPrefix: 23
```
then pod ip may show: 
```
IPs:
IP: fd01:0:0:5::1e6
IP: 10.128.3.230
```

zhaozhanqi:
@rbbratta @weliang1 @huiran0826 @asood-rh @kakkoyun @brahaney PTAL，thanks

features/networking/ipv6_dualstack.feature
```diff
@@ -0,0 +1,78 @@
+Feature: ipv6 dual stack cluster test scenarios
+
+  # @author zzhao@redhat.com
+  # @case_id OCP-40581
+  @admin
+  Scenario: Perject should be isolation when using multitenant policy for ipv6 dual stack
```
kedark3, line 6: `Perject` -> `Project`?

kedark3: Do we want to change this to `Project should be in isolation` ?

lib/openshift/pod.rb
```diff
@@ -113,6 +114,18 @@ def ip(user: nil, cached: true, quiet: false)
       return get_cached_prop(prop: :ip, user: user, cached: cached, quiet: quiet)
     end
 
+    def ip_v6(user: nil, cached: true, quiet: false)
+      ips = get_cached_prop(prop: :ips, user: user, cached: cached, quiet: quiet)
+      ipv6 = ips.find { |ip| ip["ip"].include? ":" }
```
kedark3, line 119: Should we also check for presence of `[]` in the address? Usually IP6 addr has `[]` right?

lib/openshift/service.rb
```diff
@@ -80,6 +80,21 @@ def ip(user: nil, cached: true, quiet: false)
       spec = raw_resource(user: user, cached: cached, quiet: quiet).dig('spec')
       return (spec.dig('portalIP') || spec.dig('clusterIP'))
     end
+    
+    # return service ipv6 address for dualstack cluster
+    def ip_v6(user: nil, cached: true, quiet: false)
+      spec = raw_resource(user: user, cached: cached, quiet: quiet).dig('spec')
+      ipv6 = spec['clusterIPs'].find { |ip| ip.include? ":" }
```
kedark3, line 87: And here as well about `[]` ? Or other option could be:
Check if the IP address does NOT contain `.` (dots) in that case it would be assumed to be ipv6

Would that be acceptable ?

testdata/networking/list_for_pods_ipv6.json
```diff
@@ -0,0 +1,65 @@
...
+                "labels": {
+                    "name": "test-service"
+                },
+                "name": "test-service"
+            },
+            "spec": {
+                "ports": [
+                    {
+                        "name": "http",
+                        "port": 27017,
```
kedark3, line 52: Probably this will work fine, but shall we choose something in the default range of 30000-32767? As noted here: https://kubernetes.io/docs/concepts/services-networking/service/#nodeport

`port-range flag (default: 30000-32767)`

kedark3 (changes requested):

lib/openshift/pod.rb
rbbratta: IPv6 has `[]` only for URLs, that is why `ip_url` method checks.

```ruby
      return raw_ip.include?(":") ? "[#{raw_ip}]" : raw_ip
```

we decided on the `_url` suffix for methods that return IPs for use in URLs.

features/networking/ipv6_dualstack.feature
```diff
@@ -0,0 +1,78 @@
...
+    And the output should contain "Hello"
+
+    #access pod by service ipv4 and ipv6 address
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 5 | <%= cb.service_ipv4 %>:27017 |
+    Then the step should succeed
+    And the output should contain "Hello"
+
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 5 | [<%= cb.service_ipv6 %>]:27017 |
```
rbbratta, line 49: so we have to remember to use `[]` for IPv6 when used as a URL, unless we create an `ipv6_url` method.  Or maybe a general `to_url` method.

features/networking/ipv6_dualstack.feature
```diff
@@ -0,0 +1,78 @@
...
+    Then the step should succeed
+    And the output should contain "Hello"
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 5 | [<%= cb.p1pod1ipv6 %>]:8080 |
+    Then the step should succeed
+    And the output should contain "Hello"
+
+    #access pod by service ipv4 and ipv6 address
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 5 | <%= cb.service_ipv4 %>:27017 |
```
rbbratta, line 44: I guess what I am suggesting for ports is that we get in the habit of using `service.ports[0]["port"]` to automatically use the first port from the service spec.

Maybe we need a `.port_index` method.

```ruby
    def port_index(user: nil, num:, cached: true, quiet: false)
      return rr.dig('spec', 'ports')[num]["port"]
    end
```
```ruby
| curl | --connect-timeout | 5 | <%= cb.service_ipv4 %>:<%= service.port_index(0) %> |
```

lib/openshift/service.rb
```diff
@@ -80,6 +80,21 @@ def ip(user: nil, cached: true, quiet: false)
...
+      return ipv6
+    end
+
+    # return service ipv4 address for dualstack cluster
+    def ip_v4(user: nil, cached: true, quiet: false)
+      spec = raw_resource(user: user, cached: cached, quiet: quiet).dig('spec')
+      ipv4 = spec['clusterIPs'].find { |ip| ip.include? "." }
+      return ipv4
```
rbbratta, line 95: can we inline the return value

```ruby
    # return service ipv4 address for dualstack cluster
    def ip_v4(user: nil, cached: true, quiet: false)
      spec = raw_resource(user: user, cached: cached, quiet: quiet).dig('spec')
      return spec['clusterIPs'].find { |ip| ip.include? "." }
    end

```

lib/openshift/service.rb
rbbratta: What do we think about ...

```ruby

    # return service ipv6 address as URL for dualstack cluster
    def ip_v6_url(user: nil, cached: true, quiet: false)
      return"[#{ip_v6(user: user, cached: cached, quiet: quiet)}]"
    end
```

features/networking/ipv6_dualstack.feature
```diff
@@ -0,0 +1,78 @@
...
+    #create multitetant policy in project 1
+    Given I obtain test data file "networking/networkpolicy/multitenant_policy.yaml"
+    When I run the :create admin command with:
+      | f | multitenant_policy.yaml |
+      | n | <%= cb.proj1 %>         |
+    Then the step should succeed
+
+
+    When I execute on the "<%= cb.p2pod1 %>" pod:
+      | curl | --connect-timeout | 5 | <%= cb.p1pod1ipv4 %>:8080 |
```
rbbratta, line 63: ah, so this is the targetPort?  We should make it clear by using the values from the resource.

```ruby
| curl | --connect-timeout | 5 | <%= cb.p1pod1ipv4 %>:<%= service.ports[0]["targetPort"] %> |
```

features/networking/ipv6_dualstack.feature
zhaozhanqi: thanks KK, updated.

lib/openshift/pod.rb
zhaozhanqi: > Should we also check for presence of `[]` in the address? Usually IP6 addr has `[]` right?

ipv6 do not have `[]`, However if we want to access it, it should be like `curl [fd01:0:0:3::5e]:8080`

lib/openshift/service.rb
zhaozhanqi: add one function ip_v6_url, see
```
    def ip_v6_url(user: nil, cached: true, quiet: false)
      ipv6 = self.ip_v6(user: user, cached: cached, quiet: quiet)
      port = self.ports(user: user, cached: true, quiet: quiet)[0]["port"]
      "[#{ipv6}]:#{port}"
    end
```

features/networking/ipv6_dualstack.feature
zhaozhanqi: yes, this should be more stable.  updated this, thanks.

features/networking/ipv6_dualstack.feature
zhaozhanqi: update the ip_v6_url:
```
    def ip_v6_url(user: nil, cached: true, quiet: false)
      ipv6 = self.ip_v6(user: user, cached: cached, quiet: quiet)
      port = self.ports(user: user, cached: true, quiet: quiet)[0]["port"]
      "[#{ipv6}]:#{port}"
    end
```

testdata/networking/list_for_pods_ipv6.json
zhaozhanqi: 30000-32767 is for nodeport scope,  Here we use 'clusterIp' type

testdata/networking/networkpolicy/multitenant_policy.yaml
```diff
@@ -0,0 +1,55 @@
...
+spec:
+  podSelector:
+  ingress: []
+
+--- 
+
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-from-openshift-ingress
```
weliang1, line 16: Just curious, if networkpolicy/ingress not allow ingress and monitoring, what will happen in the cluster?

zhaozhanqi: if that. route will cannot be accessed and all pods 
openshift-monitoring cannot be accessed the test pods, means prometheus cannot collect the data of this test pods

weliang1: /lgtm

zhaozhanqi: if that. route will cannot be accessed and all pods 
openshift-monitoring cannot be accessed the test pods, means prometheus cannot collect the data of this test pods

PR #1931, opened 2021-03-24
Add test files for fibrechannel OCP-15499

chao007:
Adding test files for OCP-15499


testdata/storage/fibrechannel/pvc1.yaml
```diff
@@ -0,0 +1,17 @@
+{
+    "apiVersion": "v1",
+    "kind": "PersistentVolumeClaim",
```
liangxia, line 3: Do we need a pvc for each kind of storage ? Can't we reuse exiting one ?

chao007: Move these test files to cucushift. Close this pr. Thanks.

PR #1937, opened 2021-03-29
Add func to check if cluster is SNO

kasturinarra:
Add func to check if cluster is SNO

### DIFF
--- a/features/step_definitions/env.rb
+++ b/features/step_definitions/env.rb
@@ -76,3 +76,13 @@ Given /^the cluster is running on OpenStack$/ do
   end
 end
 
+Given(/^the cluster is Single Node Openshift$/) do
+  ensure_admin_tagged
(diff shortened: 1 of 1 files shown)


kasturinarra:
@geliu2016 could you please help review the function here ?

geliu2016:
LGTM,thanks

features/step_definitions/env.rb
```diff
@@ -76,3 +76,13 @@
   end
 end
 
+Given(/^I have a single node cluster$/) do
```
liangxia, line 79: Perhaps `Given(/^the cluster is Single Node OpenShift$/)`

The current name make me think it's another cluster beside the current using one.

kasturinarra: Have updated, could you please help check ? Thanks !!

PR #1942, opened 2021-03-31
removing flay tag from logging cases as they're not really flaky

QiaolingTang:
For logging cases, most of the time, they fail due to the packagemanifests are not ready or the limited resources in the cluster. These issues can't be resolved in scenario level. I remove the tag `@flaky` from these cases to make sure we can cover these cases in errata testing and logging CI pipeline.

/cc @anpingli @jhou1

### DIFF
--- a/features/logging/cluster_logging_operator.feature
+++ b/features/logging/cluster_logging_operator.feature
@@ -6,7 +6,6 @@ Feature: cluster-logging-operator related test
   @admin
   @destructive
   @commonlogging
-  @flaky
   Scenario: ServiceMonitor Object for collector is deployed along with cluster logging
     Given I wait for the "fluentd" service_monitor to appear
     Given the expression should be true> service_monitor('fluentd').service_monitor_endpoint_spec(server_name: "fluentd.openshift-logging.svc").port == "metrics"
@@ -24,7 +23,6 @@ Feature: cluster-logging-operator related test
   # @case_id OCP-21907
   @admin
   @destructive
-  @flaky
   Scenario: Deploy elasticsearch-operator via OLM using CLI
     Given logging operators are installed successfully
 


anpingli:
@QiaolingTang  @jhou1  Could we run the flaky via cucushift or Runer-v3 job? Does the flaky only can't be executed in CI testing?

anpingli (commented):
/lgtm

liangxia:
@jhou1  Any objections ?

jhou1:
There was a https://github.com/openshift/verification-tests/pull/1957 to cover logging basic tests, and the `@flaky` tag does not impact the test execution in logging ci. Is the removal still necessary?
 
To view it as a whole(test env, test scenario combined), just the fact the test fails frequently is caused by the poor test env doesn't really mean this test is not flaky, the end result that a test keeps failing is the reason why we gate it in the automation. Having a `@flaky` tag here is not because we urge you to fix the test itself, we wanted to encourage you to improve the underlying issues by all means, which eventually contributes to the success of continuous reliable errata testings.

QiaolingTang:
I found when using Jenkins job `ocp-common/job/Runner` to run automation testing, there has `16:58:27  + bash -l -c 'bundle exec cucumber -p polarshift   --tags "not @flaky"'` even when I set `TEST_MODE` to `all`. I'm not sure if there has same setting in `Runner-v3`.

liangxia:
> I found when using Jenkins job `ocp-common/job/Runner` to run automation testing, there has `16:58:27 + bash -l -c 'bundle exec cucumber -p polarshift --tags "not @flaky"'` even when I set `TEST_MODE` to `all`. I'm not sure if there has same setting in `Runner-v3`.

Runner-v3 is obsoleted, so we did not introduce the new changes. The job will be removed eventually.

anpingli:
@jhou1 @liangxia Could you merge this PR? The other cases pass.  These critical and simple cases must can pass too. while as the flaky tag, they cannot be executed.

anpingli:
@jhou1 By the way we had a jira issue to solve the unstable issue https://issues.redhat.com/browse/OCPQE-4442.

PR #1944, opened 2021-04-02
added methods to support getting daemon set properties

pruan-rht:
helper methods for https://github.com/openshift/cucushift/pull/8479


lihongan:
/lgtm
Thank you @pruan-rht

PR #1949, opened 2021-04-09
Remove OCP-24776 as it's no longer valid

liangxia:
It's blocking batch syncing scripts/tags from code back to Polarion.
```
{"error":"some test cases do not exist","description":"cases do not exist: OCP-24776"}
```

/cc @zhaozhanqi @jhou1


zhaozhanqi:
/lgtm
thanks

PR #1950, opened 2021-04-09
Scenario name change to avoid conflict

asood-rh:
Change the post upgrade check scenario description for test case OCP-38751 to resolve conflict another scenario name.

@zhaozhanqi 
@pruan-rht


asood-rh:
Peter, this scenario name change only, can it be reviewed and merged.

/assign @pruan-rht

pruan-rht (commented):
@asood-rh there's an extra space at the end of line 168...also, did you update Polarion?

asood-rh:
@pruan-rht space is fixed. It is fixed in polarion as well in the automation script field.

pruan-rht (commented):
/lgtm

PR #1952, opened 2021-04-12
Fix scheduler

kasturinarra:
With the existing code, cannot restore the cluster object to the right state if any additional child elements like tlsSecurityProfile is added to the spec, so modified the code a little to handle the same.

### DIFF
--- a/features/step_definitions/scheduler.rb
+++ b/features/step_definitions/scheduler.rb
@@ -21,7 +21,14 @@ Given /^the CR #{QUOTED} named #{QUOTED} is restored after scenario$/ do |crd, n
   patch_json = org_scheduler.to_json
   _admin = admin
   teardown_add {
-    opts = {resource: 'scheduler', resource_name: name, p: patch_json, type: 'merge' }
+    # Added code to support removal of tlsSecurityProfile while restoring
+    @result = admin.cli_exec(:get, resource: crd, resource_name: name, o: 'yaml')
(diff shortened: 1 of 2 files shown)


kasturinarra:

zhouying7780:
@liangxia Could you help review and merge , thanks  .

features/step_definitions/scheduler.rb
```diff
@@ -21,7 +21,14 @@
   patch_json = org_scheduler.to_json
   _admin = admin
   teardown_add {
-    opts = {resource: 'scheduler', resource_name: name, p: patch_json, type: 'merge' }
+    # Added code to support removal of tlsSecurityProfile while restoring
+    @result = admin.cli_exec(:get, resource: crd, resource_name: name, o: 'yaml')
+    if @result[:success] and @result[:parsed]['spec']['tlsSecurityProfile']
+      patch_json = [{"op": "remove","path": "/spec/tlsSecurityProfile"}].to_json
+      opts = {resource: crd, resource_name: name, p: patch_json, type: 'json' }
```
liangxia, line 28: We only update CRD in this case, or we want to update CRD first, then scheduler later ?

kasturinarra: @liangxia the existing function did not work to update the CRD when the spec contains 'tlsSecurityProfile'. so modified the func to get that working. In this case we only update the CRD.

liangxia: Ack

PR #1955, opened 2021-04-13
Add tool to sync tags back to polarion in batch

liangxia:
@jhou1 @duanwei33 

The usage is `[verification-tests]$ bash  tools/sync_tc_tags_to_polarion.sh`


jhou1:
Thanks for having this Liang! LGTM

PR #1957, opened 2021-04-13
remove flack and destructive cases

anpingli:
For some design limitation, all logging cases are belong to destructive cases. thus logging cases may be skipped in some testing.
For some known issues(resource limitation, unstable mirror registry in disconnect cluster,  no mirror images in some cluster), some critical logging cases are tagged as flaky and won't be executed in our automated framework. 

To minimal the risk  in openshift and openshift-logging testing.  We want to make this acceptance case to be the only one non-destructive case and can be running one any platform. As this is the only one case in no-destructive bucket.  I suggest we didn't  flaky it unless it block the other cases.  Logging QE will treat this case failure as high priority.  We will continue to enhance it and  integrate more feature into it.

@QiaolingTang  @gkarager  @wsun1  @jhou1  @liangxia   Looking forward to your idea.

### DIFF
--- a/features/logging/logging_acceptance.feature
+++ b/features/logging/logging_acceptance.feature
@@ -3,8 +3,6 @@ Feature: Logging smoke test case
   # @author gkarager@redhat.com
   # @case_id OCP-37508
   @admin
-  @destructive
-  @flaky
   Scenario: One logging acceptance case for all cluster
 # Deploy cluster-logging operator via web console
     Given logging channel name is stored in the :logging_channel clipboard	
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -196,7 +196,6 @@ end
 
 # to cleanup OLM installed clusterlogging
 Given /^logging service is removed successfully$/ do
-  ensure_destructive_tagged
   ensure_admin_tagged
   # remove namespace
   clo_proj_name = "openshift-logging"


jhou1:
Test cases got a `@flaky` tag not because they are destructive, but because they have been failing most of the time, which made us believe they were either not suitable for automation, or the automation needs rework. The [data](https://docs.google.com/spreadsheets/d/1RJN4QhuRj3lL0qzee6d8Wn0Ie1FYO3T6JBXG2zeIZy0/edit?ts=606c3e75#gid=0) showed it had passed twice and failed 81 times, and it'd be better to have this test executed manually.

If you'd like to keep it automated and inspect the failures every time, it's fine by me to remove this tag as I don't want to block your workflow. I'd appreciate enhancements being made to the mirror/resource limitations.

anpingli:
@jhou1   I'd like to delete flaky tag from it to see if  that can be more stable than before.

anpingli:
My plan to design one or two acceptance cases for OCP Nightly/Upgrade testing in future. the other cases can be covered in Logging Pipeline.

features/logging/logging_acceptance.feature
```diff
@@ -3,8 +3,6 @@ Feature: Logging smoke test case
   # @author gkarager@redhat.com
   # @case_id OCP-37508
   @admin
-  @destructive
```
QiaolingTang, line 6: @anpingli The `@destructive` tag can't be removed because the scenario calls https://github.com/openshift/verification-tests/blob/master/features/logging/logging_acceptance.feature#L11, and this step must have `@destructive` label: https://github.com/openshift/verification-tests/blob/master/features/step_definitions/logging.rb#L199

anpingli: @QiaolingTang  What does ensure_destructive_tagged do? Could we delete it in logging.rb#L199?

QiaolingTang: @anpingli same as `@destructive`:  https://github.com/openshift/verification-tests/blob/master/doc/admin_features.adoc#tags
I added this method to some steps because deploy/remove operators/clusterlogging can affect all the logging cases. From my side, removing this method won't affect anything as all the logging scenarios are tagged with `@destructive`. 
@pruan-rht  any thoughts about removing `ensure_destructive_tagged` from https://github.com/openshift/verification-tests/blob/master/features/step_definitions/logging.rb#L199 and https://github.com/openshift/verification-tests/blob/master/features/step_definitions/logging.rb#L31?

jhou1: It's good to have `ensure_destructive_tagged` in the step definitions if the step is destructive to your other tests

anpingli: I think that is OK to remove this tag. First:  the step is only called by this acceptance case.  Second:  We have destructive tag for all senearios expect this one. Third: we are running caess in sequnce in Logging Pipeline. so that won't make trouble to Logging testing.

pruan-rht (commented):
/lgtm

PR #1960, opened 2021-04-15
Fix small mtu test

kedark3:
# Purpose

Developers pointed out in the bz https://bugzilla.redhat.com/show_bug.cgi?id=1930446 that we should not need to use NetworkManager during modification of MTU. In order to align with that, I updated the test case.

# Current state

The Test works as expected, it is able to modify the MTU successfully on a new cluster. But as soon as we modify the MTU, the cluster breaks and starts acting weird(sometimes, more likely in the case where your test picks master node to modify mtu). What that means is we cannot actually reliably execute the test successfully. The bug associated with the test is open and awaiting fix. https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/ocp-common/job/Runner/28355/console

cc @rbbratta @weliang1 @asood-rh @zhaozhanqi @huiran0826  @brahaney @pruan-rht

### DIFF
--- a/features/networking/service.feature
+++ b/features/networking/service.feature
@@ -434,7 +434,7 @@ Feature: Service related networking scenarios
     And the output should contain "Hello OpenShift"
     """
 
-  # @author anusaxen@redhat.com, kkulkarni@redhat.com
+  # @author anusaxen@redhat.com
   # @case_id OCP-24694
   @admin
   @destructive
@@ -469,8 +469,8 @@ Feature: Service related networking scenarios
     #Reset the MTU using nmcli
     Given I use the "<%= cb.subject_node %>" node
     And I run commands on the host:
-    | nmcli con modify "<%= cb.nmcli_active_con_uuid %>" ethernet.mtu <%= cb.mtu_actual %> |
-    | nmcli dev reapply  <%= cb.default_interface %> |
+      | nmcli con modify "<%= cb.nmcli_active_con_uuid %>" ethernet.mtu <%= cb.mtu_actual %> |
+      | nmcli dev reapply  <%= cb.default_interface %> |
     Then the step should succeed
     And the node's MTU value is stored in the :mtu_actual_redeployed clipboard
     And the expression should be true> cb.mtu_actual == cb.mtu_actual_redeployed
--- a/features/networking/service.feature
+++ b/features/networking/service.feature
@@ -434,23 +434,25 @@ Feature: Service related networking scenarios
     And the output should contain "Hello OpenShift"
     """
 
-  # @author anusaxen@redhat.com
+  # @author anusaxen@redhat.com, kkulkarni@redhat.com
   # @case_id OCP-24694
   @admin
   @destructive
   Scenario: Taint node with too small MTU value
     Given the default interface on nodes is stored in the :default_interface clipboard
     And the node's MTU value is stored in the :mtu_actual clipboard
+    And the node's active nmcli connection is stored in the :nmcli_active_con_uuid clipboard
     And evaluation of `node.name` is stored in the :subject_node clipboard
     And I run commands on the host:
-      | systemctl stop NetworkManager                        |
-      | ip link set mtu 1300 dev <%= cb.default_interface %> |
+      | nmcli con modify  "<%= cb.nmcli_active_con_uuid %>"  ethernet.mtu  1300  |
+      | nmcli dev reapply <%= cb.default_interface %> |
     Then the step should succeed
     Given I register clean-up steps:
     """
     Given I use the "<%= cb.subject_node %>" node
     And I run commands on the host:
-      | systemctl start NetworkManager |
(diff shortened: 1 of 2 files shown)


features/step_definitions/networking.rb
```diff
@@ -716,7 +716,7 @@
   end
   # OVN uses `br-ex` and `-` is not a word char, so we have to split on whitespace
   cb[cb_name] = @result[:response].split("\n").first.split[4]
-  logger.info "The node's default interface is stored in the #{cb_name} clipboard."
+  logger.info "The node's default interface is stored in the #{cb_name} clipboard as #{cb[cb_name]}."
```
kedark3, line 719: adding `as <variable>` to the logger. This helps readability of the logs to determine if the correct value is being saved to the clipboard. Very much improvement in Quality of Life with this addition.

features/networking/service.feature
```diff
@@ -464,10 +466,11 @@ Feature: Service related networking scenarios
     Then the step should succeed
     And the output should contain "mtu-too-small"
     """
-    #Starting NetworkManager to roll out original system MTU
+    #Reset the MTU using nmcli
     Given I use the "<%= cb.subject_node %>" node
     And I run commands on the host:
-      | systemctl start NetworkManager |
+    | nmcli con modify "<%= cb.nmcli_active_con_uuid %>" ethernet.mtu <%= cb.mtu_actual %> |
```
pruan-rht, line 472: indentation

features/step_definitions/networking.rb
```diff
@@ -1303,3 +1303,12 @@
   default_network = network_operator.default_network(user: admin)
   raise "env doesn't have IPSec enabled" unless default_network["ipsecConfig"]
 end
+
+Given /^the node's active nmcli connection is stored in the#{OPT_SYM} clipboard$/ do |cb_name|
+  ensure_admin_tagged
+  cb_name = "nmcli_active_con_uuid" unless cb_name
```
pruan-rht, line 1309: @kedark3 perhaps shorten the default cb_name

features/networking/service.feature
```diff
@@ -434,23 +434,25 @@ Feature: Service related networking scenarios
     And the output should contain "Hello OpenShift"
     """
 
-  # @author anusaxen@redhat.com
+  # @author anusaxen@redhat.com, kkulkarni@redhat.com
```
pruan-rht, line 437: @kedark3 not sure adding two authors like this will impact any existing script that parses `@author` field.

kedark3: I can remove it, I added it so that reader could know who all touched the test, but that also could be done in git blame. Ok if I remove my name?

features/networking/service.feature
kedark3: fixing.

features/step_definitions/networking.rb
kedark3: ok I will make it shorter :)

features/networking/service.feature
pruan-rht: @kedark3 yeah, or you can replace Anurag as the author while he's on leave. :). Yeah, git-blame will do give you credit

asood-rh (commented):
@kedark3 Could you include runner job that shows test executes fine with changes?

features/networking/service.feature
```diff
@@ -441,16 +441,18 @@ Feature: Service related networking scenarios
   Scenario: Taint node with too small MTU value
     Given the default interface on nodes is stored in the :default_interface clipboard
     And the node's MTU value is stored in the :mtu_actual clipboard
+    And the node's active nmcli connection is stored in the :nmcli_active_con_uuid clipboard
     And evaluation of `node.name` is stored in the :subject_node clipboard
     And I run commands on the host:
-      | systemctl stop NetworkManager                        |
-      | ip link set mtu 1300 dev <%= cb.default_interface %> |
+      | nmcli con modify  "<%= cb.nmcli_active_con_uuid %>"  ethernet.mtu  1300  |
+      | nmcli dev reapply <%= cb.default_interface %> |
```
kedark3, line 448: @asood-rh test only runs till this line and then it fails and there is a bug open for the same.  https://bugzilla.redhat.com/show_bug.cgi?id=1930446

features/networking/service.feature
```diff
@@ -441,16 +441,18 @@ Feature: Service related networking scenarios
   Scenario: Taint node with too small MTU value
     Given the default interface on nodes is stored in the :default_interface clipboard
```
asood-rh, line 442: nit - node is singular here

asood-rh: nit, ok to ignore

kedark3: Thanks Arti. As discussed offline, since its not a line I touched in this PR, I would like to not change it, as it would widen the scope of the PR, having to update the corresponding step definition. There are 16 other tests I can see using this same `incorrect` plural.

features/networking/service.feature
kedark3: I removed my name :) np.

asood-rh (approved):
/lgtm

PR #1962, opened 2021-04-16
Auto OCP-41132 OCP-41179

huiran0826:
Auto OCP-41132 OCP-41179

Logs:
https://privatebin-it-iso.int.open.paas.redhat.com/?748c12a2010d55a1#23GZVV2RSSVsWMN9Us3SJ4Cxmfokj9GgyTPXkb8WMk17
https://privatebin-it-iso.int.open.paas.redhat.com/?9b1b8a78f8f6e10b#BDZzPtVW6ta8Fpqkyi2oQPSv4s9FzcQ8CUfP3B3qvuey

Could you help take a look? Thanks!
@rbbratta @weliang1 @zhaozhanqi  @asood-rh @kakkoyun @brahaney

### DIFF
--- a/testdata/networking/ovn-egressfirewall/egressfirewall-policy5.yaml
+++ b/testdata/networking/ovn-egressfirewall/egressfirewall-policy5.yaml
@@ -1,0 +1,72 @@
+apiVersion: k8s.ovn.org/v1
+kind: EgressFirewall
+metadata:
+  name: default
+spec:
+  egress:
+  - type: Allow
+    to:
+      cidrSelector: 192.168.0.0/16
+  - type: Allow
+    to:
+      cidrSelector: 10.188.28.0/24
+  - type: Allow
+    to:
+      cidrSelector: 10.9.0.0/16
+  - type: Allow
+    to:
+      dnsName: yahoo.com
+  - type: Allow
+    to:
+      dnsName: github.com
+  - type: Allow
+    to:
+      dnsName: www.baidu.com
+  - type: Allow
+    to:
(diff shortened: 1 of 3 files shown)


testdata/networking/ovn-egressfirewall/egressfirewall-policy5.yaml
```diff
@@ -0,0 +1,72 @@
...
+    to:
+      dnsName: api.segment.io
+  - type: Allow
+    to:
+      dnsName: login.ibm.com
+  - type: Allow
+    to:
+      dnsName: marketscan2.truvenhealth.com
+  - type: Allow
+    to:
```
zhaozhanqi, line 65: Above Allow dnsName is needed for this testing? saw case only use 'www.amihealthy.com'

zhaozhanqi, line 65: Above Allow dnsName is needed for this testing? saw case only use 'www.amihealthy.com'

huiran0826: In the original issue, only first 5 dnsName rules were applied, and no status displaying when "oc get egressfirewall".  I don't think we need to check each dnsName for the curl result, so here just check egressfirewall status to make sure the egressfirewall policy was applied and the last dnsName curl result was correct.

huiran0826: Same as above.

PR #1964, opened 2021-04-20
New test case for Node Tuning Operator OCP-27491

skordas:
New test case for Node Tuning Operator:
Node tuning operator: tuning is working - add profile


features/node/nto.feature
```diff
@@ -0,0 +1,98 @@
+Feature: Node Tuning Operator related scenarios
+
+Background: Cleaning before if some previous test run failed
```
pruan-rht, line 3: @skordas you don't need to do this...what are you trying to accomplish?

features/node/nto.feature
```diff
@@ -0,0 +1,98 @@
...
+  And I run the :delete admin command with:
+      | object_type       | tuned                                 |
```
pruan-rht, line 5: @skordas 2-spaces indentation for cucumber steps

skordas: Fixed

features/node/nto.feature
skordas: Please check line 34
it will create custom resource (new tuned) which will be removed in line 66
if test will fail (somewhere between) - then any other scenario which is using the same tuned will failed (leftovers from other tuned tests). This is first one, but it will be more :)

features/node/nto.feature
pruan-rht: @skordas we don't use the `Background` directive at all.  I don't know how the system will react to this approach.  If you just want to make sure the object `tuned` is gone, there's an [existing step](https://github.com/openshift/verification-tests/blob/master/features/step_definitions/resources.rb#L128) to do that already.

features/node/nto.feature
```diff
@@ -0,0 +1,98 @@
...
+      | resource     | node/<%= cb.pod_node %>        |
+      | oc_opts_end  |                                |
+      | exec_command | chroot                         |
+      | exec_command | /host                          |
+      | exec_command | sysctl                         |
+      | exec_command | net.netfilter.nf_conntrack_max |
+    And the output should match:
+      | net.netfilter.nf_conntrack_max = 1048578 |
+    # Removing custom tuned profile
+    When I run the :delete admin command with:
```
pruan-rht, line 66: you can use the step https://github.com/openshift/verification-tests/blob/master/features/step_definitions/resources.rb#L128

skordas:
@pruan-rht PTAL

features/node/nto.feature
```diff
@@ -0,0 +1,92 @@
+Feature: Node Tuning Operator related scenarios
+
+  # @author skordas@redhat.com
+  # @case_id OCP-27491
+  @admin
```
pruan-rht, line 5: @skordas I think this is a destructive test?

features/node/nto.feature
```diff
@@ -0,0 +1,92 @@
...
+    And I run the :logs admin command with:
+      | resource_name | <%= cb.tuned_pod %>                    |
+      | n             | openshift-cluster-node-tuning-operator |
+      | tail          | 10                                     |
+    And the output should match:
+      | tuned.daemon.daemon: static tuning from profile 'nf-conntrack-max' applied |
+    """
+    # Profile verification
+    And I run the :get admin command with:
+      | resource      | profile                                |
```
pruan-rht, line 47: @skordas I see two calls to `profile` verification, which suggest we should make that an openshift object by adding `lib/openshift/profile.rb`

lib/openshift/tuned.rb
```diff
@@ -0,0 +1,7 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class Tuned < ClusterResource
+    RESOURCE = 'tuned'
```
pruan-rht, line 5: we pluralized the object `RESOURCE`. For this case it would be `RESOUCE='tuneds'`

features/node/nto.feature
```diff
@@ -0,0 +1,92 @@
...
+    """
+    And I run the :logs admin command with:
+      | resource_name | <%= cb.tuned_pod %>                    |
+      | n             | openshift-cluster-node-tuning-operator |
+      | tail          | 10                                     |
+    And the output should match:
+      | tuned.daemon.daemon: static tuning from profile 'openshift-node' applied |
+    """
+    # Profile verification
+    And I run the :get admin command with:
```
pruan-rht, line 76: @skordas once the `lib/openshift/profile.rb` is created with the proper method(s) defined, you can replace lines 76-82 with ```And the expression should be true> profile(cb.pod_node).tuned_profile(cached: false) == 'openshift-node'```

features/node/nto.feature
pruan-rht: if you have Profile class defined properly, then you can replace lines 46-52 with the following
`And the expression should be true> profile(cb.pod_node).tuned_profile == 'nf-conntrack-max'`

features/node/nto.feature
skordas: No, is not :)

Even test will pass or fail, cluster should be 100% operational. Additional step `Given admin ensures "nf-conntrack-max" tuned is deleted from the "openshift-cluster-node-tuning-operator" project after scenario` should remove custom NTO resource after test.
NTO is a Second Line Operator - OpenShift will work correctly even NTO will be deactivated.
In worse case scenario - kernel in one of working node will be working with different `nf-conntrack-max` value (and is only 2 bytes from 1Mb)

lib/openshift/profile.rb
```diff
@@ -0,0 +1,13 @@
+require 'openshift/cluster_resource'
+
+module BushSlicer
+  class Profile < ClusterResource
+    RESOURCE = 'profiles'
+
+    def tuned_profile(user: nil, cached: false, quiet: false)
```
pruan-rht, line 7: @skordas, `cached: true` should be used since this is not a dynamic element, it's a spec.

features/node/nto.feature
```diff
@@ -0,0 +1,80 @@
...
+    """
+    And I run the :logs admin command with:
+      | resource_name | <%= cb.tuned_pod %>                    |
+      | n             | openshift-cluster-node-tuning-operator |
+      | tail          | 10                                     |
+    And the output should match:
+      | tuned.daemon.daemon: static tuning from profile 'openshift-node' applied |
+    """
+    # Profile verification
+    And the expression should be true> profile(cb.pod_node).tuned_profile == 'openshift-node'
```
pruan-rht, line 70: @skordas if we do agree to use `cache: true` for the method `tuned_profile`, then you need to update this like to `And the expression should be true> profile(cb.pod_node).tuned_profile(cached: false) == 'openshift-node'` in order to refresh the cached item from the earlier call in line 46.

features/node/nto.feature
pruan-rht: so, as long as we can run them in parallel.

lib/openshift/profile.rb
skordas: Changed as suggested.

pruan-rht:
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3087/console
/lgtm

PR #1969, opened 2021-04-22
Customize the function so that it can be used with any CR

kasturinarra:
with the previous way the function can only be used with kube-scheduler, but with the change in this PR we should be able to be use it with any CR.


kasturinarra:

zhouying7780:
@liangxia Could you please help review and merge , thanks.

PR #1971, opened 2021-04-25
kata automation support for 4.8 release

pruan-rht:
@liangxia @cpmeadors PTAL

### DIFF
--- a/lib/openshift/kata_config.rb
+++ b/lib/openshift/kata_config.rb
@@ -6,7 +6,7 @@ module BushSlicer
     RESOURCE = 'kataconfigs'
 
     def install_completed_node_count(user: nil, cached: true, quiet: false)
-      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'installationStatus', 'completed', 'completedNodesList')
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'installationStatus', 'completed', 'completedNodesList')&.count || 0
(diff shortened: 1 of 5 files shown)


lib/openshift/kata_config.rb
```diff
@@ -6,7 +6,12 @@ class KataConfig < ClusterResource
     RESOURCE = 'kataconfigs'
 
     def install_completed_node_count(user: nil, cached: true, quiet: false)
-      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'installationStatus', 'completed', 'completedNodesCount')
+      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'installationStatus', 'completed', 'completedNodesList')
```
akostadinov, line 9: Maybe needed to use `#count` here.

```
      raw_resource(user: user, cached: cached, quiet: quiet).dig('status', 'installationStatus', 'completed', 'completedNodesList')&.count || 0
```

akostadinov: Otherwise seems good.

pruan-rht: @akostadinov, good suggestion, while doing this PR, it actually discovered a [regression](https://issues.redhat.com/browse/KATA-730).  So this method will go back to `completeNodesCount` eventually.

pruan-rht:
@akostadinov ok, updated.  Thx again!

PR #1973, opened 2021-04-26
fix OCP-41179

huiran0826:
When creating the EgressFirewall, it may takes some time to generate the rule, but in auto run, the step is fast, so add timeout here to wait for the rule generated and status updated.

And run it multiple times for the fix and it passed.
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3067/console

@zhaozhanqi @anuragthehatter @rbbratta @weliang1


pruan-rht (commented):
/lgtm

PR #1976, opened 2021-04-27
Added CLI OLM installation for Kata and other fixes.

pruan-rht:
1. added support for  OLM CLI installation
2. removed 4.6 and 4.7 support.
3. removed and moved around some input YAML


pruan-rht:
@cpmeadors per our slack discussion, I've removed the 4.6 & 4.7 OCP support.  I've added and tested the CLI installation as well as moved some files around.  @liangxia  PTAL as well.  Thx!

pruan-rht:
@liangxia PTAL when you have a chance.  thx!

features/step_definitions/kata.rb
```diff
@@ -109,26 +86,22 @@
...
+  kata_ns ||= "sandboxed-containers-operator-system"
+  kata_config_name = "example-kataconfig"
+  step %Q/I store master major version in the :master_version clipboard/
+  cb.channel = cb.master_version
+  raise "Kata operator OLM installation only supported for OCP >= 4.8" unless cb.master_version >= "4.8"
   install_method ||= 'CLI'
   if install_method == 'GUI'
     package_name = 'kata-operator'
-    if cluster_ver >= 4.8
+    if cb.master_version >= "4.8"
```
liangxia, line 97: Duplicate check with line 93 ?

pruan-rht: @liangxia not really, just putting a check there just in case 4.9 or something is different.

pruan-rht: @liangxia that's the get the `catalog_name` to be versioned... technically we don't really need this check since we are saying only  OCP 4.8 or greater is supported...that's just for incase 4.9 or something will have a different name...probably not though.  I can take out the if statement

pruan-rht: @liangxia ok, took that `if` statement check.  PTAL

liangxia:
LGTM

pruan-rht:
@liangxia can you pls merge it?  Cameron has reviewed the previous PR and it was his suggestion to take out the 4.6/4.7 support.

PR #1978, opened 2021-04-27
Update pv_metering.yaml

pruan-rht:
metering was failing with the sharedPVC profile which turns out to be no enought PVCs.  Adding more to satisfy the requirements.


pruan-rht:
thanks again @duanwei33 for your suggesitions!

PR #1980, opened 2021-04-27
OVN: error if no leader

rbbratta:
On Single Node OpenShift there is no OVN leader, `leader_id` is nil.

Instead of

```
NoMethodError: undefined method `[]' for nil:NilClass
```
use the regular error message

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1043,7 +1043,7 @@ Given /^I store the ovnkube-master#{OPT_QUOTED} leader pod in the#{OPT_SYM} clip
   raise "Failed to execute network command!" unless cluster_state != nil
   # for some reason "oc rsh" output contains CR, so we have to remove them
   leader_id = cluster_state.match(/Leader:\s+(\S+)/)
-  if leader_id[1] == "unknown"
+  if leader_id.nil? || leader_id[1] == "unknown"
     raise "Unknown leader"
   end
   servers = cluster_state.match(/Servers:\n(.*)/m)


rbbratta:
needs re-test on SNO.

pruan-rht (commented):
/lgtm

rbbratta:
on SNO the leader will still be `self`, except when we kill the leaders during the Raft stress, then there will temporarily be no leader.

```
    Given I store the ovnkube-master "north" leader pod in the clipboard                                      # features/step_definitions/networking.rb:1006
      [12:18:24] INFO> Shell Commands: oc exec ovnkube-master-n6fhj  --kubeconfig=ocp4_admin.kubeconfig -n openshift-ovn-kubernetes  --container=northd -i -- ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound
      f960
      Name: OVN_Northbound
      Cluster ID: 2955 (2955d841-b6c7-4cff-9d69-58aa04e0aee0)
      Server ID: f960 (f960d290-22c6-47fc-a331-6cadf371b717)
      Address: ssl:10.0.0.5:9643
      Status: cluster member
      Role: leader
      Term: 2
      Leader: self
      Vote: self

      Last Election started 88728823 ms ago, reason: timeout
      Last Election won: 88728823 ms ago
      Election timer: 10000
      Log: [6685, 12133]
      Entries not yet committed: 0
      Entries not yet applied: 0
      Connections:
      Disconnections: 0
```

But it's still better to have the specific error instead of `NilClass` when there is no leader.

kedark3 (approved):

PR #1986, opened 2021-04-30
[hotfix] for scenario skipping fix

akostadinov:
fix for #1985

### DIFF
--- a/tools/openshift-ci/Dockerfile.centos
+++ b/tools/openshift-ci/Dockerfile.centos
@@ -1,4 +1,4 @@
-FROM centos:7
+FROM quay.io/centos/centos:7
 
 LABEL vendor="Red Hat inc."
 LABEL maintainer="AOS QE Team"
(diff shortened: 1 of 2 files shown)


tools/openshift-ci/Dockerfile.centos
```diff
@@ -1,4 +1,4 @@
-FROM centos:7
+FROM quay.io/centos/centos:7
```
akostadinov, line 1: This is to avoid

> build error: failed to pull image: After retrying 2 times, Pull image still failed due to error: while pulling "docker://centos:7" as "centos:7": Error initializing source docker://centos:7: Error reading manifest 7 in docker.io/library/centos: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

I'm not sure anybody is using this image but anyway.

PR #1987, opened 2021-05-04
backporting of Cucumber 5, ReportPortal and other recent changes

akostadinov:
@liangxia, I didn't test this change. I just went through recent pull requests
and added what I identifyied as necessary to backport + some conflict fixes.

HTH


liangxia:
I had several try today, but interrupted by Jenkins instability. I'll continue to test it tomorrow.

akostadinov:
This will not try to backport scenario and step regexp fixes because discrepancy between master and v3 branch is too big in that regard.

PR #1988, opened 2021-05-05
Fix image name

kasturinarra:
With the previous image the test case was hitting docker pull limit, so changing the image so that it can be pulled from quay.io


pruan-rht (commented):
/lgtm

PR #1990, opened 2021-05-06
Use OCP4 by default and as examples

liangxia:
master branch is for OCP 4, so use OCP4 by default, and in the examples.

/cc @pruan-rht @jhou1 @JianLi-RH


jhou1:
/lgtm
Thanks for having this update!

PR #1991, opened 2021-05-06
[WIP] fixes transform in which the arg is an Integer

pruan-rht:
still need more testing.

### DIFF
--- a/config/cucumber.yml
+++ b/config/cucumber.yml
@@ -18,7 +18,7 @@
   end
 %>
 
-default: --profile devel
+default: --publish-quiet --profile devel
 debug: --profile _debug --profile devel
 devel: --profile _devel <%= default_args %>
 tcms: BUSHSLICER_TEST_CASE_MANAGER=tcms <%= tcms_args %>
--- a/lib/world.rb
+++ b/lib/world.rb
@@ -482,6 +482,8 @@ module BushSlicer
         table( x.raw.map { |row| row.map { |cell| transform_value(cell) } } )
       elsif x.respond_to? :gsub
         x.gsub(/<%=(.+?)%>/m) { |c| eval $1 }
+      elsif x.respond_to? :to_s
+        x.to_s
       else
         raise ArgumentError, "Unexected argument: #{x.inspect}"
       end


config/cucumber.yml
```diff
@@ -18,7 +18,7 @@
   end
 %>
 
-default: --profile devel
+default: --publish-quiet --profile devel
```
akostadinov, line 21: what is this for?

lib/world.rb
```diff
@@ -482,6 +482,8 @@ def transform_value(x)
         table( x.raw.map { |row| row.map { |cell| transform_value(cell) } } )
       elsif x.respond_to? :gsub
         x.gsub(/<%=(.+?)%>/m) { |c| eval $1 }
+      elsif x.respond_to? :to_s
+        x.to_s
```
akostadinov, line 486: I wonder why we could get anything but a string here. Maybe some default processing by Cucumber 5? And also it doesn't happen always as it seems :/

I wonder if we can disable such cucumber automatic stuff.

config/cucumber.yml
pruan-rht: already done with https://github.com/openshift/verification-tests/pull/1989...so redundant.

lib/world.rb
pruan-rht: @akostadinov it comes down to the step definition.  If the calling step parameter is Integer, that's when we get into trouble.  Example failure that this PR fixes http://ci-qe-openshift.usersys.redhat.com/userContent/cucushift/v3/2021/05/05/04:45:50/PV_PVC_bind_in_a_reasonable_time_when_PVC_is_created_before_PV_while_PVC_pre-bind_to_PV/console.html

pruan-rht: one of the failures this PR won't fix is something like below, which deals with nesting RegExp
```    Given I store all machines in the "openshift-machine-api" project to the :machines clipboard # features/step_definitions/resources.rb:209
      Your block takes 5 arguments, but the Regexp matched 4 arguments. (Cucumber::Glue::ArityMismatchError)
      features/step_definitions/resources.rb:209:in `/^(I|admin) stores? all (\w+)( in the(?: "(.+?)")? project)? to the(?: :(\S+))? clipboard$/'```

Comment (pruan-rht) [reply lib/world.rb:486]: @akostadinov for the above failure, perhaps the step itself needs to be re-written.

Comment (pruan-rht) [reply lib/world.rb:486]: @akostadinov the regexp checks out with rubular.com https://rubular.com/r/KYwSHJjM7hRiYI  so I'm not sure why cuke5 is complaining

Comment (rbbratta) [reply lib/world.rb:486]: There is another failure mode with

https://github.com/openshift/cucushift/blob/master/web/admin-console/home.feature#L60

```
Given /^I repeat the following steps for each #{SYM} in cb\.([\w]+):$/ do |varname, cbsym, steps_str|
  eval_regex = /\#\{(.+?)\}/
  eval_found = steps_str =~ eval_regex
  begin
    logger.dedup_start
    cb[cbsym].each { |x|
      cb[varname] = x
      if eval_found
        steps steps_str.gsub(eval_regex) { |s| "<%= #{$1} %>"}
```

These meta-steps won't work.  `steps_str` is transformed before we get to to `steps_str.gsub`

File: config/cucumber.yml (hunk 1 above)
Comment (akostadinov) [reply config/cucumber.yml:21]: If this is this [1], then I don't think it is a good idea - to publish our logs to a public service.

[1]  https://cucumber.io/blog/open-source/cucumber-reports/

Comment (pruan-rht) [reply config/cucumber.yml:21]: @akostadinov #1989 already disabled printing that marketing stuff.  That line just disables cucumber from printing that marketing text at the end of the runs.

File: lib/world.rb (hunk 2 above)
Comment (rbbratta) [reply lib/world.rb:486]: I guess it's bad syntax.  Should be  `#{}` instead of `<%= %>` in the meta-step?

```ruby
    When I repeat the following steps for each :resource in cb.resources:
    """
    When I perform the :goto_resource_page_under_default_project web action with:
      | resource_url_name   | #{cb.resource.downcase.to_s.delete(" ")+"s"} |
    Then the step should succeed
```

Maybe fixed with https://github.com/openshift/cucushift/pull/8530

akostadinov: Proper syntax. This is to evaluate the expression by each iteration.

akostadinov:
The diff with #1992 is that there we handle only numerics. Here it is more general. Which is better depends on whether you want to make sure you handle expected cases vs catching more cases but virtually impossible to catch any argument type issues.

lib/world.rb
akostadinov: wrt `Your block takes 5 arguments, but the Regexp matched 4 arguments.`, please tell the exact scenario failing. It is very strange why it happens in any case.

pruan-rht:
ok, I'll keep this PR up since there are some conversations about other issues we are seeing... Wil close it out once the other issues have been resolved.

lib/world.rb
akostadinov: @rbbratta , you are correct,  I read your message improperly.

pruan-rht:
@akostadinov @liangxia  what's your opinion about the Regexp failure one?  Rewrite the existing step?  I don't know why the regexp match is different....

lib/world.rb
pruan-rht: @akostadinov this is the step that's failing ```Given I store all machines in the "openshift-machine-api" project to the :machines clipboard```

akostadinov:
@pruan-rht , can you provide exact scenario, I didn't find that.

lib/world.rb
pruan-rht: @akostadinov Rubular ouptut  https://rubular.com/r/KYwSHJjM7hRiYI

pruan-rht:
@akostadinov offending feature features/tierN/web/admin-console/node.feature:92

lib/world.rb
akostadinov: @pruan so any call to the step or only inside a loop step? Do you have exact scenario?

akostadinov: @pruan-rht , ok, it seems like when regular expression matches something from the cucumber's built-in parameter types, then things tend to break. In this situation step definition contained expression `\w+`. So I modified step definition like:
```
Given /^(I|admin) stores? all (#{WORD})( in the#{OPT_QUOTED} project)? to the#{OPT_SYM} clipboard$/
```
And now it seems to catch it. Please try and push such a change. I'll be gone for the rest of the day but will check in the evening for further issue.

i.e. replaced `\w+` with `#{WORD}`

akostadinov:
@pruan-rht , please see my other reply. And summarize any remaining issues in an email or something, I'll check tonight. Sorry, I can't look at stuff more right now. Thanks!

PR #1994, opened 2021-05-08
automate OCP-27615: Clone a pvc with filesystem VolumeMode in the same namespace is successfully	

JianLi-RH:
@jhou1 @liangxia @pruan-rht

### DIFF
--- a/features/storage/csi_clone.feature
+++ b/features/storage/csi_clone.feature
@@ -1,0 +1,44 @@
+Feature: CSI clone testing related feature
+  """
+  This test can only work on a Cluster which has CSI driver deployed, OSP has csi deploy by default from 4.7. 
+  From 4.7, OSP CSI driver was installed by default. Its storage class name is standard-csi.
+  Use "oc get sc" can find all deployed stroage class.
+  """
+  
+  # @author jianl@redhat.com
+  # @case_id OCP-27615
+  Scenario: Clone a PVC and verify data consistency
+    # Step 1
+    Given the master version >= "4.7"
+    Given I have a project
+    # Step 2
+    When I obtain test data file "storage/csi/pod_pvc_sc.yaml"
+    And I run the :create client command with:
+      | f | pod_pvc_sc.yaml |
+    Then the pod named "mypod" becomes ready
+    And the "mypvc" PVC becomes :bound
+    # Step 3
+    Given 1 pod becomes ready with labels:
+      | name=frontendhttp |
+    When I execute on the pod:
+      | sh | -c | echo "clone test" > /mnt/local/testfile |
+    Then the step should succeed
+    # Because the execution is too fast, we have to sync test file to avoid nothing to clone.
+    When I execute on the pod:
+      | sh | -c | sync -f /mnt/local/testfile |
+    Then the step should succeed
+    # Step 4
+    When I obtain test data file "storage/csi/pod_clone.yaml"
+    And I run the :create client command with:
+      | f | pod_clone.yaml |
+    Then the pod named "mypod-clone" becomes ready
+    And the "mypvc-clone" PVC becomes :bound
+    # Step 5
+    Given 1 pod becomes ready with labels:
+      | name=frontendhttp-clone |
+    When I execute on the pod:
+      | sh | -c | cat /mnt/local/testfile |
+    Then the output should contain "clone test"
+    When I execute on the pod:
+      | sh | -c | echo "clone test" > /mnt/local/testfile_2nd |
+    Then the step should succeed
--- a/testdata/storage/csi/pod_clone.yaml
+++ b/testdata/storage/csi/pod_clone.yaml
@@ -1,0 +1,36 @@
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+    name: mypvc-clone
+spec:
+  accessModes:
+  - ReadWriteOnce
+  storageClassName: "standard-csi"
+  resources:
+    requests:
+      storage: 2Gi
+  dataSource:
+    kind: PersistentVolumeClaim
+    name: mypvc
+---
+kind: Pod
+apiVersion: v1
+metadata:
+  name: mypod-clone
+  labels: 
+    name: frontendhttp-clone
+spec:
+  containers:
+    - name: myfrontend
+      image: quay.io/openshifttest/storage@sha256:a05b96d373be86f46e76817487027a7f5b8b5f87c0ac18a246b018df11529b40
+      ports:
+        - containerPort: 80
+          name: "http-server"
+      volumeMounts:
+      - mountPath: "/mnt/local"
+        name: local
+  volumes:
+    - name: local
+      persistentVolumeClaim:
+        claimName: mypvc-clone
--- a/testdata/storage/csi/pod_pvc_sc.yaml
+++ b/testdata/storage/csi/pod_pvc_sc.yaml
@@ -1,0 +1,33 @@
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: mypvc
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 2Gi
+  storageClassName: "standard-csi"
+---
+kind: Pod
+apiVersion: v1
+metadata:
+  name: mypod
+  labels: 
+    name: frontendhttp
+spec:
+  containers:
+    - name: myfrontend
+      image: quay.io/openshifttest/storage@sha256:a05b96d373be86f46e76817487027a7f5b8b5f87c0ac18a246b018df11529b40
+      ports:
+        - containerPort: 80
+          name: "http-server"
+      volumeMounts:
+      - mountPath: "/mnt/local"
+        name: local
+  volumes:
+    - name: local
+      persistentVolumeClaim:
+        claimName: mypvc


features/storage/csi_clone.feature
```diff
@@ -0,0 +1,67 @@
...
+    When I execute on the pod:
+      | sh | -c | touch /mnt/local/testfile |
+    Then the step should succeed
+    """
+    And I wait for the steps to pass:
+    """
+    When I execute on the pod:
+      | sh | -c | echo "clone test" > /mnt/local/testfile |
+    Then the step should succeed
+    """
```
jhou1, line 43: Line 32~42 can be shortened as.
```
And I wait for the steps to pass:
"""
When I execute on the pod:
      | sh | -c | echo "clone test" > /mnt/local/testfile |
Then the step should succeed
"""
```

The `echo` can create the content and the file if the file is not present, thus `touch` is not needed.

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,67 @@
...
+    And I wait for the steps to pass:
+    """
+    When I execute on the pod:
+      | sh | -c | more /mnt/local/testfile |
+    Then the step should succeed
+    """
+    And the output should contain "clone test"
```
jhou1, line 50: Line 44~50 can be removed. We don't really have to verify the "clone test" is written to `/mnt/local/testfile`. The `echo` command does its job very reliably

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,67 @@
+Feature: CSI clone testing related feature
+  """
+  This test can only work on a Cluster which has CSI driver deployed, OSP has csi deploy by default from 4.7. 
+  When deploy a cluster which has no csi deployed, please use enable_tpnu_featureset: "yes" in LAUNCHER_VARS.
+  From 4.7, OSP CSI driver was installed by default. Its storage class name is standard-csi.
+  Use below command can temporarily add cs class(oc get sc, can find all deployed stroage class):
+  oc patch featuregate cluster -p '{"spec": {"featureSet": "TechPreviewNoUpgrade"}}' --type merge
+  """
```
jhou1, line 8: Thanks for this comment, I like to see this in our tests, this is a good practice.

jhou1 (changes requested):
Generally looks good to me, some redundant steps can be removed. @duanwei33 Can you also PTAL?

features/storage/csi_clone.feature
JianLi-RH: In my impression you are correct and actually I tried it as you said, but unfortunately it doesn't work. It take me hours to fix it... @jhou1

features/storage/csi_clone.feature
duanwei33: @JianLi-RH That's good to add such info, as we supposed that csi driver is installed already, we could only mention the platform, OCP verison and storageclass name, I think no need to mention the feature gate and "enable_tpnu_featureset" (a flag of flexy-installation template)

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,67 @@
...
+  
+  # @author jianl@redhat.com
+  # @case_id OCP-27615
+  @admin
+  Scenario: Clone a pvc on OSP in the same namespace is successfully
+    Given the master version >= "4.8"
+    Given I switch to cluster admin pseudo user
+    # Step 1
+    # From 4.7, OSP CSI driver was installed by default. The storage class name is standard-csi
```
duanwei33, line 17: Let's remove the "step 1" as the we will only run it on the platform which have already installed the CSI driver, and all the info is added in your previous comments.
And restart the other steps from "step 1".

features/storage/csi_clone.feature
JianLi-RH: @duanwei33 OK, i will remove "When deploy a cluster which has no csi deployed, please use enable_tpnu_featureset: "yes" in LAUNCHER_VARS." from comments.

features/storage/csi_clone.feature
JianLi-RH: @duanwei33 Sure

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,67 @@
...
+    # Step 2
+    When admin creates a project with:
+      | project_name | clone |
+    Then the step should succeed
+    Given I use the "clone" project
```
duanwei33, line 22: Usually we use the <random> project name instead of hard coding it, it will avoid the conflict with other cases which happen to use the same project, or we can use "clone-<random>" if you want to make it meaningful.

features/storage/csi_clone.feature
JianLi-RH: sure

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,67 @@
...
+    When I obtain test data file "storage/csi/pod_pvc_sc.yaml"
```
duanwei33, line 23: To avoid maintaining too many yaml files, we can use some existing one if it matches your requirement.
Like storage/misc/pvc.json and storage/misc/pod.yaml, only add one pvc yaml file for cloned pvc.

features/storage/csi_clone.feature
duanwei33: BTW, maybe the "And I wait for the steps to pass:" is not needed, we just execute cmd and the step should pass.

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,67 @@
...
+    """
+    And I wait for the steps to pass:
+    """
+    When I execute on the pod:
+      | sh | -c | more /mnt/local/testfile |
+    Then the step should succeed
+    """
+    And the output should contain "clone test"
+    # Step 4
+    When I obtain test data file "storage/csi/pod_clone.yaml"
```
duanwei33, line 52: Same comments with previous one, maybe need to add one file for cloned pvc.

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,67 @@
...
+    And I run the :create client command with:
+      | f | pod_clone.yaml |
+    Then the step should succeed
+    And the pod named "mypod-clone" becomes ready
+    And the "mypvc-clone" PVC becomes :bound
+    # Step 5
+    Given 1 pod becomes ready with labels:
+      | name=frontendhttp-clone |
+    And I wait for the steps to pass:
+    """
```
duanwei33, line 62: I think "And I wait for the steps to pass:" could be removed.

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,67 @@
...
+    When I execute on the pod:
+      | sh | -c | more /mnt/local/testfile |
+    Then the step should succeed
+    """
+    And the output should contain "clone test"
```
duanwei33, line 67: How about add another step to write another file in this cloned pvc? That's will cover more for the clone test.

features/storage/csi_clone.feature
duanwei33: @JianLi-RH Please also remove the line6~7 for the cmd. It will not be covered in our test case.

features/storage/csi_clone.feature
duanwei33: @JianLi-RH Is it possible to add some word in the comments for the step, I guess only "step x" may be not so helpful.

features/storage/csi_clone.feature
JianLi-RH: @duanwei33 Line 6-7 just a reminder for whose environment does not have storage class installed, so I think keep it here is proper, do you think so?

features/storage/csi_clone.feature
JianLi-RH: Good suggestion. I did not know that step before...

features/storage/csi_clone.feature
JianLi-RH: This step is used to verify if the file "/mnt/local/testfile" copied correctly, so the step should be kept.

features/storage/csi_clone.feature
duanwei33: @JianLi-RH Hmm, as I said, we could add the information we supported(like platform, OCP verison and storageclass name), means we will not cover the untouched cases,  and will not perform such command in our case.

features/storage/csi_clone.feature
JianLi-RH: @duanwei33 ok, i will try it, if it can pass stable, i will remove it.

features/storage/csi_clone.feature
duanwei33: Sorry I notice that github remove some word in my comment, I mean to use the project with a random name, or clone+random name.
The step will give you a project with a random name
> Given I have a project

features/storage/csi_clone.feature
JianLi-RH: I just want to keep all steps separated clearly, it depends on Polarion test steps. No more meaning... @duanwei33

features/storage/csi_clone.feature
JianLi-RH: @duanwei33 I don't want to share files with others, because I have experience that someone in a team changes a public file, it blocked the others...

features/storage/csi_clone.feature
jhou1: The step `Given I have a project` will do, in most cases.

features/storage/csi_clone.feature
JianLi-RH: @duanwei33 I want to keep them...

features/storage/csi_clone.feature
duanwei33: @JianLi-RH
When the pod running with the cloned pvc, the file should exist, just use following step to check but no need using wait loop.  
>     When I execute on the pod:
>       | sh | -c | more /mnt/local/testfile |
>     Then the step should succeed

JianLi-RH: @duanwei33 Yes, you are right.

features/storage/csi_clone.feature
JianLi-RH: @jhou1 Yes, i am trying it now.

features/storage/csi_clone.feature
JianLi-RH: @duanwei33 Yes, its my fault, i did not noticed that. I will remove these lines.

features/storage/csi_clone.feature
duanwei33: Here I mean add one check point to write something new into the cloned pvc, not just check the existing file.

JianLi-RH: Sure

features/storage/csi_clone.feature
JianLi-RH: I tried to remove the Wait line...it made the test failed...So I have to keep it here.

features/storage/csi_clone.feature
jhou1: Where is the log for the failures?

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,50 @@
...
+    And I run the :create client command with:
+      | f | pod_clone.yaml |
+    Then the pod named "mypod-clone" becomes ready
+    And the "mypvc-clone" PVC becomes :bound
+    # Step 4
+    When I execute on the pod:
+      | sh | -c | more /mnt/local/testfile |
+    Then the output should contain "clone test"
+    And I execute on the pod:
+      | sh | -c | echo "clone test" > /mnt/local/testfile_2nd |
```
duanwei33, line 50: @JianLi-RH Thanks for the update, could you please add one more step:
**_Then the step should succeed_**

features/storage/csi_clone.feature
JianLi-RH: No error log in these steps, the latest step of verify testfile could not output "clone test" because there is no file cloned to the new pod.

duanwei33: @JianLi-RH Do you mean that when only using the following steps,  the testfile is not written to clone pvc when checking? 
> When I execute on the pod:
>       | sh | -c | echo "clone test" > /mnt/local/testfile |
> Then the step should succeed

From my view:
1. "And I wait for the steps to pass:" This step could be removed safely, we don't use the waiting loop for a simple step which the result will not change.
2. When you get the fail, we are not sure if it fails because no data writing into origin pvc or it fails because the clone feature, could you update a bit as below, it will help us to catch what is not correct. 
> When I execute on the pod:
>       | sh | -c | echo "clone test" > /mnt/local/testfile; ls /mnt/local/testfile |
> Then the step should succeed

JianLi-RH: @duanwei33 after failure, I logged into the origin pod and found the testfile is created with correct content, and then logged into the cloned pod I found there is no such file. I tried same step manually, they all pass. So I assume there is something delayed before creating the cloned pvc.
BTW, to run same code on my local machine, it is very stable. On Jenkins,  it is not stable.

duanwei33: Hmm, from my view, if failed case run as you said, I prefer to treat it as a bug. I think you can update with our comments, our script is not to avoid the failure, and we can debug what caused that, of course maybe by PSI unstable, but it is acceptable. WDYT?

jhou1: I'm curious why that had failed too. Let's be more communication efficient. Next time we talk about failures, let's make sure we attach logs so that relevant owners can know what the underlying problem is. Logs are more helpful than the language here.

JianLi-RH: @duanwei33 @jhou1 here is one failed run: [https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/ocp-common/job/Runner/47232/console](url),  you can see it could not find testfile on the clone pod:
0:34:23      When I execute on the pod:                                       # features/step_definitions/pod.rb:216
10:34:23        | sh | -c | more /mnt/local/testfile |
10:34:23        [02:34:22] INFO> Shell Commands: oc exec mypod-clone  --kubeconfig=/home/jenkins/ws/workspace/ocp-common/Runner/workdir/ocp4_testuser-0.kubeconfig -n sqkbk  -i -- sh -c more\ /mnt/local/testfile
10:34:23        
10:34:23        STDERR:
10:34:23        more: stat of /mnt/local/testfile failed: No such file or directory

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,58 @@
...
+    Then the step should succeed
+    """
+    And I wait for the steps to pass:
+    """
+    When I execute on the pod:
+      | sh | -c | more /mnt/local/testfile |
+    Then the output should contain "clone test"
+    """
+    # Step 3
+    Given I obtain test data file "storage/csi/pod_clone.yaml"
```
jhou1, line 46: You are having Pod and PVC specs together in one file, you need to name the file to reflect that. The name `pod_clone.yaml` looks like a yaml for Pod, when it also specifies a PVC.

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,58 @@
...
+  """
+  This test can only work on a Cluster which has CSI driver deployed, OSP has csi deploy by default from 4.7. 
+  From 4.7, OSP CSI driver was installed by default. Its storage class name is standard-csi.
+  Use "oc get sc" can find all deployed stroage class.
+  """
+  
+  # @author jianl@redhat.com
+  # @case_id OCP-27615
+  @admin
+  Scenario: Clone a pvc on OSP in the same namespace is successfully
```
jhou1, line 11: Grammar error, rename scenario: `Clone a PVC and verify data consistency`

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,56 @@
...
+    # Step 2
+    Given 1 pod becomes ready with labels:
+      | name=frontendhttp |
+    # Because the execution is too fast, we have to split test file creation and writing text into 2 commands: 
+    # touch and echo, and use the wait step.
+    When I execute on the pod:
+      | sh | -c | touch /mnt/local/testfile |
+    Then the step should succeed
+    When I execute on the pod:
+      | sh | -c | echo "clone test" > /mnt/local/testfile |
```
liangxia, line 31: Try to call `sync` to force write the memory into disk.

features/storage/csi_clone.feature
JianLi-RH: @jhou1 the file name is same as it in bug, [https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-27615](url)

JianLi-RH: The file name and structure are totally against the test case.

features/storage/csi_clone.feature
JianLi-RH: @jhou1 um...I used the description of the test case...but you are right, we should use correct English grammar.

features/storage/csi_clone.feature
JianLi-RH: @liangxia Thanks, sync fixed the issue.

features/storage/csi_clone.feature
JianLi-RH: Sure

features/storage/csi_clone.feature
jhou1: OK, don't worry about that test case in polarion, let's make sure our automation files and scripts have clean and consistent naming.

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,45 @@
...
+      | f | pod_pvc_sc.yaml |
+    Then the pod named "mypod" becomes ready
+    And the "mypvc" PVC becomes :bound
+    # Step 3
+    Given 1 pod becomes ready with labels:
+      | name=frontendhttp |
+    When I execute on the pod:
+      | sh | -c | echo "clone test" > /mnt/local/testfile |
+    Then the step should succeed
+    # Because the execution is too fast, we have to sync test file to avoid nonthing to clone.
```
jhou1, line 27: nonthing -> nothing

features/storage/csi_clone.feature
```diff
@@ -0,0 +1,45 @@
+Feature: CSI clone testing related feature
+  """
+  This test can only work on a Cluster which has CSI driver deployed, OSP has csi deploy by default from 4.7. 
+  From 4.7, OSP CSI driver was installed by default. Its storage class name is standard-csi.
+  Use "oc get sc" can find all deployed stroage class.
+  """
+  
+  # @author jianl@redhat.com
+  # @case_id OCP-27615
+  @admin
```
liangxia, line 10: Could you point where is the admin privilege needed ?

JianLi-RH: Good question, actually I removed it some days ago, but it failed, and at that time I could not figure out what caused that failure.  But right now we know the failure is caused by storage's strategy. I will remove @admin tag if it does not block the test. Till now i have no cluster ready, i will verify it when my cluster setup. @jhou1

jhou1: Sure, since removing the tag doesn't touch any test scenario, feel free to skip the test to save your time.
To explain it in more detail, `@admin` enables the user to create/update resources outside of the current namespace. I looked at the test that the operations of creating pods/pvc are within the user's namespace, so `@admin` is not necessary, thanks Liang for pointing that out.

JianLi-RH: Yes, because at very beginning, I try to deploy csi with existing code, it requires admin permission, so the admin tag left here.

jhou1 (approved):

PR #1995, opened 2021-05-09
fix automation to work with new kata plugin image

pruan-rht:
@liangxia @cpmeadors @valtair1891 PTAL

### DIFF
--- a/testdata/kata/release-4.8/deployment.yaml
+++ b/testdata/kata/release-4.8/deployment.yaml
@@ -1,24 +1,24 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: sandboxed-containers-operator-system
+  name: openshift-sandboxed-containers-operator
 ---
 apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
 metadata:
-  name: sandboxed-containers-operator-system
-  namespace: sandboxed-containers-operator-system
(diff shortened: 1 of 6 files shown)


testdata/kata/release-4.8/deployment.yaml
```diff
@@ -1,24 +1,24 @@
...
 ---
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
 metadata:
   name: kata-operator
-  namespace: sandboxed-containers-operator-system
+  namespace: openshift-sandboxed-containers-operator
 spec:
-  channel: "4.8"
+  channel: "1.0"
```
liangxia, line 21: In above code, `  cb.channel = cb.master_version`, then
```
    step %Q/I perform the :set_custom_channel_and_subscribe web action with:/, table(%{
      | update_channel    | #{cb.channel} |
      | install_mode      | OwnNamespace  |
      | approval_strategy | Automatic     |
```
Which one are we expecting ?

pruan-rht: in the old way it was following OCP version.  I'm getting it from the subscription now.  I'll take that old line out.

pruan-rht: fixed

PR #1997, opened 2021-05-10
update rules

yapei:
pass log job/ocp-common/job/Runner/46728/console


lib/rules/web/console/base/login.xyaml
```diff
@@ -25,7 +25,7 @@ login_sequence:
       selector:
         # title RE not working w/ ff legacy
         # text: !ruby/regexp '/Login - OpenShift/'
-        xpath: '//title[starts-with(.,"Login")]'
```
yapei, line 28: This change affect all console cases

PR #2003, opened 2021-05-12
update embed function

yapei:
WIP


yapei:
closing since another PR is fixing the same issue and it is working #2025

PR #2004, opened 2021-05-12
Auto case OCP-41789

weliang1:
Auto a test case for Bug 1944678 - Whereabouts IPAM CNI duplicate IP addresses assigned to pods

Test log: https://privatebin-it-iso.int.open.paas.redhat.com/?7fe0f76aa1d2f400#4XxavvWoeGDwqauqihZnQbyxKUWudWN5XmuWHQuHtQ26


features/networking/multus.feature
```diff
@@ -1533,3 +1533,59 @@ Feature: Multus-CNI related scenarios
...
+    Given I have a project
+    Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/Bug-1944678.yaml"
+    When I run oc create as admin over "Bug-1944678.yaml" replacing paths:
+      | ["metadata"]["namespace"] | <%= project.name %> |
+    Then the step should succeed
+    
+    # Create a pod absorbing above net-attach-def
+    Given I obtain test data file "networking/multus-cni/Pods/generic_multus_pod.yaml"
+    When I run oc create over "generic_multus_pod.yaml" replacing paths:
+      | ["metadata"]["name"]                                       | macvlan-bridge-pod1 |
```
pruan-rht, line 1553: @weliang1 table `|` alignment

weliang1: @pruan-rht Fixed. Thanks!

zhaozhanqi:
@weliang1 sorry I should be missed this PR. BTW,  can we create one deployment or replicas for pod to make the pods number is 3.  is there any difference with create 3 pods one by one for that bug?

weliang1:
> @weliang1 sorry I should be missed this PR. BTW, can we create one deployment or replicas for pod to make the pods number is 3. is there any difference with create 3 pods one by one for that bug?

@zhaozhanqi The verification steps is to check if there are any duplicate IPs assigned to the pods from whereabouts IPAM, it's easy to check pod's IP one by one. For bug itself, I do not think there are any difference between creating pod one by one or using deployment or replicas. For automation steps, checking pod one by one is easy.

zhaozhanqi:
> > @weliang1 sorry I should be missed this PR. BTW, can we create one deployment or replicas for pod to make the pods number is 3. is there any difference with create 3 pods one by one for that bug?
> 
> @zhaozhanqi The verification steps is to check if there are any duplicate IPs assigned to the pods from whereabouts IPAM, it's easy to check pod's IP one by one. For bug itself, I do not think there are any difference between creating pod one by one or using deployment or replicas. For automation steps, checking pod one by one is easy.

if so.  I think we can use one rc to create 3 pods and get the each ip. 
and check the are different with like this kind of step `Then the expression should be true> cb.pod1ip!=cb.pod2ip!=cb.pod3ip`
thus we do not need to check the ip ( hardcode).

weliang1:
> > > @weliang1 sorry I should be missed this PR. BTW, can we create one deployment or replicas for pod to make the pods number is 3. is there any difference with create 3 pods one by one for that bug?
> > 
> > 
> > @zhaozhanqi The verification steps is to check if there are any duplicate IPs assigned to the pods from whereabouts IPAM, it's easy to check pod's IP one by one. For bug itself, I do not think there are any difference between creating pod one by one or using deployment or replicas. For automation steps, checking pod one by one is easy.
> 
> if so. I think we can use one rc to create 3 pods and get the each ip.
> and check the are different with like this kind of step `Then the expression should be true> cb.pod1ip!=cb.pod2ip!=cb.pod3ip`
> thus we do not need to check the ip ( hardcode).

@zhaozhanqi whereabouts dhcp IP range is needed and hardcoded in the NAD yaml file. Beside checking duplicated IP address, using hardcoded pods' IP will also check below features at same time:  1. Make sure the pods' IP should be assigned from dhcp IP range.  2. Make sure the pods' IP should be assigned in the order from that range.  To meet above three testing purpose, hardcoded IP is needed.

weliang1:
Any comments for this PR?

testdata/networking/multus-cni/NetworkAttachmentDefinitions/Bug-1944678.yaml
```diff
@@ -0,0 +1,15 @@
...
+  config: '{
+      "cniVersion": "0.3.1",
+      "type": "macvlan",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+        "range": "10.199.199.0/24",
+        "range_start": "10.199.199.100"
+      }
+  }'
```
kedark3, line 15: Add a newline at the end of file.

kedark3 (approved):

testdata/networking/multus-cni/NetworkAttachmentDefinitions/Bug-1944678.yaml
weliang1: @kedark3 Fixed. Thanks!

kedark3 (approved):

features/networking/multus.feature
```diff
@@ -1533,3 +1533,59 @@ Feature: Multus-CNI related scenarios
...
+      | ["metadata"]["annotations"]["k8s.v1.cni.cncf.io/networks"] | macvlan-bridge      |
+      | ["spec"]["containers"][0]["name"]                          | macvlan-bridge      |
+    Then the step should succeed
+    And the pod named "macvlan-bridge-pod1" becomes ready
+    # Check the created pod has correct ip
+    When I execute on the "macvlan-bridge-pod1" pod:
+      | bash | -c | ip -4 -brief a |
+    Then the output should contain:
+      | 10.199.199.100 |
```
anuragthehatter, line 1562: Thanks @weliang1. Are we sure that multus would assign dynamic IPs in sequential order as we are intentionally checking 100,101 and 102? Normally DHCP takes a random IP from the specified range, AFAIK.

anuragthehatter: Oh I see , (from above description)2. Make sure the pods' IP should be assigned in the order from that range.

weliang1: > Thanks @weliang1. Are we sure that multus would assign dynamic IPs in sequential order as we are intentionally checking 100,101 and 102? Normally DHCP takes a random IP from the specified range, AFAIK.

@anuragthehatter So far, the multus/whereabouts assign the ip addresses to the pods in sequential order, the testing results approved this in several multus test cases.

PR #2008, opened 2021-05-13
fix some rules

yapei:
/assign @yanpzhan


yapei:
@liangxia  Could you help review and merge if no objections?

PR #2009, opened 2021-05-14
use #attach method insted of #embed to get rid of deprecation warning.

pruan-rht:
This addresses [OCPQE-3985](https://issues.redhat.com/browse/OCPQE-3985) @yapei @liangxia PTAL.

### DIFF
--- a/lib/log.rb
+++ b/lib/log.rb
@@ -158,9 +158,8 @@ module BushSlicer
 
     # supports embedding content similar with same semantics as Cucumber
     def embed(src, mime_type, label)
-      if self.class.runtime.respond_to? :embed
-        info "embedding #{label.inspect}"
-        self.class.runtime.embed src, mime_type, label
+      if self.class.runtime.respond_to? :attach
+        self.class.runtime.attach src, mime_type
       else
         if src.kind_of?(String)
           if src.empty?
--- a/lib/webauto/web4cucumber.rb
+++ b/lib/webauto/web4cucumber.rb
@@ -3,7 +3,6 @@ require 'psych'
 require 'uri'
 require 'watir'
 require "base64"
-
 require_relative 'chrome_extension'
 
   class Web4Cucumber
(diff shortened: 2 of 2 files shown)


lib/log.rb
```diff
@@ -158,9 +158,8 @@ def print(msg)
 
     # supports embedding content similar with same semantics as Cucumber
     def embed(src, mime_type, label)
-      if self.class.runtime.respond_to? :embed
-        info "embedding #{label.inspect}"
-        self.class.runtime.embed src, mime_type, label
+      if self.class.runtime.respond_to? :attach
+        self.class.runtime.attach src, mime_type
```
liangxia, line 162: @pruan-rht Thank you for the fix.
@yapei  Could you help to test the changes ? Thanks.

pruan-rht: @liangxia I tested locally by using a bad xyaml...the screen capture took place and the html is saved.

yapei:
I tested and found some issues, copying my messages in Slack here
```
There are some issues:
We still need this line info "embedding #{label.inspect}"  so that we can find the correct screenshot for specific failure when more than 1 screenshot is saved
The screenshot is saved with name test case execution attachment  and text  format where we .png  file is expected
```

pruan-rht:
@yapei I tried to run the `embed` methods as what the current state of github repo and the screen shot is just one big one.  Perhaps this is related to cuke5.  Also, I checked in a patch to get rid of the warning.

yapei:
> @yapei I tried to run the `embed` methods as what the current state of github repo and the screen shot is just one big one. Perhaps this is related to cuke5. Also, I checked in a patch to get rid of the warning.
`the screen shot is just one big one`  yes, is it saved as name `test case execution attachment` which should be `2021-xx-xxT07:00:40+00:00-screenshot.png`

pruan-rht:
superseded by #2025, will open another one for the browser warnings.

PR #2014, opened 2021-05-17
fix failure when checking data

QiaolingTang:
Fix failure when checking data in external ES, e.g.:  https://url.corp.redhat.com/08cbbc8

/cc @anpingli


anpingli (approved):
/lgtm

PR #2015, opened 2021-05-17
Get config infrastructure

lihongyan1:
Used by PR:
https://github.com/openshift/cucushift/pull/8545


lib/openshift/infrastructure.rb
```diff
@@ -12,5 +12,10 @@ def infra_name(user: nil, cached: true, quiet: false)
     def platform(user: nil, cached: true, quiet: false)
       status_raw(user: user, cached: cached, quiet: quiet).dig("platform")
     end
+
+    def infra_topo(user: nil, cached: true, quiet: false)
```
pruan-rht, line 16: @lihongyan1 how about just use `infra_topology` instead to be more specific?

PR #2016, opened 2021-05-17
[upgrade] Spot/preemptible instances should not block upgrade

sunzhaohua2:
Auto OCP-41175 Spot/preemptible instances should not block upgrade @jhou1 @miyadav please help to take a look.
Before upgrade:
```
      [06:46:51] INFO> === End After Scenario: Spot/preemptible instances should not block upgrade - prepare ===

1 scenario (1 passed)
14 steps (14 passed)
4m1.068s
```

After upgrade:
```
      [15:46:04] INFO> === End After Scenario: Spot/preemptible instances should not block upgrade ===

1 scenario (1 passed)
8 steps (8 passed)
1m40.336s
```

### DIFF
--- a/features/upgrade/machines/machines.feature
+++ b/features/upgrade/machines/machines.feature
@@ -79,6 +79,77 @@ Feature: Machine-api components upgrade tests
 
   @upgrade-prepare
   @admin
+  @destructive
+  Scenario Outline: Spot/preemptible instances should not block upgrade - prepare
+    Given I have an IPI deployment
+    And I switch to cluster admin pseudo user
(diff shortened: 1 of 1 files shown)


miyadav:
@sunzhaohua2 , how are we passing prices ?

sunzhaohua2:
@miyadav We only need to pass price on aws and azure. On gcp spot instance will be terminated after 24 hours. On aws and azure, if the price is not passed, the default price is On-Demand price.  https://github.com/openshift/enhancements/blob/master/enhancements/machine-api/spot-instances.md#cloud-provider-implementation-specifics

PR #2017, opened 2021-05-17
OCP-22504 quick fix

anuragthehatter:
Seems like the upstream/downtream multus has finally corrected this minor english mistake and backported to earlier releases. Hence test case needs to be adjusted accordingly to avoid CI failures
Re-ran to make sure : http://pastebin.test.redhat.com/964530

@openshift/team-sdn-qe


pruan-rht:
@anuragthehatter have you tried it against older version of OCP?

PR #2018, opened 2021-05-18
add jitter to wait_for delay to reduce contention

rbbratta:
The `I wait up to 30 seconds for the steps to pass`
step calls `wait_for` but does not allow for changing
the interval.

The proper way to do retry loops in a distributed system is
to use jitter and/or exponential backoff as described in the Amazon paper.

https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

Implement the Decorrelated Jitter algorithm in the wait_for
method.

### DIFF
--- a/lib/base_helper.rb
+++ b/lib/base_helper.rb
@@ -138,20 +138,46 @@ module BushSlicer
       # @param seconds [Numeric] the max number of seconds to try operation to
       #   succeed
       # @param interval [Numeric] the interval to wait between attempts
+      # @param stats [Hash] collect stats
+      # @param jitter_multiplier [Numeric] (value >= 1) used to generate
+      #   random jitter in the wait interval using the decorrelated jitter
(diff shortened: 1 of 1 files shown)


rbbratta:
#1649

rbbratta:
```
ArgumentError: time interval must not be negative
verification-tests/lib/base_helper.rb:165:in `sleep'
verification-tests/lib/base_helper.rb:165:in `wait_for'
```

pruan-rht:
@rbbratta thanks for reviving this PR.  TBH, currently we are still dealing with the cuke5 migration and might not have time to review this one.  There are a couple of questions to keep in mind for this RFE 
1. Since we use `wait_for` a lot, this is a fundamental change that needs to be tested throughly.
2. Does the benefit out-weights the risks?  What do we gain from this enhancement?  Again, the potential for breakage is pretty big.
Just a few points to think about...

rbbratta:
not needed

PR #2025, opened 2021-05-19
Use attach to replace deprecated embed

liangxia:
/cc @yapei @yanpzhan @pruan-rht @jhou1 @JianLi-RH  for review.

### DIFF
--- a/features/support/save_to_dir_embedding_formatter.rb
+++ b/features/support/save_to_dir_embedding_formatter.rb
@@ -34,8 +34,20 @@ module BushSlicer
 
     # @note #embed has been replaced by #attach in Cucumber 5.x
     def attach(src, media_type)
-      if media_type != 'text/x.cucumber.log+plain'
-        embed(src, media_type, "test case execution attachment")
+      if media_type == 'text/x.cucumber.log+plain'
+        puts src
+      elsif media_type == 'image/png;base64'
+        filename = Time.now.iso8601 + '-screenshot.png'
+        puts "embedding #{filename}"
+        embed(src, media_type, "#{filename}")
+      elsif media_type == 'text/html;base64'
+        filename = Time.now.iso8601 + '.html'
+        puts "embedding #{filename}"
+        embed(src, media_type, "#{filename}")
+      else
+        filename = 'test case execution attachment'
+        puts "embedding #{filename}"
+        embed(src, media_type, "#{filename}")
       end
     end
 
--- a/features/support/save_to_dir_embedding_formatter.rb
+++ b/features/support/save_to_dir_embedding_formatter.rb
@@ -34,9 +34,7 @@ module BushSlicer
 
     # @note #embed has been replaced by #attach in Cucumber 5.x
     def attach(src, media_type)
-      if media_type == 'text/x.cucumber.log+plain'
-        puts src
-      elsif media_type == 'image/png;base64'
+      if media_type == 'image/png;base64'
         filename = Time.now.iso8601 + '-screenshot.png'
         puts "embedding #{filename}"
         embed(src, media_type, "#{filename}")
--- a/lib/log.rb
+++ b/lib/log.rb
@@ -158,9 +158,8 @@ module BushSlicer
 
     # supports embedding content similar with same semantics as Cucumber
     def embed(src, mime_type, label)
-      if self.class.runtime.respond_to? :embed
-        info "embedding #{label.inspect}"
-        self.class.runtime.embed src, mime_type, label
+      if self.class.runtime.respond_to? :attach
+        self.class.runtime.attach src, mime_type
       else
         if src.kind_of?(String)
           if src.empty?
--- a/lib/webauto/web4cucumber.rb
+++ b/lib/webauto/web4cucumber.rb
@@ -660,11 +660,10 @@ require_relative 'chrome_extension'
     # another option would be to save to file but not sure where to store
     #   as most formatters do not relocate the file
     def take_screenshot
-      filename = Time.now.iso8601
       # screenshot = %{data:image/png;base64,#{browser.screenshot.base64}}
       # logger.embed screenshot, "application/octet-stream", "#{filename}-screenshot.png"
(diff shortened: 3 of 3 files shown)


liangxia:
Tested with Runner/56461/

features/support/save_to_dir_embedding_formatter.rb
```diff
@@ -34,8 +34,17 @@ def initialize(config)
...
-      if media_type != 'text/x.cucumber.log+plain'
-        embed(src, media_type, "test case execution attachment")
+      if media_type == 'text/x.cucumber.log+plain'
+        puts src
+      elsif media_type == 'image/png;base64'
+        filename = Time.now.iso8601 + '-screenshot.png'
+        embed(src, media_type, "#{filename}")
+      elsif media_type == 'text/html;base64'
+        filename = Time.now.iso8601 + '.html'
+        embed(src, media_type, "#{filename}")
```
yapei, line 44: maybe we don't need to save html file

lib/log.rb
```diff
@@ -158,9 +158,8 @@ def print(msg)
 
     # supports embedding content similar with same semantics as Cucumber
     def embed(src, mime_type, label)
-      if self.class.runtime.respond_to? :embed
-        info "embedding #{label.inspect}"
-        self.class.runtime.embed src, mime_type, label
+      if self.class.runtime.respond_to? :attach
+        self.class.runtime.attach src, mime_type
```
yapei, line 162: Could you keep this line `info "embedding #{label.inspect}"` because it will be really helpful finding the correct screenshot when too many screenshot is saved

features/support/save_to_dir_embedding_formatter.rb
liangxia: It comes from https://github.com/openshift/verification-tests/pull/2025/files#diff-4e2de23aa38bbb77f3e035cf102da99897869c1bae03520d60ae31664cd7819dL667
Just keep it to work as the old fashion.

If you persist it's not needed, I can remove them.

pruan-rht:
@liangxia @yapei how did we end up with 3 different PRs doing the same thing? :)

1. https://github.com/openshift/verification-tests/pull/2003
2. https://github.com/openshift/verification-tests/pull/2025
3. https://github.com/openshift/verification-tests/pull/2009 (this one has another fix about another warning)

lib/log.rb
liangxia: The label here is empty ATM. Keep it does not help much.

liangxia: Move/add them in https://github.com/openshift/verification-tests/pull/2025/files#diff-fcf1e86594bb0416d6eecb2efacfc0de19c92dba55fdcf6848142a697e211657R41

yapei:
@pruan-rht I will close my fix PR since it doesn't work. You can just scope your PR to fix another warning IMO, HDYT?

features/support/save_to_dir_embedding_formatter.rb
```diff
@@ -34,8 +34,20 @@ def initialize(config)
 
     # @note #embed has been replaced by #attach in Cucumber 5.x
     def attach(src, media_type)
-      if media_type != 'text/x.cucumber.log+plain'
-        embed(src, media_type, "test case execution attachment")
+      if media_type == 'text/x.cucumber.log+plain'
+        puts src
+      elsif media_type == 'image/png;base64'
+        filename = Time.now.iso8601 + '-screenshot.png'
+        info "embedding #{filename}"
```
yapei, line 41: With the new changes, it reports error when `info` is called seems it is only accessible in `log.rb`

```
05-20 16:33:49.206        undefined method `info' for #<BushSlicer::SaveToDirEmbeddingFormatter:0x0000000005a81f38> (NoMethodError)
05-20 16:33:49.206        /home/jenkins/ws/workspace/ocp-common/Runner/features/support/save_to_dir_embedding_formatter.rb:41:in `attach'
05-20 16:33:49.207        /home/jenkins/ws/workspace/ocp-common/Runner/lib/log.rb:162:in `embed'
```

features/support/save_to_dir_embedding_formatter.rb
```diff
@@ -34,8 +34,20 @@ def initialize(config)
...
+        embed(src, media_type, "#{filename}")
+      elsif media_type == 'text/html;base64'
+        filename = Time.now.iso8601 + '.html'
+        info "embedding #{filename}"
+        embed(src, media_type, "#{filename}")
```
yapei, line 46: maybe we don't need save `.html` files to reduce space

features/support/save_to_dir_embedding_formatter.rb
liangxia: Use `puts` to replace `info`

features/support/save_to_dir_embedding_formatter.rb
yapei: I'm ok to keep it and we can still remove it when we think the large artifact is a problem

features/support/save_to_dir_embedding_formatter.rb
yapei: sorry I I missed your above comment

yapei:
tested with Runner/57371, there is one minor issue and I'm not sure what change caused it. 
web action logs were printed twice(see my screenshot for details)

Screen Shot 2021-05-20 at 6 01 54 PM

Here is the log screenshot can be used as comparison before the fix
Screen Shot 2021-05-20 at 6 10 46 PM

liangxia:
> But one minor regression issue and I'm not sure what change caused it.
> web action logs were printed twice(see my screenshot for details)
> 

@yapei  Good catch, fix in https://github.com/openshift/verification-tests/pull/2025/commits/d2f03df2bbcfcaf71c61f5cbeb3ba2179f997ce9

pruan-rht:
@yapei let's just use this PR.  I'll close mine and file another one for the other warning.

pruan-rht (commented):
/lgtm

PR #2030, opened 2021-05-20
Fix scheduler restore logic

kasturinarra:
Currently when the scheduler cluster has profile in the spec, restore logic does not handle it the right way, so fixing the same.


kasturinarra:

Run logs: https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/ocp-common/job/Runner/57232/console

kasturinarra:

PR #2034, opened 2021-05-25
Reorder the log info

liangxia:
/cc @yapei @yanpzhan

### DIFF
--- a/features/support/save_to_dir_embedding_formatter.rb
+++ b/features/support/save_to_dir_embedding_formatter.rb
@@ -34,18 +34,10 @@ module BushSlicer
 
     # @note #embed has been replaced by #attach in Cucumber 5.x
     def attach(src, media_type)
-      if media_type == 'image/png;base64'
-        filename = Time.now.iso8601 + '-screenshot.png'
-        puts "embedding #{filename}"
-        embed(src, media_type, "#{filename}")
-      elsif media_type == 'text/html;base64'
-        filename = Time.now.iso8601 + '.html'
-        puts "embedding #{filename}"
-        embed(src, media_type, "#{filename}")
-      else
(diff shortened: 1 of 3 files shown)


features/support/save_to_dir_embedding_formatter.rb
```diff
@@ -34,18 +34,10 @@ def initialize(config)
...
-      elsif media_type == 'text/html;base64'
-        filename = Time.now.iso8601 + '.html'
-        puts "embedding #{filename}"
-        embed(src, media_type, "#{filename}")
-      else
-        filename = 'test case execution attachment'
-        puts "embedding #{filename}"
-        embed(src, media_type, "#{filename}")
+      if media_type == 'image/png;base64' || media_type == 'text/html;base64'
+        embed(src, media_type, "#{$screenshot_label}")
```
yapei, line 38: Do you think it is a good time to remove saving of html files? They are not needed at all

features/support/save_to_dir_embedding_formatter.rb
```diff
@@ -34,18 +34,10 @@ def initialize(config)
...
+      elsif media_type != 'text/x.cucumber.log+plain'
+        embed(src, media_type, "test case execution attachment")
```
yapei, line 40: that means we only run `embed(src, media_type, "#{$screenshot_label}")` for `image/png;base64` , HDYT?

yapei: The original code is:

```
    def attach(src, media_type)
      if media_type != 'text/x.cucumber.log+plain'
        embed(src, media_type, label)
      end
    end
```

yapei (approved):
/lgtm

PR #2036, opened 2021-05-25
rules added for OCP-36165

yapei:


lib/rules/web/admin_console/4.7/actions.xyaml
```diff
@@ -153,8 +153,8 @@ confirm_cancel_action:
   action:
     if_param:
       cancel: 'false'
-    ref: click_not_cancel
-click_not_cancel:
+    ref: click_cancel_in_modal
```
yapei, line 156: rename the action to be more readable

yapei:
@liangxia Could you help review and merge if no objections?

pruan-rht (commented):
/lgtm

PR #2038, opened 2021-05-25
rules added for OCP-40405

yapei:
feature PR cucushift#8566


yapei:
@yanpzhan  @XiyunZhao  Could you help review?

PR #2041, opened 2021-05-26
[ONEOCPDEPL-93] Class names renamed as requested

pjanouse:


pruan-rht:
@pjanouse looks good..just curious, who requested this change?

pjanouse:
@pruan-rht coming from [here](https://github.com/openshift/verification-tests/pull/1919#discussion_r630445985).

PR #2029, opened 2021-05-20
Hongyli sno 1

lihongyan1:
Related to PR
https://github.com/openshift/cucushift/pull/8545

PR #2042, opened 2021-05-26
helper step to check for number of replicas for a cluster resource

pruan-rht:
example usage `Given there are 6 "daemon_set" replicas in the "openshift-monitoring" project`.  Should help with PR #2029

@lihongyan1 this should help with your PR

### DIFF
--- a/features/step_definitions/replicas.rb
+++ b/features/step_definitions/replicas.rb
@@ -1,0 +1,20 @@
+# replicas related steps
+#
+
+## example usage Given there are 6 "daemon_set" replicas in the "openshift-monitoring" project
+Given /^there are #{NUMBER} #{QUOTED} replicas in the #{QUOTED} project$/ do |expected, resource_type, namespace|
+  ensure_admin_tagged
+  expected = expected.to_i
+  clazz = resource_class(resource_type)
+  if BushSlicer::ProjectResource > clazz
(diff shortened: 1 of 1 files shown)


features/step_definitions/replicas.rb
```diff
@@ -0,0 +1,20 @@
...
+  clazz = resource_class(resource_type)
+  if BushSlicer::ProjectResource > clazz
+    list = clazz.list(user: user, project: project(namespace))
+  else
+    list = clazz.list(user: user)
+  end
+  list.each do | cl |
+    desired_count = cl.replica_counters(cached: false)[:desired]
+    if desired_count != expected.to_i
+      raise "#{resource_type} count is #{desired_count}, but is expected to be #{expected} instead"
```
lihongyan1, line 17: Better change as the following to provide more information to merge
 raise "#{resource_type} #{cl.name} count is #{desired_count}, but is expected to be #{expected} instead"

pruan-rht: @lihongyan1 fixed and edited for grammar error as well.
```      daemon_set node-exporter count is 1, but is expected to be 2 instead (RuntimeError)```

pruan-rht:
@liangxia PTAL and merge if there is no objections as @lihongyan1's PR https://github.com/openshift/verification-tests/pull/2029  is dependent on this

PR #2049, opened 2021-05-27
SDN-1340 test case step definition

asood-rh:
Step definition to obtain endpoint publishing strategy.
OCP-40908
OCP-41944

/cc @openshift/team-sdn-qe

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -857,6 +857,13 @@ Given /^the env is using "([^"]*)" networkType$/ do |network_type|
   raise "the networkType is not #{network_type}" unless network_operator.network_type(user: _admin) == network_type
 end
 
+Given /^the cluster has "([^"]*)" endpoint publishing strategy$/ do |ep_pub_strategy|
+  ensure_admin_tagged
+  _admin = admin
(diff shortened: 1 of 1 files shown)


features/step_definitions/networking.rb
```diff
@@ -857,6 +857,13 @@
   raise "the networkType is not #{network_type}" unless network_operator.network_type(user: _admin) == network_type
 end
 
+Given /^the cluster has "([^"]*)" endpoint publishing strategy$/ do |ep_pub_strategy|
```
pruan-rht, line 860: @asood-rh probably we don't need to step since the ingress_controller class is already defined [here](https://github.com/openshift/verification-tests/blob/master/lib/openshift/ingress_controller.rb#L15)

asood-rh: @pruan-rht The step obtains the endpoint publishing strategy and fails the test script if it is not what was passed in by the script. This is more than a getter method.

asood-rh: @pruan-rht https://github.com/openshift/cucushift/pull/8572 PR that uses the step. Could you please approve thisone?

Thanks.

pruan-rht: /lgtm

PR #2053, opened 2021-06-01
Add code to remove forcedeploymentreason

kasturinarra:
currently when etcd cluster is patched to redeploy openshift-etcd pods there is no way to restore the etcd cluster back after the test, adjusting the function here to do the same.


kasturinarra:
@zhouying7780 could you please help review ? More context from here [1]

[1] https://github.com/openshift/cucushift/pull/8576
[2] Runner log: https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3130/console

PR #2055, opened 2021-06-02
[WIP] implement nesting of templates

akostadinov:
Without any testing. This should more or less allow nesting of templates into
other templates.

Two precaustions:

* templates can interfere if they redefine the same variable names
* reusing of a nested variables file is not possible (should be possible to implement but IMO it will be confusing to have multiple variables files.. and we are already ocnfusing enough)

Let me know what you think and whether we should allow nesting or not after all.

/assign @pjanouse

cc @apodhrad

### DIFF
--- a/tools/jenkins/slaves/Dockerfile.rhel7
+++ b/tools/jenkins/slaves/Dockerfile.rhel7
@@ -38,7 +38,7 @@ ADD . $HOME/verification-tests
 # RUN scl enable rh-git218 "git clone --depth 1 --single-branch 'https://github.com/openshift/verification-tests.git' '$HOME/verification-tests'"
 
 RUN scl enable rh-ruby27 "$HOME/verification-tests/tools/install_os_deps.sh"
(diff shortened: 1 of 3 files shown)


pjanouse:
I like the idea although I don't know if there is a demand for it. Have you any insight @abraverm?

One thing I don't like here is that this MR contains (if I understand it correctly) also unrelated changes (Dockerfile's) which should be processed by another MR(s).

pjanouse:
@apodhrad Have you chance to try it already? I don't know exact syntax how to include one part to other, so probably this impl. is based on your request and agreement between you and @akostadinov how to implement it.

akostadinov:
Strange about dockerfiles. Perhaps ignore them. Not sure why they show up.

abraverm:
I think this might help with making the templates a bit more moduler and will make it easier for teams to reuse working templates. This is partially related to https://issues.redhat.com/browse/ONEOCPDEPL-71

PR #2062, opened 2021-06-04
fix automation error for tzhou's cases

yasun1:
fix automation error for tzhou's cases


yasun1:
@xueli181114 Please help review.

yasun1:
@liangxia  As the runner-v3 is hung. I can not get the executed log, but these changes have been passed at my local. Could you help merge these? thanks.

PR #2063, opened 2021-06-04
Add OWNERS file for OCM

liangxia:
/cc @xueli181114 @yasun1 

I'm just adding an sample there, you can update the list at any time.


pruan-rht (commented):
/lgtm

PR #2069, opened 2021-06-08
extend timeout for install and uninstall

pruan-rht:
Extend the timeout value to 1000 seconds instead of 900.  Also print the kata operator image information

### DIFF
--- a/features/step_definitions/kata.rb
+++ b/features/step_definitions/kata.rb
@@ -41,7 +41,7 @@ Given /^I wait for #{QUOTED} (uninstall|install) to start$/ do | kc_name, mode |
 end
 
 Given /^I wait until number of completed kata runtime nodes match#{OPT_QUOTED} for #{QUOTED}$/ do |number, kc_name|
-  ready_timeout = 900
+  ready_timeout = 1000
   matched = kata_config(kc_name).wait_till_installed_counter_match(
     user: user, seconds: ready_timeout)
   unless matched[:success]
@@ -59,7 +59,7 @@ Given /^I remove kata operator from the#{OPT_QUOTED} namespace$/ do | kata_ns |
   # 1. remove kataconfig first
   project(kata_ns)
   kataconfig_name = BushSlicer::KataConfig.list(user: admin).first.name
-  step %Q/I ensure "#{kataconfig_name}" kata_config is deleted within 900 seconds/
+  step %Q/I ensure "#{kataconfig_name}" kata_config is deleted within 1000 seconds/
   # 2. remove namespace
   step %Q/I ensure "#{kata_ns}" project is deleted/
 end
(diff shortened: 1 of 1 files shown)


features/step_definitions/kata.rb
```diff
@@ -166,7 +166,12 @@
     logger.info("There's already an existing 'kataconfig' resuing it...")
     project(kata_ns)
     step %Q/I switch to cluster admin pseudo user/
+    # make sure kata-operator is running first before installing the kataconfig
+    step %Q/a pod is present with labels:/, table(%{
+      | control-plane=controller-manager |
+    })
```
cpmeadors, line 172: This check seems really generic.  Are you sure that this check is specific to the sandboxed-container-operator and would not be satisfied falsely by something else?

features/step_definitions/kata.rb
```diff
@@ -41,7 +41,7 @@
 end
 
 Given /^I wait until number of completed kata runtime nodes match#{OPT_QUOTED} for #{QUOTED}$/ do |number, kc_name|
-  ready_timeout = 900
+  ready_timeout = 1000
```
jensfr, line 44: this is ~16 minutes, which should be enough for the typical cluster sizes we get in our public cloud clusters. Not sure if this will ever run on bare-metal clusters. If yes, this could be not enough depending on how many nodes the machine pool has.

pruan-rht: @jensfr yeah for our acceptance test, I saw a couple of times 900 seconds was not enough, that's why I extended it to 1000 seconds.  I believe that should be sufficient.

features/step_definitions/kata.rb
```diff
@@ -166,7 +166,12 @@
     logger.info("There's already an existing 'kataconfig' resuing it...")
     project(kata_ns)
     step %Q/I switch to cluster admin pseudo user/
+    # make sure kata-operator is running first before installing the kataconfig
+    step %Q/a pod is present with labels:/, table(%{
```
jensfr, line 170: What if the pod is present but in a bad state?

pruan-rht: @jensfr then the step will fail and test will mark as such.

cpmeadors (commented):
Timeout change looks good.  Not sure about the readiness check for the operator.

features/step_definitions/kata.rb
pruan-rht: @cpmeadors yes, that's the first and only pod with that label in the project.  We could request the pod to have a more specific label, but that has to be done kata itself.

cpmeadors (approved):
Comments resolved.  Everything looks good.

features/step_definitions/kata.rb
pruan-rht: @jensfr I changed it to use a step to wait for the pod to become `ready` instead...

pruan-rht:
@liangxia I'm going to merge this one since I need this change to unblock the reporting portion.

PR #2024, opened 2021-05-19
[v3] update some critical routing cases to use images from quay

lihongan:
We found some v3 test cases were failed due to image pull rate limit issue of docker.io, since the old  `openshift-qe/v3-testfiles` repo is deprecated, so update the test cases to use the testdata (master branch) that already consumed the images from `quay.io/openshifttest`.

@quarterpin @jechen0648 Please help take a look. Thanks.

PR #2076, opened 2021-06-09
[v3] update pod-for-ping to use image from quay.io

lihongan:
follow up https://github.com/openshift/verification-tests/pull/2024

@liangxia please take a look.

cc @zhaozhanqi


zhaozhanqi:
ah,  we missed this fish :)
/lgtm

PR #2077, opened 2021-06-10
fix volume_snapshot readiness check

pruan-rht:
Current logic is wrong.  We just need to check for  status.readyToUse to be true to be considered 'ready'.  Also added method 'wait_til_ready'
```
     Then the expression should be true> volume_snapshot("mysnapshot").wait_till_ready(user: user) # features/step_definitions/common.rb:133
      [01:00:52] INFO> After 0 iterations and 21 seconds:
```      
@duanwei33, PTAL,

### DIFF
--- a/lib/openshift/volume_snapshot.rb
+++ b/lib/openshift/volume_snapshot.rb
@@ -9,27 +9,22 @@ module BushSlicer
       rr.dig('spec', 'snapshotDataName')
     end
 
-    # @return [BushSlicer::ResultHash] with :success depending on
-    #   condition type=Ready and status=True
     def ready?(user:, quiet: false, cached: false)
-      if cached && props[:raw]
-        res = { instruction: "get cached volume snapshot #{name} readiness",
-                response: props[:raw].to_yaml,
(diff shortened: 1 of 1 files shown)


lib/openshift/volume_snapshot.rb
```diff
@@ -9,27 +9,22 @@ def snapshot_data_name(user: nil, cached: true, quiet: false)
...
+    def wait_till_ready(user: nil, seconds: 30)
+      res = nil
+      iterations = 0
+      start_time = monotonic_seconds
+
+      success = wait_for(seconds) {
+        res = ready?(user: user, quiet: true)
+      }
+      duration = monotonic_seconds - start_time
+      logger.info "After #{iterations} iterations and #{duration.to_i} " <<
```
liangxia, line 26: The `iterations` will always `0` ?

pruan-rht: my mistake... fixed with https://github.com/openshift/verification-tests/pull/2078  PTAL. thx!

PR #2081, opened 2021-06-11
add support for OCP-33656

yaoli-redhat:
PR https://github.com/openshift/cucushift/pull/8597 depends on this and this should be merged firstly. 
Pass log: Runner/75195/console
@xingxingxia Please help to review when free. Thanks.

### DIFF
--- a/lib/base_helper.rb
+++ b/lib/base_helper.rb
@@ -392,6 +392,13 @@ module BushSlicer
         end
       end
 
+      def get_oauthaccesstoken(token)
+        if token.include? "sha256~"
+          "sha256~" + Base64.urlsafe_encode64(OpenSSL::Digest::SHA256.digest(token.gsub("sha256~","")),padding: false)
+        else
+          raise "provided token is not valid and not started with sha256"
+        end
+      end
 
     end
 


features/step_definitions/oauth_client.rb
```diff
@@ -2,3 +2,13 @@
   cb.cluster_resource_to_recreate = o_auth_client(name)
   step 'hidden recreate cluster resource after scenario'
 end
+
+Given(/^convert token (.*) to oauthaccesstoken successfully$/) do |token|
```
pruan-rht, line 6: @yaoli-redhat can you pls given an example of how a user would use this step?  Probably need to be more explicit in the step name that it's implicitly saving the converted result into the clipboard

pruan-rht: also probably missing a pronoun there, like `I` or `admin` before `convert token`

yaoli-redhat: @pruan-rht this is to support OCP-33656 auto case. You can check line 31 in https://github.com/openshift/cucushift/pull/8597 and the function is to conver user token to oauthaccesstoken.

yaoli-redhat: @pruan-rht update the step name, please help to review. Thanks. Pass log: https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/ocp-common/job/Runner/80291/console

xingxingxia:
Please keep in mind: as synced in Slack, any company **internal info** should not be exposed in public (GH / BZ etc.).
This repo is a PUBLIC repo, so please not expose the internal test case URL, not expose Jenkins URL (though they are not accessible from public).

xingxingxia:
In addition, as shared in Slack, try to add it into lib/base_helper.rb :
```
    module BaseHelper
      def get_oauthaccesstoken(token)
...
```
Then use it in any place including *.feature file, like <%= get_oauthaccesstoken(user.cached_tokens.first) %>

lib/base_helper.rb
```diff
@@ -392,6 +392,15 @@ def find_follow(*paths)
...
       end
 
+      def get_oauthaccesstoken(token)
+        if token.include? "sha256~"
+          oauthaccesstoken = token.gsub("sha256~","")
+          oauthaccesstoken = "sha256~" + Base64.urlsafe_encode64(OpenSSL::Digest::SHA256.digest(oauthaccesstoken),padding: false)
+        else
+          raise "provided user token is not valid"
+        end
+        return oauthaccesstoken
```
xingxingxia, line 402: In ruby, "return" is not needed

lib/base_helper.rb
```diff
@@ -392,6 +392,15 @@ def find_follow(*paths)
         end
       end
 
+      def get_oauthaccesstoken(token)
+        if token.include? "sha256~"
+          oauthaccesstoken = token.gsub("sha256~","")
+          oauthaccesstoken = "sha256~" + Base64.urlsafe_encode64(OpenSSL::Digest::SHA256.digest(oauthaccesstoken),padding: false)
+        else
+          raise "provided user token is not valid"
```
xingxingxia, line 400: From good code perspective, add several words to explicitly tell for which it is not valid.

yaoli-redhat:
@xingxingxia thanks for the review and all comments are resolved.

xingxingxia:
/lgtm
@pruan-rht @JianLi-RH @liangxia , help review / merge, thanks!
Thanks

pruan-rht (commented):
/lgtm

PR #2083, opened 2021-06-11
Automate snapshot case 

duanwei33:
[pass log](https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3158/console)
@jhou1 @chao007  PTAL


features/step_definitions/volume_snapshot.rb
```diff
@@ -48,3 +48,11 @@
     raise "Failed to create VolumeSnapshotClass."
   end
 end
+
+
+Given /^the#{OPT_QUOTED} volumesnapshot becomes ready(?: within (\d+) seconds)?$/ do |name, timeout|
+  timeout = timeout ? timeout.to_i : 60
+  @result = wait_for(timeout) {
```
duanwei33, line 55: @liangxia @pruan-rht could you kindly help take a look on this step, not sure if something need to be improved although test passed.
In **lib/openshift/volume_snapshot.rb**[1], Peter helped add the **wait_till_ready** with the **wait_for**, but I'm not sure I still have to add the waiting loop here, otherwise the step does not really wait for the expected ready status (only one loop and passed (After 1 iterations and 1 seconds:))
[1]https://github.com/openshift/verification-tests/blob/master/lib/openshift/volume_snapshot.rb#L12-L31

features/storage/csi_snapshot.feature
```diff
@@ -0,0 +1,57 @@
+Feature: Volume snapshot test
+
+  # @author wduan@redhat.com
+  @admin
+  Scenario Outline: Volume snapshot create and restore test
+    Given I have a project  
+    Given I obtain test data file "storage/misc/pvc.json"
+    When I create a dynamic pvc from "pvc.json" replacing paths:
+      | ["metadata"]["name"]                         | mypvc-ori |
+      | ["spec"]["storageClassName"]                 | <csi-sc>  |
```
chao007, line 10: Redundant space

features/step_definitions/volume_snapshot.rb
jhou1: https://github.com/openshift/verification-tests/blob/d0d4e5db1b9eadeceecac460282b9071e55846c8/lib/openshift/volume_snapshot.rb#L17, `wait_till_ready` accepts a param as seconds to wait for.

So you coud do something like `volume_snapshot(name).wait_till_ready(user: user, seconds timeout)`

duanwei33: hmm, I tried by this way, but doesn't work. So I have to add the wait_for here, and ask for help.

pruan-rht: @duanwei33 I totally missed this comment... in short it's my fault.  The support method has a bug.   The`wait_for` loop will exit when the last statement is `true`.  So currently, `iteration = iteration +1` will evaluate to true. Solution is to move the `iteration` step above the `res = ready?(user: user, quiet: true)
` line.

duanwei33: @pruan-rht Thanks a lot, it works now.

>         And the "mysnapshot" volumesnapshot becomes ready                     # features/step_definitions/volume_snapshot.rb:53
>       [10:34:16] INFO> After 24 iterations and 50 seconds:

@jhou1 Your suggested way works now.

features/storage/csi_snapshot.feature
duanwei33: Updated.

features/step_definitions/volume_snapshot.rb
pruan-rht: @duanwei33  we would need to file a PR to fix the method.  Let me do that

duanwei33: @pruan-rht I have updated **lib/openshift/volume_snapshot.rb** according to your suggestion, or I miss something else? 

>       success = wait_for(seconds) {
>         iterations = iterations + 1
>         res = ready?(user: user, quiet: true)
>       }

PR #2084, opened 2021-06-12
a few fixes for kata testing.

pruan-rht:
1. create a method to get the runtimeClass from a pod's spec
2. added missing testdata file for the case.
3. changed to use the newly created method.

### DIFF
--- a/testdata/kata/webhook/example-fedora.yaml
+++ b/testdata/kata/webhook/example-fedora.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example-fedora
+  labels:
+    app: example-fedora-app
+spec:
+  containers:
+    - name: example-fedora
+      image: fedora:30
+      ports:
+        - containerPort: 8080
+      command: ["python3"]
+      args: [ "-m", "http.server", "8080"]
+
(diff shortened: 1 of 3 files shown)


testdata/kata/webhook/example-fedora.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
liangxia, line 1:
```
$ cat testdata/kata/example-fedora-kata.yaml
apiVersion: v1
kind: Pod
metadata:
  name: example-fedora-kata
  labels:
    app: example-fedora-kata-app
spec:
  containers:
    - name: example-fedora
      image: fedora:30
      ports:
        - containerPort: 8080
      command: ["python3"]
      args: [ "-m", "http.server", "8080"]
  runtimeClassName: kata
```

cpmeadors (approved):

testdata/kata/webhook/example-fedora.yaml
liangxia: We already have a file with the same content there. Perhaps we should fix the filename in the scenario instead of create a new file with the exact content.

pruan-rht: @liangxia those two files are different if you look carefully or diff them
```
$ diff example-fedora-kata.yaml webhook/example-fedora.yaml
4c4
<   name: example-fedora-kata
---
>   name: example-fedora
6c6
<     app: example-fedora-kata-app
---
>     app: example-fedora-app
15d14
<   runtimeClassName: kata
```

PR #2088, opened 2021-06-16
Fix resource type for NetworkPolicy

liangxia:
```
networkpolicies                       netpol           networking.k8s.io/v1                          true         NetworkPolicy
```

/cc @pruan-rht @asood-rh @jhou1


asood-rh (approved):

asood-rh (commented):
/lgtm

PR #2090, opened 2021-06-17
set resource check as false  instead of raising exception when 'Unabl…

pruan-rht:
…e to connect to the server` is encountered.  Please see full description in [OCPQE-4351](https://issues.redhat.com/browse/OCPQE-4351)


dis016 (commented):
/lgtm

PR #2091, opened 2021-06-17
Fix nc connection refused issue

weliang1:
Fix nc connection refused issue in https://issues.redhat.com/browse/OCPQE-2668

Current script sometimes cause nc connection refused due to nc -k option is used in the sctp server and nc -w option is used in sctp client.
New update delete both nc options and will continue trying open nc connection in both sctp server and client sides at the same time for 60 seconds to avoid randomly nc connection refused.

Continue testing new script in both ovn and sdn clusters for more than 1 hours, no connection failure happen, the script is more reliable now. 

Testing logs:
OVN: http://file.rdu.redhat.com/~weliang/ovn-log
SDN: http://file.rdu.redhat.com/~weliang/sdn-log

/cc @openshift/team-sdn-qe

### DIFF
--- a/features/networking/sctp.feature
+++ b/features/networking/sctp.feature
@@ -207,10 +207,10 @@ Feature: SCTP related scenarios
     """
 
     # Define a networkpolicy to deny sctpclient to sctpserver
-    Given I obtain test data file "networking/sctp/default-deny.yaml"
+    Given I obtain test data file "networking/networkpolicy/defaultdeny-v1-semantic.yaml"
     When I run the :create admin command with:
-      | f | default-deny.yaml   |
-      | n | <%= project.name %> |
+      | f | defaultdeny-v1-semantic.yaml   |
+      | n | <%= project.name %>            |
     Then the step should succeed
 
     # sctpclient pod start to send sctp traffic
--- a/features/networking/sctp.feature
+++ b/features/networking/sctp.feature
@@ -28,21 +28,20 @@ Feature: SCTP related scenarios
     Then the step should succeed
     And the pod named "sctpclient" becomes ready
    
-    # sctpserver pod start to wait for sctp traffic
-    When I run the :exec background client command with:
-      | pod              | sctpserver                                |
-      | namespace        | <%= project.name %>                       |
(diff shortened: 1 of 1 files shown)


features/networking/sctp.feature
```diff
@@ -28,21 +28,20 @@ Feature: SCTP related scenarios
...
+    And I wait up to 60 seconds for the steps to pass:
+    """
     # sctpserver pod start to wait for sctp traffic
     When I run the :exec background client command with:
-      | pod              | sctpserver                                |
-      | namespace        | <%= project.name %>                       |
-      | oc_opts_end      |                                           |
-      | exec_command     | bash                                      |
-      | exec_command_arg | -c                                        |
-      | exec_command_arg | /usr/bin/nc -k -l 30102 --sctp            |
```
asood-rh, line 38: Why was -k used to begin with?

asood-rh: @weliang1 I also see default deny policy is used in the test, seems this policy resides in sctp folder when we have a similar policy YAML file in testdata/networking/networkpolicy/ folder. Any reason we need two YAML files for same policy?

weliang1: > Why was -k used to begin with?

@asood-rh I think that's the old version you looked at

weliang1: > @weliang1 I also see default deny policy is used in the test, seems this policy resides in sctp folder when we have a similar policy YAML file in testdata/networking/networkpolicy/ folder. Any reason we need two YAML files for same policy?

@asood-rh Good catch, update script to use a common policy file from networking/networkpolicy/defaultdeny-v1-semantic.yaml

weliang1: > > @weliang1 I also see default deny policy is used in the test, seems this policy resides in sctp folder when we have a similar policy YAML file in testdata/networking/networkpolicy/ folder. Any reason we need two YAML files for same policy?
> 
> @asood-rh Good catch, update script to use a common policy file from networking/networkpolicy/defaultdeny-v1-semantic.yaml

New testing log: http://file.rdu.redhat.com/~weliang/ovn-log-update

PR #2093, opened 2021-06-18
Fixed the error associated with OCP-21100

ChenruiGu0107:
As for OCP-21100, the script for the automation test creating an advanced OSD cluster, there quite a few errors that prevent the automation from successfully running, I fixed those. The first things I fixed are some small errors where the name is not up-to-date, such as I fixed the outdated name "Location" to the new name "Region". 

The next thing is I fixed the automation go to the add-on tab in the cluster overview page and select the appropriate add-on and install it. The script works only for the old UI and now it needs updates.

I also fixed the place where the add-on failed and we need to check whether there is a "Contact Support". Again, the script only works for the old UI, it needs updates

Next, the automation can't successfully selected the correct button for selecting the Machine Type when creating a OSD cluster, I also fixed this.

I also fixed the waiting time while trying to log in. Sometimes the automation cannot log in when the waiting time is short.

### DIFF
--- a/lib/rules/web/ocm_console/cluster_detail_addons_tab.xyaml
+++ b/lib/rules/web/ocm_console/cluster_detail_addons_tab.xyaml
@@ -17,9 +17,13 @@ addon_install_button_missing:
         <<: *addon_install_button
         missing: true
 click_addon_install_button:
-    element:
-        <<: *addon_install_button
-        op: click
+    elements:
+        - selector:
+            xpath: //div[text()='<addon_name>']
+          op: click
+        - selector:
+            xpath: //button[contains(@aria-label,'Install')]
+          op: click
 install_addon:
     action: click_add_ons_tab_button
     action: click_addon_install_button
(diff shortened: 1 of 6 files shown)


lib/rules/web/ocm_console/osd_creation_page.xyaml
```diff
@@ -56,15 +56,19 @@ select_compute_node_count_on_creation_page:
...
 select_machine_type:
-    element:
-        <<: *machine_type_locator
-        op: click
+    elements: 
+        - selector:
+            xpath: //*[text()='Worker node instance type']/../../..//button
+          op: click
+        - selector:
+            xpath: //span[text()='<machine_type>']
```
xueli181114, line 71: Not click here after select the machine_type

lib/rules/web/ocm_console/cluster_detail_addons_tab.xyaml
```diff
@@ -63,9 +67,12 @@ check_console_url:
            xpath: //div[text()='<addon_name>']/../div/a/button[text()='View in console']
 
 check_contact_support:
-    element:
-        selector:
-            xpath: //div[text()='<addon_name>']/..//a[@href='https://access.redhat.com/support/cases/#/case/new' and text()='Contact support']
+    elements:
+        - selector:
+            xpath: //div[text()='<addon_name>']
```
xueli181114, line 72: As so many place need to click add-on tab first, you'd better to make click add-on name as am action and then write it like below in other action calls.
action: click_addon_card

PR #2096, opened 2021-06-19
more verbose eval print

pruan-rht:
Added support for more verbose printing when eval is called and the result is false.

Current printout:
```
    Then the expression should be true> cb.tmp.size > 5                       # features/step_definitions/common.rb:133
      expression "cb.tmp.size > 5" returned non-positive status: false
```

New more verbose printout:
```
      expression "cb.tmp.size > 5" returned non-positive status: false
      left_operand: 5, right_operand: 5
```


dis016 (commented):
/lgtm

PR #2097, opened 2021-06-21
rules updated

yapei:
/assign @yanpzhan


lib/rules/web/admin_console/4.8/goto.xyaml
```diff
@@ -313,7 +313,10 @@ goto_operator_all_instance_page:
   action: wait_box_loaded
 
 
-
+# machines
+goto_machines_page:
+  url: /k8s/ns/openshift-machine-api/machine.openshift.io~v1beta1~Machine
+  action: wait_box_loaded
```
yapei, line 319: use `goto_machines_page` instead of clicking navigation

PR #2098, opened 2021-06-21
Keep dir /var/tmp/ in image cucushift-public

liangxia:
/cc @pruan-rht @jhou1 @JianLi-RH @dis016


dis016 (commented):
/lgtm

pruan-rht:
@liangxia what's the reason for keeping that?  Just curious...

liangxia:
> @liangxia what's the reason for keeping that? Just curious...

@pruan-rht 
Without the fix,
```
$ podman run -it --rm cucushift-public:latest bash
Cannot create a temporary file: /var/tmp/scljmGEQq
```

PR #2101, opened 2021-06-21
formatting changes and use '&&' instead of 'and' for comparison

pruan-rht:
found these issues while doing PR https://github.com/openshift/verification-tests/pull/2096


anuragthehatter:
Thanks

/lgtm

PR #2103, opened 2021-06-22
fix some rules

yapei:
feature update https://github.com/openshift/cucushift/pull/8611

### DIFF
--- a/lib/rules/web/admin_console/4.5/operator_hub.xyaml
+++ b/lib/rules/web/admin_console/4.5/operator_hub.xyaml
@@ -353,37 +353,55 @@ wait_filter_section_loaded:
 choose_one_checkbox:
   element:
     selector:
-      xpath: //input[contains(@class, 'pf-c-check__input') and @title='<checkbox_text>']
+      xpath: //input[contains(@class, 'pf-c-check__input') and contains(@title,'<checkbox_text>')]
     op: click
-check_label_and_value_in_modal:
+check_property_label_and_value:
   element:
     selector:
-      xpath: //h5[contains(@class,'properties-side-panel-pf-property-label') and text()='<label>']/../div[contains(@class,'properties-side-panel-pf-property-value')]/div[text()='<value>']
+      xpath: //h5[contains(@class,'property-label') and contains(text(),'<label>')]/../div[contains(@class,'property-value')]/div[contains(text(),'<value>')]
+check_property_label_missing:
+  element:
+    selector:
+      xpath: //h5[contains(@class,'property-label') and contains(text(),'<label>')]
+    missing: true
+check_validsubscription_property_missing:
+  params:
+    label: Valid Subscriptions
+  action: check_property_label_missing
+check_infrastructure_property_missing:
+  params:
+    label: Infrastructure features
+  action: check_property_label_missing
 check_disconnected_infra_value:
   params:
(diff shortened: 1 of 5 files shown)


lib/rules/web/admin_console/4.5/operator_hub.xyaml
```diff
@@ -353,12 +353,12 @@ wait_filter_section_loaded:
...
   element:
     selector:
-      xpath: //input[contains(@class, 'pf-c-check__input') and @title='<checkbox_text>']
+      xpath: //input[contains(@class, 'pf-c-check__input') and contains(@title,'<checkbox_text>')]
     op: click
 check_label_and_value_in_modal:
   element:
     selector:
-      xpath: //h5[contains(@class,'properties-side-panel-pf-property-label') and text()='<label>']/../div[contains(@class,'properties-side-panel-pf-property-value')]/div[text()='<value>']
+      xpath: //h5[contains(@class,'properties-side-panel-pf-property-label') and text()='<label>']/../div[contains(@class,'properties-side-panel-pf-property-value')]/div[contains(text(),'<value>')]
```
JianLi-RH, line 361: The xpath looks a bit complicated, but it should works. Can it be shortened?

lib/rules/web/admin_console/4.8/operator_hub.xyaml
```diff
@@ -383,12 +383,12 @@ wait_filter_section_loaded:
...
   element:
     selector:
-      xpath: //input[contains(@class, 'pf-c-check__input') and @title='<checkbox_text>']
+      xpath: //input[contains(@class, 'pf-c-check__input') and contains(@title,'<checkbox_text>')]
     op: click
 check_label_and_value_in_modal:
   element:
     selector:
-      xpath: //h5[contains(@class,'properties-side-panel-pf-property-label') and text()='<label>']/../div[contains(@class,'properties-side-panel-pf-property-value')]/div[text()='<value>']
+      xpath: //h5[contains(@class,'properties-side-panel-pf-property-label') and text()='<label>']/../div[contains(@class,'properties-side-panel-pf-property-value')]/div[contains(text(),'<value>')]
```
yanpzhan, line 391: How about also update to use "contains" for the label text?

lib/rules/web/admin_console/4.5/operator_hub.xyaml
yapei: I think so, good idea @JianLi-RH  let me update

lib/rules/web/admin_console/4.8/operator_hub.xyaml
yapei: Yeah I agree @yanpzhan  let me update

JianLi-RH: Please be careful with the contais method @yanpzhan and @yapei , if there are some similar labels, it may cause unexpected issue. In this case, all labels are different, so don't worry.

yapei: yeah, we only use `contains` in very certain conditions @JianLi-RH

yapei:
@liangxia  Could you help approve?

pruan-rht (commented):
/lgtm

PR #2104, opened 2021-06-22
Automate OCP-37564/OCP-24546/OCP-24572

duanwei33:
pass log: 
[OCP-37564](https://privatebin-it-iso.int.open.paas.redhat.com/?73dcee5ba6be18d5#GWaVy7BRMdqEjvrJMXACnhtRm2HVeN9m8uv5X7vxvQui) 
[OCP-24546/OCP-24572](https://privatebin-it-iso.int.open.paas.redhat.com/?e55679f92b5a8649#AVAuSYRNGGPAbBno33V3dMK6RWa6beDwzEi38yFrWsxv)
        
@chao007 @jhou1 PTAL

### DIFF
--- a/features/storage/csi.feature
+++ b/features/storage/csi.feature
@@ -197,4 +197,82 @@ Feature: CSI testing related feature
       | standard-csi  | ext4   | # @case_id OCP-37558
 
 
+  # @author wduan@redhat.com
+  Scenario Outline: CSI dynamic provisioning with block
+    Given I have a project
+    Given I obtain test data file "storage/misc/pvc.json"
+    When I create a dynamic pvc from "pvc.json" replacing paths:
+      | ["metadata"]["name"]         | mypvc     |
+      | ["spec"]["storageClassName"] | <sc_name> |
+      | ["spec"]["volumeMode"]       | Block     |
+    Then the step should succeed
+    Given I obtain test data file "storage/misc/pod-with-block-volume.yaml"
+    When I run oc create over "pod-with-block-volume.yaml" replacing paths:
+      | ["metadata"]["name"]                                         | mypod       |
+      | ["spec"]["volumes"][0]["persistentVolumeClaim"]["claimName"] | mypvc       |
+      | ["spec"]["containers"][0]["volumeDevices"][0]["devicePath"]  | /dev/dblock |
+    Then the step should succeed
+    And the pod named "mypod" becomes ready
+    And the "mypvc" PVC becomes :bound
+    When I execute on the pod:
+      | sh | -c | [[ -b /dev/dblock ]] |
+    Then the step should succeed
+    When I execute on the pod:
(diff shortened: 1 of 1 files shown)


features/storage/csi.feature
```diff
@@ -197,4 +197,81 @@ Feature: CSI testing related feature
...
+    Given I obtain test data file "storage/misc/deployment.yaml"
+    When I run oc create over "deployment.yaml" replacing paths:
+      | ["metadata"]["name"]                                                             | mydep      |
+      | ["spec"]["template"]["metadata"]["labels"]["action"]                             | storage    |
+      | ["spec"]["template"]["spec"]["volumes"][0]["persistentVolumeClaim"]["claimName"] | mypvc      |
+      | ["spec"]["template"]["spec"]["containers"][0]["volumeMounts"][0]["mountPath"]    | /mnt/local |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | action=storage |
+    And the "mypvc" PVC becomes :bound
```
chao007, line 262: Suggest to check pv volume size and volume type

features/storage/csi.feature
```diff
@@ -197,4 +197,81 @@ Feature: CSI testing related feature
...
+      | cp | /hello | /mnt/local |
+    Then the step should succeed
+    When I execute on the pod:
+      | /mnt/local/hello |
+    Then the step should succeed
+    And the output should contain "Hello OpenShift Storage"
+
+    Examples:
+      | sc_name | type | size  |
+      | gp2-csi | sc1  | 125Gi | # @case_id OCP-24546
```
jhou1, line 276: Is 125Gi too big? Do our cloud providers have enough capacity to support the stability of such tests?

features/storage/csi.feature
duanwei33: @chao007, thanks for the suggestion, seems the volume type like sc/st1 doesn't exist in the pvc or pv into. I added the check for pvc size and passed. 

>         And the "mypvc" PVC becomes :bound                                           # features/step_definitions/pvc.rb:3
>       [03:10:08] INFO> 
>       [03:10:08] INFO> After 1 iterations and 2 seconds:
>       matched status for persistentvolumeclaims mypvc: 'bound' while expecting '[:bound]'
>         And the expression should be true> pvc.capacity == "125Gi"                   # features/step_definitions/common.rb:133
> waiting for operation up to 3600 seconds..
>         When I execute on the pod:                                                   # features/step_definitions/pod.rb:216
>           | sh | -c | echo "test" > /mnt/local/testfile |

features/storage/csi.feature
duanwei33: 125Gi is the minimum value for sc1/st1 type.  Currently we have enough capacity with our test account.

duanwei33:
@chao007 @jhou1 Hi, please let me know if any other comments after the update, thanks.

PR #2106, opened 2021-06-23
Adding test for BZ-1953063

miyadav:
@sunzhaohua2 @jhou1  , PTAL ..

This is to cover test for -https://bugzilla.redhat.com/show_bug.cgi?id=1953063 

Here are the results : https://url.corp.redhat.com/42346


pruan-rht (commented):
/lgtm

PR #2108, opened 2021-06-23
rewrite the case OCP-25353

yasun1:
Update the rules for OCP-25353. Testing is passed on local with the log http://pastebin.test.redhat.com/974067


pruan-rht (commented):
/lgtm

PR #2114, opened 2021-06-25
fix destructive rules

yapei:
feature updates PR cucushift#8620


yapei:
@liangxia  Could you help approve if it looks good to you?

PR #2117, opened 2021-06-25
Add steps to modify the policy post upgrade to verify it works well

asood-rh:
The network policy is tweaked to change the pod labels to ensure ACLs are recreated if the policy is modified.
Pod label controlled by policy is hello-idle instead of test-pods.
pod4->pod3ip passed prior to modification but fails as expected
pod5->pod2ip failed but succeeds post modification.
pod5->pod3ip (positive test case step)
Testing:-
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3206/console (Upgrade prepare)
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3206/console (Post upgrade)

/cc @openshift/team-sdn-qe

### DIFF
--- a/features/upgrade/sdn/policy-upgrade.feature
+++ b/features/upgrade/sdn/policy-upgrade.feature
@@ -209,3 +209,29 @@ Feature: SDN compoment upgrade testing
     Then the step should succeed 
     And the output should contain "Hello"
     """
+    #Steps to modify policy post upgrade for bug 1973679
+    When I obtain test data file "networking/networkpolicy/allow-ns-and-pod.yaml"
+    And I replace lines in "allow-ns-and-pod.yaml":
+      | test-pods | hello-idle |
+    And I run the :replace admin command with:
+      | f | allow-ns-and-pod.yaml |
+      | n | policy-upgrade1       |
+    Then the step should succeed
+
+    And I wait up to 10 seconds for the steps to pass:
+    """
+    When I execute on the "<%= cb.pod4 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod3ip %>:8080 |
+    Then the step should fail
+    And the output should not contain "Hello"
+    When I execute on the "<%= cb.pod5 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod2ip %>:8080 |
+    Then the step should succeed
+    And the output should contain "Hello"
+    When I execute on the "<%= cb.pod5 %>" pod:
+      | curl | -s | --connect-timeout | 5 | <%= cb.pod3ip %>:8080 |
+    Then the step should succeed
+    And the output should contain "Hello"
+    """
+
+
--- a/features/upgrade/sdn/policy-upgrade.feature
+++ b/features/upgrade/sdn/policy-upgrade.feature
@@ -233,5 +233,3 @@ Feature: SDN compoment upgrade testing
     Then the step should succeed
     And the output should contain "Hello"
     """
-
-


anuragthehatter:
@asood-rh Looks good to me. Is there any way you can obtain logs for this change? You don't need to upgrade for this but let say just run on OVN cluster. Can you also check consistency of this use case like running it 2-3 times post upgrade back to back. I have seen issues in past regarding network policy not clearing up past ACLs and messing up with next due to caching issues.

asood-rh:
@anuragthehatter A separate test cases can be added for caching issue where we the network policy is changed in a loop. I am not sure of value of obtaining the logs when the policy change is validated by accessing the pods. If you could approve this PR I ca add a test case that validates the modified ACL due to policy change. The test case can also be added as part of ACL logging  if validation needs to be done with looking at logs.

anuragthehatter:
> @anuragthehatter A separate test cases can be added for caching issue where we the network policy is changed in a loop. I am not sure of value of obtaining the logs when the policy change is validated by accessing the pods. If you could approve this PR I ca add a test case that validates the modified ACL due to policy change. The test case can also be added as part of ACL logging if validation needs to be done with looking at logs.

Ahh you already had it there https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3206/console. Thats what I was trying to find. Overlooked it. Thanks. yea so we already have a caching case on network policy. But my concern was running post-upgrade check couple of times to make sure it passes every time (Not in automation script but just while verifying your script).

asood-rh:
@anuragthehatter When this test starts running as part of upgrade it will be watched.  Are you suggesting running this test manually? Is the PR review contingent on manual test? Thanks!

asood-rh:
@anuragthehatter I have executed the post upgrade part of the script several times after reverting the policy back to original state and then running the script that changes the policy. It passed all the time locally. Let me know if understood the manual test that you requested.
Original policy
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: allow-ns-and-pod
spec:
  podSelector:
    matchLabels:
      name: test-pods
  ingress:
    - from:
      - namespaceSelector:
          matchLabels:
            team: operations
        podSelector:
          matchLabels:
            name: test-pods
Changed policy replaces test-pods with hello-rc.

anuragthehatter:
> @anuragthehatter When this test starts running as part of upgrade it will be watched. Are you suggesting running this test manually? Is the PR review contingent on manual test? Thanks!

So such tests have been found flaky in past, not necessarily this. Specially the network policy cases post upgrade.There were bugs reported by edge team that and those issues considered to be as flaky. Just wanted to make sure if you run it multiple times it doesn't flake due to a reason which is difficult to catch. Anyways that looks good,

anuragthehatter:
> @anuragthehatter I have executed the post upgrade part of the script several times after reverting the policy back to original state and then running the script that changes the policy. It passed all the time locally. Let me know if understood the manual test that you requested.
> Original policy
> kind: NetworkPolicy
> apiVersion: networking.k8s.io/v1
> metadata:
> name: allow-ns-and-pod
> spec:
> podSelector:
> matchLabels:
> name: test-pods
> ingress:
> - from:
> - namespaceSelector:
> matchLabels:
> team: operations
> podSelector:
> matchLabels:
> name: test-pods
> Changed policy replaces test-pods with hello-rc.

Thanks for checking locally to avoid any flakes.

PR #2118, opened 2021-06-26
added support to remove all kata pods in a cluster

pruan-rht:
1. optimized pods in all namespace for kata. It was printing out all of the pods in all namespaces which resulted in a massive print so the run log is over 4Meg as a result.  Now we filter out using jsonpath, using two `oc` calls.  Not sure if it's possible to get multiple values in one `oc` call
2. added methods for getting `metadata.ownerReferences` for project resource objects.  This is needed to determine the proper resource to remove in order to delete a pod.  For example, pods that are generated via `Deployment` or `DeploymentConfig`, we'll need to delete that resource instead of the kata pod to prevent it from regenerated

3. Added extra duty during the remove kataconfig step to remove all existing kata-pods in the cluster instead of raising an exception and exit.

runlog: job/Runner/91382/console


dis016 (commented):
/lgtm

PR #2127, opened 2021-06-29
Change location of policy files to public repo and use multus pod from naster node

asood-rh:
The network policies moved from private repo to public repo as these are going to be used for upgrade scripts.
Added a step to obtain multus pod on master nodes to ensure allow from hostnetwork policy is actually tested.
If the test pod end up on same node from where multus pod is scheduled then the network policy does not have anything to do so it is not really tested. This step makes sure multus pod used for testing is running on one of master node.

Test logs can be found here:-
vSphere
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3223/console OCP-41944
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3224/console (OCP-40898 and OCP-40899)

https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3225/console (OCP-40898 and OCP-40899)
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3228/console  OCP-40908

/cc @openshift/team-sdn-qe

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1322,7 +1322,7 @@ end
 
 Given /^I save multus pod on master node to the#{OPT_SYM} clipboard$/ do | cb_name |
   ensure_admin_tagged
-  cb_name = "multus_pod" unless cb_name
+  cb_name ||= :multuspod
   master_nodes = env.nodes.select { |n| n.schedulable? && n.is_master? }
   master_node_names = master_nodes.collect { |n| n.name }
   cb[cb_name] = BushSlicer::Pod.get_labeled("app=multus", project: project("openshift-multus", switch: false), user: admin) { |pod, hash|
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1322,6 +1322,7 @@ end
 
 Given /^I save multus pod on master node to the#{OPT_SYM} clipboard$/ do | cb_name |
   ensure_admin_tagged
+  cb_name = "multus_pod" unless cb_name
   master_nodes = env.nodes.select { |n| n.schedulable? && n.is_master? }
   master_node_names = master_nodes.collect { |n| n.name }
   cb[cb_name] = BushSlicer::Pod.get_labeled("app=multus", project: project("openshift-multus", switch: false), user: admin) { |pod, hash|
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1319,3 +1319,13 @@ Given /^the node's active nmcli connection is stored in the#{OPT_SYM} clipboard$
   cb[cb_name] = @result[:response].split("\n").first
   logger.info "Node's active nmcli connection uuid is stored in the #{cb_name} clipboard as #{cb[cb_name]}"
 end
+
+Given /^I save multus pod on master node to the#{OPT_SYM} clipboard$/ do | cb_name |
+  ensure_admin_tagged
+  master_nodes = env.nodes.select { |n| n.schedulable? && n.is_master? }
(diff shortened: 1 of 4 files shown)


features/step_definitions/networking.rb
```diff
@@ -1319,3 +1319,13 @@
   cb[cb_name] = @result[:response].split("\n").first
   logger.info "Node's active nmcli connection uuid is stored in the #{cb_name} clipboard as #{cb[cb_name]}"
 end
+
+Given /^I save multus pod on master node to the#{OPT_SYM} clipboard$/ do | cb_name |
```
pruan-rht, line 1323: @asood-rh, I think you need to handle the case in which the user didn't specific the clipboard name, you need to default it to something.

asood-rh: @pruan-rht If the clipboard variable name is omitted, the user would get an error as below:-
undefined method `to_sym' for nil:NilClass (NoMethodError)
      /home/asood/github-projects/verification-tests/features/step_definitions/networking.rb:1327:in `/^I save multus pod on master node to the(?: :(\S+))? clipboard$/'
      features/tierN/networking/network-policy.feature:2351:in `I save multus pod on master node to the clipboard'
Also script uses the variable for validation. It is required. Do you still need default?

asood-rh: @pruan-rht 
Anyways, incorporated your feedback
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3240/console

features/step_definitions/networking.rb
```diff
@@ -1322,6 +1322,7 @@
 
 Given /^I save multus pod on master node to the#{OPT_SYM} clipboard$/ do | cb_name |
   ensure_admin_tagged
+  cb_name = "multus_pod" unless cb_name
```
pruan-rht, line 1325: @asood-rh you can use this format instead `cb_name ||= "multus_pod"` or ` cb_name ||= :multus_pod`

pruan-rht: @asood-rh seeing that you are calling it like ```Given I save multus pod on master node to the :multuspod clipboard``` so to be consistent you can change that statement to `cb_name ||= :multspod`

features/step_definitions/networking.rb
asood-rh: @pruan-rht How does providing default value helps?

features/step_definitions/networking.rb
asood-rh: Done - https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3241/console

features/step_definitions/networking.rb
pruan-rht: @asood-rh since your step definition is using the regexp `OPT_SYM`, which mean `Optional Symbol`,  you give the user the choice of not putting in the clipboard name.  So you need to handle that situation when the user use the step `Given I save multus pod on master node to the clipboard` which is acceptable according to your step definition.

asood-rh: @pruan-rht  Got it, I could also it required instead.

pruan-rht: @asood-rh no need, handling the optional `cb_name` will take care of it.  It's fine to leave as it's

PR #2132, opened 2021-07-01
Fix OCP-33058

sunzhaohua2:
increase the time to create the machineset @miyadav @jhou1 please help to take a look.
```
      [04:13:04] INFO> === End After Scenario: Implement defaulting machineset values for azure ===
2 scenarios (2 passed)
30 steps (30 passed)
15m16.543s
```


pruan-rht (commented):
/lgtm

PR #2137, opened 2021-07-02
method `query_test_cases_smart` only accepts 4 parameters `[:project_…

pruan-rht:
…id, :case_query, :template_id, :custom_fields]`, need to delete the other keys from the input hash when doing the query.  @eurijon this should fix the issue you mentioned so you don't need to modify the input YAML to remove the un-supported keys.


eurijon:
/lgtm

I've tested the PR locally and it worked, thanks @pruan-rht for fixing it!

PR #2152, opened 2021-07-07
Fix the v3 script to use the same args with 4.x

duanwei33:
1. Update the fsType.feature to make sure the args are as same as in the polarion
2. Update the cleanup step, in previous code, only deleting pv (which is bound to pvc) doesn't work. It always hung for the timeout.

[pass log](https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3282/console)

@chao007 @jhou1 PTAL


duanwei33:
@liangxia @pruan-rht Could you help take a look? Thanks.

PR #2153, opened 2021-07-08
Add dockerfile for unify CI

liangxia:
@jhou1 @JianLi-RH @dis016

### DIFF
--- a/tools/openshift-ci/Dockerfile
+++ b/tools/openshift-ci/Dockerfile
@@ -1,0 +1,36 @@
+FROM quay.io/centos/centos:7
+
+LABEL vendor="Red Hat inc."
+LABEL maintainer="AOS QE Team"
+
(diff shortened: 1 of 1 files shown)


tools/openshift-ci/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM quay.io/centos/centos:7
+
+LABEL vendor="Red Hat inc."
+LABEL maintainer="AOS QE Team"
```
pruan-rht, line 4: perhaps we should change it to `OCP QE Team` instead... `AOS` is a legacy name that's not commonly used anymore.

liangxia: @pruan-rht Fix that in https://github.com/openshift/verification-tests/pull/2154

PR #2156, opened 2021-07-12
Remove smoke tag

duanwei33:
Remove some cases' smoke tag which are duplicated or not suitable or no need for 3.11 release test.


duanwei33:
@chao007 PTAL. thanks.

PR #2160, opened 2021-07-12
Adding cucushift

jhou1:
The image built from cucushift repo will be used as the base image for this one, I tested this locally by replacing the `FROM` with cucushift image. 

@liangxia PTAL, thank you!

### DIFF
--- a/tools/openshift-ci/Dockerfile
+++ b/tools/openshift-ci/Dockerfile
@@ -3,10 +3,12 @@ FROM quay.io/centos/centos:7
 LABEL vendor="Red Hat inc."
 LABEL maintainer="OCP QE Team"
 
-WORKDIR /root
 USER root
 
+ADD . /verification-tests/
+
 RUN set -x && \
+    mv /tierN/ /verification-tests/tierN/ && \
     yum -y update && \
     SCL_BASE_PKGS="centos-release-scl scl-utils-build" && \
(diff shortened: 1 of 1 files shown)


tools/openshift-ci/Dockerfile
```diff
@@ -3,34 +3,26 @@ FROM quay.io/centos/centos:7
...
+ADD . /verification-tests/
+
+RUN set -x && \
+    mv /tierN/ /verification-tests/tierN/
+
 RUN set -x && \
     yum -y update && \
     SCL_BASE_PKGS="centos-release-scl scl-utils-build" && \
-    INSTALL_PKGS="rh-ruby27 rh-ruby27-ruby-devel bsdtar git openssh-clients" && \
+    INSTALL_PKGS="rh-ruby26 rh-ruby26-ruby-devel rh-git218 bsdtar" && \
```
liangxia, line 16: Perhaps use ruby27 ? as the image we currently use is using ruby27.

liangxia:
/lgtm

/hold
in case you do not want to merge it now.

tools/openshift-ci/Dockerfile
jhou1: Thanks Liang, fixed that

jhou1:
Thanks, Liang, it's fine to unhold. Current openshift/release is building the Dockerfile.centos

PR #2165, opened 2021-07-16
allow user to specify how much information to extract from a test-run

pruan-rht:
This will allow user to get more information about a run.


dis016 (commented):
/lgtm

PR #2172, opened 2021-07-21
Execute all multicast cases in verification-tests

weliang1:
Multicast is a k8s upstream feature, it makes no sense to keep openshift multicast cases separately in cucushift and verification-tests.

This PR will make below changes:
1. Copy OCP-12928/OCP-12929/OCP-12966 from cucushift to verification-tests, after this PR merging and updating Automation path in polarion, all the multicast cases will be executed from verification-tests.
2. Update OCP-12928/OCP-12929/OCP-12966  to be executed in both SDN and OVN cluster, right now these three cases can only be executed in SDN.
3. Update OCP-12928/OCP-12929/OCP-12966  to expect not all 5 packets lost in the client side due to Bug 1879241 - [OVN] Pods always lose one multicast packets
4.  Update OCP-12928/OCP-12929/OCP-12966  to allow 10 seconds for the multicast igmp group information shown up when execute "netstate -ng"
5. Add a step to disable multicast in default namespace for  OCP-12928 and OCP-12966
6. Fix indentations in all the cases

SDN log:
http://file.rdu.redhat.com/~weliang/multicast-SDN
OVN log:
http://file.rdu.redhat.com/~weliang/multicast-OVN

/cc @openshift/team-sdn-qe

### DIFF
--- a/features/networking/multicast.feature
+++ b/features/networking/multicast.feature
@@ -52,10 +52,10 @@ Feature: testing multicast scenarios
 
     # check the omping result on third pod
     When I run the :exec background client command with:
-      | pod              | <%= cb.pod3 %> |
-      | oc_opts_end      |                |
-      | exec_command     | sh             |
-      | exec_command_arg | -c             |
+      | pod              | <%= cb.pod3 %>                                                                     |
+      | oc_opts_end      |                                                                                    |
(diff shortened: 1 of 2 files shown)


pruan-rht (commented):
/lgtm

PR #2176, opened 2021-07-22
commit for automation testcase ocp42112

XiyunZhao:
commit for automation testcase ocp42112


lib/rules/web/admin_console/4.8/deployment.xyaml
```diff
@@ -149,4 +149,31 @@ select_replica_count_and_selector:
   params:
     text: Replica count and selector
   action: click_checkbox_input_by_text
-
+check_workloads_metrics:
```
yapei, line 152: We can add `action: click_metrics_tab` into `check_workloads_metrics` 
```
check_workloads_metrics:
  action: click_metrics_tab
  xxxxxxx
```

XiyunZhao: Done

PR #2180, opened 2021-07-27
update channel name and fix some issue when deleting clusterlogging

QiaolingTang:
Set default channel to `stable-5.2` when testing logging on OCP 4.9 and fix some issues when deleting clusterlogging.
runner job log: https://url.corp.redhat.com/a52ef61

/cc @anpingli

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -340,19 +340,27 @@ Given /^I delete the clusterlogging instance$/ do
   if cluster_logging("instance").exists?
     @result = admin.cli_exec(:delete, object_type: 'clusterlogging', object_name_or_id: 'instance', n: 'openshift-logging')
     raise "Unable to delete instance" unless @result[:success]
-  end
-  step %Q/I wait for the resource "deployment" named "kibana" to disappear/
-  step %Q/I wait for the resource "elasticsearch" named "elasticsearch" to disappear/
-  step %Q/I wait for the resource "cronjob" named "curator" to disappear/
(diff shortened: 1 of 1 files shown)


anpingli (commented):
/lgtm

features/step_definitions/logging.rb
```diff
@@ -340,19 +340,27 @@
...
-    | component=elasticsearch |
-  })
-  step %Q/all existing pods die with labels:/, table(%{
-    | component=kibana |
-  })
-  if BushSlicer::PersistentVolumeClaim.list(user: user, project: project).count > 0
-    admin.cli_exec(:delete, object_type: 'pvc', l: 'logging-cluster=elasticsearch', n: 'openshift-logging')
+    unless cluster_logging("instance").log_store_spec(cached: true).nil?
+      step %Q/I wait for the resource "elasticsearch" named "elasticsearch" to disappear/
+      success = wait_for(180, interval: 10) {
```
pruan-rht, line 345: @QiaolingTang why do you `wait` again here?  The previous step does that already, correct?

QiaolingTang: @pruan-rht no, the previous step is to wait for CR elasticsearch/elasticsearch to disappear, these steps wait for the deployments which have label `cluster-name=elasticsearch` to disappear. 
I had hit an issue: the elasticsearch/elasticsearch was deleted, but the ES pods were not. So here I want to add these steps to check if the deployments can be removed or not, since the deployment names are dynamic, so I use label here.

pruan-rht (commented):
/lgtm

PR #2186, opened 2021-07-30
Update pod image for sctp testing

weliang1:
Currently, sctp use ubi image plus ["dnf install -y nc && sleep inf"] for sctpserver and sctpclient pod, in disconnected cluster, pod can not execute  ["dnf install -y nc && sleep inf"] and sctp testing will fail in disconnected cluster.

This PR is to update pod image to use quay.io/openshifttest/centos-network which include nc utility being used for sctp testing.

Test log from a disconnected cluster/edge50 bare metal machine:
http://file.rdu.redhat.com/~weliang/sctp.log

/cc @openshift/team-sdn-qe

### DIFF
--- a/testdata/networking/sctp/sctpclient.yaml
+++ b/testdata/networking/sctp/sctpclient.yaml
@@ -7,7 +7,4 @@ metadata:
 spec:
   containers:
     - name: sctpclient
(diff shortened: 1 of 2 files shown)


pruan-rht (commented):
/lgtm

PR #2189, opened 2021-08-03
OCP-11286: don't restart kubelet to avoid test failures

rbbratta:
restarting kubelet seems to be an old 3.x step
that seems to fail frequently and might be of dubious utility

remove the kubelet restart step and just restart the network service.

The iptables rules should be re-created already as part of the test.

Also replace `puts` with `logger.info` as per deprecation notice.


rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2021-08-03-12-54-40.log

PR #2192, opened 2021-08-04
OCP-41926: view past logs

schituku:
Runner:
 https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/ocp-common/job/Runner/131408/console


lib/rules/web/admin_console/4.8/logs.xyaml
```diff
@@ -106,5 +106,31 @@ toggle_log_wraps:
...
     op: click
+check_logs_history_not_present:
+  elements:
+  - selector:
+      xpath: //button[@disabled]//span[text()='Current log']
+  - selector:
+      xpath: //div[@class='pf-c-select']
+    op: hover
+  - selector:
+      xpath: //div[@class='pf-c-tooltip__content' and contains(text(),'<message>')]
```
yapei, line 117: use `contains` when matching element classes since it can easily change

lib/rules/web/admin_console/4.8/logs.xyaml
```diff
@@ -106,5 +106,31 @@ toggle_log_wraps:
...
+check_logs_history_is_present:
+  elements:
+  - selector:
+      xpath: //button//span[text()='<log_current>']
+  - selector:
+      xpath: //div[@class='pf-c-select']
+    op: click
+  - selector:
+      xpath: //button[@class='pf-c-select__menu-item pf-m-selected' and contains(text(),'<log_current>')]
```
yapei, line 126: same comment as above

lib/rules/web/admin_console/4.8/logs.xyaml
```diff
@@ -106,5 +106,31 @@ toggle_log_wraps:
...
+  - selector:
+      xpath: //button[@class='pf-c-select__menu-item' and contains(text(),'<log_previous>')]
```
yapei, line 128: same as above

schituku:
@yapei I did not change the matching classes as it will give false positives. So it has to be exact match, if it changes later then we have to change accordingly. Any ways the rules will give the ability to maintain them per release so we can change per release.
  For example, in this test i want to make sure that "Current log" is Selected and "Previous log" is not selected. If I choose
contains then it will return true for @class='pf-c-select__menu-item' and @class='pf-c-select__menu-item pf-m-selected'. Basically it will pass for both selected and un-selected item. Hence not making the change.

lib/rules/web/admin_console/4.8/logs.xyaml
schituku: see combined comment for all class text matching comment below

lib/rules/web/admin_console/4.8/logs.xyaml
schituku: see combined comment for all class text matching comment below

lib/rules/web/admin_console/4.8/logs.xyaml
schituku: see combined comment for all class text matching comment below

yapei:
> @yapei I did not change the matching classes as it will give false positives. So it has to be exact match, if it changes later then we have to change accordingly. Any ways the rules will give the ability to maintain them per release so we can change per release.
> For example, in this test i want to make sure that "Current log" is Selected and "Previous log" is not selected. If I choose
> contains then it will return true for @Class='pf-c-select__menu-item' and @Class='pf-c-select__menu-item pf-m-selected'. Basically it will pass for both selected and un-selected item. Hence not making the change.

Ok, we can leave them as current

yapei (approved):
/lgtm

PR #2195, opened 2021-08-06
Deactivate some nfs tests

jhou1:
As a follow up of storage test cases review, marking tests as 'inactive' in automation as well. There polarion test case has been deactivated.
@duanwei33 @chao007 @ropatil010 @xgu-redhat @Phaow PTAL


duanwei33:
From my side, ReadWriteMany is key feature for nfs volume and is worthy to be tested, but this test case should be updated as it could not make sure the two pods are scheduled to different nodes.

jhou1:
I'm closing this PR as we conclude this is a needed scenario.

PR #2196, opened 2021-08-09
update scheme to tls

QiaolingTang:
Fix the scheme in URL when forwarding with TLS.

/cc @anpingli 

runner job log: https://url.corp.redhat.com/c6384ad


anpingli (commented):
/lgtm

PR #2200, opened 2021-08-11
Implementation of ovirt modules

santos1709:


pjanouse:
LGTM however I can't verify by myself.

liangxia:
@pruan-rht  PTAL

lib/launchers/ovirt.rb
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env ruby
+require 'ovirtsdk4'
```
pruan-rht, line 2: @santos1709 this doesn't look like a standard ruby gem, correct?  If that's the case, then it needs to be added to the top level Gemfile.

lib/launchers/ovirt.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    def trust_ovirt_certificate(ca_path=nil)
+      if ca_path
+        cmd = "cp #{ca_path} /tmp/ovirt.pem"
+      else
+        # ovirt credentials exposed here to get the ca, for testing only!
+        cmd = "curl -k '#{@pem_url}' -o /tmp/ovirt.pem"
+      end
+        system("#{cmd}")
```
pruan-rht, line 103: @santos1709 indentation issue.

lib/launchers/ovirt.rb
```diff
@@ -0,0 +1,125 @@
...
+      svc = @connection.system_service.send("#{service}_service")
+      if query
+        return svc.list(search: query)
+      else
+        return svc.list
+      end
+    end
+
+    def get_cluster_id(name)
+      cluster = get_resource(service='clusters', query="name=#{name}")[0]
```
pruan-rht, line 55: Guilherme De Oliveira Santos, do you think we need to check if the `cluster` returned is a valid object?...in order words, you are alway getting the first element of the `get_resource` return value...can it be empty?   if so then call to `cluster.id` will probably cause issue.

santos1709: I see @pruan-rht , in fact `get_resource` should return a list o elements and indeed the list can be empty (It will always return a list as long as it not fails). In this case (empty list), do you have any specific suggestion on how to handle it? I'm thinking on raise an error with a proper message

pruan-rht: @santos1709 probably check for the size of the return list to see if it's empty first instead to make sure it's not `nil`..otherwise you will get ```NoMethodError: undefined method `id' for nil:NilClass``` when you do cluster.id.   Whether or not to raise an exception depends on your use case.

santos1709: Okay, it should always return something or it will break ahead.

santos1709: @pruan-rht Changes done, once you have some time, PTAL, thanks!

santos1709: I see the job here fails when trying to install ovirt-engine-sdk gem. This gem needs the following requirements: `gcc`,
`libcurl-devel`, `libxml2-devel`. Maybe the job runner doesn't have one of them? @pruan-rht

tools/os_deps/deps.yum.RHEL8
```diff
@@ -21,4 +23,4 @@ sqlite-devel
 which
 Xvfb
 zlib
-zlib-devel
+izlib-devel
```
pruan-rht, line 26: @santos1709 what's the difference between this and the old `zllib-devel`

pruan-rht: @santos1709  the PR looks good, but since it changes the lib requirements will need to consult with others to make sure they are ok with adding these cc: @jhou1, @liangxia do you have objects to these library addition?

santos1709: @pruan-rht ohhh this is actually a typo

PR #2201, opened 2021-08-11
Removed inactive cases , they were failing

miyadav:
… and fails

https://url.corp.redhat.com/OCP-26248 , https://url.corp.redhat.com/OCP-26250 , https://url.corp.redhat.com/OCP-26110,https://url.corp.redhat.com/OCP-26110

@sunzhaohua2  @jhou1  PTAL


miyadav:
Thanks @jhou1 for review  , we have the new alerts now , for which we have tests , these alerts are no longer present at all . 
https://github.com/openshift/machine-api-operator/blob/master/install/0000_90_machine-api-operator_04_alertrules.yaml , will check further on this ..

sunzhaohua2:
/lgtm
thanks @miyadav, we can add some new cases for these alerts and automate them https://github.com/openshift/machine-api-operator/blob/master/install/0000_90_machine-api-operator_04_alertrules.yaml

PR #2212, opened 2021-08-24
Finish the osd cluster creation part with new wizard in OCM

xueli181114:


xueli181114:
https://github.com/openshift/cucushift/pull/8690

PR #2217, opened 2021-08-26
Automation case update for OCP-36912,OCP-37098,OCP-25765

XiyunZhao:
Update rules for OCP-36912,OCP-37098,OCP-25765


yapei (approved):
/lgtm

PR #2223, opened 2021-08-30
auto OCP-43052 and fixed cases and supplement OCP-26380

LiangquanLi930:
auto OCP-43052 and fixed OCP-42677, OCP-38887, OCP-25175, OCP-30121, OCP-26297 and supplement OCP-26380


yasun1:
As I said in https://github.com/openshift/cucushift/pull/8703, only 1 commit in a PR. So need to do squash.

PR #2224, opened 2021-08-30
Adding @aws-ipi tag to tests

jhou1:
@liangxia PTAL

Case query using https://gitlab.cee.redhat.com/jhou/ci-profiles/-/blob/prow/prow/query_49_aws_ipi_tests.json

### DIFF
--- a/features/admin/pod.feature
+++ b/features/admin/pod.feature
@@ -2,6 +2,7 @@ Feature: pod related features
 
   # @author xiuli@redhat.com
   # @case_id OCP-15808
+  @aws-ipi
   Scenario: Endpoints should update in time and no delay
     Given I have a project
     Given I obtain test data file "networking/list_for_pods.json"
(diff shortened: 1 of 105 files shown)


features/admin/pod.feature
```diff
@@ -102,6 +102,7 @@ Feature: pod related features
   # @case_id OCP-11752
   @admin
   @inactive
+  @aws-ipi
```
liangxia, line 105: @jhou1 Perhaps we do not need to add the tag when it already has tag inactive ?

jhou1: @liangxia I think if someone activates the test case and removes the `@inactive` tag from the code base, then the prow tags are out of sync. I suggest we still add the tags anyway in case someone wants to reenable them in the future. We would therefore have less effort keeping things up to date.
WDYT?

jhou1: I also made a mistake in the test case query, inactive test cases should not be filtered in my query anyway, fixing everything now...

liangxia: Make sense now.    I think the mistake makes most of the test cases added with tag aws-ipi also have inactive tag, that's why the concern comes from.

PR #2222, opened 2021-08-29
Kata tests refactoring and data collect feature

valtair1891:

PR #2229, opened 2021-08-31
Kata feature refactoring and OO framework

valtair1891:


pruan-rht:
@valtair1891 same comments from  https://github.com/openshift/verification-tests/pull/2222/files still needs to be addressed

valtair1891:
Breaking down to a smaller PRs

PR #2232, opened 2021-09-01
Add proxy tag

jhou1:
The `@proxy` tag needs to be added to distinguish tests that can only be run in a cluster proxy deployment.

@liangxia @pruan-rht PTAL


jhou1:
No destructive operations in this PR, fast merging to get things tested overnight.

PR #2239, opened 2021-09-02
Rewrite rules for the new ocm creation wizard

xueli181114:
/assign @yasun1

### DIFF
--- a/lib/rules/web/ocm_console/osd_wizard_cloud_provider.xyaml
+++ b/lib/rules/web/ocm_console/osd_wizard_cloud_provider.xyaml
@@ -6,20 +6,40 @@ select_cloud_provider_wizard:
             xpath: //article[@data-test-id='<cloud_provider>-provider-card']
         op: click
 
+click_validate_button:
+    element:
+        selector:
+            xpath: //button[text()='Validate']
+        op: click
 # -------------------- AWS credential related --------------------
 check_no_input_errors_to_required_aws_items:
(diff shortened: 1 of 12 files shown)


lib/rules/web/ocm_console/osd_wizard_cloud_provider.xyaml
```diff
@@ -6,20 +6,40 @@ select_cloud_provider_wizard:
...
-            xpath: //div[@id='access_key_id-helper' and text()=' Field is required']
+            xpath: //*[@id='access_key_id']
+          op: click
         - selector:
-            xpath: //div[@id='secret_access_key-helper' and text()=' Field is required']
-check_error_to_cluster_name_on_creation_page:
-    element:
-        selector:
-            xpath: //div[@id='name-helper' and contains(text(),'<error_message>')]
+            xpath: //*[text()='AWS access key ID']
```
yasun1, line 29: why using '*' but not 'span'?

yasun1 (commented):
others lgtm

lib/rules/web/ocm_console/osd_wizard_cloud_provider.xyaml
xueli181114: Because the block type changes too frequently.  And I just need some place to click to call out the error message. I don't want to debug failed cases all the time due to the block type changed. * can be used to locate a unique element. That's expected.

PR #2248, opened 2021-09-06
Fix lastTriggeredImageID issue

xiuwang:
lastTriggeredImageID change from spec to status
@wewang58 Please help review, thanks


pruan-rht (commented):
/lgtm

PR #2256, opened 2021-09-08
Make sure EXECUTOR_NAME always end with proper char

liangxia:
`EXECUTOR_NAME` is generated via below code in `lib/bushslicer.rb`. And it's used as part of project name and workdir. We need to make sure it ends with a proper char which is valid for project name and workdir.
```
  if ENV["NODE_NAME"]
    # likely a jenkins environment
    EXECUTOR_NAME = "#{ENV["NODE_NAME"]}-#{ENV["EXECUTOR_NUMBER"]}".freeze
  else
    EXECUTOR_NAME = "#{HOSTNAME.split('.')[0]}-#{LOCAL_USER}".freeze
  end

```

### DIFF
--- a/lib/bushslicer.rb
+++ b/lib/bushslicer.rb
@@ -50,6 +50,7 @@ module BushSlicer
   else
     EXECUTOR_NAME = "#{HOSTNAME.split('.')[0]}-#{LOCAL_USER}".freeze
   end
+  EXECUTOR_NAME = "#{EXECUTOR_NAME}-e"
 
   START_TIME = Time.now
   TIME_SUFFIX = [
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -439,7 +439,7 @@ module BushSlicer
       unless @service_project
         # if the cluster set the default scheduler, set the project running debug pod node-selector=''
         # to overwrite the default scheduler, or the pod can not be run successfully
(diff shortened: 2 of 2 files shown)


lib/bushslicer.rb
```diff
@@ -50,6 +50,7 @@ module BushSlicer
   else
     EXECUTOR_NAME = "#{HOSTNAME.split('.')[0]}-#{LOCAL_USER}".freeze
   end
+  EXECUTOR_NAME = "#{EXECUTOR_NAME}-e"
```
jhou1, line 53: What about `EXECUTOR_NAME = EXECUTOR_NAME.chomp('-')`, if the string ends with `-` then it's not modified at all.

liangxia: We can not guarantee `LOCAL_USER` end with a valid char for project name and workdir, too

liangxia: E.g, 
```
$ sudo useradd test-lxia-
[sudo] password for lxia: 

$ tail -3 /etc/passwd
systemd-timesync:x:975:975:systemd Time Synchronization:/:/sbin/nologin
flatpak:x:974:974:User for flatpak system helper:/:/sbin/nologin
test-lxia-:x:1001:1002::/home/test-lxia-:/bin/bash

$ ls /home/
lost+found  lxia  test-lxia-
```

PR #2266, opened 2021-09-09
[WIP] Upload logs to upshift s3 bucket

pruan-rht:
Please note, that the `presigned_url` is only valid for 7 days max.  We can keep it as is, or have a dedicated web service in which we feed the s3 object key and have it generate and redirect the user to the new generated-url.  That would mean one more service to keep though..

### DIFF
--- a/lib/launchers/amz.rb
+++ b/lib/launchers/amz.rb
@@ -133,7 +133,7 @@ module BushSlicer
 
     # given a s3 object key, return a valid URL to the key to be accessible
     # via web
-    def s3_generate_url(key: nil, bucket_name: nil, expires_in_seconds: 604800)
+    def s3_generate_url(key: nil, bucket_name: 'cucushift-html-logs', expires_in_seconds: 604800)
       Aws::S3::Object.new(key: key, bucket_name: bucket_name).presigned_url(:get, expires_in: expires_in_seconds)
     end
 
@@ -169,6 +169,7 @@ module BushSlicer
       file_name = File.basename(local_log)
       object_key =  File.join(dst_base_path, file_name)
       local_log = File.join(local_log, "console.html")
+      logger.info("s3 object key: #{object_key}")
       res = s3_upload_file(bucket: bucket_name, file: local_log, target: object_key)
     end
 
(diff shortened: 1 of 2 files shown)


pruan-rht:
to see an example of the, pls use polarshift and enter run id `OS-20210712-2336` to see the link generated is correct and be able to display the html log.

dis016 (commented):
/lgtm

pruan-rht:
1. added support to generate the presigned_url dyanmically
2. construct the URL as a static value which serves as a parameter to a web server
3. depends on https://gitlab.cee.redhat.com/aosqe/cucushift-internal/-/merge_requests/10

pruan-rht:
@jhou1 @liangxia @JianLi-RH @dis016 

Screen Shot 2021-09-18 at 11 20 02 AM

jhou1:
@pruan-rht once we start to upload datahub, can we stop copying logs to the qe.ci host?

pruan-rht:
@jhou1 only one method of upload is active currently.  We can do both if that is what you want.  Once this merged the default is the datahub storage.   I was thinking perhaps to have both active in the beginning to make sure the webserver can handle the translation by heavy loads

jhou1:
Having only the default(datahub) will be good. I see people reporting issues about qe.ci now and then, so time for us to move forward :)

liangxia:
/lgtm

/hold
so other can share their thoughts in case the bot merge it too fast.

pruan-rht:
@jhou1 @liangxia do you know if other teams are looking specifically for the `ci-qe` server logs?  This will import those if there are anything that relies on the location.

jhou1:
I don't know any team that is looking specifically for the logs from this server.

pruan-rht:
@jhou1 ok, I've sent an email out to the group stating the consequence of this PR.  If no objections, perhaps we can merge it this Friday.

dis016 (commented):
/lgtm

pruan-rht:
@jhou1 thanks...with all the asking, I just remember that the `ci-monitor-helper` does interface with the html-logs.   I will need to fix that script. :)

PR #2274, opened 2021-09-13
Moving kata features under feature folder

valtair1891:


pruan-rht:
@valtair1891 it looks like you created these new files instead of moving the original?  You should do a move instead, so to avoid duplication.

PR #2281, opened 2021-09-15
update support

yapei:


yapei:
@liangxia Could you help review and approve if no objections?

lib/openshift/subscription.rb
```diff
@@ -12,6 +12,11 @@ def startingCSV(user: nil, quiet: false, cached: false)
       return rr.dig('status', 'currentCSV')      
     end
 
+    def installedCSV(user: nil, quiet: false, cached: false)
```
liangxia, line 15: Perhaps make quiet/cached defaults to true ?

yapei: @liangxia good point, I agree `quiet` should be `true` by default, however I'm not sure if `cached` should be true or false by default, anyway let me update all to `true` by default and see if any issues

PR #2291, opened 2021-09-16
Simplify docker build

liangxia:
The checkpoint of the scenario is: verify source secret is mounted into builder container when using dockerstrategy to build.
Removing the `yum install` docker step does not help/harm the checkpoints. But it do saves test execution time and produce less execution logs.

From my testing, the scenario passed both with and without the `yum install` step.
But the execution time decreased from `2m22.360s` to `1m16.346s`, and the log size decrease from `30K (577 lines)` to `24K (475 lines)`


liangxia:
Tag for review /cc @jhou1 @JianLi-RH @dis016 @pruan-rht

PR #2292, opened 2021-09-16
Fix several failures in PROW

liangxia:


features/images/jenkins.feature
```diff
@@ -17,7 +17,7 @@ Feature: jenkins.feature
       |  id | -u |
     Then the step should succeed
     #Check that the user is not root, or 0 id
-    Then the expression should be true> Integer(@result[:response]) > 0
+    Then the expression should be true> Integer(@result[:stdout]) > 0
```
liangxia, line 20: Previously failures,
```
Message:

    invalid value for Integer(): "1001170000\n\nSTDERR:\nE0916 07:00:35.823470   47484 v2.go:105] read /dev/stdin: invalid argument\n" (ArgumentError)
(eval):1:in `Integer'
(eval):1:in `block in <top (required)>'
./features/step_definitions/common.rb:134:in `eval'
./features/step_definitions/common.rb:134:in `/^(?:the )?expression should be true> (.+)$/'
features/images/jenkins.feature:77:20:in `the expression should be true> Integer(@result[:response]) > 0'
```

features/service_catalog_tsb/tsb.feature
```diff
@@ -12,51 +12,3 @@ Feature: Template service broker related features
     Then the output should contain "template-service-broker"
     Given cluster service classes are indexed by external name in the :csc clipboard
     Then the expression should be true> cb.csc.values.find {|c| c.cluster_svc_broker_name == "template-service-broker"}
-
-
-  # @author zitang@redhat.com
-  # @case_id OCP-14477
```
liangxia, line 18: Test case for 3.x only.
Screenshot from 2021-09-16 17-25-54

PR #2297, opened 2021-09-17
AWS snapshot and Azure csi resize

ropatil010:
Adding the case id


ropatil010:
Closing and reopening with individual commits.

PR #2300, opened 2021-09-17
[Automate][OCP-41452][Azure CSI Resize] offline volume expansion

ropatil010:
Description: To check the pvc support for offline volume expansion.
Steps:
1) Create pvc, deployment.
2) Apply the patch to resize pvc size
3) Scale the dep to 0, wait untill pvc starts resizing.
4) Scale back the dep to 1, check the pvc size, write data.

### DIFF
--- a/features/storage/csi_resize.feature
+++ b/features/storage/csi_resize.feature
@@ -75,3 +75,85 @@ Feature: CSI Resizing related feature
     Examples:
       | sc_name |
       | gp2-csi | # @case_id OCP-25809
+
+  # @author ropatil@redhat.com
+  @admin
+  Scenario Outline: [CSI Resize] offline volume expansion from 1Gi to 2Gi
+    Given I have a project
+ 
+    # Create pvc with csi storage class 
+    Given I obtain test data file "storage/misc/pvc.json"
+    When I create a dynamic pvc from "pvc.json" replacing paths:
+      | ["metadata"]["name"]                         | mypvc     |
+      | ["spec"]["storageClassName"]                 | <sc_name> |
+      | ["spec"]["resources"]["requests"]["storage"] | 1Gi       |
+    Then the step should succeed    
+  
+    # Create deployment with pvc and check for pvc bound status
+    Given I obtain test data file "storage/misc/deployment.yaml"
+    When I run oc create over "deployment.yaml" replacing paths:
+      | ["metadata"]["name"]                                                             | mydep        |
+      | ["spec"]["template"]["metadata"]["labels"]["action"]                             | storage      |
+      | ["spec"]["template"]["spec"]["volumes"][0]["persistentVolumeClaim"]["claimName"] | mypvc        |
+      | ["spec"]["template"]["spec"]["containers"][0]["volumeMounts"][0]["mountPath"]    | /mnt/storage |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | action=storage |
+    And the "mypvc" PVC becomes :bound
+
+    # Apply the patch command to resize the pvc storage size 
+    When I run the :patch client command with:
+      | resource      | pvc                                                   |
+      | resource_name | mypvc                                                 |
+      | p             | {"spec":{"resources":{"requests":{"storage":"2Gi"}}}} |
+    Then the step should succeed
+    And the output should match:   
+      | persistentvolumeclaim/mypvc patched |
+    
+    # Scale the deployment and wait till it gets disappeared
+    When I run the :scale admin command with:
+      | resource | deployment          |
+      | name     | mydep               |
+      | replicas | 0                   |
+      | n        | <%= project.name %> |
+    Then the step should succeed
+    And I wait for the resource "pod" named "<%= pod.name %>" to disappear   
+
+    #Check the pvc status whether its ready to enable pvc resize
+    And I wait up to 60 seconds for the steps to pass:
+    """
+    When I run the :describe client command with:
+      | resource | pvc   |
+      | name     | mypvc |
+    Then the step should succeed
+    And the output should contain:      
+      | FileSystemResizePending |
+    """
+ 
+    # Scale the deployment and check for ready status
+    When I run the :scale client command with:
+      | resource | deployment          |
+      | name     | mydep               |
+      | replicas | 1                   |
+      | n        | <%= project.name %> |
+    Then the step should succeed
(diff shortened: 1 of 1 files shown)


ropatil010:
/cc @duanwei33 @chao007 @Phaow @jhou1 

Attaching logs:
https://privatebin-it-iso.int.open.paas.redhat.com/?accb7011c04893a2#4nyCy31ZoKB458fvFNjLRfszUxHGK17NQAPwZWYEaH7H

duanwei33:
@ropatil010  Please always make the title clear, like automate OCP-xxx, it will help us get to know what it really does.

features/storage/csi_resize.feature
```diff
@@ -75,3 +75,86 @@ Feature: CSI Resizing related feature
     Examples:
       | sc_name |
       | gp2-csi | # @case_id OCP-25809
+
+  # @author ropatil@redhat.com
+  @admin
+  Scenario Outline: [CSI Resize] offline volume expansion from 1Gi to 2Gi
+    Given the master version >= "4.8"
```
duanwei33, line 82: Azure Disk CSI driver is not installed by default until 4.10, so I think we could let it run after 4.10(using polarion version), and remove this line.

features/storage/csi_resize.feature
```diff
@@ -75,3 +75,86 @@ Feature: CSI Resizing related feature
...
+    Given I have a project
+ 
+    # Create pvc with csi storage class 
+    Given I obtain test data file "storage/misc/pvc.json"
+    When I create a dynamic pvc from "pvc.json" replacing paths:
+      | ["metadata"]["name"]                         | mypvc         |
+      | ["spec"]["storageClassName"]                 | <sc_name>     |
+      | ["spec"]["accessModes"][0]                   | ReadWriteOnce |
```
duanwei33, line 90: Let's remove this line.

features/storage/csi_resize.feature
```diff
@@ -75,3 +75,86 @@ Feature: CSI Resizing related feature
...
+    # Scale the deployment and wait till it gets disappeared
+    When I run the :scale admin command with:
+      | resource | deployment          |
+      | name     | mydep               |
+      | replicas | 0                   |
+      | n        | <%= project.name %> |
+    Then the step should succeed
+    And I wait for the resource "pod" named "<%= pod.name %>" to disappear   
+
+    # Run mount on node for syncup
```
duanwei33, line 124: The following steps are not for syncup, it is to make sure the volume is "really" detached, as from our test, resize did not support resize the un-detached volume, it will fail.
Please update the comment.

features/storage/csi_resize.feature
```diff
@@ -75,3 +75,86 @@ Feature: CSI Resizing related feature
...
+      | resource | deployment          |
+      | name     | mydep               |
+      | replicas | 1                   |
+      | n        | <%= project.name %> |
+    Then the step should succeed
+    And a pod becomes ready with labels:
+      | action=storage |
+
+    # Check the pv,pvc capacity size 
+    And I wait up to 800 seconds for the steps to pass:
```
duanwei33, line 145: Do we need to wait so many time? Do you have the test result? 
We can reduce the wait time, if it really fails, we don't need to wait such a long time.

ropatil010: I guess, no need to wait that much time. From execution logs attached as well,
[12:33:41] INFO> Shell Commands: oc get persistentvolumes pvc-53eb0fa3-ec11-4be5-8a55-686c2f179116 --output=yaml --kubeconfig=/var/root/workdir/ropatil-mac-rohitpatil/ocp4_admin.kubeconfig
[12:33:45] INFO> Exit Status: 0
Since in previous reference code, gave as 800 sec, so by defaulty i added the same.

duanwei33: Yes, let's make it less, no one likes to wait so much time if it really get failed. Does 180s make more sense?

features/storage/csi_resize.feature
duanwei33: It has no side impact, but if we use the default value and there is no specific meaning or different from others, no need to list it.
For another storage size, although 1G is the default value, it will help up to understand the resize from 1G to 2G, so I think it is meaningful here.

features/storage/csi_resize.feature
ropatil010: Yeah sure. Any possibility of adding other platform if they supporting this feature below 4.10? 
If no will change this to 4.10.

duanwei33: I have no idea, I prefer to using polarion parameter when it is "Scenario Outline" which need consider other cases.

jhou1: We could add tags to the examples if each of the examples supports a different version. See example here - https://github.com/openshift/cucushift/pull/8709/files

Please also add a `@4.10` tag. This tag will be useful for executing test scenarios in Prow.

jhou1:
Please update this PR and make the summary, description, commit message more informative about what this does.

features/storage/csi_resize.feature
ropatil010: @duanwei33 Instead of mount option, can i use the below code which will be more easy to understand? Let me know your opinion. 
    When I run the :describe client command with:
      | resource | pvc   |
      | name     | mypvc |
    Then the step should succeed
      | FileSystemResizePending |

jhou1:
@ropatil010 Once you update this PR summary and commit message, we can get this merged.

features/storage/csi_resize.feature
duanwei33: @ropatil010 Do you want to use the 
```
When I run the :describe client command with:
| resource | pvc |
| name | mypvc |
Then the step should succeed
| FileSystemResizePending |
```
to replace the following line?
```
    And I wait up to 60 seconds for the steps to pass:
    """
    Given I use the "<%= pod.node_name %>" node
      | mount |
    Then the output should not contain:
      | <%= pvc.volume_name %> |
    """
```
It should also work, but I the umount/detach is the root cause.
But I  don't object your update if you test it and passed.

ropatil010:
Attaching logs: 
https://privatebin-it-iso.int.open.paas.redhat.com/?8c2e37284a9d648f#9AmhF6T8JfPJvFPQjmgm357FwxH5vrLP2XEkd8Bk8xsT

PR #2312, opened 2021-09-22
auto OCP-44251 OCP-44250

huiran0826:
Auto OCP-44251 OCP-44250 which are from customer bugs.

https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3521/console
https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3518/console

@openshift/team-sdn-qe


features/networking/ovn-egress-ip.feature
```diff
@@ -643,3 +643,194 @@ Feature: OVN Egress IP related features
...
+
+    # Now scale up CNO pod to 1
+    Given I run the :scale admin command with:
+      | resource | deployment                 |
+      | name     | network-operator           |
+      | replicas | 1                          |
+      | n        | openshift-network-operator |
+    Then the step should succeed
+    # A minimum wait for 30 seconds is tested to reflect CNO deployment to be effective which will then re-spawn ovn pods
+    Given 30 seconds have passed
```
pruan-rht, line 823: @huiran0826 hardcoded wait is discouraged.  Isn't the next step atline 824 sufficient?

huiran0826: @pruan-rht The next step 824 is not enough under this condition. 
 After  CNO pod scaled to 1, if only step 824, it may pass immediately as all the ovnkube-node-XXX pods are ready status and no ovnkube-master-XXX created yet.  We need to wait ovnkube-master-XXX pods to be recreated.

liangxia: Try with
```
    Given I use the "xxx" project
    And a pod becomes ready with labels:
      | foo=bar |
    Given I run the :scale admin command with:
      | resource | deployment                 |
      | name     | network-operator           |
      | replicas | 1                          |
      | n        | openshift-network-operator |
    Then the step should succeed
    Given I wait for the resource "pod" named "<%= pod.name %>" to disappear
    And admin waits for all pods in the "xxx" project to become ready up to 150 seconds
```

huiran0826: We don't need " Given I wait for the resource "pod" named "<%= pod.name %>" to disappear" here,  at that time, still no ovnkube-master-XXX pods, I mean need wait ovnkube-master-XXX pods to be created.
I thought generally we don't use hard code time just avoid to waste waiting time in case it may get ready before the hard code time. Actually in this condition, 30s is not wasted. As you see, next step is still waiting the pods ready.

liangxia: It is to make sure old pod is gone. And next step make sure new pod is up.

We have thousands of test cases. We can not bear test case useless waiting.

huiran0826: The old pods was gone check not in this step. It was covered in previous steps. 
Anyway, I will check and try to replace this hard code.

huiran0826: Fixed it, thanks!

PR #2314, opened 2021-09-22
[Automate][OCP-41449][Azure-csi-snapshot]with default parameters

ropatil010:
Description: Volume snapshot create and restore test
Steps: 
1) Create pvc from csi storage class(managed-csi) , pod from pvc. 
2) Check the pod status and write data.
3) Create a Volumesnapshot class, volumesnapshot and check the status.
4) Create a snapshot pvc from volumesnapshot, pod from snapshot pvc. 
5) Check the snapshot pvc/pod status, data content in the pod.

### DIFF
--- a/features/storage/csi_snapshot.feature
+++ b/features/storage/csi_snapshot.feature
@@ -57,7 +57,7 @@ Feature: Volume snapshot test
       | csi-sc       |
       | gp2-csi      | # @case_id OCP-27727
       | standard-csi | # @case_id OCP-37568
-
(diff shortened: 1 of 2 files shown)


ropatil010:
/cc @duanwei33 @chao007 @Phaow 

Logs: https://privatebin-it-iso.int.open.paas.redhat.com/?2867447fcc79a908#B7re8F7sT6cteJKxSUqXTfW6rKXBeusjtfqpHAFUmwnP

PR #2331, opened 2021-09-24
fix psap automation failures

pruan-rht:
1. labels for OCP < 4.8 are different
2. fix timing issue when creating namespace from OperatorHub is slower, so added a statement to make sure the namespace is present before doing anything else.
3. for OCP > 4.8, the nfd-master.yaml needs to be updated to include a section workerConfig https://github.com/openshift/cucushift/pull/8785


wabouhamad (commented):
/lgtm

PR #2341, opened 2021-09-29
Add ensure_console_tagged

liangxia:
Add method to force console scenarios are tagged with console tag, to support tests in Prow.


jhou1:
Nice approach!
/lgtm
/approve

PR #2348, opened 2021-09-30
Add tag 4.10

liangxia:

We just map all 4.9 test cases into 4.10 as we do not have that info in Polarion yet, so we can try out 4.10 jobs in Prow.


liangxia:
/hold

for https://github.com/openshift/cucushift/pull/8804 to merge first

PR #2349, opened 2021-09-30
[WIP]construct an initialization function for baremetal iaas

wangke19:
Currently we don't support baremetal iaas, but the iaas type will be checked by some cases with baremetal cluster
We construct an empty initialization function for baremetal in verification-tests/lib/iaas/iaas.rb, let test case 
check iaas type with baremetal without exception and exit.


lib/iaas/iaas.rb
```diff
@@ -16,6 +16,8 @@ def self.select_provider(env)
           return {:type => "gcp", :provider => self.init_gcp(env)}
         when "Azure"
           return {:type => "azure", :provider => self.init_azure(env)}
+        when "None"
+          return {:type => "baremetal", :provider => self.init_baremetal(env)}
```
jhou1, line 20: Can we set return as `{:type => "none", :provider => nil}`. This aligns with `when "None"` from the iaas type.

PR #2350, opened 2021-10-06
add tags vsphere-upi,4.9

dis016:
[OCPQE-5689](https://issues.redhat.com/browse/OCPQE-5689) /cc @lxia @jhou1  @jianl @pruan


dis016:
tags added with https://github.com/openshift/verification-tests/pull/2372

PR #2363, opened 2021-10-14
Remove specific codes for 3.x in master(for 4.x)

liangxia:
For version < 3.3, the command is `oadm`
For version >= 3.3, the command is `oc adm`
In master branch, we can assume the version >= 4.1

/cc @jhou1 @pruan-rht @xiuwang


liangxia:
> Looks good, but what's the harm of keeping these conditions? Which problem does this fix?

No harm. Just cleanup useless codes.

jhou1:
OK, thanks for the clarification.
/lgtm
/approve

PR #2364, opened 2021-10-14
Add tag baremetal-ipi to support tests in Prow

liangxia:

Co-works with https://github.com/openshift/cucushift/pull/8823


jhou1:
Do we also have to add `@4.9` and `@4.10` versions?

PR #2371, opened 2021-10-15
Enhance tag tool with grouping tags

liangxia:
@jhou1 @JianLi-RH @dis016 @pruan-rht


jhou1:
Does this work well with `Scenario`, `Scenario Outline`, `Examples`?

PR #2391, opened 2021-10-20
New testcase OCP-42205

XiyunZhao:


XiyunZhao:
For the latest comment "It looks like the case is applicable for >= 4.8 so we still need add 4.8 rules"
I have updated the master version to 4.9, after checking the case and confirming on Console-2386

PR #2417, opened 2021-10-28
get the leader node from config_map

pruan-rht:
@jechen0648 this should work for you.


jechen0648:
@liangxia Can you help review this PR?  I am waiting for it being merged so I can submit my PR that uses this function, thank you!

PR #2425, opened 2021-11-02
use stable before logging 5.3 is released

anpingli:


anpingli:
@liangxia  @QiaolingTang  PTAK

PR #2432, opened 2021-11-03
Adding 4.7 upgrade sanity tests for EUS

jhou1:
Adding tags to run upgrade-sanity tests for EUS upgrades. @liangxia @pruan-rht @JianLi-RH @dis016 PTAL PTAL

### DIFF
--- a/doc/prow/4.7/query_upgrade_sanity.json
+++ b/doc/prow/4.7/query_upgrade_sanity.json
@@ -1,0 +1,3 @@
+{
+  "case_query": "products.KEY:ocp AND NOT status:inactive AND type:testcase AND caseautomation.KEY:automated AND automation_script:cucushift AND NOT upstream.KEY:yes AND caseimportance.KEY:critical AND version.KEY:4_7 AND NOT tags:(prod_only stage_only)"
+}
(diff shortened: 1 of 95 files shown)


doc/prow/4.7/query_upgrade_sanity.json
```diff
@@ -0,0 +1,3 @@
+{
+  "case_query": "products.KEY:ocp AND NOT status:inactive AND type:testcase AND caseautomation.KEY:automated AND automation_script:cucushift AND NOT upstream.KEY:yes AND env_install_method.KEY:(__all__ ipi) AND caseimportance.KEY:critical AND version.KEY:4_7 AND NOT tags:(prod_only stage_only)"
```
liangxia, line 2: Perhaps reflect ipi in the file name ?

jhou1: Thanks, I'll correct the query and test cases, the query should include critical tests, regardless of installation methods.

jhou1: Updated, the query result did not change, so no new tags added

PR #2447, opened 2021-11-05
namespace fix in upgrade case

anuragthehatter:
Found ```nodeport-upgrade``` namespace already exist in other feature so better aligning this usecase to ```conntrack-upgrade``` namespace. Found its causing conflict. Just a quick fix. @pruan-rht Can you help merging it quick is possible? Thanks


asood-rh (approved):
/lgtm

PR #2448, opened 2021-11-06
added support to clone run by subteam for all status, not just PASSED…

pruan-rht:
…/FAILED as currently the case

for example, to clone all `networking` case regardless of status
```
./tools/polarshift.rb clone-run 20210612-0147 --subteam="networking" --status='ALL'
```
to clone all `networking` cases for only the FAILED ones
```
./tools/polarshift.rb clone-run 20210612-0147 --subteam="networking" --status='FAILED'
```
@zhaozhanqi I think that's what you wanted.

### DIFF
--- a/tools/polarshift.rb
+++ b/tools/polarshift.rb
@@ -530,7 +530,7 @@ module BushSlicer
         filtered_cases = query_result['records']["TestRecord"]
         if opts[:subteam]
           filtered_cases, subteams = filtered_cases_by_subteam(subteam_opt: opts[:subteam], cases: filtered_cases)
-          run_title ||= "Clone of '#{case_status}' #{subteams}  cases for #{test_run_id}"
+          run_title ||= "Clone of #{subteams}  cases for #{test_run_id}"
         end
         if filtered_cases.count > 0
           tc_ids = extract_test_case_ids(filtered_cases)
--- a/tools/polarshift.rb
+++ b/tools/polarshift.rb
@@ -404,10 +404,10 @@ module BushSlicer
       @opts = opts
     end
 
-    # given a query_result from a call to get_smart_run()
+    # given a query_result from a call to get_smart_run() Array of testcase Hash
     # @return Array of testcase IDs
     def extract_test_case_ids(query_result)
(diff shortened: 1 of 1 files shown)


zhaozhanqi:
Thank you @pruan-rht 
/lgtm

pruan-rht:
@liangxia PTAL.  thx!

tools/polarshift.rb
```diff
@@ -497,14 +497,19 @@ def clone_run(test_run_id: nil, options: nil)
       template_hash = {}
       case_status = options.status
       run_title = options.title
-      valid_statuses = ['Passed', 'Failed', 'Waiting', 'Running', 'Blocked']
+      # 'All' statuss mean we ignore the status and just filter on subteam name
+      valid_statuses = ['Passed', 'Failed', 'Waiting', 'Running', 'Blocked', 'All']
```
liangxia, line 501: `All` does not sounds like a valid case status.

Perhaps we can try, when user does not specify case status, then we select test cases that with a status in any of that in valid_statuses, which means `all`, but should not introduce a new invalid status `All`.

pruan-rht: @liangxia ok, now if nothing is given then it will return all.  PTAL

tools/polarshift.rb
```diff
@@ -497,28 +509,36 @@ def clone_run(test_run_id: nil, options: nil)
...
+          exit
+        end
+        run_title ||= "Clone of '#{case_status}'  cases for #{test_run_id}"
       else
-        tc_ids = extract_test_case_ids(query_result)
-        run_title ||= "Clone of #{test_run_id}"
+        filtered_cases = query_result['records']["TestRecord"]
+        if opts[:subteam]
+          filtered_cases, subteams = filtered_cases_by_subteam(subteam_opt: opts[:subteam], cases: filtered_cases)
+          run_title ||= "Clone of '#{case_status}' #{subteams}  cases for #{test_run_id}"
```
liangxia, line 533: We do not have `case_status` in the `else` block.

pruan-rht: @liangxia good catch, fixed now.

PR #2451, opened 2021-11-08
Length limit of EXECUTOR_NAME as we have 63 chars limit for project name in OCP

liangxia:
Fix issue in https://coreos.slack.com/archives/C01UHADN91U/p1636349302059500

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -439,7 +439,7 @@ module BushSlicer
       unless @service_project
         # if the cluster set the default scheduler, set the project running debug pod node-selector=''
         # to overwrite the default scheduler, or the pod can not be run successfully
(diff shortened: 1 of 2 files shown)


lib/environment.rb
```diff
@@ -439,7 +439,7 @@ def service_project
       unless @service_project
         # if the cluster set the default scheduler, set the project running debug pod node-selector=''
         # to overwrite the default scheduler, or the pod can not be run successfully
-        project_name = "proj-" + EXECUTOR_NAME.downcase
+        project_name = "pj-" + EXECUTOR_NAME.downcase
```
jhou1, line 442: What about `prj`, `pj` looks like a name short for `prow job` ;)

liangxia: Updated

PR #2456, opened 2021-11-09
Reorganize clone-run

liangxia:
As a follow up of https://github.com/openshift/verification-tests/pull/2448

/cc @pruan-rht @zhaozhanqi @jhou1


pruan-rht (commented):
/lgtm

PR #2463, opened 2021-11-10
remove curator from clusterlogging

QiaolingTang:
To fix https://issues.redhat.com/browse/OCPQE-6808 
The issue is: the curator was deployed, but before checking the job status, we didn't update the schedule of cj/curator. Since  curator is deprecated and we imported index management from logging 4.5, here I removed the curator from clusterlogging. 

/cc @anpingli


anpingli (commented):
/lgtm

PR #2139, opened 2021-07-02
Automate OCP-37568/OCP-37569/OCP-27727 and some fix

duanwei33:
Automate snapshot with block volume and add cinder cases: 
[pass log](https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3274/console) with cinder
[pass log](https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/Runner-v3-smoke/3273/console) with aws

PR #2464, opened 2021-11-10
Fix OCP-27727

liangxia:
Looks like the OCP-27727 is mistakenly copy/paste in https://github.com/openshift/verification-tests/pull/2139

/cc @duanwei33 @chao007 @jhou1 @pruan-rht


jhou1:
Thank you @liangxia 
/lgtm
/approve

PR #2478, opened 2021-11-16
Do not show secrets info by default

liangxia:

Tested with `features/storage/storage_class.feature:241` on a 4.10 AWS cluster locally, and the secrets info are not show.

### DIFF
--- a/lib/openshift/secret.rb
+++ b/lib/openshift/secret.rb
@@ -8,19 +8,19 @@ module BushSlicer
     RESOURCE = "secrets"
 
     # see #raw_resource
-    def type(user: nil, cached: true, quiet: false)
+    def type(user: nil, cached: true, quiet: true)
       obj = raw_resource(user: user, cached: cached, quiet: quiet)
       return obj['type']
     end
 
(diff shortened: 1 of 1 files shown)


pruan-rht (commented):
/lgtm
/approve

lib/openshift/secret.rb
```diff
@@ -8,32 +8,32 @@ class Secret < ProjectResource
...
-    def type(user: nil, cached: true, quiet: false)
+    def type(user: nil, cached: true, quiet: true)
       obj = raw_resource(user: user, cached: cached, quiet: quiet)
       return obj['type']
     end
 
     # @param user [BushSlicer::User] the user to run cli commands with if needed
-    def bearer_token?(user: nil, cached: true, quiet: false)
+    def bearer_token?(user: nil, cached: true, quiet: true)
       type(user: user, cached: cached, quiet: quiet).include?('service-account-token') && raw_resource.dig('data', 'token')
```
jhou1, line 18: Can we hard code it to `quiet: true` just in case the caller passes in the value as `false`?

PR #2482, opened 2021-11-16
Do not show the expression info when contains secret

liangxia:

### DIFF
--- a/features/step_definitions/common.rb
+++ b/features/step_definitions/common.rb
@@ -141,11 +141,15 @@ Then /^(?:the )?expression should be true> (.+)$/ do |expr|
       comparison_operators = /===|==|>=|<=|!=|>|<|&&|\|\|&|\|/
       op = expr.scan(comparison_operators).first
       if op
-        e1, e2 = expr.split(op)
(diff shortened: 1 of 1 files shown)


features/step_definitions/common.rb
```diff
@@ -141,19 +141,23 @@
       comparison_operators = /===|==|>=|<=|!=|>|<|&&|\|\|&|\|/
       op = expr.scan(comparison_operators).first
       if op
-        e1, e2 = expr.split(op)
+        expr_left, expr_right = expr.split(op)
         # XXX: we may not catch everything, wrap it around begin/rescue and print
         # out blank if some complex comparison is in the expression.
         begin
-          eval_details = "\nleft_operand: #{eval(e1)}, right_operand: #{eval(e2)}\n"
+          if ( (expr_left.include? "secret") || (expr_right.include? "secret") )
```
jhou1, line 148: Maybe adding a list of keywords to the watch and masking them. Such as `secrets, token, key, creds, credential, password etc.`
Suggest to use regexp to also capture upper and lower cases

liangxia: Fixed.

PR #2483, opened 2021-11-17
Add a step for saving CR directly

lwan-wanglin:
Add a step for saving CR to a file directly without outputing

@liangxia @jhou1 please help review this one and https://github.com/openshift/cucushift/pull/8886

### DIFF
--- a/features/step_definitions/common.rb
+++ b/features/step_definitions/common.rb
@@ -130,6 +130,14 @@ Given /^evaluation of `(.+?)` is stored in the#{OPT_SYM} clipboard$/ do |what, c
   cb[clipboard_name] = eval(what)
 end
 
+Given /^I save custom resource `(.+?)` to file>(.+)$/ do |what,filepath|
+  ensure_admin_tagged
+
+  resource = eval(what)
+  txt = resource.raw_resource(quiet: true).to_yaml
+  File.write(File.expand_path(filepath.strip), txt)
+end
+
 Then /^(?:the )?expression should be true> (.+)$/ do |expr|
   res = eval expr
   eval_details = ""


features/step_definitions/cco.rb
```diff
@@ -0,0 +1,10 @@
+Given /^save of custom resource `(.+?)` to file>(.+)$/ do |what,filepath|
```
jhou1, line 1: Since most of our current steps have a *subject* in the sentence, Do you mind updating the step to `/^I save the custom resource `(.+?)` to file>(.+)$/` (or `Admin save .... to`)

features/step_definitions/cco.rb
```diff
@@ -0,0 +1,10 @@
...
+  ensure_admin_tagged
+
+  resource = eval(what)
+  TXT = resource.raw_resource(quiet: true).to_yaml
```
jhou1, line 5: rename the variable to `txt`. Variable in UPPERCASE is usually constant.

features/step_definitions/cco.rb
lwan-wanglin: sure, wait for cucushift case review, then i will fix the both

liangxia: Good approach. Thanks.

The step is mainly for CCO currently, but it's a common step. Would you mind put it into `features/step_definitions/common.rb` ?

lwan-wanglin: sure

features/step_definitions/common.rb
```diff
@@ -130,6 +130,14 @@
   cb[clipboard_name] = eval(what)
 end
 
+Given /^(I|admin) save custom resource `(.+?)` to file>(.+)$/ do |what,filepath|
```
liangxia, line 133: A new regexp `I|admin` introduced, but not deal with. Thus 3 regexp but only 2 params, which will cause error. 
I think we can simply use `I` here to make it simple.

liangxia: Also, there is a small typo in the commit message `Add a step for saving CR to a file ^Crectly`

lwan-wanglin: oops, thanks, i will use just `I`

lwan-wanglin:
rerun the job pass:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/223498/console

@liangxia @jianping-shu  Please help review again, thanks!

PR #2484, opened 2021-11-17
Reenable the tests

jhou1:
As we have silenced the secret printing in debug messages, we are good to reenable these tests.
@liangxia @duanwei33 @pruan-rht PTAL

### DIFF
--- a/features/machine/machine_misc.feature
+++ b/features/machine/machine_misc.feature
@@ -2,7 +2,6 @@ Feature: Machine misc features testing
 
   # @author miyadav@redhat.com
   # @case_id OCP-34940
-  @inactive
   @admin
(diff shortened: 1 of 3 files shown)


features/machine/machine_misc.feature
```diff
@@ -2,7 +2,6 @@ Feature: Machine misc features testing
 
   # @author miyadav@redhat.com
   # @case_id OCP-34940
-  @inactive
```
liangxia, line 5: This one is not safe yet.
```
    When I run the :get admin command with:
      | resource      | secret        |
      | resource_name | vsphere-creds |
      | n             | kube-system   |
      | o             | yaml          |
```

PR #2486, opened 2021-11-18
update curl in step store an available router IP

lihongan:
the test-client-pod image hasn't installed bash command yet but `bash` is required in this step, so revert to pot-for-ping for now.

@liangxia Please take a look. Thanks.

cc @quarterpin @melvinjoseph86 @ShudiLi

### DIFF
--- a/features/step_definitions/route.rb
+++ b/features/step_definitions/route.rb
@@ -139,7 +139,7 @@ Given /^I store an available router IP in the#{OPT_SYM} clipboard$/ do |cb_name|
   step %Q/I expose the "selector-service" service/
   step %Q/I have a test-client-pod in the project/
   step %Q/I execute on the pod:/, table(%{
-    | bash | -c | curl http://<%= route("selector-service", service("selector-service")).dns(by: user) %>/ --connect-timeout 10 -I -v -s |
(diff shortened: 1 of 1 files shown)


quarterpin:
Looks good to me.

features/step_definitions/route.rb
```diff
@@ -137,7 +137,7 @@
     | f |  #{BushSlicer::HOME}/testdata/networking/service_with_selector.json |
   })
   step %Q/I expose the "selector-service" service/
-  step %Q/I have a test-client-pod in the project/
+  step %Q/I have a pod-for-ping in the project/
   step %Q/I execute on the pod:/, table(%{
     | bash | -c | curl http://<%= route("selector-service", service("selector-service")).dns(by: user) %>/ --connect-timeout 10 -I -v -s |
```
liangxia, line 142: `bash -c` is not a must there.
You can try `| curl | -foo | -bar | my_url |`

lihongan: Yes, you're right. Using curl is also working. 
see: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/224974/console

PR #2487, opened 2021-11-18
Updated verification-tests by replacing pod-for-ping with test-client…

ShudiLi:
@lihongan @quarterpin @melvinjoseph86, request for a review. Thanks.

The modification covers below:
haproxy-router.feature: OCP-11903,OCP-11130,OCP-10043,OCP-11679
route.feature OCP-12562,OCP-12564,OCP-9651,OCP-9650,OCP-11036,OCP-13753

A successful jenkins Runner is below:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/224923/

### DIFF
--- a/features/routing/haproxy-router.feature
+++ b/features/routing/haproxy-router.feature
@@ -24,7 +24,7 @@ Feature: Testing haproxy router
     When I expose the "service-unsecure" service
     Then the step should succeed
 
-    Given I have a pod-for-ping in the project
+    Given I have a test-client-pod in the project
     #access the route without cookies
     Given I wait for the steps to pass:
     """
@@ -85,7 +85,7 @@ Feature: Testing haproxy router
       | service | service-unsecure |
     Then the step should succeed
 
-    Given I have a pod-for-ping in the project
+    Given I have a test-client-pod in the project
     # access the route without cookies
     Given I wait for the steps to pass:
     """
@@ -268,7 +268,6 @@ Feature: Testing haproxy router
   @upgrade-sanity
   Scenario: Set balance leastconn for passthrough routes
     Given I have a project
-    And I store an available router IP in the :router_ip clipboard
     Given I obtain test data file "routing/web-server-1.yaml"
     When I run the :create client command with:
       | f | web-server-1.yaml |
@@ -293,15 +292,13 @@ Feature: Testing haproxy router
       | overwrite | true |
     Then the step should succeed
 
-    Given I have a pod-for-ping in the project
+    Given I have a test-client-pod in the project
     And I use the "service-secure" service
     Given I wait for the steps to pass:
     """
     When I execute on the pod:
       | curl                                                   |
       | -ksS                                                   |
-      | --resolve                                              |
-      | <%= route("route-pass").dns(by: user) %>:443:<%= cb.router_ip[0] %> |
       | https://<%= route("route-pass").dns(by: user) %> |
     Then the step should succeed
     And the output should contain "Hello-OpenShift web-server-2"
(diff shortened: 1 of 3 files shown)


features/routing/haproxy-router.feature
```diff
@@ -352,12 +347,17 @@ Feature: Testing haproxy router
       | /tmp/cookies |
     Then the step should succeed
     And the output should contain "Hello-OpenShift"
+    Given I run the steps 10 times:
+    """
     When I execute on the pod:
-      | bash | -c | for i in {1..10} ; do curl -sS  http://<%= route("service-unsecure", service("service-unsecure")).dns(by: user) %>/ -b /tmp/cookies ; done |
```
lihongan, line 356: Don't remove the loop `for i in {1..10`}, it is required for this case.
Since bash is required for the loop and curl, so please don't update this case.

lihongan:
@ShudiLi same question, seems the `timeout.feature` is not covered in the PR, would you like to add it to this or create another PR?

features/routing/haproxy-router.feature
ShudiLi: Sure. But the bash doesn't work on the test-client-pod, maybe another PR is needed for this test case.

ShudiLi:
@lihongan For the timeout.feature, since tests are failed with whether pod-for-ping or test-client-pod(curl: (56) OpenSSL SSL_read: error:1408F119:SSL routines:ssl3_get_record:decryption failed or bad record mac, errno 0), so it is better to create a new PR for this case.

ShudiLi:
@lihongan Thanks for the review, I will modify them or create new PRs.

features/routing/haproxy-router.feature
lihongan: just please keep the original pod-for-ping for this kind of cases, no need to update them for now.

features/routing/haproxy-router.feature
```diff
@@ -352,12 +347,17 @@ Feature: Testing haproxy router
...
+      | curl |
+      | -sS |
+      | http://<%= route("service-unsecure", service("service-unsecure")).dns(by: user) %>/ |
+      | -b |
+      | /tmp/cookies |
     Then the step should succeed
-    And the output should contain:
-      | Hello-OpenShift web-server-1 |
-      | Hello-OpenShift web-server-2 |
```
lihongan, line 360: please keep the original checkpoint as well.

ShudiLi: Yes, I have kept the original checkpoint. Just remove the test-client-pod and add pod-for-ping before this checkpoint, thanks.

ShudiLi:
Jenkins run for cases:OCP-11679,OCP-11635:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/225236

features/routing/haproxy-router.feature
liangxia: Try,
1. Replace `bash -c` with `sh -c`.
2. Remove `bash -c`

ShudiLi: @lihongan Yes, I keep the pod-for-ping for the bash. And for the first curl request before the bash, I use the test-client-pod. I also remove test-client-pod before creating pod-for-ping.

@liangxia Thanks. I tried sh -c yesterday, but it didn't work with the full command "sh -c for i in {1..10} ; ...".

liangxia: {1..10} is the expansion from bash. try `for (( i=1; i<=10; ++i ))`

lihongan: We're planning to rebuild the nginx-alpine image and to install more binaries/packages and enable more functions on it, so I prefer to update this case later with the new image that `bash` is built-in.

ShudiLi: @lihongan Thanks for the review and the new plan, I have rolled back case OCP-11679 to the original.

lihongan:
/lgtm
Thank you @ShudiLi

ShudiLi:
@lihongan Thank you for the review with with patience.

PR #2491, opened 2021-11-19
Simplify the machines upgrade codes

liangxia:
```
[lxia@lxiabox upgrade]$ grep -r 'Scenario' machines/machines.feature 
  Scenario Outline: Cluster operator should be available after upgrade - prepare
  Scenario Outline: Cluster operator should be available after upgrade
  Scenario: Cluster operator should be available after upgrade - prepare
  Scenario Outline: Cluster operator should be available after upgrade
```


jhou1:
Thank you
/lgtm
/approve

PR #2495, opened 2021-11-22
update tag customer

zhaozhanqi:
@openshift/team-sdn-qe ^^

### DIFF
--- a/features/networking/egress-ip.feature
+++ b/features/networking/egress-ip.feature
@@ -131,6 +131,7 @@ Feature: Egress IP related features
 
   # @author huirwang@redhat.com
   # @case_id OCP-21812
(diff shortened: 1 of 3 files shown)


jhou1:
Thank you @zhaozhanqi . Are they customer scenarios or test cases with customer info/data? In case they are the former, it's we don't need these tags. (Just in case the Jira didn't clearly state that.)

zhaozhanqi:
> Thank you @zhaozhanqi . Are they customer scenarios or test cases with customer info/data? In case they are the former, it's we don't need these tags. (Just in case the Jira didn't clearly state that.)

Yes, I just misunderstanding the tag meaning.

I think this PR can be closed

PR #2502, opened 2021-11-24
Move specific commands which contains sensitive info to logger.debug

liangxia:
Previously we move all the commands to logger.debug as some of the commands contains sensitive info, https://github.com/openshift/verification-tests/pull/2494


liangxia:
Tested with runner-v3-smoke/3746/console
The command `oc login` and `oc get secret` are not show. While `oc version`, `oc config` and `oc get infrastructure` do show the command which is helpful for failure analysis.

PR #2503, opened 2021-11-24
Add security tag for cases are likely to expose sensitive information

sunzhaohua2:
According to https://issues.redhat.com/browse/OCPQE-7097, add security tag for cases are likely to expose sensitive information. @jhou1 @miyadav @huali9 please help to take a look.

### DIFF
--- a/features/machine/machine.feature
+++ b/features/machine/machine.feature
@@ -67,6 +67,7 @@ Feature: Machine features testing
 
 
   # @author jhou@redhat.com
+  @security
   @admin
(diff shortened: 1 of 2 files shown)


miyadav:
/lgtm 
Both cases should have the tag , thanks

jhou1:
The framework has handle situation like this. @miyadav @sunzhaohua2 when you are watching automation, could you please pay attention to see if sensitive information is not printed? If this has been well handled then we don't need this PR now.
/hold

PR #2505, opened 2021-11-24
Allow to skip waiting for OSD metrics

apodhrad:
Sometimes, the metrics are synced after 3 hrs. The bug is reported at
https://issues.redhat.com/browse/OHSS-2640

Some test cases do not require these metrics.
So, we should allow to skip the waiting for them.


pruan-rht (commented):
/lgtm

PR #2506, opened 2021-11-24
Protect sensitive information from logger

jhou1:
This will protect sensitive information in logger. The logger format it generates will look like. 
@liangxia @pruan-rht @JianLi-RH @dis016 PTAL

```
     apiVersion: v1
      data:
        azure_client_id: [PROTECTED_DATA]
        azure_client_secret: [PROTECTED_DATA]
```

### DIFF
--- a/lib/log.rb
+++ b/lib/log.rb
@@ -115,7 +115,7 @@ module BushSlicer
         timestamp = ''
       end
 
-      m = {msg: msg, level: level, timestamp: timestamp}
+      m = {msg: censor(msg), level: level, timestamp: timestamp}
       if @dup_buffer
         @dup_buffer.push(m)
       else
@@ -181,6 +181,21 @@ module BushSlicer
 
     def reset_dedup
       @dup_buffer.reset if @dup_buffer
+    end
+
+    private
+    def censor(msg)
+      secured_lines = []
+      lines = msg.split("\n")
+      censor_kw = %w[client_id client-id client_secret client-secret subscription_id tenant_id access_key_id secret_access_key secret authorization username password oauth token .dockercfg .dockerconfigjson kubeconfig htpasswd ca service-ca tls service-account service_account serviceaccount cloud pull_secret pull-secret cred key]
+      lines.each do |line|
+        if censor_kw.any? { |kw| line.downcase.match(/^.*#{kw}.*:.*/) }
+          secured_lines << line.downcase.gsub(/:.*/, ': [PROTECTED_DATA]')
+          next
+        end
+        secured_lines << line
+      end
+      secured_lines.join("\n")
     end
 
     # map messages to unique characters, then run a regular expression to
--- a/features/machine/machine_misc.feature
+++ b/features/machine/machine_misc.feature
@@ -2,7 +2,6 @@ Feature: Machine misc features testing
 
   # @author miyadav@redhat.com
   # @case_id OCP-34940
-  @inactive
   @admin
   @destructive
   @4.10 @4.9
--- a/features/storage/csi_snapshot.feature
+++ b/features/storage/csi_snapshot.feature
@@ -1,7 +1,6 @@
 Feature: Volume snapshot test
 
   # @author wduan@redhat.com
-  @inactive
   @admin
   @4.8 @4.7 @4.10 @4.9
   Scenario Outline: Volume snapshot create and restore test
@@ -74,8 +73,6 @@ Feature: Volume snapshot test
       | standard-csi | standard-csi |# @case_id OCP-37568
 
   # @author wduan@redhat.com
-  @admin
-  @inactive
   @admin
   @4.8 @4.7 @4.10 @4.9
   Scenario Outline: Volume snapshot create and restore test with block
--- a/features/storage/storage_class.feature
+++ b/features/storage/storage_class.feature
@@ -107,7 +107,6 @@ Feature: storageClass related feature
       | aws-ebs     | st1         | us-east-1d    | false      | 500Gi | # @case_id OCP-10424
 
   # @author lxia@redhat.com
-  @inactive
   @admin
   @destructive
   @4.8 @4.7 @4.10 @4.9


lib/log.rb
```diff
@@ -183,6 +183,21 @@ def reset_dedup
       @dup_buffer.reset if @dup_buffer
     end
 
+    private
+    def sensor(msg)
+      secured_lines = []
+      lines = msg.split("\n")
+      sensor_kw = %W{client_id  client_secret  subscription_id  tenant_id  access_key_id  secret_access_key secret authorization password oauth token}
```
liangxia, line 190: .dockercfg
.dockerconfigjson
kubeConfig
htpasswd
ca.crt / service-ca.crt
tls.crt
tls.key
service-account.key

jhou1: Thank you Liang!
Is there an example of one ore more of the above format available? In the code, I was assuming the texts were in this form`client_id: xxxx`, just wanted to make sure we can cover more forms and keywords.

liangxia: Yes, all of them are in format `foo: bar`. E.g,
```
- apiVersion: v1
  data:
    kubeConfig: YXBpVmVyc2lvbjogdjE***
- apiVersion: v1
  data:
    .dockerconfigjson: eyJhdXRocyI6eyJicmV3LnJlZ2lzdHJ***
```

lib/log.rb
```diff
@@ -183,6 +183,21 @@ def reset_dedup
       @dup_buffer.reset if @dup_buffer
     end
 
+    private
+    def sensor(msg)
+      secured_lines = []
+      lines = msg.split("\n")
+      sensor_kw = %W{client_id client_secret subscription_id tenant_id access_key_id secret_access_key secret authorization password oauth token .dockercfg .dockerconfigjson kubeconfig htpasswd ca.crt service-ca.crt tls.crt tls.key service-account.key}
```
liangxia, line 190: One more `token-secret.value`,
```
  kind: Secret
  metadata:
    annotations:
      kubernetes.io/service-account.name: deployer
      kubernetes.io/service-account.uid: ed96a7b7-cbc4-4b0f-9b1a-fbdc677a8cdc
      openshift.io/token-secret.name: deployer-token-6j4dk
      openshift.io/token-secret.value: eyJhbGciOiJSUzI1***
```

Perhaps sort them ?

lib/log.rb
```diff
@@ -183,6 +183,21 @@ def reset_dedup
       @dup_buffer.reset if @dup_buffer
     end
 
+    private
+    def sensor(msg)
```
xingxingxia, line 187: Minor nag :) : do you want to mean "审计" "敏感"? Seems this word means none of that :)

lib/log.rb
jhou1: +1, the keword `secret` and the updated regexp `/^.*#{kw}.*:.*/` has caught them
```
irb(main):004:0> str="openshift.io/token-secret.value: abcxxxxxxx"
=> "openshift.io/token-secret.value: abcxxxxxxx"
irb(main):005:0> str.match(/^.*secret.*:.*/)
=> #<MatchData "openshift.io/token-secret.value: abcxxxxxxx">
```

lib/log.rb
```diff
@@ -183,6 +183,21 @@ def reset_dedup
...
+      secured_lines = []
+      lines = msg.split("\n")
+      sensor_kw = %W{client_id client_secret subscription_id tenant_id access_key_id secret_access_key secret authorization password oauth token .dockercfg .dockerconfigjson kubeconfig htpasswd ca service-ca tls tls service-account service_account serviceaccount sa}
```
xingxingxia, line 190: If there is client-id, how to cover? Suggestion: maybe using `id\b?` has fuller coverage? Tested:
```
irb(main):026:0> str="client_id.xxx: abc"
=> "client_id.xxx: abc"
irb(main):028:0> str.match(/id\b?.*:.*/)
=> #<MatchData "id.xxx: abc">

irb(main):030:0> str2="client_id_xxx.xxx: abc"
=> "client_id_xxx.xxx: abc"
irb(main):031:0> str2.match(/id\b?.*:.*/)
=> #<MatchData "id_xxx.xxx: abc">
```
Same for the remaining: just use key\b? secret\b? token\b? account\b? etc.

xingxingxia (changes requested):

lib/log.rb
jhou1: Thanks! let me update and test it.

jhou1: Having shorter keywords like `id` would overprotect data in the log, as I tested, this may not help troubleshooting, for example: 
```
      metadata:
        annotations:
          openshift.io/description: ""
          openshift.io/display-name: ""
          openshift.io/requester: testuser-0
          openshift.io/sa.scc.mcs: [PROTECTED_DATA]
          openshift.io/sa.scc.supplemental-groups: [PROTECTED_DATA]
          openshift.io/sa.scc.uid-range: [PROTECTED_DATA]
        creationTimestamp: "2021-11-25T06:13:55Z"
        labels:
          kubernetes.io/metadata.name: sxgj9
        name: sxgj9
        resourceVersion: "50033"
        uid: [PROTECTED_DATA]
```

xingxingxia: Ah, OK, thanks for showing this. Given shorter keywords id, sa are not good, **then \b is not needed either** (\b is used to include xxxid like client-/_id but exclude idxxx like ideas)

jhou1: Sure, thank you

lib/log.rb
jhou1: I mean censoring sensitive information :), I miss spelled it, thanks for catching that! Should be `censor`

jhou1:
@liangxia @xingxingxia @pruan-rht @duanwei33 I've tested against aws/gcp/azure/vsphere/osp cloud providers, data are hidden. If this look ok to you, we could re-enable the tests and monitor the ci next few days.
/unhold

lib/log.rb
```diff
@@ -115,7 +115,7 @@ def log(msg, level=INFO, show_datetime='time')
         timestamp = ''
       end
 
-      m = {msg: msg, level: level, timestamp: timestamp}
+      m = {msg: censor(msg), level: level, timestamp: timestamp}
```
liangxia, line 118: A few lines below, we have `print(construct(m))`, m calls censor, and in the print method, we calls again censor, is it dup ?

jhou1: Nice catch!

jhou1: Fixed

PR #578, opened 2020-01-02
rules added for OCP-23688

yapei:
@xiaocwan  @yanpzhan  Please help review the changes

PR #568, opened 2019-12-30
Auto OCP-19196 and OCP-16495

xiuwang:
@openshift/devexp-qe please help review, thx
Depend on https://github.com/openshift/verification-tests/pull/566

PR #2507, opened 2021-11-24
Network policy ACL Logging Upgrade test (Networking Quality OKR)

asood-rh:
Tested in upgrade path from 4.9->4.10
[
  {
    "completionTime": "2021-11-24T16:01:44Z",
    "image": "registry.ci.openshift.org/ocp/release:4.10.0-0.nightly-2021-11-24-030137",
    "startedTime": "2021-11-24T14:50:19Z",
    "state": "Completed",
    "verified": false,
    "version": "4.10.0-0.nightly-2021-11-24-030137"
  },
  {
    "completionTime": "2021-11-24T13:20:40Z",
    "image": "registry.ci.openshift.org/ocp/release@sha256:9f143e47e6cf2e880af243ab3e23e0b6db0f75559c1782f9edf911906a50878e",
    "startedTime": "2021-11-24T12:52:36Z",
    "state": "Completed",
    "verified": false,
    "version": "4.9.0-0.nightly-2021-11-24-022504"
  }
]

Test logs:-
http://file.rdu.redhat.com/~asood/OCP-44978-pre.log
http://file.rdu.redhat.com/~asood/OCP-44978-post.log

@openshift/team-sdn-qe

### DIFF
--- a/features/upgrade/sdn/policy-upgrade.feature
+++ b/features/upgrade/sdn/policy-upgrade.feature
@@ -517,18 +517,18 @@ Feature: SDN compoment upgrade testing
     """
 
     When I run the :new_project client command with:
-      | project_name | policy-upgrade6 |
+      | project_name | policy-acl-upgrade1 |
     Then the step should succeed
-    And evaluation of `"policy-upgrade6"` is stored in the :proj1 clipboard
+    And evaluation of `"policy-acl-upgrade1"` is stored in the :proj1 clipboard
     When I run the :annotate admin command with:
       | resource     | namespace                                                |
       | resourcename | <%= cb.proj1 %>                                          |
       | keyval       | k8s.ovn.org/acl-logging={"deny":"alert","allow":"alert"} |
     Then the step should succeed
     When I run the :new_project client command with:
-      | project_name | policy-upgrade7 |
+      | project_name | policy-acl-upgrade2 |
     Then the step should succeed
-    And evaluation of `"policy-upgrade7"` is stored in the :proj2 clipboard
+    And evaluation of `"policy-acl-upgrade2"` is stored in the :proj2 clipboard
     When I run the :label admin command with:
       | resource | namespace       |
       | name     | <%= cb.proj2 %> |
@@ -601,8 +601,8 @@ Feature: SDN compoment upgrade testing
     Given I switch to cluster admin pseudo user
     Given the env is using "OVNKubernetes" networkType
     And evaluation of `"rsyslog-server"` is stored in the :proj0 clipboard
-    And evaluation of `"policy-upgrade6"` is stored in the :proj1 clipboard
-    And evaluation of `"policy-upgrade7"` is stored in the :proj2 clipboard
+    And evaluation of `"policy-acl-upgrade1"` is stored in the :proj1 clipboard
+    And evaluation of `"policy-acl-upgrade2"` is stored in the :proj2 clipboard
     Given I use the "<%= cb.proj0 %>" project
     Given I wait for the "rsyslogserver" service to become ready
     Given a pod becomes ready with labels:
--- a/features/upgrade/sdn/policy-upgrade.feature
+++ b/features/upgrade/sdn/policy-upgrade.feature
@@ -576,7 +576,7 @@ Feature: SDN compoment upgrade testing
     
     Given I use the "<%= cb.proj2 %>" project
     When I execute on the "<%= cb.p2pod2name %>" pod:
-      | curl | -I | <%= cb.p1pod1ip %>:8080 | --connect-timeout | 2 |
+      | curl | -I | <%= cb.p1pod1ip %>:8080 | --connect-timeout | 20 |
     Then the output should not contain "200 OK"
     And I wait up to 20 seconds for the steps to pass:
     """
@@ -644,7 +644,7 @@ Feature: SDN compoment upgrade testing
     
     Given I use the "<%= cb.proj2 %>" project
     When I execute on the "<%= cb.p2pod2name %>" pod:
-      | curl | -I | <%= cb.p1pod1ip %>:8080 | --connect-timeout | 2 |
+      | curl | -I | <%= cb.p1pod1ip %>:8080 | --connect-timeout | 20 |
     Then the output should not contain "200 OK"
     Then the step should fail
     And I wait up to 20 seconds for the steps to pass:
--- a/features/upgrade/sdn/policy-upgrade.feature
+++ b/features/upgrade/sdn/policy-upgrade.feature
@@ -477,3 +477,169 @@ Feature: SDN compoment upgrade testing
     And the output should not contain "<%= cb.host_pod1.ip %>"
     """
 
+  # @author asood@redhat.com
+  @admin
+  @upgrade-prepare
+  @4.10 @4.9
+  @aws-ipi @gcp-ipi @vsphere-ipi @azure-ipi @baremetal-ipi @openstack-ipi
+  @aws-upi @gcp-upi @openstack-upi
+  @destructive
+  @network-ovnkubernetes
+  Scenario: Check network policy ACL logging works post upgrade -prepare
+    Given I switch to cluster admin pseudo user
+    Given the env is using "OVNKubernetes" networkType
+    When I run the :new_project client command with:
+      | project_name | rsyslog-server |
+    Then the step should succeed
+    And evaluation of `"rsyslog-server"` is stored in the :proj0 clipboard
+    Given I obtain test data file "networking/syslog/config-map.yaml"
+    Given I obtain test data file "logging/clusterlogforwarder/rsyslog/insecure/rsyslogserver_deployment.yaml"
+    Given I create the resources for the receiver with:
+      | namespace       | <%= cb.proj0 %>               |
+      | receiver_name   | rsyslogserver                 |
+      | configmap_file  | config-map.yaml               |
+      | deployment_file | rsyslogserver_deployment.yaml |
+      | pod_label       | component=rsyslogserver       |
+    And evaluation of `service("rsyslogserver").ip` is stored in the :svc_ip clipboard
+    And evaluation of `pod(0).name` is stored in the :rsyslog_server_pod clipboard
+    Then the step should succeed
+
+    Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with:
+      | {"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"policyAuditConfig": {"destination": "udp:<%= cb.svc_ip %>:514" }}}}} |
+    And I wait up to 300 seconds for the steps to pass:
+    """
+      Given the status of condition "Progressing" for network operator is :False
+      And the status of condition "Available" for network operator is :True
+    """
+
+    When I run the :new_project client command with:
+      | project_name | policy-upgrade6 |
+    Then the step should succeed
+    And evaluation of `"policy-upgrade6"` is stored in the :proj1 clipboard
+    When I run the :annotate admin command with:
+      | resource     | namespace                                                |
+      | resourcename | <%= cb.proj1 %>                                          |
+      | keyval       | k8s.ovn.org/acl-logging={"deny":"alert","allow":"alert"} |
+    Then the step should succeed
+    When I run the :new_project client command with:
+      | project_name | policy-upgrade7 |
+    Then the step should succeed
+    And evaluation of `"policy-upgrade7"` is stored in the :proj2 clipboard
+    When I run the :label admin command with:
+      | resource | namespace       |
+      | name     | <%= cb.proj2 %> |
+      | key_val  | team=operations |
+    Then the step should succeed
+    Given I use the "<%= cb.proj1 %>" project
+    Given I obtain test data file "networking/list_for_pods.json"
+    When I run oc create over "list_for_pods.json" replacing paths:
+      | ["items"][0]["spec"]["replicas"] | 1 |
+    Then the step should succeed
+    Given a pod becomes ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(1).ip_url` is stored in the :p1pod1ip clipboard
+    Given I obtain test data file "networking/networkpolicy/allow-ns-and-pod.yaml"
+    When I run the :create admin command with:
+      | f | allow-ns-and-pod.yaml |
+      | n | <%= cb.proj1 %>       |
+    Then the step should succeed
+    Given I use the "<%= cb.proj2 %>" project
+    Given I obtain test data file "networking/list_for_pods.json"
+    When I run oc create over "list_for_pods.json" replacing paths:
+      | ["items"][0]["spec"]["replicas"] | 1 |
+    Then the step should succeed
+    Given a pod becomes ready with labels:
+      | name=test-pods |
+    And evaluation of `pod(2).name` is stored in the :p2pod1name clipboard
+    Given I obtain test data file "rc/idle-rc-1.yaml"
+    When I run oc create over "idle-rc-1.yaml" replacing paths:
+      | ["items"][0]["spec"]["replicas"] | 1 |
+    Then the step should succeed
+    Given a pod becomes ready with labels:
+      | name=hello-idle |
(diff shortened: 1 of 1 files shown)


features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -477,3 +477,169 @@ Feature: SDN compoment upgrade testing
...
+      | configmap_file  | config-map.yaml               |
+      | deployment_file | rsyslogserver_deployment.yaml |
+      | pod_label       | component=rsyslogserver       |
+    And evaluation of `service("rsyslogserver").ip` is stored in the :svc_ip clipboard
+    And evaluation of `pod(0).name` is stored in the :rsyslog_server_pod clipboard
+    Then the step should succeed
+
+    Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with:
+      | {"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"policyAuditConfig": {"destination": "udp:<%= cb.svc_ip %>:514" }}}}} |
+    And I wait up to 300 seconds for the steps to pass:
```
zhaozhanqi, line 509: Not sure if network operator will update the status immediately after the patch. 
if we can add one step before this check:
```
 """
 Given the status of condition "Progressing" for network operator is :True
 """
And I wait up to 300 seconds for the steps to pass:
"""
Given the status of condition "Progressing" for network operator is :False
..
..
```

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -477,3 +477,173 @@ Feature: SDN compoment upgrade testing
...
+  # @author asood@redhat.com
+  # @case_id OCP-44978
+  @admin
+  @upgrade-check
+  @4.10 @4.9
+  @aws-ipi @gcp-ipi @vsphere-ipi @azure-ipi @baremetal-ipi @openstack-ipi
+  @aws-upi @gcp-upi @openstack-upi
+  @destructive
+  @network-ovnkubernetes
+  Scenario: Check network policy ACL logging works post upgrade -prepare
```
anuragthehatter, line 596: ``` - prepare ``` not needed here

features/upgrade/sdn/policy-upgrade.feature
asood-rh: @zhaozhanqi I have added the wait in the upgrade prepare script. I do have reservations because all the local runs I executed script did not fail. 
I did see consistent failure if the wait is added to upgrade check so I am not adding wait there.

http://file.rdu.redhat.com/~asood/OCP-44978-pre-0.log
http://file.rdu.redhat.com/~asood/OCP-44978-post-0.log

oc get clusterversion -o json|jq ".items[0].status.history"
[
  {
    "completionTime": "2021-11-29T16:56:27Z",
    "image": "registry.ci.openshift.org/ocp/release:4.10.0-0.nightly-2021-11-28-164900",
    "startedTime": "2021-11-29T15:48:29Z",
    "state": "Completed",
    "verified": false,
    "version": "4.10.0-0.nightly-2021-11-28-164900"
  },
  {
    "completionTime": "2021-11-29T15:31:06Z",
    "image": "registry.ci.openshift.org/ocp/release@sha256:e589c2f68177d986e50c023f7599c43e8c846a6437e406734f79a2236d4cb6ad",
    "startedTime": "2021-11-29T15:09:35Z",
    "state": "Completed",
    "verified": false,
    "version": "4.9.0-0.nightly-2021-11-29-063111"
  }
]

anuragthehatter: @asood suggesting to check this above 30 second step to makes sure it satisfy 30 seconds everytime. This might change based on platform we are using. I believe we may not need it.

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -477,3 +477,173 @@ Feature: SDN compoment upgrade testing
...
+      | name=test-pods |
+    And evaluation of `pod(2).name` is stored in the :p2pod1name clipboard
+    Given I obtain test data file "rc/idle-rc-1.yaml"
+    When I run oc create over "idle-rc-1.yaml" replacing paths:
+      | ["items"][0]["spec"]["replicas"] | 1 |
+    Then the step should succeed
+    Given a pod becomes ready with labels:
+      | name=hello-idle |
+    And evaluation of `pod(3).name` is stored in the :p2pod2name clipboard
+    And I wait up to 20 seconds for the steps to pass:
```
anuragthehatter, line 565: @asood-rh If you think particular step in this block expects a wait, please add there. With 20 seconds wait for whole block and couple of --connect-timeouts, doesn't making sense at this point IMO.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: @anuragthehatter easy for eyes to distinguish between two phases.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: @anuragthehatter Need the wait as there are network policies that are added to the project. Need the --connect-timeout so that curl requests terminate after specified time where curl is expected to fail because of policy.
I have removed the -connect-timeout where curl request will succeed.

features/upgrade/sdn/policy-upgrade.feature
anuragthehatter: @asood-rh Your both phases are ```- prepare``` :)

features/upgrade/sdn/policy-upgrade.feature
anuragthehatter: @asood-rh The flows typically takes milliseconds to be effective. I still suggest keeping timeouts in both cases for at least 2 seconds. You are creating a pod at line #559 (which will trigger some ovn flows) and immediately a curl after that without timeout. Your curl might get lucky sometimes or not depending on platforms configuration/cpu/mem etc.

As per my understanding you need a specific time loop at 572-574 and 582-584 only. And connect timeouts of 2s should suffice in both cases as discussed.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: @anuragthehatter Fixed.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: @anuragthehatter  So now both --connect-timeout are fine but need two time loops? Would not add double the time in script? Please clarify.

anuragthehatter: >So now both --connect-timeout are fine

Suggesting --connect-timeout 2 at least in both #568 and #578 based on above discussion

> Would not add double the time in script

That seems logically right Its a sequential execution flow as we know. If we are putting connect-timeouts wait under another timer loop,  We are kinda creating time { time } loop. If you think /var/log will take 20 seconds to populate keyword you are looking for, its okay to loop that for 20s (Thats kinda would be strange time in load tests) Same for following /val/log. Just my 3 cents :)

anuragthehatter: Sure.Thanks @asood-rh  And I also meant following logic for both /var/log steps. Its good idea to keep them separated as discussed above. It would also make reviewer understand that 20 second is for /var/log steps but not for curl loops.

``` 
And I wait up to 20 seconds for the steps to pass:
    """
        When I execute on the "<%= cb.rsyslog_server_pod %>" pod:
      | grep | -w | verdict=allow | /var/log/messages |
    Then the output should contain "verdict=allow"
    """
```

asood-rh: @anuragthehatter As discussed earlier breaking the block into two blocks adds more time to the script and I do want retry the curl if it fails the first time in the given time for the steps to pass.

anuragthehatter: We are creating normal pods across 2 projects incorporating an allow-ns-and-pod policy. If the curl expects to work in upto 20 seconds across the pods, its definitely a bug as this will get worse in perf test iterations. Also breaking the blocks won't increase the script time but should be same 20 seconds or less in PASS'ing situation (if you have waited 20 sec for verdict=drop, verdict=allow would also populate by that time so next 20 sec wait wont wait fully). In fail case, which we expect once in a while it might go up to 40 seconds and that will help rule out issue easily bte curls vs syslog.

Anyways leaving upto you to decide, just a suggestion semantically..

asood-rh: @anuragthehatter Two blocks as two set of steps are not related but retries needed for curl as well as the syslog log messages.

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -477,3 +477,181 @@ Feature: SDN compoment upgrade testing
...
+      | curl | -I | <%= cb.p1pod1ip %>:8080 | --connect-timeout | 5 | 
+    Then the output should contain "200 OK"
+    Then the step should succeed
+    Given I use the "<%= cb.proj0 %>" project
+    When I execute on the "<%= cb.rsyslog_server_pod %>" pod:
+      | grep | -w | verdict=allow | /var/log/messages |
+    Then the output should contain "verdict=allow"
+    """
+    
+    And I wait up to 20 seconds for the steps to pass:
```
zhaozhanqi, line 646: can we make line 633 and 646 in one loop `And I wait up to 20 seconds for the steps to pass:`

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -477,3 +477,181 @@ Feature: SDN compoment upgrade testing
...
+      Given the status of condition "Progressing" for network operator is :True
+    """
+    And I wait up to 300 seconds for the steps to pass:
+    """
+      Given the status of condition "Progressing" for network operator is :False
+      And the status of condition "Available" for network operator is :True
+    """
+
+    When I run the :new_project client command with:
+      | project_name | policy-upgrade6 |
```
zhaozhanqi, line 520: I suggest we can use one meaningful name instead of `policy-upgrade6`

features/upgrade/sdn/policy-upgrade.feature
asood-rh: @zhaozhanqi That is how I started with one block but Anurag insisted on having two blocks, I agreed as two steps in each block are related i.e with curl there ACL messages would not logged.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: All the test I coded are related to network policy so I just followed the pattern of using sequence. I can change to policy_acl_upgrade1 and policy_acl_upgrade2 if that works. It is easy change as I have all names defined in beginning.

features/upgrade/sdn/policy-upgrade.feature
zhaozhanqi: hmm,  if using two blocks, I think we better remove this part from the block
```
    Given I use the "<%= cb.proj2 %>" project
    When I execute on the "<%= cb.p2pod2name %>" pod:
      | curl | -I | <%= cb.p1pod1ip %>:8080 | --connect-timeout | 2 |
    Then the output should not contain "200 OK"
    Then the step should fail
```
the block only contain
```
And I wait up to 20 seconds for the steps to pass:
 """
    When I execute on the "<%= cb.rsyslog_server_pod %>" pod:
      | grep | -w | verdict=drop | /var/log/messages |
    Then the output should contain "verdict=drop"
    """
```

asood-rh: @zhaozhanqi curl request causes these ACL messages to be logged so we need the request.

zhaozhanqi: we should not make `    Given I use the "<%= cb.proj2 %>" project` in the loop. 

the is my suggestion:
```
    Given I use the "<%= cb.proj2 %>" project
    When I execute on the "<%= cb.p2pod1name %>" pod:
      | curl | -I | <%= cb.p1pod1ip %>:8080 | --connect-timeout | 5 | 
    Then the step should succeed
    Then the output should contain "200 OK"
    Given I use the "<%= cb.proj0 %>" project
    And I wait up to 20 seconds for the steps to pass:
    """
    When I execute on the "<%= cb.rsyslog_server_pod %>" pod:
      | grep | -w | verdict=allow | /var/log/messages |
    Then the step should succeed
    Then the output should contain "verdict=allow"
    """

    Given I use the "<%= cb.proj2 %>" project
    When I execute on the "<%= cb.p2pod2name %>" pod:
      | curl | -I | <%= cb.p1pod1ip %>:8080 | --connect-timeout | 2 |
    Then the output should not contain "200 OK"
    Then the step should fail

    Given I use the "<%= cb.proj0 %>" project
    And I wait up to 20 seconds for the steps to pass:
    """
    When I execute on the "<%= cb.rsyslog_server_pod %>" pod:
      | grep | -w | verdict=drop | /var/log/messages |
    Then the step should succeed
    Then the output should contain "verdict=drop"
    """
```

asood-rh: Need to have curl in the block so that it can be retried if it fails. if it fails there is no point waiting for syslog messages as they will not be any.

I am not sure how switching the project can be avoided in this case.

zhaozhanqi: I do not think we need to add curl into loop to retry since we have `--connect-timeout`,  we can set to 20 if 5 is not enough.

asood-rh: -connect-timeout does not ensure retry it is only one request with retry success of the test is ensured.

zhaozhanqi: I think with --connect-timeout is better and enough. 
we need to make sure one session can be logged.   if curl many times.  how to make sure the first session is logged.

features/upgrade/sdn/policy-upgrade.feature
asood-rh: Fixed

features/upgrade/sdn/policy-upgrade.feature
anuragthehatter: This is same suggestion as https://github.com/openshift/verification-tests/pull/2507#discussion_r759538832. @asood-rh Why dont you try with same logic and report any issues here. Then we can go from there i guess

asood-rh: @anuragthehatter There is suggestion about having all the steps in one block as well from which it is being insisted to move away. The concern is about switching the project in the loop. As stated earlier curl needs to be retried to ensure ACL messages are generated. There is consideration to make test reliable not just pass in local run.

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -477,3 +477,180 @@ Feature: SDN compoment upgrade testing
...
+    And evaluation of `"policy-acl-upgrade2"` is stored in the :proj2 clipboard
+    Given I use the "<%= cb.proj0 %>" project
+    Given I wait for the "rsyslogserver" service to become ready
+    Given a pod becomes ready with labels:
+      | appname=rsyslogserver |
+    And evaluation of `service("rsyslogserver").ip` is stored in the :svc_ip clipboard
+    And evaluation of `pod(0).name` is stored in the :rsyslog_server_pod clipboard
+    Then the step should succeed
+
+    Given as admin I successfully merge patch resource "networks.operator.openshift.io/cluster" with:
```
zhaozhanqi, line 613: do we still need this patch since it already be done before upgrade?

asood-rh: Yes, because post upgrade syslog server pod is recreated with different IP address.

anuragthehatter:
Thanks @asood-rh for incorporating few changes.

/lgtm

features/upgrade/sdn/policy-upgrade.feature
asood-rh: Thanks @zhaozhanqi for the review. 
Fixed it.

zhaozhanqi: Thanks for updating this. it looks good to me

PR #2511, opened 2021-11-25
Add @users=upuser1,upuser2 to the uprade prepare & check

kasturinarra:
As of today test was failing due to the error "no specification for user index 0 in a scenario @users tag (RuntimeError)", added @users=upuser1,upuser2 to fix the same.

### DIFF
--- a/features/upgrade/workloads/descheduler-upgrade.feature
+++ b/features/upgrade/workloads/descheduler-upgrade.feature
@@ -1,11 +1,12 @@
 Feature: Descheduler major upgrade should work fine
   # @author knarra@redhat.com
-  @upgrade-prepare
   @admin
+  @upgrade-prepare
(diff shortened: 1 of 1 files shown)


kasturinarra:
@zhouying7780 could you please help review ? Below are the run logs for upgrade-prepare & upgrade-check

upgrade-prepare: https://privatebin-it-iso.int.open.paas.redhat.com/?6f05efaa23b7e23a#4TAWvhC447Ea1rrVxtVpie2L7dBZaag4QXrBNae58XDY
upgrade-check: https://privatebin-it-iso.int.open.paas.redhat.com/?1f30c1a53dc21a54#7s85HKxou8uifZkHNXuBbNrF4zeRnmauBtD43CfK5dYZ

PR #2514, opened 2021-11-26
enhance case to fix issue OCPQE-7171

lihongyan1:
Enhanced rule file to fix issue
https://issues.redhat.com/browse/OCPQE-7171


lihongyan1:
@juzhao Please help review

PR #2515, opened 2021-11-26
Optimize OCP-40620

liangxia:
` wait up to xx seconds for the steps to pass` loop the steps inside it until reach the desired status, or fail due to timeout reached.
While the operation to switch to pseudo admin and switch project only need to do once, so move it out of the loop step.


lihongan:
/lgtm
Thank you!

asood-rh (approved):
/lgtm

PR #2516, opened 2021-11-26
Fix undefined method `split` for Array

liangxia:
Fix issue in https://coreos.slack.com/archives/CH76YSYSC/p1637906272082100?thread_ts=1637724031.447100&cid=CH76YSYSC

/cc @jhou1 @dis016 @JianLi-RH @pruan-rht

### DIFF
--- a/lib/log.rb
+++ b/lib/log.rb
@@ -186,22 +186,28 @@ module BushSlicer
     private
     def censor(msg)
       return if msg.nil?
-      secured_lines = []
-      lines = msg.split("\n")
       censor_kw = [
         'data:',
         '"data":',
         'kind: secret',
         '"kind": "secret"'
       ]
-      lines.each do |line|
-        if censor_kw.any? { |kw| line.downcase.include?(kw) }
-          secured_lines.clear()
-          break
+      secured_lines = []
+      if (msg.is_a?(String))
+        lines = msg.split("\n")
+        lines.each do |line|
+          if censor_kw.any? { |kw| line.downcase.include?(kw) }
+            secured_lines.clear()
+            break
+          end
+          secured_lines << line
         end
-        secured_lines << line
+        secured_lines.join("\n")
+      elsif (msg.is_a?(Array))
+        msg.map do |str|
+          censor(str)
+        end
       end
-      secured_lines.join("\n")
     end
 
     # map messages to unique characters, then run a regular expression to


lib/log.rb
```diff
@@ -186,22 +186,28 @@ def reset_dedup
...
+        lines = msg.split("\n")
+        lines.each do |line|
+          if censor_kw.any? { |kw| line.downcase.include?(kw) }
+            secured_lines.clear()
+            break
+          end
+          secured_lines << line
+        end
+        secured_lines.join("\n")
+      elsif (msg.is_a?(Array))
```
jhou1, line 206: Do we have some tests with condition `msg` is an array? Thanks!

jhou1: Asking because there is no handling of return value when the `msg` is an array

liangxia: Tested with the scenario reported in Slack,
```
    And I store an available router IP in the :router_ip clipboard           # features/step_definitions/route.rb:135
      [06:43:41] INFO> Shell Commands: oc create -f /home/lxia/code-bases/github.com/verification-tests/testdata/networking/service_with_selector.json --kubeconfig=/home/lxia/workdir/lxiabox-lxiax/ocp4_testuser-43.kubeconfig
      service/selector-service created
      [06:43:42] INFO> Exit Status: 0
...
    When I execute on the "hello-pod" pod:                                   # features/step_definitions/pod.rb:216
      | curl |
      | --resolve |
      | zhao-ingress.example.com:443:<%= cb.router_ip[0] %> |
      | https://zhao-ingress.example.com/ |
      | --cacert |
      | /tmp/ca-test.pem |
      [06:44:49] INFO> Shell Commands: oc exec hello-pod  --kubeconfig=/home/lxia/workdir/lxiabox-lxiax/ocp4_testuser-43.kubeconfig -n 4mro7  -i -- curl --resolve zhao-ingress.example.com:443:***.***.242.30 https://zhao-ingress.example.com/ --cacert /tmp/ca-test.pem
      Hello-OpenShift web-server-1 http-8080
      
      [06:44:52] INFO> Exit Status: 0
    Then the step should succeed                                             # features/step_definitions/common.rb:4
    And the output should contain "Hello-OpenShift"                          # features/step_definitions/common.rb:113
```

jhou1: OK, thanks. I think the return value here in this `elsif` condition is the original array object(`msg`), in order to return the updated value, I think we need `msg.map do |str| ... end`.

liangxia: Updated.
Retest the scenario passed.

PR #2518, opened 2021-11-26
Simplify OCP-29797

liangxia:
/cc @lihongyan1 @juzhao

### DIFF
--- a/features/upgrade/monitoring/monitoring-upgrade.feature
+++ b/features/upgrade/monitoring/monitoring-upgrade.feature
@@ -24,19 +24,10 @@ Feature: cluster monitoring related upgrade check
     Given I switch to cluster admin pseudo user
     And I use the "openshift-monitoring" project
     Given all pods in the project are ready
-    Then the step should succeed
-    When I run the :get client command with:
-      | resource      | clusteroperator |
-      | resource_name | monitoring      |
-    And evaluation of `@result[:stdout].split(/\n/)[1].split(/\s+/)[4]` is stored in the :degreaded_status clipboard
-    Then the expression should be true> cb.degreaded_status == "False"
+    And the expression should be true> cluster_operator('monitoring').condition(type: 'Degraded')['status'] == "False"
 
     #check retention time
-    When I run the :get client command with:
-      | resource      | prometheus |
-      | resource_name | k8s        |
-      | o             | yaml       |
-    Then the expression should be true> YAML.load(@result[:stdout])["spec"]["retention"] == "3h"
+    Given the expression should be true> prometheus('k8s').retention == "3h"
 
     # get sa/prometheus-k8s token
     When evaluation of `secret(service_account('prometheus-k8s').get_secret_names.find {|s| s.match('token')}).token` is stored in the :sa_token clipboard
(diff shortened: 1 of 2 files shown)


features/upgrade/monitoring/monitoring-upgrade.feature
```diff
@@ -24,19 +24,10 @@ Feature: cluster monitoring related upgrade check
...
-    Then the expression should be true> cb.degreaded_status == "False"
+    And the expression should be true> cluster_operator('monitoring').condition(type: 'Degraded')['status'] == "False"
 
     #check retention time
-    When I run the :get client command with:
-      | resource      | prometheus |
-      | resource_name | k8s        |
-      | o             | yaml       |
-    Then the expression should be true> YAML.load(@result[:stdout])["spec"]["retention"] == "3h"
+    Given the expression should be true> prometheus('k8s').retention == "3h"
```
lihongyan1, line 30: suppose should be changed as prometheus('k8s').retention(cached: false) == "3h"

features/upgrade/monitoring/monitoring-upgrade.feature
```diff
@@ -24,19 +24,10 @@ Feature: cluster monitoring related upgrade check
     Given I switch to cluster admin pseudo user
     And I use the "openshift-monitoring" project
     Given all pods in the project are ready
-    Then the step should succeed
-    When I run the :get client command with:
-      | resource      | clusteroperator |
-      | resource_name | monitoring      |
-    And evaluation of `@result[:stdout].split(/\n/)[1].split(/\s+/)[4]` is stored in the :degreaded_status clipboard
-    Then the expression should be true> cb.degreaded_status == "False"
+    And the expression should be true> cluster_operator('monitoring').condition(type: 'Degraded')['status'] == "False"
```
lihongyan1, line 27: Have you execute the code? I didn't find the definition of method condition

features/upgrade/monitoring/monitoring-upgrade.feature
liangxia: `cached:false` should be used only for values that change frequently. E.g, status.

features/upgrade/monitoring/monitoring-upgrade.feature
liangxia:
```
    def conditions(user: nil, cached: true, quiet: false)
      rr.dig('status', 'conditions')
    end

    # returns a hash simliar to
    # {"lastTransitionTime"=>2019-03-23 07:33:34 UTC, "status"=>"False", "type"=>"Progressing"}
    def condition(type: nil, user: nil, cached: true, quiet: false)
      res = self.conditions(user: user, cached: cached, quiet: quiet).select { |c| c['type'] == type }
      res.first
    end
```

features/upgrade/monitoring/monitoring-upgrade.feature
lihongyan1: By default the retention is a different value, I just configure it as 3h and check the value, not sure it will get the new value.

features/upgrade/monitoring/monitoring-upgrade.feature
lihongyan1: thanks, I find it.

features/upgrade/monitoring/monitoring-upgrade.feature
liangxia: Feel free to raise a PR to fix that if it uses the cached value wrongly.

lihongyan1: It should be ok if no other tests read the retention before I change configuration.

PR #2519, opened 2021-11-27
Add right image and params in data file

kasturinarra:
I see that the test case fails intermittently so when checked the old data file[1]  i see that the image there was a sigcapture and not hello-openshift and few other params were different, so updating the data file to match the same and hopefully test will pass 100%

[1] https://raw.githubusercontent.com/openshift-qe/v3-testfiles/master/pods/graceful-delete/10.json

### DIFF
--- a/testdata/pods/graceful-delete/10.json
+++ b/testdata/pods/graceful-delete/10.json
@@ -12,25 +12,11 @@
     "containers": [
       {
         "name": "hello-openshift",
-        "image": "quay.io/openshifttest/hello-openshift@sha256:0ed6e7163863e52438a5ed189d82e572a6c2ddd1b4db224c2c4b0688fee2c2cc",
+        "image": "quay.io/openshifttest/sigcapture@sha256:e9c7ecf75c57def20f1964a119e858a7a45de9d3b21a384f5a32d9562a15270b",
         "ports": [
           {
             "containerPort": 8080,
             "protocol": "TCP"
-          }
-        ],
-        "env": [         
-          {
-            "name": "PORT",
-            "value": ""
-          },
-          {
-            "name": "SECOND_PORT",
-            "value": ""
-          },
-          {
-            "name": "TRAP_SIGNALS",
-            "value": "yes"
           }
         ],
         "resources": {},


kasturinarra:
@zhouying7780 could you please help review the same ? Thanks !!

kasturinarra:

aleskandro:
This would be a revert to the previous non multi-arch image, not a fix.

Could you please share the test cases that are failing?

The hello-openshift image in the affected pod is an extended version that should fit the sigcapture image's behavior.

Its code is available at https://gitlab.cee.redhat.com/aosqe/tests-images/-/blob/master/hello-openshift/hello_openshift.go

If the image doesn't fit for some reason we could either fix the code or change to another image.

However, If you want to change the image, it should be (1) runnable both on arm and amd64, and (2) versioned by the tests-images repository.

kasturinarra:
> This would be a revert to the previous non multi-arch image, not a fix.
> 
> Could you please share the test cases that are failing?
> 
> The hello-openshift image in the affected pod is an extended version that should fit the sigcapture image's behavior.
> 
> Its code is available at https://gitlab.cee.redhat.com/aosqe/tests-images/-/blob/master/hello-openshift/hello_openshift.go
> 
> If the image doesn't fit for some reason we could either fix the code or change to another image.
> 
> However, If you want to change the image, it should be (1) runnable both on arm and amd64, and (2) versioned by the tests-images repository.

ah, okay, completely forgot about it. Test is failing because pod does not get terminated with in 30 seconds, it takes more than that. could not get the log as it is expired.

aleskandro:
@kasturinarra, the log you sent is using the original image, and it seems to pass.

```
11-29 11:04:41.761          - env:
11-29 11:04:41.761            - name: PORT
11-29 11:04:41.761            - name: SECOND_PORT
11-29 11:04:41.761            - name: TRAP_SIGNALS
11-29 11:04:41.761              value: "yes"
11-29 11:04:41.761            image: quay.io/openshifttest/hello-openshift@sha256:0ed6e7163863e52438a5ed189d82e572a6c2ddd1b4db224c2c4b0688fee2c2cc
11-29 11:04:41.761            imagePullPolicy: IfNotPresent
11-29 11:04:41.761            name: hello-openshift
11-29 11:04:41.761            ports:
```

aleskandro:
Moreover, I don't understand how in this piece of code the pod should not be "scheduled":
Are the nodes set unschedulable before?

```ruby
  # @author chezhang@redhat.com
  # @case_id OCP-11753
  @4.10 @4.9
  Scenario: Pod should be immediately deleted if it's not scheduled even if graceful termination period is set
    Given I have a project
    Given I obtain test data file "pods/graceful-delete/10.json"
    When I run the :create client command with:
      | f | 10.json |
    Then the step should succeed
    Given the pod named "grace10" becomes ready
    When I run the :delete background client command with:
      | object_type       | pods    |
      | object_name_or_id | grace10 |
    Then the step should succeed
    Given the pod named "grace10" becomes terminating
    Then I wait for the resource "pod" named "grace10" to disappear within 30 seconds

```

kasturinarra:
@aleskandro yes, nodes are set unschedulable before that, you are right, I am anyway closing this as we do not need to change any params here, i will see how can i get this fixed, thanks !!

PR #2524, opened 2021-11-29
Hide sensitive oc commands' output from framework improvement

xingxingxia:
This PR is added for https://github.com/openshift/cucushift/pull/8915#issuecomment-981426788 .
Once this is merged, https://github.com/openshift/cucushift/pull/8915 is not needed.
This only deals with one command for figuring out the place to hide sensitive oc commands' output via framework level, such that it can merge quicker. For more commands, separate PR could follow up, I'd like to leave to core auto guys : )
Pass log: https://url.corp.redhat.com/runner-v3-smoke-3756-console
@jhou1 @liangxia @dis016 @JianLi-RH @pruan-rht : please review / merge, thanks ~ !
CC @kasturinarra @zhouying7780 @wangke19 @lihongyan1 @XiyunZhao of original PR.

### DIFF
--- a/lib/cli_executor.rb
+++ b/lib/cli_executor.rb
@@ -305,8 +305,13 @@ module BushSlicer
 
       add_proxy_env_opt(user.env, opts)
       cli_tool = tool_from_opts!(opts)
+      if key == :serviceaccounts_get_token
+        # no worry if opts already has set it; below will override
+        opts << [:_quiet, true]
+      end
+      new_opts = Common::Rules.merge_opts(logged_users[user.id], opts)
       executor(cli_tool: cli_tool).
-        run(key, Common::Rules.merge_opts(logged_users[user.id], opts))
+        run(key, new_opts)
     end
 
     def clean_up


lib/cli_executor.rb
```diff
@@ -305,8 +305,13 @@ def exec(user, key, opts={})
 
       add_proxy_env_opt(user.env, opts)
       cli_tool = tool_from_opts!(opts)
+      if key == :serviceaccounts_get_token
+        # no worry if opts already has set it; below will override
+        opts << [:_quiet, true]
+      end
+      new_opts = Common::Rules.merge_opts(logged_users[user.id], opts)
```
xingxingxia, line 312: FYI, added below to debug when developing. Now removed below:
```
print "key is ", key, "\n"
print "opts is ", new_opts, "\n"
```

features/mytesting.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Test feature
```
liangxia, line 1: Test scenarios should goes to `features/test/`

xingxingxia: Removed it. Need not add it as test scenario. Only for quicker test for pass log.

PR #2526, opened 2021-11-30
Hide sensitive info from CMDs/logs

liangxia:
Thanks for @xingxingxia's idea.

@jhou1 @xingxingxia  @dis016 @JianLi-RH @pruan-rht : please help review, thanks ~ !

CC @kasturinarra @zhouying7780 @wangke19 @lihongyan1 @XiyunZhao of original PR.

Related with https://github.com/openshift/cucushift/pull/8915 and https://github.com/openshift/verification-tests/pull/2524

### DIFF
--- a/lib/cli_executor.rb
+++ b/lib/cli_executor.rb
@@ -10,6 +10,10 @@ module BushSlicer
 
     RULES_DIR = File.expand_path(HOME + "/lib/rules/cli")
     LOGIN_TIMEOUT = 20 # seconds
+    CENSOR_LOGS = [
+      'serviceaccounts_get_token',
+      'whoami',
+    ]
 
     attr_reader :opts
 
@@ -305,6 +309,9 @@ module BushSlicer
 
       add_proxy_env_opt(user.env, opts)
       cli_tool = tool_from_opts!(opts)
+      if CENSOR_LOGS.any? { |cmd| key.match?(cmd) }
+        opts << [:_quiet, true]
+      end
       executor(cli_tool: cli_tool).
         run(key, Common::Rules.merge_opts(logged_users[user.id], opts))
     end
--- a/lib/local_process.rb
+++ b/lib/local_process.rb
@@ -10,6 +10,12 @@ module BushSlicer
     include Common::Helper
 
     DEFAULT_TIMEOUT = 3600
+    CENSOR_CMDS = [
+      'oc exec alertmanager',
+      'oc exec elasticsearch',
+      'oc exec prometheus',
+      'oc login',
+    ]
 
     attr_reader :pid
     attr_accessor :exit_status, :wait_thread, :cmd, :opts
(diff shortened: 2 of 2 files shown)


lib/cli_executor.rb
```diff
@@ -305,6 +309,9 @@ def exec(user, key, opts={})
 
       add_proxy_env_opt(user.env, opts)
       cli_tool = tool_from_opts!(opts)
+      if CENSOR_LOGS.any? { |cmd| key.match?(cmd) }
+        opts << [:_quiet, true]
+      end
       executor(cli_tool: cli_tool).
         run(key, Common::Rules.merge_opts(logged_users[user.id], opts))
     end
```
xingxingxia, line 317: After more looking at the stacks from https://github.com/openshift/verification-tests/pull/2524/files, checked `executor.run` of lib/rules_command_executor.rb, it has `cmd, cmd_options, exec_options`, it is better place to censor the cmd, cmd_options than here because it can use RE to match the cmd like like get-token, whoami -t, get secret -o yaml/json, etc.

xingxingxia: But I then searched :response in `lib/local_process.rb`, find that is **even further better** place to censor, we can censor command before `logger.plain(result[:response]...`. Via this way, the shell command can be still printed, which is useful, only shell command result is hidden. Like:
```
          if cmd matched censored commands like get-token, whoami -t, get secret -o yaml/json, etc.
            is_censored_cmd = true
          end
          unless opts[:quiet] || is_censored_cmd
            logger.plain(result[:response], false)
```
Like demo https://github.com/xingxingxia/verification-tests/pull/1/files , its test log: https://...ocp-c1.prod.psi.redhat.com/job/runner-v3-smoke/3760/console , it can censor any auto commands, not only oc.
Could you help to elaborate based on this? (No further time on it; need allocate time on epics test : ) )
Thanks!

liangxia: We can iterate to better and better.
Currently the tests on Prow are all disabled/removed due to those sensitive info in the logs.

liangxia:
BTW, secure the sensitive info is one of the key responsibility for everyone in the team. We can not only rely on specific people/sub-team.

lib/local_process.rb
```diff
@@ -10,6 +10,12 @@ class LocalProcess
     include Common::Helper
 
     DEFAULT_TIMEOUT = 3600
+    CENSORT_CMDS = [
```
jhou1, line 13: `CENSORT_CMDS` -> `CENSOR_CMDS`

lib/local_process.rb
```diff
@@ -30,13 +36,7 @@ def initialize(*cmd, **opts)
...
       end
       log_text << result[:command]
-      censor_cmds = [
-        'oc exec alertmanager',
-        'oc exec elasticsearch',
-        'oc exec prometheus',
-        'oc login',
-      ]
-      if censor_cmds.any? { |cmd| log_text.downcase.include?(cmd) }
+      if CENSORT_CMDS.any? { |cmd| log_text.downcase.include?(cmd) }
```
jhou1, line 39: `CENSORT_CMDS` -> `CENSOR_CMDS`

lib/local_process.rb
liangxia: Good catch. Fixed.

jhou1:
Merging to unblock Prow, we can also revisit the approach added by Xingxing. This will be a continuous effort ;)

/lgtm
/approve

PR #2544, opened 2021-12-03
OVN: fix find leader for IPv6

rbbratta:
Fix the regex to match

```
Servers:
  6e24 (6e24 at ssl:[fd2e:6f44:5dd8::81]:9643) (self) next_index=11214 match_index=11517
  90fb (90fb at ssl:[fd2e:6f44:5dd8::68]:9643) next_index=11518 match_index=11517 last msg 350 ms ago
```

match first string in the parens, the everything from the first colon to a colon digit
close-paren sequence


rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/ovn-ipv6.log

PR #2545, opened 2021-12-06
update click_first_item_in_grid_cell_list

yapei:
We only want to click the first item in table no matter its header,  `Name` or `Image name` or something else, so removed `check_name_column_header` check


XiyunZhao (approved):

yapei:
approving since only web changes and got team member approval
/approve

PR #2548, opened 2021-12-06
modification for recreating hello pod for ping

melvinjoseph86:
After the actual upgrade it is notice from CI run, the test-client-pod is not present in the cluster. So the step is modified to include creation of Hello pod and then executing in it.


melvinjoseph86:
@lihongan  @quarterpin @ShudiLi please have  a look. Thanks.

lihongan:
/lgtm
/approve
Thank you @melvinjoseph86

PR #2551, opened 2021-12-06
OCP-46285

anuragthehatter:
Customer sceario https://bugzilla.redhat.com/show_bug.cgi?id=2009873
Logs: http://pastebin.test.redhat.com/

@openshift/team-sdn-qe

### DIFF
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -609,7 +609,8 @@ Feature: OVN related networking scenarios
   
   Given I use the "<%= cb.apiVIP_node %>" node
   And the host is rebooted and I wait it up to 600 seconds to become available
-  #Make sure after the reboot apiVIp switches to new node and only 1 entry correspond to apiVIP exist in NB db (shouldn't be any stale or duplicates)
+  #Make sure after the reboot apiVIp switches to new node and only 1 entry correspond to apiVIP 
+  #exist in NB db (shouldn't be any stale or duplicates)
   Given I store the ovnkube-master "north" leader pod in the clipboard
   And admin executes on the pod "northd" container:
     | bash | -c | ovn-nbctl find logical_router_policy \| grep -B 5 -A 10 <%= cb.apiVIP %> |
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -582,53 +582,53 @@ Feature: OVN related networking scenarios
   @network-ovnkubernetes
   @vsphere-ipi
   Scenario: Logical Router Policies and Annotations for a given node should be current
-  Given the env is using "OVNKubernetes" networkType
-  #Find apiVIP address of the cluster
-  When I run the :get admin command with:
-    | resource | infrastructure                                                         |
-    | o        | jsonpath={.items[*].status.platformStatus.vsphere.apiServerInternalIP} |
-  Then the step should succeed
-  And evaluation of `@result[:response]` is stored in the :apiVIP clipboard
-  Given I store the ovnkube-master "north" leader pod in the clipboard
-  And admin executes on the pod "northd" container:
-    | bash | -c | ovn-nbctl find logical_router_policy \| grep -B 5 -A 10 <%= cb.apiVIP %> |
-  Then the step should succeed
-  And evaluation of `@result[:response].match(/\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}/)[0]` is stored in the :apiVIP_node clipboard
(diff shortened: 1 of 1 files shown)


features/networking/ovn.feature
```diff
@@ -573,3 +573,57 @@ Feature: OVN related networking scenarios
...
+  #Checking annotation exist for the node
+  When I run the :describe admin command with:
+    | resource | node                  |
+    | name     | <%= cb.apiVIP_node %> |
+  Then the step should succeed
+  And the output should contain:
+    | k8s.ovn.org/host-addresses |
+    | <%= cb.apiVIP %>           |
+  
+Given I use the "<%= cb.apiVIP_node %>" node
```
pruan-rht, line 611: @anuragthehatter indentation incorrect

features/networking/ovn.feature
```diff
@@ -573,3 +573,57 @@ Feature: OVN related networking scenarios
...
     And the output should contain "0% packet loss"
+  
+  # @author anusaxen@redhat.com
+  # @case_id OCP-46285
+  @admin
+  @destructive
+  @4.10 @4.9 @4.8
+  @network-ovnkubernetes
+  @vsphere-ipi
+  @vsphere-upi
```
zhaozhanqi, line 584: just want to confirm, this can be tested on UPI, right

anuragthehatter: Nope as apiVIP wont be there on UPI. Fixed it. Thanks

features/networking/ovn.feature
anuragthehatter: Yep, fixed

features/networking/ovn.feature
```diff
@@ -573,3 +573,57 @@ Feature: OVN related networking scenarios
...
+  And evaluation of `@result[:response].scan(/\/* ([^\/]*) /)[1][0]` is stored in the :apiVIP_node_new clipboard
+  #Checking annotation exist for the node
+  When I run the :describe admin command with:
+    | resource | node                      |
+    | name     | <%= cb.apiVIP_node_new %> |
+  Then the step should succeed
+  And the output should contain:
+    | k8s.ovn.org/host-addresses |
+    | <%= cb.apiVIP_new %>       |
+  And the expression should be true> cb.apiVIP_node!=cb.apiVIP_node_new
```
huiran0826, line 629: Looks like this step only check the new node has the annotation, but cannot ensure this annotation is not existing in other nodes.

anuragthehatter: @huiran0826 Thats right. This step would ensure only the annotation presence on node which is reflected in NB db as well. We both saw during verification that duplicate entries do exist for sometime but not permanently as apiVIP takes sometime to switchover to new node. Thats not a bug though as we know and may be no need to check here as well. The aim is to check absense of stale entries and current in NB db and annotation to reflect current. Let me know if you think otherwise and i can definitely tweak script to add all nodes filter for that annotation.

huiran0826: Yeah, then for checking NB db, I guess it may need some timeout. Sometimes, there will be two entries during the apiVIP switching for a short time.   Without timeout, that may also cause the case temporary failing.

anuragthehatter: Sounds good Huiran. NB dbs entries are fastly written and deletes as compare to annotations update so should be okay IMO but certainly we can add 120 seconds loop at #618 to make sure loop exit when there is 1 entry only in NB. If not it will exit as non-zero.

anuragthehatter: Done and re-ran http://pastebin.test.redhat.com/1013865

features/networking/ovn.feature
```diff
@@ -573,3 +573,62 @@ Feature: OVN related networking scenarios
...
+  
+  # @author anusaxen@redhat.com
+  # @case_id OCP-46285
+  @admin
+  @destructive
+  @4.10 @4.9 @4.8
+  @network-ovnkubernetes
+  @vsphere-ipi
+  Scenario: Logical Router Policies and Annotations for a given node should be current
+  Given the env is using "OVNKubernetes" networkType
```
zhaozhanqi, line 585: @anuragthehatter seems the indentation for the whole steps for this case is incorrect.

anuragthehatter: Thats right. Seems like my auto indent is not working fine post fedora upgrade. Thanks. Done for now

pruan-rht (commented):
/lgtm
/approve

PR #2557, opened 2021-12-07
add wait to make sure the update works

zhouying7780:
The case also failed with: Error from server (Conflict): Operation cannot be fulfilled on deploymentconfigs.apps.openshift.io "hooks": the object has been modified; please apply your changes to the latest version and try again 
when it run fast . so I added the wait time. 

@kasturinarra  Please help review , thanks .

### DIFF
--- a/features/cli/deploy.feature
+++ b/features/cli/deploy.feature
@@ -207,12 +207,15 @@ Feature: deployment related features
       | "type": "ImageChange"     |
       | "replicas": 2             |
       | "value": "Plqe5Wevchange" |
+    And I wait for the steps to pass:
+    """
     When I run the :rollback client command with:
       | deployment_name         | hooks-1 |
     Then the output should contain:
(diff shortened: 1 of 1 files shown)


zhouying7780:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/245690/console

kasturinarra:
@zhouying7780 one question, when the update fails i.e a conflict occurs does it still keep retrying with out fail when enclosed in the above wait ?

zhouying7780:
@kasturinarra  I don't kind will have the  ' a conflict occurs' , in our case, should no conflict , only one user update the deploy.

kasturinarra:
@zhouying7780 should we wait for a specific time instead of saying "I wait for the steps to pass" ?

PR #2571, opened 2021-12-09
add nodeport case for ipv6

zhaozhanqi:

### DIFF
--- a/features/networking/ipv6_dualstack.feature
+++ b/features/networking/ipv6_dualstack.feature
@@ -3,10 +3,9 @@ Feature: ipv6 dual stack cluster test scenarios
   # @author zzhao@redhat.com
   # @case_id OCP-40581
   @admin
-  @4.8 @4.7 @4.10 @4.9
-  @vsphere-ipi @openstack-ipi @gcp-ipi @baremetal-ipi @azure-ipi @aws-ipi
-  @vsphere-upi @openstack-upi @gcp-upi @azure-upi @aws-upi
-  @upgrade-sanity
+  @network-ovnkubernetes
+  @4.10 @4.9
+  @baremetal-ipi
   Scenario: Project should be in isolation when using multitenant policy for ipv6 dual stack
     # create project and pods
     Given I have a project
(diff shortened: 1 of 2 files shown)


zhaozhanqi:
logs: http://pastebin.test.redhat.com/1014284

features/networking/ipv6_dualstack.feature
```diff
@@ -80,3 +79,35 @@ Feature: ipv6 dual stack cluster test scenarios
...
+    Given I obtain test data file "networking/nodeport_test_pod.yaml"
+    When I run the :create client command with:
+      | f | nodeport_test_pod.yaml |
+    Then the step should succeed
+    When I obtain test data file "networking/nodeport_test_service.yaml"
+    When I run oc create over "nodeport_test_service.yaml" replacing paths:
+      | ["spec"]["ports"][0]["nodePort"] | <%= cb.port %> |
+      | ["spec"]["ipFamilyPolicy"] | "RequireDualStack" |
+    Then the step should succeed
+    Given a pod becomes ready with labels:
```
weliang1, line 102: Should move this check to line 96, if this pod check fail, then no need to run line 97-101

features/networking/ipv6_dualstack.feature
```diff
@@ -80,3 +79,35 @@ Feature: ipv6 dual stack cluster test scenarios
...
+    Given I have a project
+    And evaluation of `rand(30000..31000)` is stored in the :port clipboard
+    Given I obtain test data file "networking/nodeport_test_pod.yaml"
+    When I run the :create client command with:
+      | f | nodeport_test_pod.yaml |
+    Then the step should succeed
+    When I obtain test data file "networking/nodeport_test_service.yaml"
+    When I run oc create over "nodeport_test_service.yaml" replacing paths:
+      | ["spec"]["ports"][0]["nodePort"] | <%= cb.port %> |
+      | ["spec"]["ipFamilyPolicy"] | "RequireDualStack" |
```
weliang1, line 100: Existing networking/nodeport_test_service.yaml do not have ipFamilyPolicy option, should you include new updated networking/nodeport_test_service.yaml in this PR?

features/networking/ipv6_dualstack.feature
```diff
@@ -80,3 +79,35 @@ Feature: ipv6 dual stack cluster test scenarios
     Then the step should fail
 
 
+  # @author zzhao@redhat.com
+  # @case_id OCP-46816
+  @admin
+  @network-ovnkubernetes
+  @4.10 @4.9
+  @baremetal-ipi
+  Scenario: ipv6 for nodeport service
```
weliang1, line 88: Look like you only cover ipv6 testing here, for dual stack cluster, suggest to cover both ipv4 and ipv6 testing at same time in this script. Otherwise you need another script to cover dual stack testing

features/networking/ipv6_dualstack.feature
zhaozhanqi: yes, replace can add this option

features/networking/ipv6_dualstack.feature
zhaozhanqi: Yes, correct.  updated.

features/networking/ipv6_dualstack.feature
zhaozhanqi: hmm, it's also ok.  since there is another nodeport is for ipv4.  I just thought it can be covered that.  but sure, we can check both again in this case
updated.

PR #2574, opened 2021-12-10
fix OCP-45955

lihongan:
Usually the "hello-pod" will be removed after the node rebooting during upgrade, but it might be still running in abnormal scenario, so adding the step to ensure it is deleted in upgrade-prepare then it can be recreated in upgrade-check.

@quarterpin @melvinjoseph86 @ShudiLi please help take a look. thanks.


pruan-rht (commented):
/lgtm
/approve

PR #2575, opened 2021-12-10
Shorten logs generate by OCP-19470

liangxia:
The original command generate too many logs.
```
$ oc get clusterroles.rbac -o yaml | wc
  17610   31457  339285
```
/cc @wangke19 @xingxingxia @jhou1


xingxingxia:
Yes, an issue that should not haapen. Thanks!
/lgtm

PR #2584, opened 2021-12-17
add nodeport case from bug and update the existing cases

zhaozhanqi:
auto case OCP-47087 and Move nodePort case from cucushift repo to verification-tests https://github.com/openshift/cucushift/pull/8901

http://pastebin.test.redhat.com/1016453


zhaozhanqi:
I will approve this due to https://github.com/openshift/cucushift/pull/8901/files already be merged.

zhaozhanqi:
please let me know if you need any suggestion. I can update it later as well. thanks.

PR #2589, opened 2021-12-21
update external ES configurations

QiaolingTang:
Correcting external ES configurations.


QiaolingTang:
To avoid affecting our current test, I didn't update the `Automation Script` field in polarion, so I only executed these cases in my local machine:
```
$ bundle exec cucumber features/logging/forward_logs_to_external_elasticsearch.feature 
Using the default, devel and _devel profiles...
@clusterlogging
Feature: Cases to test forward logs to external elasticsearch
......
      [01:53:02] INFO> Exit Status: 0
      [01:53:03] INFO> === End After Scenario: Forward to external ES with username/password ===

5 scenarios (5 passed)
108 steps (108 passed)
23m3.688s
[01:53:03] INFO> === At Exit ===
```

features/step_definitions/logging.rb
```diff
@@ -993,16 +993,14 @@
   pod_label = "app=elasticsearch-server"
 
   # create secret/elasticsearch-server for ES if needed
-  if transport_ssl_enabled == "true" || scheme == "https"
+  if scheme == "https"
```
anpingli, line 996: I like transport_ssl_enabled more than scheme

features/step_definitions/logging.rb
```diff
@@ -959,15 +959,15 @@
 # version: 6.8 or 7.12
 # project_name: where the external ES deployed
 # scheme: http or https
-# transport_ssl_enabled: true or false
+# client_auth: true or false
```
anpingli, line 962: Could we explain here?  
client_auth: true or false. When true, client crendential(Eeither certs files or user/passwd or both) must be provided to access ES.

features/step_definitions/logging.rb
anpingli, line 996: I'd like use http_ssl_enabled instead of scheme

features/step_definitions/logging.rb
QiaolingTang: OK

features/step_definitions/logging.rb
QiaolingTang: Using `http_ssl_enabled` makes it hard to figure out the external ES url when creating the CLF/instance:
```
    When I process and create:
      | f | <file> |
      | p | URL=<scheme>://elasticsearch-server.<%= cb.es_proj.name %>.svc:9200 |
    Then the step should succeed
```

anpingli: I mean using ( transport_ssl_enabled == "true" ). I saw you had init transport_ssl_enabled  at L981. 
https://github.com/openshift/verification-tests/blob/edb0cc137ede922e4d0ff1e76e7cce55b9fa11bf/features/step_definitions/logging.rb#L981

anpingli: I am thinking the scheme can quit the scene once we init transport_ssl_enabled

QiaolingTang: Oh, I didn't notice that, will update the step, thanks.

PR #2609, opened 2022-01-05
fix some rules

yapei:


XiyunZhao (approved):

XiyunZhao (approved):

lib/openshift/quota_limits.rb
```diff
@@ -18,6 +19,10 @@ def storage_requests_raw
       @storage_requests
     end
 
+    def cpu_raw
```
pruan-rht, line 22: @yapei I think `cup` and `cpu_raw` can be consolidated into just `cpu`

yapei: got it, let me update

yapei: updated, could you help review again ? @pruan-rht

pruan-rht (commented):
/lgtm
/approve

PR #2614, opened 2022-01-06
Support for Jaya's cucushift #8988

xingxingxia:
This is framework support for https://github.com/openshift/cucushift/pull/8988's upcoming fix change (i.e. [this review suggestion](https://github.com/openshift/cucushift/pull/8988#discussion_r779550497)), which needs several times of curl watches in background and needs check them.

### DIFF
--- a/features/step_definitions/cli.rb
+++ b/features/step_definitions/cli.rb
@@ -198,9 +198,15 @@ Given /^I terminate last background process$/ do
   @result = @bg_processes.last.result
 end
 
-Given /^I check status of last background process$/ do
+Given /^I check status of( and pop)? last background process$/ do |pop|
   @bg_processes.last.wait(10)
   @result = @bg_processes.last.result
+  if pop
+    raise "0 background process left" unless @bg_processes.length > 0
+    print "Background processes number: ", @bg_processes.length, " Last background process pid: ", @bg_processes.last.pid, "\n"
+    @bg_processes.last.kill_tree
+    @bg_processes = @bg_processes[0...-1]
+  end
 end
 
 # This step needs flexibly specify table.


xingxingxia:
@jmekkatt @wangke19 @gangwgr @liangxia @pruan-rht help review / merge, thx!

jmekkatt:
I have tested the change w.r.t https://github.com/openshift/cucushift/pull/8988#issuecomment-1009779566 and it is successful.
/lgtm

features/step_definitions/cli.rb
```diff
@@ -198,9 +198,15 @@
   @result = @bg_processes.last.result
 end
 
-Given /^I check status of last background process$/ do
+Given /^I check status of( and pop)? last background process$/ do |pop|
```
liangxia, line 201: Perhaps `I check status( and pop)? of last background process` ?

features/step_definitions/cli.rb
```diff
@@ -198,9 +198,15 @@
...
   @bg_processes.last.wait(10)
   @result = @bg_processes.last.result
+  if pop
+    raise "0 background process left" unless @bg_processes.length > 0
```
wangke19, line 205: `@bg_processes.length` should >= 0, seems the condition always is unable to meet. Replace with `if  @bg_processes.length == 0`, the raise will be executed.

xingxingxia: @wangke19 `unless @bg_processes.length > 0` is equivalent to `if @bg_processes.length == 0`, so I prefer to not change it.

features/step_definitions/cli.rb
liangxia: @xingxingxia Forget it. Let's leave it as it is now.

pruan-rht: @xingxingxia the phrasing of this step is confusing.  Do you mean what does `pop` mean here?

xingxingxia: The `pop` is the `push` opposite, which means removing an entry from one stack-like object like the `@bg_processes` array.

PR #2624, opened 2022-01-14
[WIP]update acceptance case to use CLF

QiaolingTang:
Update the acceptance case, adding steps to test CLF.

/cc @anpingli


QiaolingTang:
Log: https://url.corp.redhat.com/bda8c16

QiaolingTang:
@anpingli PTAL

anpingli (commented):
/lgtm

features/step_definitions/logging.rb
```diff
@@ -391,11 +391,11 @@
...
       cb[cb_name] = version
     when '4.7'
-      cb[cb_name] = "5.0"
-    when '4.8'
       cb[cb_name] = "stable-5.1"
-    when '4.9'
+    when '4.8'
       cb[cb_name] = "stable-5.2"
+    when '4.9'
+      cb[cb_name] = "stable-5.3"
```
liangxia, line 398: Perhaps reverse the order from 4.10 to 4.1, as we test the latest versions more often.

QiaolingTang: For logging, `stable` channel always points to the latest version. Currently the latest version is `logging 5.4`, but it hasn't been released yet. 

For pre-release CI, we have post actions to add QE managed catsrc, then the `stable` channel points to `logging 5.4`. For Prow CI, since we can't do the post actions, the `stable` channel points to `logging 5.3` which is the latest released version.  So here I think these configuration matches `test the latest versions more often`

liangxia: I should have make it more clear.

I mean to use
```
when '4.10'
..
when '4.9'
...
when '4.8'
...
```
As we test 4.10 ( the latest ) more often. So it get match for the first one, rather than iterate over 4.1, 4.2, ... 4.10

QiaolingTang: Oh, I see, I'll update it

QiaolingTang: Done

PR #2627, opened 2022-01-14
Update app name creation

xiuwang:
```    
error: ImageStream.image.openshift.io "cjx6--vh" is invalid: metadata.name: Invalid value: "cjx6--vh": must match "[a-z0-9]+(?:[._-][a-z0-9]+)*"
```
When use dns952 type to generate string, non-hyphen may be joined together, it will cause application creation failed.

Passed on my local host. @wewang58 Please help review, thanks

### DIFF
--- a/features/cli/create.feature
+++ b/features/cli/create.feature
@@ -192,7 +192,7 @@ Feature: creating 'apps' with CLI
     Given I have a project
     When I git clone the repo "https://github.com/openshift/ruby-hello-world"
     Then the step should succeed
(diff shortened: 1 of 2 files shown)


pruan-rht:
@xiuwang there's newly introduced conflicts, please resolve them.  thx!

liangxia:
If you check the files, https://github.com/openshift/verification-tests/pull/2627/files, there are unexpected files introduced.

xiuwang:
@liangxia Removed the unexpected files, thanks

PR #2628, opened 2022-01-14
Fix git url

pruan-rht:


features/cli/create.feature
```diff
@@ -90,10 +90,10 @@ Feature: creating 'apps' with CLI
     #Check git github url
     Given an 8 character random string of type :dns952 is stored into the :appname3 clipboard
     When I run the :new_app client command with:
-      | code         | git://github.com/openshift/ruby-hello-world |
-      | image_stream | openshift/ruby                              |
-      | name         | <%= cb.appname3 %>                          |
-      | env          | MYSQL_ROOT_PASSWORD=test                    |
+      | code         | git@github.com:openshift/ruby-hello-world |
```
liangxia, line 93: Format `git@xxx` requires SSH key. Perhaps we can use https for them ?

pruan-rht: @liangxia the the reason I did that was the comment `#Check git github url`.

pruan-rht: @liangxia I've updated it now to use `https` instead. PTAL.

features/cli/create.feature
```diff
@@ -90,10 +90,10 @@ Feature: creating 'apps' with CLI
     #Check git github url
     Given an 8 character random string of type :dns952 is stored into the :appname3 clipboard
     When I run the :new_app client command with:
-      | code         | git://github.com/openshift/ruby-hello-world |
-      | image_stream | openshift/ruby                              |
-      | name         | <%= cb.appname3 %>                          |
-      | env          | MYSQL_ROOT_PASSWORD=test                    |
+      | code         | https://github.com:openshift/ruby-hello-world |
```
liangxia, line 93: Looks like you forgot to update the `:` in the URL.

features/cli/create.feature
```diff
@@ -247,10 +247,10 @@ Feature: creating 'apps' with CLI
     #Check git github url
     Given an 8 character random string of type :dns952 is stored into the :appname3 clipboard
     When I run the :new_app client command with:
-      | code         | git://github.com/openshift/ruby-hello-world |
-      | image_stream | openshift/ruby                              |
-      | name         | <%= cb.appname3 %>                          |
-      | env          | MYSQL_ROOT_PASSWORD=test                    |
+      | code         | git@github.com:openshift/ruby-hello-world.git |
```
liangxia, line 250: The line still use git format, which will requires SSH key.

pruan-rht: @liangxia I've taken that check out since it's doing `git github url` which if we use the git format, then it runs into the issue of not being allowed.  I think we can just remove this section

PR #2642, opened 2022-01-21
OCP-11286: remove docker0 rule and log failures for debug

rbbratta:
The old `docker0` rule was removed
`OPENSHIFT-FIREWALL-ALLOW -i docker0 -m comment --comment "from docker to localhost" -j ACCEPT`

https://github.com/openshift/sdn/pull/345

Also for step failures, `@result[:message]` is never displayed, use `logger.info`.
Use `info` level because we expect this to fail at one step in the test.


rbbratta:
http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2022-01-21-10-23-23.log

we now print the missing rules
```
      [15:24:56] INFO> Exit Status: 0
      [15:24:56] INFO> The filter table verification failed, missing [OPENSHIFT-FIREWALL-ALLOW -p udp -m udp --dport 4789 -m comment --comment "VXLAN incoming" -j ACCEPT]
    # Removing individual rules will not trigger automatic repair on < 4.3
    # the check should fail
    Then the step failed                                                         # features/step_definitions/common.rb:10
    # >= 4.3 we have to flush all the rules and tables to trigger a repair
```

@openshift/team-sdn-qe PTAL

zhaozhanqi:
Thanks @rbbratta 
/lgtm
/approve

PR #2644, opened 2022-01-21
Add sgw-lgw-migration-externalIP case

weliang1:
The the deny externalIP policy in this case will break some features tested in sgw-lgw-migration.feature, so create this separate
sgw-lgw-migration-externalIP.feature

Test logs:
SGW <-> LGW in vsphere cluster:
http://file.rdu.redhat.com/~weliang/SGW-LGW-Migration-vsphere
SGW <-> LGW in baremetal cluster:
http://file.rdu.redhat.com/~weliang/SGW-LGW-Migration-baremetal

@openshift/team-sdn-qe PTAL

### DIFF
--- a/features/networking/sgw-lgw-migration-externalIP.feature
+++ b/features/networking/sgw-lgw-migration-externalIP.feature
@@ -1,0 +1,84 @@
+Feature: SGW<->LGW migration related scenarios
+
+  # @author weliang@redhat.com
+  # @case_id OCP-48066
+  @4.10
+  @admin
+  @destructive
+  @network-ovnkubernetes
+  @vsphere-ipi
+  @baremetal-upi
+  Scenario: [SDN-2290] SGW <-> LGW migration scenarios for externalIP	
+    Given the env is using "OVNKubernetes" networkType
+
+    ######## Prepare Data Pre Migration for multiple use cases############
+    #OCP-24669 - externalIP defined in service with set ExternalIP in allowedCIDRs
+    Given I have a project
+    And SCC "privileged" is added to the "system:serviceaccounts:<%= project.name %>" group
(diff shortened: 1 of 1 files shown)


features/networking/sgw-lgw-migration-externalIP.feature
```diff
@@ -0,0 +1,84 @@
+Feature: SGW<->LGW migration related scenarios
```
anuragthehatter, line 1: Not sure why we are creating this as a different feature file. May be Just add this as a third scenario in same ```sgw-lgw-migration.feature```? This way it will always run as a separate use case

features/networking/sgw-lgw-migration-externalIP.feature
```diff
@@ -0,0 +1,84 @@
...
+
+  # @author weliang@redhat.com
+  # @case_id OCP-48066
+  @4.10
+  @admin
+  @destructive
+  @network-ovnkubernetes
+  @vsphere-ipi
+  @baremetal-upi
```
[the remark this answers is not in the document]
weliang1: > @weliang1 Thanks for including this scenario in migration. We also need to ass baremetal in your Polarion test case to align it?

@anuragthehatter Just add baremetal in Polarion for this case

features/networking/sgw-lgw-migration-externalIP.feature
weliang1: > Not sure why we are creating this as a different feature file. May be Just add this as a third scenario in same `sgw-lgw-migration.feature`? This way it will always run as a separate use case

@anuragthehatter That's sounds a good workaround,I will give it a try, thanks!

weliang1:
Close this PR and open a new https://github.com/openshift/verification-tests/pull/2655

PR #2647, opened 2022-01-24
OVN Raft: increase wait due to increase in election timer

rbbratta:
election timer was 5 seconds, but now is 16 seconds and convergence also depends
on the number of masters. 40 seconds was not enough to converge with a 5 master
cluster, increase wait.

ref:
BZ 1883662 - [sbdb][raft] Tune out of the box timer to be 16sec

### DIFF
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -444,7 +444,10 @@ Feature: OVN related networking scenarios
     Then the step should succeed
 
     # election timer is 1 second by default but the RAFT JSON-RPC probe might take 5 seconds to notice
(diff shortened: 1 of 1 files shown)


rbbratta:
Failure example: http://file.rdu.redhat.com/~rbrattai/logs/OCP-26138-ovn-raft-timeout-5-masters

With 120 second wait: http://file.rdu.redhat.com/~rbrattai/logs/OCP-26138-with-120-second-wait

@openshift/team-sdn-qe PTAL

asood-rh (approved):
/lgtm

pruan-rht (commented):
/lgtm
/approve

PR #2652, opened 2022-01-28
setting routing cases as inactive temporarily for debugging purpose

lihongan:
will revert this after the debugging.

@quarterpin @melvinjoseph86 @ShudiLi  PTAL, thanks


ShudiLi:
/lgtm
Thanks

PR #2656, opened 2022-01-28
fix slack notification error.

pruan-rht:
Outputing summary to team-qe channel instead to lessen the noise in forum-qe

@mffiedler can you pls take a look when you have a chance.  I've tested it already and it works.  I want to merge it today if possible to cut down the noise in `#forum-qe` in slack.


pruan-rht:
/lgtm
/approve
merging myself for it to take effect...nothing OCP testing related so it should be fine.

PR #2659, opened 2022-01-31
default values to create machineset for clusters 

miyadav:
@pruan-rht , have a look if time permits ..

The clusters which are installed behind proxy , have a slightly different yaml for machineset 

where subnet id can be got using something like 
`x.spec.template.spec.providerSpec.value.subnet.id` whereas for all other clusters configs it is like `x.spec.template.spec.providerSpec.value.subnet.filters` and also the testdata file that we use have list for subnet , whereas it is just key value pair for proxy clusters.

the changes done using this PR will make it work for all profiles where proxy is not used .(It was earlier failing for non-proxy as well) 
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/296030/console 

Please provide if any suggestion I could try , basically need condition to modify subnet values by identifying the profile ( proxy or non proxy) to be able to use a different test data file or may be a seperate step definition for the step - 
`  | ["spec"]["template"]["spec"]["providerSpec"]["value"]["subnet"]["filters"][0]["values"]   | <%= cb.default_subnet[0].values[1] %>       |` 
or 
`  | ["spec"]["template"]["spec"]["providerSpec"]["value"]["subnet"]["id"]   | <%= cb.default_subnet %>       |`

### DIFF
--- a/features/machine/machine.feature
+++ b/features/machine/machine.feature
@@ -332,12 +332,12 @@ Feature: Machine features testing
     And I switch to cluster admin pseudo user
     And I use the "openshift-machine-api" project
     Then admin ensures machine number is restored after scenario
-
-    Given I store the last provisioned machine in the :machine clipboard
-    When evaluation of `machine(cb.machine).aws_ami_id` is stored in the :default_ami_id clipboard
-    And evaluation of `machine(cb.machine).aws_availability_zone` is stored in the :default_availability_zone clipboard
-    And evaluation of `machine(cb.machine).aws_subnet` is stored in the :default_subnet clipboard
-    And evaluation of `machine(cb.machine).aws_iamInstanceProfile` is stored in the :default_iamInstanceProfile clipboard
+    
+    Given I pick a earliest created machineset and store in :machineset clipboard
+    When evaluation of `machine_set(cb.machineset).aws_machineset_ami_id` is stored in the :default_ami_id clipboard
+    And evaluation of `machine_set(cb.machineset).aws_machineset_availability_zone` is stored in the :default_availability_zone clipboard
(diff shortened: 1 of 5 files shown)


miyadav:
/unhold 

@jhou1 @huali9 @sunzhaohua2  PTAL once you guys are back , just made refractoring to keep simple and cover all configurations , looks good with these changes to work on all configs that we run for CI ( proxy , disconnected , windows etc.)

https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/296699/console

features/machine/machine.feature
```diff
@@ -332,12 +332,12 @@ Feature: Machine features testing
...
     And I use the "openshift-machine-api" project
     Then admin ensures machine number is restored after scenario
-
-    Given I store the last provisioned machine in the :machine clipboard
-    When evaluation of `machine(cb.machine).aws_ami_id` is stored in the :default_ami_id clipboard
-    And evaluation of `machine(cb.machine).aws_availability_zone` is stored in the :default_availability_zone clipboard
-    And evaluation of `machine(cb.machine).aws_subnet` is stored in the :default_subnet clipboard
-    And evaluation of `machine(cb.machine).aws_iamInstanceProfile` is stored in the :default_iamInstanceProfile clipboard
+    
+    Given I pick a earliest created machineset and store in :machineset clipboard
```
sunzhaohua2, line 336: @miyadav can you tell me why we need this step? Is there any difference between getting value from machineset and machine?

miyadav: If we have a cluster with windows machines we need to make use of this, in that case otherwise we have to capture the value from master machine, which I thought of avoiding as mapi doesnt have anything to do with master machines.

PR #2660, opened 2022-02-01
Randomize name of the service project/namespace created by test framework.

jmekkatt:
@xingxingxia @wangke19 @gangwgr In recent test case failure analysis (Refer https://issues.redhat.com/browse/OCPQE-8433 - Part B) exposed that , the service project or namespace created by test framework for hosting a debug pod for running node commands were always same for all the tests . This may leads to potential issues like failing the test case execution one after other due to existence of namespace(as described in ticket) especially when the service namespace fails or slow in cleanup, deletion etc. So this PR would help us to pick unique name for service project or namespace.
Jenkins job against the PR -  Runner-v3-smoke/3907/console
Kindly review the same and let me know your feedback. Thanks.

### DIFF
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -450,7 +450,9 @@ module BushSlicer
       unless @service_project
         # if the cluster set the default scheduler, set the project running debug pod node-selector=''
         # to overwrite the default scheduler, or the pod can not be run successfully
-        project_name = "prj-" + EXECUTOR_NAME.downcase
+        # project_suffix creates non identical names for namespace or project running debug pod in test cases.
+        project_suffix = [*"a".."z", *"1".."9"]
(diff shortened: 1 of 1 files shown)


lib/environment.rb
```diff
@@ -450,7 +450,8 @@ def service_project
       unless @service_project
         # if the cluster set the default scheduler, set the project running debug pod node-selector=''
         # to overwrite the default scheduler, or the pod can not be run successfully
-        project_name = "prj-" + EXECUTOR_NAME.downcase
+        project_suffix = [*"a".."z", *"1".."9"]
```
gangwgr, line 453: @jmekkatt Please add the comment here regarding why you are adding this step? so that other people can understand.

gangwgr:
@jmekkatt Run multiple cases of different components. So that it does not impact others case.

gangwgr:
@pruan-rht  Please review this PR and help us to merge.

PR #2661, opened 2022-02-02
Update ocm-cli default version to 0.1.61

apodhrad:
There is an issue with 0.1.59 which blocks users to log in.

### DIFF
--- a/lib/launchers/o_c_m_cluster.rb
+++ b/lib/launchers/o_c_m_cluster.rb
@@ -161,7 +161,7 @@ module BushSlicer
     def download_ocm_cli
       url = ENV['OCM_CLI_URL']
       unless url
-        url_prefix = ENV['OCM_CLI_URL_PREFIX'] || 'https://github.com/openshift-online/ocm-cli/releases/download/v0.1.59'
+        url_prefix = ENV['OCM_CLI_URL_PREFIX'] || 'https://github.com/openshift-online/ocm-cli/releases/download/v0.1.61'
         if OS.mac?
           url = "#{url_prefix}/ocm-darwin-amd64"
         elsif OS.linux?


apodhrad:
Hi @akostadinov do you have any idea why the prow job is still failing? It seems like a problem during building a docker image

apodhrad:
@akostadinov similar issue was hit by the RHOAM team. They fixed it by
https://github.com/integr8ly/integreatly-operator/pull/2442/files
maybe it could be the same root cause.

akostadinov:
@apodhrad , I'm in another team now. I checked the log and I see that it is using centOS appstream now. When I last touched imaage it has been still based on RHEL if IIRC. So I have zero idea why it fails. Something with repos it seems. Perhaps check `git blame` to see who last edited file.

apodhrad:
Hi @akostadinov it would be nice if the will not suggest you as a potential assignee ;-)

```
To complete the pull request process, please assign akostadinov after the PR has been reviewed.
```

apodhrad:
@dis016 @JianLi-RH do you know what is the root cause of the failing docker build?

dis016:
@apodhrad looks same issue as https://issues.redhat.com/browse/OCPQE-8472. PR  https://github.com/openshift/verification-tests/pull/2662 is filed for the fix. /cc @pruan-rht

pruan-rht (commented):
/lgtm
/approve

pruan-rht:
@akostadinov let me know if you really want to be removed from the OWNERS file...I'm against it TBH, but it's up to you.

akostadinov:
As I said, I feel sadness about it but this seems to be the only way to avoid confusion. Otherwise I will pop-up to people as suggested approver.

pruan-rht:
@akostadinov ok, sounds good... https://github.com/openshift/verification-tests/pull/2666/files

PR #2663, opened 2022-02-02
Adding flaky tag so that test wont be picked up during prepare

kasturinarra:
One of the upgrade-prepare test case always gets picked up even after marking it 'Not Automated' and removing content from 'AutomationScript' field. As suggested by vadim adding @flaky tag so that test wont be picked up. Having this test case run is causing kube-scheduler upgrades to fail.


kasturinarra:
@pruan-rht could you please help review & merge ? Thanks !!

PR #2664, opened 2022-02-02
Automate OCP-46116

weliang1:
Automate OCP-46116: [BZ 1897431] CIDR support for additional network attachment with the bridge CNI plug-in

Test log:
http://file.rdu.redhat.com/~weliang/OCP-46116.txt

@openshift/team-sdn-qe PTAL


pruan-rht (commented):
/lgtm
/approve

PR #2665, opened 2022-02-02
Fixes OCP-15731 and OCP-15744, reactivated by #2372

aleskandro:
@pruan-rht as discussed, this removes the automation tags for the inactive cases OCP-15731 and OCP-15744 in #2372

### DIFF
--- a/features/networking/build.feature
+++ b/features/networking/build.feature
@@ -114,9 +114,6 @@ Feature: Testing the isolation during build scenarios
     Then the output should not contain "Hello OpenShift"
     Then the output should contain "Connection timed out after"
     """
-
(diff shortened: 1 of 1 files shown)


pruan-rht:
/lgtm
/approve
@liangxia perhaps your tag optimized script needs to take into account cases with the `@inactive` tag

liangxia:
> /lgtm /approve @liangxia perhaps your tag optimized script needs to take into account cases with the `@inactive` tag

@pruan-rht   The inactive test cases are exclude in the query. E.g, https://github.com/openshift/verification-tests/blob/master/doc/prow/4.10/query_aws_ipi_tests.json
I guess those tags are added before we finalize the process.

PR #2666, opened 2022-02-02
Remove alex from owner

pruan-rht:


akostadinov:
symbolic PR number... :smiling_imp:

akostadinov:
Well that's exactly what makes the situation really creepy..

PR #2668, opened 2022-02-03
OCPQE-8485: Fix OCP-10743

aleskandro:


aleskandro:
This also refers ARMOCP-173

PR #2669, opened 2022-02-03
fix pricing issue and packet price lookup logic

pruan-rht:
we should really do a DB lookup instead of hard-coding.


pruan-rht:
@mffiedler can you please give a quick LGTM.  I've tested locally and it works.

pruan-rht:
@dis016 PTAL so I can merge it before the next run

dis016 (commented):
/lgtm

PR #2675, opened 2022-02-09
Fix urls for cucushift in logs

liangxia:
Per discussion in https://coreos.slack.com/archives/CPJQCSN49/p1644325801134459

/cc @jmekkatt @pruan-rht @dis016 @JianLi-RH @jhou1


pruan-rht (commented):
/lgtm

PR #2540, opened 2021-12-02
Reimplementation of ovirt modules and dependencies

santos1709:
commit https://github.com/openshift/verification-tests/pull/2200 was merged and reverted, this is another try of re-implementing the ovirt  modules.

PR #2677, opened 2022-02-10
Add packages to support OVirt

liangxia:
/cc @JianLi-RH @dis016 @pruan-rht @jhou1 

This is needed by PR https://github.com/openshift/verification-tests/pull/2540

I want to bring in oc411 as soon as possible and also want to avoid some duplicate effort on image building, so using this PR to moving ahead.


dis016 (commented):
/lgtm

PR #2683, opened 2022-02-14
fix upgrade failure

lihongyan1:
http://virt-openshift-05.lab.eng.nay.redhat.com/buildcorp/ocp_upgrade/5012.html
http://virt-openshift-05.lab.eng.nay.redhat.com/buildcorp/ocp_upgrade/5056.html
http://virt-openshift-05.lab.eng.nay.redhat.com/buildcorp/ocp_upgrade/5075.html
http://virt-openshift-05.lab.eng.nay.redhat.com/buildcorp/ocp_upgrade/5137.html

Fixed failures in above jobs: node-exporter daemonset is highly depend on status of node, if machineset config, the daemon set will be created or removed when a node is added or removed, which will lead to an old daemonset will never be found. I will update automation test case to only check existing node-exporter pods.

runner job:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/313071/

### DIFF
--- a/features/upgrade/monitoring/monitoring-upgrade.feature
+++ b/features/upgrade/monitoring/monitoring-upgrade.feature
@@ -23,7 +23,10 @@ Feature: cluster monitoring related upgrade check
   Scenario: upgrade cluster monitoring along with OCP
     Given I switch to cluster admin pseudo user
     And I use the "openshift-monitoring" project
+    And I wait up to 120 seconds for the steps to pass:
+    """
     Given all pods in the project are ready
+    """
     And the expression should be true> cluster_operator('monitoring').condition(type: 'Degraded')['status'] == "False"
 
     #check retention time


lihongyan1:
@juzhao please help review

lihongyan1:
@liangxia please approve

features/upgrade/monitoring/monitoring-upgrade.feature
```diff
@@ -23,7 +23,10 @@ Feature: cluster monitoring related upgrade check
   Scenario: upgrade cluster monitoring along with OCP
     Given I switch to cluster admin pseudo user
     And I use the "openshift-monitoring" project
+    And I wait up to 120 seconds for the steps to pass:
+    """
     Given all pods in the project are ready
+    """
```
liangxia, line 29:
```
# use this with care and it just all of the pods in the project, better to filter out the pods with labels if possible
Given /^all pods in the project are ready$/ do
  pods = project.pods(by:user)
  ready_timeout = 15 * 60
  logger.info("Number of pods: #{pods.count}")
  pods.each do |pod|
    cache_pods(pod)
    @result = pod.wait_till_ready(user, ready_timeout)
    unless @result[:success]
      raise "pod #{pod.name} did not become ready within allowed time"
    end
  end
end
```
The step itself contains wait logic (for 15*60 seconds), I'm afraid adding another wrap for 120 seconds won't help there.

lihongyan1: It does help here, I need get pods again from the project for exporter pod is just deleted and I should get the deleted information, that is, wait for until the exporter pod is gone.

liangxia: Where is the deletion step of export pod ?

lihongyan1: the delete and adding of exporter pod is automatically triggered by node number change, when there is mcp config change, node change will happen. One node had one exporter pod.
mcp config change belong to other team's test case

liangxia: Thanks for the explanation.

PR #2689, opened 2022-02-16
update some unstable cases

zhaozhanqi:
small update:
1. expand time since the worker will be reboot when creating vfio-pci type/mlx VF
2. remove rootdevice since the pciaddress maybe change to another when worker reboot
3. update logs with the correct show


zhaozhanqi:
@yingwang-0320 ^^  PTAL

PR #2690, opened 2022-02-16
Removing version tags so that test does not get executed

kasturinarra:
Removing version tags from upgrade-prepare & upgrade-check so that tests does not get executed. When these tests are being executed in an upgrade path which contains 4.10 kube-scheduler always goes to crashLoopBackOff state due to a recent change introduced in 4.10 (Removal of scheduler-policy)


kasturinarra:
@zhouying7780 @liangxia could you please help review & merge ? thanks !!

liangxia:
/approve

It also lgtm.  Not adding `/lgtm` here together with `/approve` so that sub-team have time to review/aware the changes, or else add those two tags together will trigger the bot to merge the PR immediately.

PR #2693, opened 2022-02-17
[WIP]fixing step definition for SNO

gangwgr:
Raise this PR because step not accepting variable integer value which cause below issue. @liangxia Please review this PR
```
Given <%= cb.wait_time1 %> seconds have passed                                  # features/tierN/apiserver/kube-apiserver.feature:267
   Undefined step: "<%= cb.wait_time1 %> seconds have passed" (Cucumber::Core::Test::Result::Undefined)
   features/tierN/apiserver/kube-apiserver.feature:267:in `<%= cb.wait_time1 %> seconds have passed'
```
Test run - http://pastebin.test.redhat.com/1030580

### DIFF
--- a/lib/openshift/resource.rb
+++ b/lib/openshift/resource.rb
@@ -49,6 +49,11 @@ module BushSlicer
       elsif result[:response] =~ /Unable to connect to the server/
         logger.info(result[:response])
         return false
+      # Adding below because when in Single node cluster node restarted then SNO not have other nodes to communicate which cause connection timeout issue.
+      # And to prevent premature exit as describe in OCPQE-8043.
+      elsif result[:response] =~ /connection to the server \S+ was refused - did you specify the right host or port/ && env.nodes.length == 1
+        logger.info(result[:response])
+        return false
       else
         # e.g. when called by user without rights to list Resource
         raise "error getting #{self.class.name} '#{name}' existence: #{result[:response]}"
--- a/features/step_definitions/common.rb
+++ b/features/step_definitions/common.rb
@@ -121,7 +121,7 @@ Then /^the output should equal #{QUOTED}$/ do |value|
   end
 end
 
-Given /^([0-9]+?) seconds have passed$/ do |num|
+Given /#{NUMBER} seconds have passed$/ do |num|
   sleep(num.to_i)
 end
 


jhou1:
What about other steps that are invoking this step using in String?

gangwgr:
@jhou1 As i checked other case invoke like  below for steps. I run for one case 
Tes tun - http://pastebin.test.redhat.com/1030594
```
 grep -inr "seconds have passed" .                      
./features/test/cli.feature:8:    And 3 seconds have passed
./features/test/cli.feature:33:    And 5 seconds have passed
./features/test/cli.feature:111:    Given 3 seconds have passed
./features/networking/operator.feature:54:    And 20 seconds have passed
./features/networking/operator.feature:59:    Given 20 seconds have passed
```

gangwgr:
@jhou1 I'm trying other thing also, so that we don't need this. Till then we can wait.

lib/openshift/resource.rb
```diff
@@ -49,6 +49,14 @@ def exists?(user: nil, result: {}, quiet: false)
       elsif result[:response] =~ /Unable to connect to the server/
         logger.info(result[:response])
         return false
+      # Adding below because when in Single node cluster node restarted then SNO not have other nodes to communicate which cause connection timeout issue.
+      # And to prevent premature exit as describe in OCPQE-8043.
+      no_of_master = env.nodes.length
```
jhou1, line 54: Rename the variable `no_of_master` to `num_of_master`

jhou1: The variable is assigned length of nodes, so maybe `num_of_nodes` is more appropriate var name here.

lib/openshift/resource.rb
```diff
@@ -49,6 +49,14 @@ def exists?(user: nil, result: {}, quiet: false)
...
+      cluster_host = env.api_hostname
+      elsif result[:response] =~ /connection to the server \S+.qe.devcluster.openshift.com:6443 was refused - did you specify the right host or port/ && no_of_master == 1
```
jhou1, line 56: Are we certain all clusters are using a base domain as `qe.devcluster.openshift.com` and port `6443`?. This may not be the case when the cluster is created on Baremetal using a different configuration.

lib/openshift/resource.rb
```diff
@@ -49,6 +49,14 @@ def exists?(user: nil, result: {}, quiet: false)
...
+        logger.info(result[:response])
+        logger.info("#{cluster_host}:6443")
```
jhou1, line 58: Can this line be removed? `env.api_hostname` should be in the log.

xingxingxia:
Which one will be in favor of? This or 2698? And is 'seconds have passed' change still needed?

gangwgr:
@jhou1 @xingxingxia I will close this PR as @liangxia also working on same in 2698 and "seconds have passed' not needed now.

PR #2218, opened 2021-08-26
Add common step for authentication rolls out after config changes

yaoli-redhat:
@xingxingxia this PR is to add common step 'authentication successfully rolls out after config changes' based on https://github.com/openshift/cucushift/pull/8676 feedback and should be merged before that. pass log: Runner-v3-smoke/3427 for OCP-23517/OCP-23514/OCP-23515 and Runner-v3-smoke/3428 for OCP-29916. Please help to review when free. thanks!

PR #2695, opened 2022-02-21
Make co/authentication rotation robust in SNO env test

xingxingxia:
Fixing [OCPQE-8730](https://issues.redhat.com/browse/OCPQE-8730).

The failure has 2 parts to fix:
One is:
```
01:15:12 INFO> #### Operator authentication Expected conditions: {"Progressing"=>"True"}
#<RuntimeError: The authentication operator still didn't become {"Progressing"=>"True"} after 180 seconds>
```
It actually had become "Progressing"=>"True". But the poll interval is too big and the transition to False is quick in SNO env. We need use small interval for SNO env.

The other issue is:
```
01:15:12 INFO> Shell Commands: oc replace -f - --kubeconfig=/home/jenkins/ws/workspace/ocp-common/Runner/workdir/ocp4_admin.kubeconfig
oauth.config.openshift.io/cluster replaced
error: You must be logged in to the server (Unauthorized)
01:16:44 ERROR> #<RuntimeError: error getting projects by user: 'testuser-40@ocp4'>
```
When oauth.config.openshift.io/cluster is updated, auth pods are rotating and unavailable temporarily, token validation will be broken, i.e. here for request of `getting projects by user`. We need ensure auth pods rotation is complete first.
Pass logs for fix in SNO env: ocp-c1/job/Runner/322630/console
@y4sht @liangxia help review / merge, thanks!

### DIFF
--- a/features/step_definitions/operators.rb
+++ b/features/step_definitions/operators.rb
@@ -168,6 +168,7 @@ Given /^operator #{QUOTED} becomes #{NO_SPACE_STR}(?: within #{NUMBER} seconds)?
   interval_time = 5
   timeout = Integer(timeout) rescue 60
   interval_time = 20 if timeout > 100
+  interval_time = 5 if env.nodes.length == 1 # interval needs to be short because rotation is quick in SNO env
   actual_results = {}
   stats = {}
 
--- a/features/step_definitions/oauth.rb
+++ b/features/step_definitions/oauth.rb
@@ -17,11 +17,20 @@ Given /^the #{QUOTED} oauth CR is restored after scenario$/ do |name|
   org_oauth_json = org_oauth.to_json
   _admin = admin
   teardown_add {
-    puts "Original CR to restore:", org_oauth_json
+    @result = admin.cli_exec(:get, resource: 'oauth', resource_name: name, o: 'yaml')
+    if @result[:success]
+      org_oauth_now = @result[:parsed]
+    else
+      raise "Could not get OAuth: #{name}"
+    end
+    print "Original CR to restore:\n", org_oauth_json
     # Use replace instead of patch, otherwise the patch with org_oauth_json can only add or modify fields, but will not remove fields that were added
     @result = _admin.cli_exec(:replace, f: "-", _stdin: org_oauth_json)
     raise "Cannot restore OAuth: #{name}" unless @result[:success]
-    sleep 60
+    # Check 'spec' first because pods only roll out when 'spec' had been changed
+    if org_oauth_now['spec'] != org_oauth['spec']
+      step %Q/authentication successfully rolls out after config changes/
+    end
   }
 end
 


features/step_definitions/operators.rb
```diff
@@ -168,6 +168,7 @@
   interval_time = 5
   timeout = Integer(timeout) rescue 60
   interval_time = 20 if timeout > 100
+  interval_time = 5 if env.nodes.length == 1 # interval needs to be short because rotation is quick in SNO env
```
liangxia, line 171: Check line 168, `interval_time` defaults to 5

xingxingxia: @liangxia line 170 overrides it, the case has steps of timeout > 100 (180 seconds), this caused the failure, see the Description part:
```
#<RuntimeError: The authentication operator still didn't become {"Progressing"=>"True"} after 180 seconds>
```

liangxia: Perhaps that means you need a shorter timeout rather than overwrite it again

xingxingxia: We need it due to https://github.com/openshift/verification-tests/pull/2218/files#diff-8efd8ed73179d3b17bdfae8adad9cf49e0a0ad91eaf68e3b2a36766289d56616R41 .

xingxingxia: Could you help check again per previous comment?

xingxingxia: @liangxia @pruan-rht , could you help merge if no more objection given above reasons explained? Thanks!

liangxia: That's for the rollout, but here we are talking about operator in progressing.

It should work if you change 
`operator "authentication" becomes progressing within 180 seconds` to 
`operator "authentication" becomes progressing within 100 seconds`

xingxingxia: `operator "authentication" becomes progressing within 100 seconds` met failures due to https://github.com/openshift/verification-tests/pull/2218/files#diff-8efd8ed73179d3b17bdfae8adad9cf49e0a0ad91eaf68e3b2a36766289d56616R41 .

liangxia: So the workaround for the bug spreads here and there.  Any plan to cleanup those workarounds after the bug is fixed ?

xingxingxia: > here and there

Understood your concern, and frankly speaking, I hate workarounds too, just had to do.
About "here and there", I have to explain "here and there" are both needed. For "there" (i.e. let the timeout big enough), it is due to the bug [1958198](https://bugzilla.redhat.com/show_bug.cgi?id=1958198) there. For "here" (i.e. let `interval_time = 5 if env.nodes.length == 1`), it is due to SNO feature that transitions statuses quicker. Please don't worry about cleanup, when ON_QA we of course we'll remember to clean up. Thanks!

PR #2698, opened 2022-02-22
Retry when server refuse connection during restart on SNO

liangxia:
/cc @gangwgr @xingxingxia @jhou1 @dis016 @pruan-rht 

Tested on the same cluster,

Without the changes in this PR,
```
    Given operator "kube-apiserver" becomes non-progressing within <%= cb.wait_time %> seconds                        # features/step_definitions/operators.rb:164
      [05:10:30] INFO> ......
      [05:10:32] INFO> Exit Status: 0
      [05:10:52] INFO> Shell Commands: oc get clusteroperators kube-apiserver --output=yaml --kubeconfig=/home/jenkins/workdir/b86a29ab7df1-rootx/ocp4_admin.kubeconfig
      
      STDERR:
      The connection to the server api.rgangwar-21de3.qe.devcluster.openshift.com:6443 was refused - did you specify the right host or port?
      [05:10:53] INFO> Exit Status: 1
      error getting BushSlicer::ClusterOperator 'kube-apiserver' existence: 
      STDERR:
      The connection to the server api.rgangwar-21de3.qe.devcluster.openshift.com:6443 was refused - did you specify the right host or port?
       (RuntimeError)
      /verification-tests/lib/openshift/resource.rb:54:in `exists?'
      /verification-tests/lib/openshift/resource.rb:60:in `get_checked'
      /verification-tests/lib/openshift/resource.rb:127:in `get_cached_prop'
      /verification-tests/lib/openshift/resource.rb:134:in `raw_resource'
      /verification-tests/lib/openshift/resource.rb:293:in `conditions'
      /verification-tests/features/step_definitions/operators.rb:195:in `block (2 levels) in <top (required)>'
      /verification-tests/lib/base_helper.rb:160:in `wait_for'
      /verification-tests/features/step_definitions/operators.rb:193:in `/^operator "(.+?)" becomes ([\S]+|<%=.+?%>)(?: within ([0-9]+|<%=.+?%>) seconds)?$/'
      features/tierN/apiserver/kube-apiserver.feature:262:in `operator "kube-apiserver" becomes non-progressing within <%= cb.wait_time %> seconds'
```

With the changes in this PR,
```
    Given operator "kube-apiserver" becomes available/non-progressing/non-degraded within <%= cb.wait_time %> seconds # features/step_definitions/operators.rb:164
      [05:20:29] INFO> ......
      [05:22:00] INFO> Exit Status: 0
      [05:22:00] INFO> #### Operator kube-apiserver Expected conditions: {"Available"=>"True", "Progressing"=>"False", "Degraded"=>"False"}
      [05:22:00] INFO> #### After 91.10809769103071 seconds and 5 iterations operator kube-apiserver becomes: {"Available"=>"True", "Progressing"=>"False", "Degraded"=>"False"}
    # Validation for WriteRequestBodies profile setting
    When I use the "openshift-apiserver" project                                                                      # features/step_definitions/project.rb:110
      [05:22:00] INFO> Shell Commands: oc project openshift-apiserver --kubeconfig=/home/jenkins/workdir/b86a29ab7df1-rootx/ocp4_admin.kubeconfig
      Now using project "openshift-apiserver" on server "https://api.rgangwar-21de3.qe.devcluster.openshift.com:6443".
      [05:22:04] INFO> Exit Status: 0
waiting for operation up to 900 seconds..
```

### DIFF
--- a/lib/openshift/resource.rb
+++ b/lib/openshift/resource.rb
@@ -49,6 +49,9 @@ module BushSlicer
       elsif result[:response] =~ /Unable to connect to the server/
         logger.info(result[:response])
         return false
+      elsif result[:response] =~ /connection to the server \S+ was refused - did you specify the right host or port/ && env.nodes.length == 1
+        logger.info(result[:response])
+        return false
       else
         # e.g. when called by user without rights to list Resource
         raise "error getting #{self.class.name} '#{name}' existence: #{result[:response]}"


gangwgr:
@liangxia Why we are not checking error whole error.
`connection to the server \S+.qe.devcluster.openshift.com:6443 was refused - did you specify the right host or port?`
We found this issue for SNO as in this PR https://github.com/openshift/verification-tests/pull/2693. Can you please add check for SNO also?

liangxia:
> @liangxia Why we are not checking error whole error. `connection to the server \S+.qe.devcluster.openshift.com:6443 was refused - did you specify the right host or port?` We found this issue for SNO as in this PR #2693. Can you please add check for SNO also?

We use short string as the shorter string are less likely to change than the longer string, thus save us debug/maintenance work. We have see many cases that DEV optimize the sentences by add/remove a word.

Tested on SNO in Runner-v3-smoke/3970, and it works.
We didn't add SNO specific check there as it also benefits the non-SNO clusters with the connection refuse issue.

gangwgr:
@liangxia I think in my it is not picking whole string. Thanks

xingxingxia:
The error had been there since long time ago, it is better to avoid too fuzzy string pattern.
In addition, in SNO cluster the broken error is expected because only one replica is there, in HA cluster which ensures rotation gracefulness with HA replicas it is unexpected and should be raised for debugging and reporting bug instead of retried. Below epic from Apiserver subteam was the background feature:
```
API-1136
```
That is why it is commented: if the error contains `connection to the server ... was refused - did you specify the right host or port` && if it is SNO env.

gangwgr:
@liangxia Better to use "\S+" instead of "\w+". It covers only word but in cluster name we get digits, words and special characters.

liangxia:
> @liangxia Better to use "\S+" instead of "\w+". It covers only word but in cluster name we get digits, words and special characters.

@gangwgr Good catch.
The amazing part is that the tests passed with "\w+".
Let me also test it with "\S+". If the tests also passed, then I'll update it to "\S+"

liangxia:
> That is why it is commented: if the error contains `connection to the server ... was refused - did you specify the right host or port` && if it is SNO env.

Increase the string a little bit to be more descriptive. Let me know if you do want the full string.
After grepping the error string `was refused - did you specify the right host or port` for Runner logs in virt05, its really not specific for SNO, so its not added there.
```
$ grep -r 'was refused - did you specify the right host or port' | wc -l
4957
```

xingxingxia:
As said, in HA cluster the error is not expected, should not be retried, and leave it to be raised.

liangxia:
> > @liangxia Better to use "\S+" instead of "\w+". It covers only word but in cluster name we get digits, words and special characters.
> 
> @gangwgr Good catch. The amazing part is that the tests passed with "\w+". Let me also test it with "\S+". If the tests also passed, then I'll update it to "\S+"

Both "\w+" and "\S+" worked. Update it to "\S+" now

liangxia:
> As said, in HA cluster the error is not expected, should not be retried, and should leave it to be raised.

In a real world, when you lost connection to a sever, the first thing came to my mind is retry, rather than do a debug immediately.
The debug generally comes after several failure retries.  And in our case, if the connection issue persistent, the step will fail after reaching the timeout, and the connection refuse issue is still raised.
Also, about 5000 CI tests abort in the middle and leave test cases in idle status due to connection issue does not sound great to me.

gangwgr:
@liangxia @xingxingxia I tried below and it is working fine. @liangxia Can you please try like this?
```
 elsif result[:response] =~ /connection to the server \S+ was refused - did you specify the right host or port/ && env.nodes.length == 1
        logger.info(result[:response])
        return false
```

xingxingxia:
Not sure how many are HA and and how many are SNO among the 5000. Even if the ratio does not indicate something, still prefer to leave it to capture intermittently apiserver unavailable bugs in HA cluster like 1856729 1836016 1874582 1971046. Or if retried, could you help limit retry count if it can be implement.

liangxia:
Let's get the SNO issue resolved first.

xingxingxia:
> this test scenario ... also make sense... for the test step. The change to the library is global.

@jhou1 actually as per https://github.com/openshift/cucushift/pull/9025#issuecomment-1043871276 said, changing the global `exists` prevents any cases that poll (i.e. oc get) any resource from failing in SNO env, not only @gangwgr's case that polls resource clusteroperators kube-apiserver which executes command oc get clusteroperators kube-apiserver -o yaml.
/lgtm
@liangxia thx

gangwgr:
Thanks @xingxingxia @liangxia for fixing for PR.

PR #2706, opened 2022-02-24
Fix OCPQE-7477

kasturinarra:
Increasing the time as i see that the pod is still in terminating state which causes pattern hooks-2 to be found With this fix i am hoping we should not be seeing any more failures with respect to this case.


kasturinarra:

PR #2707, opened 2022-02-25
Fix OCPQE-8520

kasturinarra:
I see that test has failed twice with error below, so adding a step in the code to make sure that pod appears before moving to other step.

"failed to delete deployer/hook pod hooks-1-hook-pre for deployment #1: pods "hooks-1-hook-pre" not found


kasturinarra:
http://pastebin.test.redhat.com/1032628

PR #2716, opened 2022-03-02
Fix OCP-21034

XiyunZhao:
The error message will disappear on page when too many results found on fuzzy query, that's the reason why the case failed. So this changes is to make sure only one result can be found

### DIFF
--- a/lib/rules/web/admin_console/4.7/status.xyaml
+++ b/lib/rules/web/admin_console/4.7/status.xyaml
@@ -50,4 +50,4 @@ check_name_validation_error:
   - selector:
       xpath: //h4[contains(.,'error occur')]
   - selector:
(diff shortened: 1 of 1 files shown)


lib/rules/web/admin_console/4.7/status.xyaml
```diff
@@ -50,4 +50,4 @@ check_name_validation_error:
   - selector:
       xpath: //h4[contains(.,'error occur')]
   - selector:
-      xpath: //div[contains(.,'subdomain must consist of lower case alphanumeric character')]
+      xpath: //div[contains(text(),'subdomain must consist of lower case alphanumeric character')]
```
yapei, line 53: Could you help explain `The error message will disappear on page when too many results found on fuzzy query` ? I didn't get your point

PR #2718, opened 2022-03-03
Several changes to RHEL 8 Jenkins agent image

liangxia:
/cc @jhou1 @dis016 @pruan-rht @JianLi-RH 

We are using env USER_PATHS in our images, also we need the write permission to /etc/pki.


dis016 (commented):
/lgtm

PR #2719, opened 2022-03-03
fix OCPQE-8932

QiaolingTang:
Fix https://issues.redhat.com/browse/OCPQE-8932, the failure happens when channel name is `stable`,  below step will always fail:
```
  success = wait_for(180, interval: 10) {
    (subscription(subscription).installplan_csv.include? channel) || (subscription(subscription).installplan_csv.include? (channel.split('-')[1]))
  }
```
Here add a step to check channel name first, if it's `stable`, then only compare pre installed CSV version and new install CSV version.

/cc @anpingli

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -684,6 +684,7 @@ Given /^I upgrade the operator with:$/ do | table |
   namespace = opts[:namespace]
   step %Q/I use the "#{namespace}" project/
 
+  pre_csv = subscription(subscription).current_csv
   # upgrade operator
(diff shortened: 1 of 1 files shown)


QiaolingTang:
Tested upgrade from `stable-5.2` to `stable-5.3`, then upgrade to `stable`, here are the logs:
upgrade prepare: https://url.corp.redhat.com/f5d007a
upgrade to stable-5.3: https://url.corp.redhat.com/74d0fc6
upgrade to stable: https://url.corp.redhat.com/f24d763

PR #2721, opened 2022-03-03
introduce a case_query to exclude the kuryr network type profiles in test cases

jmekkatt:
@liangxia We have test case that needs to be excluded from the test case execution if the install profile is with network type kuryr.
As a first step to this , I would like to introduce a case_query to exclude the network type filter from the case filtration.
Once we finalize the case query, would like to introduce tag something like @exclude_network_kuryr to respective case.
Reference ticket :  OCPQE-7900.
Kindly review above change and feel free to provide comments and suggestion for the same requirement.


liangxia:
Is the issue exist in Jenkins CI or Prow ?

jmekkatt:
The PR has been outdated. I don't think the approach is right here. I would revisit on this and create new one with latest changes.

PR #2726, opened 2022-03-07
Wait for 120 seconds for checking multus status

weliang1:
Before the real multus testing steps running, https://github.com/weliang1/verification-tests/blob/master/features/step_definitions/networking.rb#L605 will check the numbers of desired_multus_replicas , available_multus_replicas and env.nodes.count,  if the conditions are not met, the script will exit with status: 0 and log " Multus is not running correctly!"

The previous running scripts(not multus cases) or other reasons May cause cluster in this bad state. Update the script to give 120 seconds for continue checking multus status before exiting the testing.

@openshift/team-sdn-qe PTAL

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -595,23 +595,23 @@ Given /^an IP echo service is setup on the master node and the ip is stored in t
   }
 end
 
-
 Given /^the multus is enabled on the cluster$/ do
   ensure_admin_tagged
-
+  multus_status_timeout = 120
   desired_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:desired]
   available_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:available]
   #storing desired_multus_replicas value in desired_multus_replicas clipboard variable as well
   cb.desired_multus_replicas = desired_multus_replicas
-  unless (desired_multus_replicas == available_multus_replicas || desired_multus_replicas > env.nodes.count) && available_multus_replicas != 0
-    daemon_set('multus', project('openshift-multus')).describe(admin, quiet:false)
-    BushSlicer::Pod.get_labeled("app=multus", user: admin, project: project("openshift-multus", switch: false)) do |pod|
-      pod.describe(admin, quiet: false)
+  wait_for(multus_status_timeout) {
+    unless (desired_multus_replicas == available_multus_replicas || desired_multus_replicas > env.nodes.count) && available_multus_replicas != 0
+      daemon_set('multus', project('openshift-multus')).describe(admin, quiet:false)
+      BushSlicer::Pod.get_labeled("app=multus", user: admin, project: project("openshift-multus", switch: false)) do |pod|
+        pod.describe(admin, quiet: false)
+      end
+      env.nodes(user:admin, refresh: true, quiet: false)
+      raise "Multus is not running correctly!"
     end
-    env.nodes(user:admin, refresh: true, quiet: false)
-    raise "Multus is not running correctly!"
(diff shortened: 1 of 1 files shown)


features/step_definitions/networking.rb
```diff
@@ -595,23 +595,23 @@
...
+  multus_status_timeout = 120
   desired_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:desired]
   available_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:available]
   #storing desired_multus_replicas value in desired_multus_replicas clipboard variable as well
   cb.desired_multus_replicas = desired_multus_replicas
-  unless (desired_multus_replicas == available_multus_replicas || desired_multus_replicas > env.nodes.count) && available_multus_replicas != 0
-    daemon_set('multus', project('openshift-multus')).describe(admin, quiet:false)
-    BushSlicer::Pod.get_labeled("app=multus", user: admin, project: project("openshift-multus", switch: false)) do |pod|
-      pod.describe(admin, quiet: false)
+  wait_for(multus_status_timeout) {
```
zhaozhanqi, line 605: I think we better use interval, like `wait_for(multus_status_timeout, interval: 10)`

features/step_definitions/networking.rb
```diff
@@ -595,23 +595,23 @@
...
+    unless (desired_multus_replicas == available_multus_replicas || desired_multus_replicas > env.nodes.count) && available_multus_replicas != 0
```
zhaozhanqi, line 606: Do you have one logs for this changes?  I think we can set one multus pod to be down by manual and then run this testing to check if this works well. 
and another issue is:
Not sure if this will print a lot of logs( pod description etc.) after this changes. if so. I think we need to one way to reduce the test logs.

features/step_definitions/networking.rb
weliang1: > I think we better use interval, like `wait_for(multus_status_timeout, interval: 10)`

@zhaozhanqi Fixed.

features/step_definitions/networking.rb
weliang1: @zhaozhanqi Update script to only print error message, not print pod description log in the wait_for loop.

Log from good multus status cluster:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4087/console

Log from bad multus status cluster:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4088/console

features/step_definitions/networking.rb
```diff
@@ -595,23 +595,27 @@
   }
 end
 
-
 Given /^the multus is enabled on the cluster$/ do
   ensure_admin_tagged
-
   desired_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:desired]
   available_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:available]
```
zhaozhanqi, line 601: seems 600-601 is not useful now?  
other lgtm

weliang1: @zhaozhanqi Oops, forgot those duplicate lines, delete those now.

New Log from bad multus status cluster:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4089/console

New Log from good multus status cluster:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4091/console

PR #2728, opened 2022-03-08
update test edge route with 2048 bit key

lihongan:
The old test edge route was using the key generated by 1024 bit and not supported in FIPS enabled cluster, so regenerate new key and cert for the edge route and update the testdata.

@quarterpin @melvinjoseph86 @ShudiLi Please help take a look. Thanks.


lihongan:
Here is log of OCP-30190 that using the updated testdata and executed on FIPS enabled cluster: 
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/344558/console

PR #2730, opened 2022-03-08
update the replicas of router pods

lihongan:
Some of vsphere and bare metal profiles scale up rhel worker nodes in the cluster, to ensure router pods can work on those nodes so this case scales router pods to make them run on all worker nodes.
But ingress often hit unscheduled worker node issue during a upgrade and blocked the upgrade, the PR is trying to mitigate the problem by decreasing the number of router pods. Let's keep an eye on upgrade CI to see if it is helpful for troubleshooting and debugging.

@quarterpin @melvinjoseph86 @ShudiLi Please help take a look. Thanks.

### DIFF
--- a/features/upgrade/routing/upgrade.feature
+++ b/features/upgrade/routing/upgrade.feature
@@ -81,17 +81,18 @@ Feature: Routing and DNS related scenarios
     # Get the number of worker nodes and scale up router pods
     Given I switch to cluster admin pseudo user
     And I store the number of worker nodes to the :num_workers clipboard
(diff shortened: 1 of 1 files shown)


lihongan:
upgrade prepare log:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/345513/consoleFull

upgrade check log:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/345514/consoleFull

PR #2734, opened 2022-03-10
OCPQE-8979: Fix ID issue

liangxia:
We run the container with a very large random uid, but it does not have name mapping to it in /etc/passwd.  And the git used in the RHEL 8 image need that.
```
sh-4.2$ id
uid=1002280000 gid=0(root) groups=0(root),1002280000

sh-4.2$ id -un
id: cannot find name for user ID 1002280000
1002280000
```


pruan-rht (commented):
/lgtm

PR #2741, opened 2022-03-14
OCP-26139: increase jitter threshold

rbbratta:
increase threshold until we only fail when there is a real error

In this case jitter was 3.26201295671534


rbbratta:
https://issues.redhat.com/browse/OCPQE-7814

PR #2743, opened 2022-03-16
OCP-10904 has been fixed on 4.11 and 4.10,remove flaky tag

xiuwang:
@liangxia Please help review this, thanks


liangxia:
Could you link the PR which fix the issue ?

xiuwang:
@liangxia It's fixed in this bug https://bugzilla.redhat.com/show_bug.cgi?id=2060605. The case only failed on 4.10 and 4.11 before, and the bug has been backport to 4.10.

PR #2748, opened 2022-03-19
fix UI to use 'stable' channel

pruan-rht:
depends on https://github.com/openshift/cucushift/pull/9076


jhou1:
Looks good to me, needs approval from test owners.

PR #2750, opened 2022-03-21
fix OCPQE-9246

QiaolingTang:
Fix https://issues.redhat.com/browse/OCPQE-9246

During upgrade, ES pods performs rolling-upgrade, the cluster health status will be `green` temporarily, so checking the cluster health and upgrade status in CR elasticsearch isn't proper. Here change to check the ES cluster health and compare the ready pods in CR elasticsearch and the pods in the logging project. 

/cc @anpingli


QiaolingTang:
Log: 
upgrade-prepare: https://url.corp.redhat.com/026f9bb
upgrade-check: https://url.corp.redhat.com/5e81454

anpingli (commented):
/lgtm

PR #2752, opened 2022-03-21
update fluentd configuration

QiaolingTang:
When testing forward logs to fluentd with logging 5.4, some infra container logs are stored into app.log, the root cause is we change the source to `/var/log/pods/` in logging 5.4. 
Update the fluentd configuration to fix this issue.


QiaolingTang:
Log: https://url.corp.redhat.com/e35ba2d

PR #2755, opened 2022-03-22
Add wait so that descheduler cluster pod comes up fine

kasturinarra:
With 4.10 there were recent changes to descheduler where the descheduler cluster pod comes up then terminates immediately to start securely, so adding some wait to make sure that test here does not fail.

### DIFF
--- a/features/upgrade/workloads/descheduler-upgrade.feature
+++ b/features/upgrade/workloads/descheduler-upgrade.feature
@@ -16,6 +16,7 @@ Feature: Descheduler major upgrade should work fine
     When I run the :create admin command with:
       | f | kubedescheduler-<%= cb.master_version %>.yaml |
     Then the step should succeed
+    Given 60 seconds have passed
     And status becomes :running of exactly 1 pods labeled:
       | app=descheduler |
     Given evaluation of `pod.name` is stored in the :pod_name clipboard


kasturinarra:
@zhouying7780 could you please help review the fix? thanks !!

kasturinarra:
@liangxia could you please help review and merge ? Thanks !!

features/upgrade/workloads/descheduler-upgrade.feature
```diff
@@ -16,6 +16,7 @@ Feature: Descheduler major upgrade should work fine
     When I run the :create admin command with:
       | f | kubedescheduler-<%= cb.master_version %>.yaml |
     Then the step should succeed
+    Given 60 seconds have passed
```
liangxia, line 19: Hard wait should always avoid unless necessary and can not be workaround.
@pruan-rht  Thoughts ?

kasturinarra: @liangxia i do not think there is a way for me to workaround since i wanted this test to work for all 4.6 to 4.10 clusters, in 4.10 there has been a fix where the cluster pod comes up, terminates and a new cluster pod comes up and that is the reason i have put a hard wait.

pruan-rht: @kasturinarra why the wait makes a difference?  In the current repo I see 
```
    Then the step should succeed
    And status becomes :running of exactly 1 pods labeled:
      | app=descheduler |
 ```
 perhaps try to use this step instead  `Given /^a pod becomes ready with labels:$/ `

kasturinarra: @pruan-rht the problem is a pod comes up with label app=descheduler then runs for around 7seconds and another pod comes up with same label app=descheduler and continues to be running state. The time it stays is very less and the same case needs to be executed for all other versions where in which this change is not present.(like pod coming, disappearing and another pod coming up and running). So i feel having a hard wait will avoid seeing failures with respect to the case and that is the reason for it.

kasturinarra: @pruan-rht could you please help check my reply and review this again ? thanks !!

pruan-rht: @kasturinarra ok, let's see if this will make it more stable.
/lgtm

PR #2756, opened 2022-03-23
don't run logging upgrade on baremetal  

anpingli:
don't run logging upgrade on baremetal  as the persistent volumes are provided by nfs pod. the nfs pod may be in the bad status. that may make Logging ES can not find the storage and thus can not be terminated.

https://issues.redhat.com/browse/OCPQE-8922

### DIFF
--- a/features/upgrade/logging/upgrade.feature
+++ b/features/upgrade/logging/upgrade.feature
@@ -7,8 +7,8 @@ Feature: Logging upgrading related features
   @upgrade-prepare
   @users=upuser1,upuser2
   @4.11 @4.10 @4.9 @4.8 @4.7 @4.6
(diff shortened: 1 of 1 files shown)


anpingli:
@duanwei33  @JianLi-RH  @QiaolingTang  PTAL

QiaolingTang:
/lgtm

@anpingli It seems these tags are used for prow CI, if we want to exclude this case from the pre-release CI when testing on bare metal, we may need to update in Polarion too. Please correct me if I'm wrong.

jhou1:
Just a note, the Polarion test case needs to be updated to remove baremetal cloud provider as well. We are syncing these tags during reach release and the data source is from Polarion.

anpingli:
The cases had been updated.

jhou1:
Thanks
/approve

PR #2757, opened 2022-03-23
disable the snapshot upgrade case which uses hostpath csi driver

duanwei33:
To fix https://issues.redhat.com/browse/OCPQE-9371
We have snapshot upgrade with AWS EBS CSI Driver, so disable this one which uses hostpath csi driver

@chao007 @ropatil010 PTAL


duanwei33:
@jhou1 PTAL, thanks.

PR #2758, opened 2022-03-23
Adding @flaky tag so that test gets skipped

kasturinarra:
I see that test is not suitable to run with 4.10 & 4.11 also if ci/user tries to upgrade from 4.7->4.10 upgrade will still fail as scheduler cannot be upgraded by having policy configmap set. So skipping the test, i will add other tests with checks to make sure that kube-scheduler pods run fine after upgrade.


kasturinarra:
@liangxia could you please help review and merge? Thanks !!

PR #2765, opened 2022-03-29
Add network related tags

liangxia:
/cc @zhaozhanqi @jhou1 @JianLi-RH @dis016 @pruan-rht

### DIFF
--- a/features/admin/pod.feature
+++ b/features/admin/pod.feature
@@ -7,6 +7,7 @@ Feature: pod related features
   @vsphere-upi @openstack-upi @gcp-upi @baremetal-upi @azure-upi @aws-upi
   @upgrade-sanity
   @singlenode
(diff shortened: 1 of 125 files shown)


zhaozhanqi:
see still only `network-ovnkubernetes`?

liangxia:
> see still only `network-ovnkubernetes`?

The PR contains tags for `@network-ovnkubernetes` and `@network-multitenant`. E.g, https://github.com/openshift/verification-tests/pull/2765/files#diff-5988860cf440c001e7daa1191c9720d193dc8c66d27333c296a37c135565c67a

We can not render new tags due to Polarion/polarshift outage, will do that in a follow up PR once they are back.

zhaozhanqi:
ah, got it, thanks your reply. @liangxia

PR #2767, opened 2022-03-29
Set skip_if_unavailable to true

liangxia:
Skip unavailable repositories

### DIFF
--- a/tools/openshift-ci/Dockerfile
+++ b/tools/openshift-ci/Dockerfile
@@ -5,6 +5,7 @@ LABEL maintainer="OCP QE Team"
 USER root
 
 RUN set -x && \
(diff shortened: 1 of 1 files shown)


jhou1:
With this change can we still successfully build the images with needed dependencies?

liangxia:
> With this change can we still successfully build the images with needed dependencies?

It depends.  If an optional repo is unavailable, then the PR workaround that issue and moves on. But if an critical repo is unavailable, it will fail several steps further due to required dependencies can not met.

liangxia:
/hold
Research with other ways.

liangxia:
Prefer https://github.com/openshift/release/pull/27362

/close

PR #2776, opened 2022-04-07
update OCP-30501 to count the number of linux workers

lihongan:
To fix https://issues.redhat.com/browse/OCPQE-9496  since router pods cannot be scheduled to windows worker nodes.

@quarterpin @melvinjoseph86 @ShudiLi  Please help take a look. Thanks.

### DIFF
--- a/features/step_definitions/node.rb
+++ b/features/step_definitions/node.rb
@@ -26,7 +26,7 @@ end
 # select a random node from a cluster.
 Given /^I select a random( windows)? node's host$/ do | windows |
   ensure_admin_tagged
-  nodes = env.nodes.select { |n| n.schedulable? && (windows ? n.is_windows_worker? : ! n.is_windows_worker?) }
+  nodes = env.nodes.select { |n| n.schedulable? && (windows ? n.is_windows? : ! n.is_windows?) }
   cache_resources *nodes.shuffle
   @host = node.host
 end
@@ -56,7 +56,7 @@ Given /^I store the( schedulable| ready and schedulable)?( windows)? (node|maste
   end
 
   ###select windows worker or not
-  cb[cbname] = cb[cbname].select { |n| windows ? n.is_windows_worker? : ! n.is_windows_worker? }
+  cb[cbname] = cb[cbname].select { |n| windows ? n.is_windows? : ! n.is_windows? }
 
   cache_resources *cb[cbname].shuffle
 end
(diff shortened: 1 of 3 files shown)


lihongan:
log: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/380950/console

lib/openshift/node.rb
```diff
@@ -99,7 +99,12 @@ def is_master?(user: nil, cached: true, quiet: false)
     def is_windows_worker?(user: nil, cached: true, quiet: false)
       rr = raw_resource(user: user, cached: cached, quiet: quiet)
       rr.dig('metadata', 'labels', 'kubernetes.io/os') == "windows"
-    end    
+    end
+
+    def is_linux?(user: nil, cached: true, quiet: false)
```
liangxia, line 104: Perhaps unify the name to `is_linux_worker` ?

lihongan: Since master node also uses the label `kubernetes.io/os=linux` so it just can identify all linux nodes but cannot identify linux worker nodes. Maybe I can update the name as `is_linux_node`, WDYT?
It is different with the one `def is_windows_worker?` because windows node cannot be configured as master for now. Indeed I don't think `is_windows_worker` is accurate.

liangxia: Then feel free to update `is_windows_worker` to the proper one and make them unify.

lihongan: sure, updated all `is_windows_worker` to `is_windows`.

features/step_definitions/node.rb
```diff
@@ -633,9 +633,15 @@
...
+Given /^I store the number of( linux| windows)? worker nodes to the#{OPT_SYM} clipboard$/ do |os_type, cb_name|
   nodes = BushSlicer::Node.list(user: admin)
-  worker_nodes = nodes.select { |n| n.is_worker? }
+  if !os_type
+    worker_nodes = nodes.select { |n| n.is_worker? }
+  elsif os_type.strip == "linux"
+    worker_nodes = nodes.select { |n| n.is_linux? && n.is_worker?}
+  else
+    worker_nodes = nodes.select { |n| n.is_windows && n.is_worker?}
+  end
```
liangxia, line 644: Since we only want the workers, perhaps we can use
```
  worker_nodes = nodes.select { |n| n.is_worker? }
  if os_type =~ /linux/
    worker_nodes = worker_nodes.select { |n| n.is_linux? }
  elsif os_type =~ /windows/
    worker_nodes = worker_nodes.select { |n| n.is_windows? }
  end
```
Not tested though.

lihongan:
logs: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/382407/

also tested `... the number of worker nodes...` and `...the number of windows worker nodes...` locally and both work well.

features/step_definitions/node.rb
lihongan: Good suggestion! updated

PR #2781, opened 2022-04-07
Special debug code to execute after scenario step in case of special test case failures.

jmekkatt:
As you see in both [OCPQE-8172](https://issues.redhat.com/browse/OCPQE-8172) and [OCPQE-8117](https://issues.redhat.com/browse/OCPQE-8117) tickets, in these cases, it is difficult for us to debug test case failures or taking longer time to identify the root cause. To minimize the effort on debugging such failures , we want to introduce certain debug code to execute right after the unexpected failures during execution.
I have investigated with cucushift framework ( with my limited knowledge :) ) and made few dummy changes and that makes execution of debug statements "after scenario" step. The intention is NOT to execute debug statements for every case but need to execute only in case of failures w.r.t kube-apiserver such as,
1. xxxx:6443 connection is refused.
2. xxxx:6443 timeout.
3. server is currently unable to handle the request (get projects.project.openshift.io)
Currently i'm unable to get the test case result or it's scope to make such decision for above code. I would need your review and suggestions to achieve the solution.
Note: These are random failures and not stick to any specific test case , Hence I see that it's needs to be handled in core instead of step implementation.
Attached local execution results for both positive and negative scenarios - http://pastebin.test.redhat.com/1043194
@wangke19 @liangxia @qe-productivity-team Please share your thoughts.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -73,6 +73,10 @@ After do |_scenario|
   debug_in_after_hook
 
   begin
+    # Calling additional debug statement and when the scenario is failed along with custom error message in execution.
+    if (self.scenario.failed?) && (logger.handled_error?)
+      print_debugging_info(localhost)
+    end
     ## raise inside block only if error can affect next scenarios execution ##
     # Manager will call clean-up including self.after_scenario
     manager.after_scenario
--- a/lib/log.rb
+++ b/lib/log.rb
@@ -19,6 +19,8 @@ module BushSlicer
     # include Term::ANSIColor
 
     attr_reader :level
+    attr_accessor :handled_error
+
     CENSOR_KW = [
       'kind: secret',
       '"kind": "secret"',
@@ -82,6 +84,10 @@ module BushSlicer
       end
     end
 
+    def handled_error?
+      handled_error && !handled_error.empty?
+    end
+
     def dedup_start
       require 'strscan'
 
@@ -106,6 +112,8 @@ module BushSlicer
     end
 
     def log(msg, level=INFO, show_datetime='time')
+      filter_custom_logs(msg)
+
       return if level > self.level
 
       ## take care of special message types
@@ -163,6 +171,12 @@ module BushSlicer
       @@puts_method.call(msg)
     end
 
+    def filter_custom_logs(msg)
+      # This is to capture the message when the trickier failure like below seen in any test case execution.
+      if (msg =~/connection to the server\S+was refused|Unable to connect to the server|Error from server \(ServiceUnavailable\)/)
+        @handled_error = msg
+      end
+    end
     # supports embedding content similar with same semantics as Cucumber
     def embed(src, mime_type, label)
       if self.class.runtime.respond_to? :attach
--- a/lib/world.rb
+++ b/lib/world.rb
@@ -50,6 +50,20 @@ module BushSlicer
       end
     end
 
+    def print_debugging_info(host)
+      # Intention of method is to execute additional debug statement to analyze unexpected trickier failures.
+      logger.handled_error = ''
+      res = admin.cli_exec(:get, resource: "nodes", _quiet: true)
+      nodes = host.exec_raw("echo '#{res[:response]}' | grep -E 'SchedulingDisabled|NotReady' | awk '(NR>1)'", quiet: true)
+      res = admin.cli_exec(:get, resource: "co", _quiet: true)
+      cos = host.exec_raw("echo '#{res[:response]}' | grep -v '.True.*False.*False' | awk '(NR>1)'", quiet: true)
+      res = admin.cli_exec(:get, all_namespaces: true, resource: "pods", _quiet: true)
+      pods = host.exec_raw("echo '#{res[:response]}' | grep -Ev 'Running|Completed' | awk '(NR>1)'", quiet: true)
+      logger.info("List of abnormal nodes :\n" + nodes[:response])
+      logger.info("List of abnormal co :\n" + cos[:response])
+      logger.info("List of abnormal pods :\n" + pods[:response])
+    end
+
     def scenario_tags
       scenario.source_tag_names
     end


lib/world.rb
```diff
@@ -50,6 +50,16 @@ def debug_in_after_hook
       end
     end
 
+    def debug_stmts_on_failures
```
pruan-rht, line 53: @jmekkatt perhaps change the method name to `print_debugging_info`

lib/world.rb
```diff
@@ -50,6 +50,16 @@ def debug_in_after_hook
...
+      # Intention of method is to execute additional debug statement to analyze most tricker failures.
+      logger.info("Get the list of pods under kubeapiserver namespace")
```
pruan-rht, line 55: @jmekkatt good start, but these prints seems testcase specific.

wangke19: @jmekkatt From the test results, it works well. Actually, when case ran into failure, we are more concerned with abnormal nodes, operators and pods, events.
`oc get node --no-headers | grep -E 'SchedulingDisabled|NotReady'`
`oc get co | grep -v '.True.*False.*False'`
`oc get pod -A | grep -Ev 'Running|Completed'`
`oc get events -A | grep -v Normal | grep " [0-9]\{2\}m " | awk '$2 <= 15m'`

wangke19: Here a way to make the cluster abnomally, you can use it to run test case, it will cause several operators Degraded.
```
oc debug node/ip-xx-0-xxx-242.us-east-2.compute.internal
sh-4.4# ip link | grep ens
sh-4.4# cat > test.sh <<EOF
ifconfig ens3 down
sleep 300
ifconfig ens3 up
EOF
sh-4.4# bash ./test.sh &
```

features/support/env.rb
```diff
@@ -29,6 +29,10 @@
   BushSlicer::World.new
 end
 
+def world
+  return BushSlicer::World.new
+end
```
akostadinov, line 34: you need to create a new world? Why not call `self` to get current world? btw note that between scenario executions, world may disappear or be inaccessible or something like that I remember. I was dealing with formatting. You can see in `Log` class that we fall back to `Kernel` when World does not exist.

lib/world.rb
akostadinov: I would suggest you to use a tag specific `After` block so that this debugging only happens with the tests that need it. This is without looking in details whether this debugging is universally useful or not. Just on the outlook it appears too much to add this unconditionally.

I leave to OpenSHift QE team to decide how to do. Just I noticed this PR in my inbox and it looked interesting so decided to comment. Cheers.

lib/world.rb
jmekkatt: Done

jmekkatt:
Updated the PR with following changes.

1. Implemented logging of pod,operator and node status when the below trickier errors (probably an unexpected and independent test cases) found during case execution.
       - The connection to the server api.qeci-34036.0222-k2w.qe.rhcloud.com:6443 was refused - did you specify the right host or port?
       - Unable to connect to the server: dial tcp 138.91.78.249:6443: i/o timeout
       - Error from server (ServiceUnavailable): the server is currently unable to handle the request (get projects.project.openshift.io)
2. Logging of pod,operator and node status would happen in case scenario is failed and above custom logs captured during test case execution.
3. Implemented additional hacks in Logger class to hide some of the unwanted logs/logging while filtering above requirement.
See execution logs [here](http://pastebin.test.redhat.com/1066659)
@wangke19 @akostadinov @pruan-rht  Kindly review when you have time , let me know your feedback with new changes. Thanks.

lib/log.rb
```diff
@@ -80,6 +80,33 @@ def initialize(level=nil)
       else
         @level = INFO
       end
+      @ignore_log = false
```
wangke19, line 83: Lines 83-85, missed indents.

wangke19: others, I'm OK.

jmekkatt: Updated and also added addition comments.

lib/world.rb
```diff
@@ -50,6 +51,25 @@ def debug_in_after_hook
...
+      # Hack(below two lines) to ignore unwanted logger messages.
+      logger.write_custom_errmsg('')
+      logger.set_log_ignore(true)
+      logger.set_log_level('PLAIN')
+      res = admin.cli_exec(:get, resource: "nodes")
+      nodes = hos.exec_raw("echo '#{res[:response]}' | grep -E 'SchedulingDisabled|NotReady' | awk '(NR>1)'")
+      res = admin.cli_exec(:get, all_namespaces: true, resource: "co")
+      cos = hos.exec_raw("echo '#{res[:response]}' | grep -v '.True.*False.*False' | awk '(NR>1)'")
+      res = admin.cli_exec(:get, all_namespaces: true, resource: "pods")
+      pods = hos.exec_raw("echo '#{res[:response]}' | grep -Ev 'Running|Completed' | awk '(NR>1)'")
```
akostadinov, line 65: I remember we had `_silent: true` options when executing commands. Maybe you can use this option instead of this custom log filtering? Maybe it was `silent: true` for `hos.exec_raw`. You can `grep` through the existing codebase to see.

btw perhaps you meant `host` instead of `hos`?

jmekkatt: I did grep search in repo about `silent`, but couldn't see any related implementation. 
To avoid confusion, I have changed the variable name from `hos` to `host` now.

akostadinov: ops, sorry, it is `quiet: true` for host commands and `_quiet: true` for oc cli commands.

jmekkatt: Wow, this is very useful and reduces the additional hacks. Thanks for the suggestion. Updated now.

lib/log.rb
```diff
@@ -80,6 +80,19 @@ def initialize(level=nil)
...
+    end
+
+    def write_custom_errmsg(msg)
+      # Capture custom error message during execution. 
+      @custom_err = msg
+    end
+
+    def custom_errmsg
+      # Returns the custom error log in executions.
+      return @custom_err
```
akostadinov, line 95:
```suggestion
attr_accessor :handled_error
```

`attr_accessor` will create automatically reader and writer methods for the specified instance variable. It saves some lines. Just put it somewhere at the top.

My suggestion is to use name `handled_error` to be more clear about the intention of the value.

lib/log.rb
```diff
@@ -80,6 +80,19 @@ def initialize(level=nil)
       else
         @level = INFO
       end
+
+      # Reference variables to evaluate custom error values.
+      @custom_err = ''
```
akostadinov, line 85:
```suggestion
      handled_error = ''
```

This will use the reader method defined by `attr_accessor`.

akostadinov: But in fact it would be a good ruby style to remove this line and write a method like

```ruby
def handled_error?
  handled_error && !handled_error.empty?
end
```

Then later, instead of checking `logger.custom_errmsg.length > 0`, you would do:
```ruby
logger.handled_error?
```

lib/log.rb
jmekkatt: Done

lib/log.rb
jmekkatt: Updated

lib/log.rb
```diff
@@ -80,6 +82,13 @@ def initialize(level=nil)
       else
         @level = INFO
       end
+
+      # Reference variables to evaluate handled error status.
+      @handled_error = ''
+    end
```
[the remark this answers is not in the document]
jmekkatt: I wasn't not sure if that get initialized every time from Logger class hence introduced this step. Anyway I totally agree your thoughts , analyzed further and removed this part. Thanks.

jmekkatt:
@akostadinov if all looks fine now , could you please add lgtm ?

jmekkatt:
@liangxia Kindly help to review the PR . Thanks

akostadinov:
It looks good now, although not ideal separation of concerns. Somebody from OCP QE needs to approve though because I have no idea about how useful is this additional debug information.

liangxia:
/hold

Let's hold the change for a few more days until we release 4.11 out, just in case.

lib/world.rb
```diff
@@ -50,6 +50,20 @@ def debug_in_after_hook
       end
     end
 
+    def print_debugging_info(host)
+      # Intention of method is to execute additional debug statement to analyze unexpected trickier failures.
+      logger.handled_error = ''
+      res = admin.cli_exec(:get, resource: "nodes", _quiet: true)
+      nodes = host.exec_raw("echo '#{res[:response]}' | grep -E 'SchedulingDisabled|NotReady' | awk '(NR>1)'", quiet: true)
+      res = admin.cli_exec(:get, all_namespaces: true, resource: "co", _quiet: true)
```
liangxia, line 58: ClusterOperator is not a namespace scope resource. `all_namespaces` is not needed here.

jmekkatt: Updated. Thank you.

lib/log.rb
```diff
@@ -163,6 +171,12 @@ def print(msg)
       @@puts_method.call(msg)
     end
 
+    def filter_custom_logs(msg)
+      # This is to capture the message when the trickier failure like below seen in any test case execution.
+      if (msg =~/connection to the server\S+was refused|Unable to connect to the server|Error from server \(ServiceUnavailable\)/)
```
liangxia, line 176: If we are `Unable to connect to the server`, will the debug step `oc get node` works ?

PR #2785, opened 2022-04-11
OVN: use pkill -9 to simulate crash and fix typo

rbbratta:
BZ 2022144: `sbdb` will CrashLoopBackOff when there is a stale PID file.

Regular `pkill` will send `SIGTERM` which allows the container to cleanup
the pid file somehow in one of the hooks.  To simulate a crash we need
to kill without chance of cleanup and leave a stale PID.

Now that https://github.com/openshift/cluster-network-operator/pull/1256
is merged we can test without breaking the cluster

s/openfolw/openflow/

### DIFF
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -136,10 +136,10 @@ Feature: OVN related networking scenarios
     Then the step should succeed
     """
     And I run the :scale admin command with:
-      | resource  | deployment                 |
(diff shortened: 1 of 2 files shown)


rbbratta:

http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2022-04-11-18-38-28.log

PR #2786, opened 2022-04-12
skip step to be honored post upgade too

anuragthehatter:
Need to add OVNK backend check in ```upgrade-check``` scenario too. This will enable ipsec tests to be skipped on non supported profiles during ```upgrade-check``` as well.

Check step was added in upgrade-prepare in the following PR
Ref:https://github.com/openshift/verification-tests/pull/2751

@zhaozhanqi @liangxia Quick merge. Thanks

### DIFF
--- a/features/upgrade/sdn/ipsec-upgrade.feature
+++ b/features/upgrade/sdn/ipsec-upgrade.feature
@@ -75,6 +75,7 @@ Feature: IPsec upgrade scenarios
   @upgrade
   @proxy @noproxy @disconnected @connected
   Scenario: Confirm node-node and pod-pod packets are ESP enrypted on IPsec clusters post upgrade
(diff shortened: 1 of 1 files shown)


features/upgrade/sdn/ipsec-upgrade.feature
```diff
@@ -75,6 +75,7 @@ Feature: IPsec upgrade scenarios
   @upgrade
   @proxy @noproxy @disconnected @connected
   Scenario: Confirm node-node and pod-pod packets are ESP enrypted on IPsec clusters post upgrade
+    Given the env is using "OVNKubernetes" networkType
```
zhaozhanqi, line 78: why not `Given /^the IPsec is enabled on the cluster$/ do` step?

PR #2787, opened 2022-04-12
Sync upgrade tags from check to prepare

liangxia:


juzhao:
missed https://github.com/openshift/verification-tests/blob/master/features/upgrade/monitoring/monitoring-upgrade.feature

liangxia:
> and we should not sync all tags from check to prepare, since they may have different tags

@juzhao Could you list some examples ? The only one I know is the upgrade-prepare/check.

liangxia:
/approve

We can fix individual ones which is missed from this PR in follow up PRs.

PR #2790, opened 2022-04-13
Added 180 seconds of wait time in upgrade.feature

radeore:
Scenario updated: Cluster operator storage should be in correct status and dynamic provisioning should work well after upgrade

Execution run log: 
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/389873/

Issue/bug number:
https://issues.redhat.com/browse/OCPQE-9247

### DIFF
--- a/features/upgrade/storage/upgrade.feature
+++ b/features/upgrade/storage/upgrade.feature
@@ -90,10 +90,14 @@ Feature: Storage upgrade tests
     Given the "storage" operator version matches the current cluster version
 
     # Check cluster operator storage should be in correct status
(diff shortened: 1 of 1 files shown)


duanwei33:
@radeore Thanks for the update, it should help the update case (impacted by another case as mentioned in Jira)

duanwei33:
@jhou1 PTAL, thanks.

jhou1:
I think we need to ensure destructive tests are executed last according to the analysis in Jira. If that is implemented, do we have to add wait here?
/hold

PR #2791, opened 2022-04-14
fixed vf number and negative log

zhaozhanqi:
@openshift/team-sdn-qe ^^


zhaozhanqi:
@yingwang-0320 PTAL?

zhaozhanqi:
since there is bug for idle service cannot be waked up from 4.6 to 4.7 for ovn cluster. and it has been marked as docs bug https://bugzilla.redhat.com/show_bug.cgi?id=2041307

so here remove 4.6 version for idle case upgrade

PR #2796, opened 2022-04-19
remove OCP-33040 as it is rewrited with golang

sunzhaohua2:
This case is rewrite with golang https://github.com/openshift/openshift-tests-private/pull/3289

### DIFF
--- a/features/machine/machine.feature
+++ b/features/machine/machine.feature
@@ -308,13 +308,6 @@ Feature: Machine features testing
       | iaas_type | machineset_name        |
       | gcp       | machineset-clone-32126 | # @case_id OCP-32126
 
-    @azure-ipi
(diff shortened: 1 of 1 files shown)


features/machine/machine.feature
```diff
@@ -308,6 +308,7 @@ Feature: Machine features testing
       | iaas_type | machineset_name        |
       | gcp       | machineset-clone-32126 | # @case_id OCP-32126
 
+    @inactive
```
liangxia, line 311: Since it is migrated, perhaps we should just remove them here, HDYT ? Currently tag inactive is used to indicate the test case is inactive in Polarion, which is not suitable for this case.

liangxia:
/lgtm

Would you mind to update PR title to reflect the changes ?

PR #2797, opened 2022-04-20
Remove unsupported release for OCP-27617

duanwei33:
Cinder CSI Driver is not released in 4.6, so remove it from Polarion and cucushift code.
@jhou1 PTAL


jhou1:
@duanwei33 please the version is removed from Polarion as well, otherwise, our next sync will add it back.

/approve

PR #2802, opened 2022-04-21
fix OCPQE-9774

QiaolingTang:
This PR:
1. add a step to check remaining CPU and memory on linux worker nodes before deploying logging to avoid hitting issue like:
```
0/8 nodes are available: 2 node(s) didn't match node selector, 3 Insufficient memory, 3 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.
```
2. log the clusterlogging and elasticsearch status before checking logging pods' status for failure debugging. 

/cc @anpingli

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -153,6 +153,9 @@ Given /^I wait for clusterlogging(?: named "(.+)")? to be functional in the#{OPT
   step %Q/I switch to cluster admin pseudo user/
   step %Q/I use the "#{proj_name}" project/
 
+  # log the clusterlogging and elasticsearch status before checking logging pods' status.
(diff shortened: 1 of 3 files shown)


anpingli (commented):
/lgtm. Many thanks

features/step_definitions/logging.rb
```diff
@@ -1289,3 +1292,23 @@
   end
   cb[collector_name] = fluentd_component_label
 end
+
+Given /^I check if the remaining_resources in woker nodes meet the requirements for logging stack$/ do
```
liangxia, line 1296: What's `remaining_resources` ? Did you mean `free resources` ?

QiaolingTang: Yes, it's defined in https://github.com/openshift/verification-tests/blob/master/lib/openshift/node.rb#L252-L263

liangxia: Thanks for the confirm.
Ack

PR #2806, opened 2022-04-22
OCPQE-9841: Add several tags for upgrade-prepare

liangxia:
/cc @geliu2016 @juzhao @dis016 @pruan-rht @jhou1 @JianLi-RH


geliu2016 (approved):
/lgtm

PR #2807, opened 2022-04-22
Fix non groupified resources

kasturinarra:
I see that some more testdata files needed fix, so raising a pr here


kasturinarra:

PR #2808, opened 2022-04-24
OCPQE-9784: Add query for architecture

liangxia:
This is to support [OCPQE-9784](https://issues.redhat.com//browse/OCPQE-9784), which will be used to add tags for amd64/arm64

/cc @aleskandro @YuLi517 @jhou1 @wsun1

### DIFF
--- a/doc/prow/query_amd64.json
+++ b/doc/prow/query_amd64.json
@@ -1,0 +1,3 @@
+{
+  "case_query": "products.KEY:ocp AND type:testcase AND caseautomation.KEY:automated AND automation_script:cucushift AND NOT status:inactive AND NOT upstream.KEY:yes AND NOT tags:(stage_only prod_only) AND arch.KEY:x8664"
+}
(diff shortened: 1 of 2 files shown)


aleskandro:
Hi, as per comment in https://github.com/openshift/cucushift/pull/9103#issuecomment-1107719891 we could need to add another field to the arch option in Polarion, "heterogeneous", to be able disabling cases that are supported for both amd64 and arm64 but can't be on heteregeneous clusters due to missing features.

Should we consider this work part of this PR, or would you open another one for it?

liangxia:
> Hi, as per comment in [openshift/cucushift#9103 (comment)](https://github.com/openshift/cucushift/pull/9103#issuecomment-1107719891) we could need to add another field to the arch option in Polarion, "heterogeneous", to be able disabling cases that are supported for both amd64 and arm64 but can't be on heteregeneous clusters due to missing features.
> 
> Should we consider this work part of this PR, or would you open another one for it?

If we need another field in Polarion, then we need to finish that in Polarion first, as the query (and adding tags) rely on Polarion.
Also that's another tag to be added, we can do it in a follow up PR.

aleskandro:
Fair enough. Thanks @liangxia 

/lgtm

aleskandro:
Refers OCPQE-9784

PR #2814, opened 2022-04-25
Sync upgrade tags from check to prepare for sample operator

liangxia:
Fixed a bug in the sync-tags script, which previously use `/` for address part in command sed, now replaced with `#`.
Since the scenario line contains multiple `/` which breaks the script, the scenario is missed at previous batch tag sync.

/cc @xiuwang @jhou1 @JianLi-RH @dis016 @pruan-rht


dis016:
LGTM

PR #2816, opened 2022-04-25
Remove special char from Scenario name and sync tags

liangxia:
The `[]` are special character in regexp, remove them from the scenario name so that  several tools rely on regexp works.

Will sync the scenario name change back to Polarion after PR merge.

/cc @ropatil010 @duanwei33 @kasturinarra @jhou1 @JianLi-RH @dis016 @pruan-rht


dis016:
LGTM

duanwei33:
Acked, thanks.

liangxia:
Polarion automation field are updated, too.

PR #2822, opened 2022-04-26
update rules for OCP-20932 and OCP-29198

yapei:


yapei:
4.11 pass log Runner/405292/console

PR #2827, opened 2022-04-28
Fix multiple failures in Prow CI

liangxia:
/cc @jhou1  @xiuwang @anpingli @zhouying7780 @kasturinarra


jhou1:
Thanks for addressing the fixes quickly @liangxia 

/lgtm
/approve

PR #2831, opened 2022-04-29
Adding another step to validate Nodeport ETP=local properly

anuragthehatter:
@zhaozhanqi PTAL. Added one mroe step to make this feature check properly as per https://github.com/openshift/ovn-kubernetes/pull/1063
The changes fails on 4.10 build but passes on cluster bot build with referenced PR


anuragthehatter:
Added comments as well for understanding

anuragthehatter:
We probably need to add this in IPv6 case as well or dual stack if exists

PR #2832, opened 2022-05-05
Adding some wait time so that test case does not fail

kasturinarra:
Currently i sometimes see that the test fail because shardB or shardA is not available. Adding in some wait time to make sure that test does not fail & also replacing docker_image with image as it has been deprecated.


kasturinarra:

https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/416812/console

PR #2835, opened 2022-05-07
skip scenario when it's not match the cluster

zhaozhanqi:


zhaozhanqi:
@openshift/team-sdn-qe ^^
@liangxia PTAL, thanks

zhaozhanqi:
@anuragthehatter ^^ PTAL, thanks

zhaozhanqi:
Add one step, if the cluster is migration from sdn to ovn or rollback,  then skip this cases

logs: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4392/console

zhaozhanqi:
@huiran0826 @weliang1 ^^

features/upgrade/sdn/egress-upgrade.feature
```diff
@@ -261,6 +263,7 @@ Feature: Egress compoment upgrade testing
   @proxy @noproxy @disconnected @connected
   @arm64 @amd64
   Scenario: Check sdn egressip is functional post upgrade
+    Given the env is using "OpenShiftSDN" networkType
```
weliang1, line 266: Should post upgrade checking be "Given the cluster is not migration from sdn plugin"?

zhaozhanqi: yes, you are right. I will add the step.
 I thought we can check this feature after sdn-migration-rollback.  but seems this resource will be deleted after migration to OVN.  So even roll back to openshift-sdn. the resource will not be back.

zhaozhanqi: added this step for unsupported features including egressip/egressfirewall/multicast. etc.

weliang1: > yes, you are right. I will add the step. I thought we can check this feature after sdn-migration-rollback. but seems this resource will be deleted after migration to OVN. So even roll back to openshift-sdn. the resource will not be back.

"Given the cluster is not migration from sdn plugin" is not used for "post upgrade - prepare".
Should line 267 be replaced by "Given the cluster is not migration from sdn plugin" and remove "Given the cluster is not migration from sdn plugin" line in 193?

zhaozhanqi: yes..good catch. updated. thanks.

PR #2837, opened 2022-05-09
fix OCPQE-7772

QiaolingTang:
To fix https://issues.redhat.com/browse/OCPQE-7772, do below steps to create an index pattern:
1. running Javascript to set the index pattern name, for example: `app`
2. using `send_keys` to add `*` to the index pattern name, then the index pattern name will be `app*`
The reason why don't use Javascript to set the index pattern to `app*` is: if set index pattern name to `app*`, the button `Next step` is not enabled, then the step will fail. Adding `*` by `send_keys` will enable this button.

### DIFF
--- a/features/logging/kibana.feature
+++ b/features/logging/kibana.feature
@@ -88,7 +88,7 @@ Feature: Kibana related features
     Given I switch to the first user
     When I login to kibana logging web console
     Then the step should succeed
-    Given I have index pattern "*app"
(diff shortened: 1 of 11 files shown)


QiaolingTang:
Log: https://url.corp.redhat.com/2443ac8

QiaolingTang:
Hi @yapei @yanpzhan , could you please help to review this PR when you get time?

PR #2839, opened 2022-05-09
Changing the mode type log check behavior for ipsec 

anuragthehatter:
Probably a bad idea to check libreswan logs which is managed mostly upstream. We can rely on `ip transform state /ip x s` utlity to check ipsec mode 
Logs: https://privatebin.corp.redhat.com/?d68fdddf02f8db32#EyVVWiU7jUfGTWDERXeb3YwVYzZyhhVc1CVZrvTFNT6N

Small fix


anuragthehatter:
cc @openshift/team-sdn-qe

PR #2841, opened 2022-05-10
fix OCP-25816,OCP-27586 rules

yapei:


yapei:
4.11 pass log /Runner/422829/

PR #2850, opened 2022-05-13
Fix apiVersion

liangxia:


dis016:
LGTM

pruan-rht:
@liangxia curious, why is it necessary to remove the `"`?  I think for XML, it should work either way.

liangxia:
> @liangxia curious, why is it necessary to remove the `"`? I think for XML, it should work either way.

Those test files set an bad example of mixture use of json/yaml in one test file. It makes future batch update more difficult.

PR #2854, opened 2022-05-18
fix OCP-30084

QiaolingTang:
Fix failure:
```
And the expression should be true> cb.index_data['docs.count'] > "0"
no implicit conversion of String into Integer (TypeError)
```

### DIFF
--- a/features/step_definitions/logging_metrics.rb
+++ b/features/step_definitions/logging_metrics.rb
@@ -43,8 +43,11 @@ When /^I wait(?: (\d+) seconds)? for the #{QUOTED} index to appear in the ES pod
   seconds ||= 10 * 60
   count = 0
   success = wait_for(seconds) {
-    step %Q/I get the "#{index_name}" logging index information from a pod with labels "#{labels}"/
-    res = cb.index_data
+    step %Q/I perform the HTTP request on the ES pod with labels "#{labels}":/, table(%{
+      | relative_url | _cat/indices?format=JSON |
+      | op           | GET                      |
+    })
+    res = @result[:parsed].select {|e| e['index'].start_with? index_name}
       # exit only health is not 'red' and index is 'open' and the docs.count > 0
       # XXX note, to be more correct, we should check that the index is not red
       # for an extended persiod.  The tricky part is how to define extended period????
@@ -56,6 +59,7 @@ When /^I wait(?: (\d+) seconds)? for the #{QUOTED} index to appear in the ES pod
     end
     count > 0
   }
+  cb.docs_count = count
   raise "Index '#{index_name}' failed to appear in #{seconds} seconds" unless success
 end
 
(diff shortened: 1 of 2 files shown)


QiaolingTang:
Logs: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/433217/console

/cc @anpingli

anpingli (commented):
/lgtm

features/step_definitions/logging_metrics.rb
```diff
@@ -75,7 +79,7 @@
     | relative_url | _cat/indices?format=JSON |
     | op           | GET                      |
   })
-  res = @result[:parsed].select {|e| e['index'].start_with? index_name}
+  res = @result[:parsed].find {|e| e['index'].start_with? index_name}
```
anpingli, line 82: @QiaolingTang  How about count the doc.count by the index by name as above. For example:
#index_name="app"
index_data={"health":"green","status":"open","index":index_name,"uuid":"G4naB-r4SQKG5dwrrWtfxA","pri":"1","rep":"0","docs.count":"0","docs.deleted":"0","store.size":"0b","pri.store.size":"0b"}
doc_count=0
doc_deleted=0
res=result.select {|e| e[:index].start_with?"app"}
#sort by index name to make use some values(uuid,health) of the last index.
res.sort{|a,b| a[:index] <=> b[:index]}.each do | i |
  #if(i["health"] == "yellow" && index_data["health" != "red" ];
  #   index_data["health"] = "yellow"
  #end
  #if(i["health"] == "red") :
  #   index_data["health"] = "red"
  #end

  doc_count += i[:'docs.count'].to_i
  doc_deleted += i[:'docs.deleted'].to_i

  index_data[:status] = i[:status]
  index_data[:uuid] = i[:uuid]
  index_data[:pri] = i[:pri]
  index_data[:rep] = i[:rep]
  index_data[:'docs.count'] = doc_count.to_s
  index_data[:'docs.deleted'] = doc_deleted.to_s
  index_data[:'store.size'] = i[:'store.size']
  index_data[:'pri.store.size'] = i[:'pri.store.size']
end

QiaolingTang: @anpingli I think we use this step to get an index's information, especially the primary/replica shard number. If we want to get the docs count of a group indices with same prefix, we can use step `I wait(?: (\d+) seconds)? for the #{QUOTED} index to appear in the ES pod(?: with labels #{QUOTED})`

QiaolingTang: For step `I wait for the "xxxx" index to appear in the ES pod with labels "es-node-master=true"`, the content of `res` is an array of indices, that's why we got `no implicit conversion of String into Integer (TypeError)`. Using `select` will return all the objects match the condition, `find` will only return one.

anpingli (commented):
/lgtm

features/step_definitions/logging_metrics.rb
anpingli: Ok.

anpingli: /hold cancel

PR #2858, opened 2022-05-23
Replace image with app_repo

kasturinarra:
I see that replacing it with image does not work with versions less than 4.11, so replacing it with app_repo so that it works with all versions of ocp


kasturinarra:

Logs: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/441354/console (4.8)
Logs: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/434125/parameters/ (4.11)

PR #2860, opened 2022-05-23
OVN: use ip6tables for IPv6 addresses

rbbratta:
For split brain we DROP packes from peer master, but for IPv6
addresses we need to use `ip6tables` command


rbbratta:

http://file.rdu.redhat.com/~rbrattai/logs/cucushift-2022-05-23-17-08-33.log

anuragthehatter:
/lgtm
Thanks @rbbratta

PR #2861, opened 2022-05-24
fix OCP-10770

huiran0826:
Fix  OCP-10770 which failed on ipv6 single stack cluster.

/cc @zhaozhanqi @anuragthehatter


huiran0826:
close it in favour of  https://github.com/openshift/verification-tests/pull/2862

PR #2862, opened 2022-05-24
IPv6 improvements for nodeport and hard delay for gw migration

anuragthehatter:
- Nodeport cases compatibility with IPv6 single stack (we discussed it on slack last week). This is failing on single stack IPv6

- 30 seconds hard wait added in gateway migration tests (comment added on why)

- A step is added into nodeport scenario to align with https://github.com/openshift/verification-tests/blob/master/features/networking/service.feature#L799. It was added an an improvement post customer bug so keeping aligned in both scenarios

@openshift/team-sdn-qe PTAL

### DIFF
--- a/features/networking/service.feature
+++ b/features/networking/service.feature
@@ -786,23 +786,23 @@ Feature: Service related networking scenarios
     Given I use the "<%= cb.masters[1].name %>" node
     #It should work because its external traffic from another node and destination node has a backend pod on it (ETP=local respected)
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.hostip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 [<%= cb.hostip %>]:<%= cb.port %>|
     Then the step should succeed
     And the output should contain:
       | Hello OpenShift! |
     #It should NOT work because its external traffic from another node and destination node DOES NOT have a backend pod on it (ETP=local respected)
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.master0_ip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 [<%= cb.master0_ip %>]:<%= cb.port %> |
     And the output should contain:
       | Connection refused |
     #It should work like ETP=cluster because its not external traffic, its within the node (ETP=local shouldn't be respected and its like ETP=cluster behaviour) 
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.master1_ip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 [<%= cb.master1_ip %>]:<%= cb.port %> |
     And the output should contain:
       | Hello OpenShift! |
     Given I ensure "hello-pod" service is deleted
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.hostip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 [<%= cb.hostip %>]:<%= cb.port %> |
     And the output should contain:
       | Connection refused |
 
@@ -834,17 +834,17 @@ Feature: Service related networking scenarios
 
     Given I use the "<%= cb.masters[1].name %>" node
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.hostip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 [<%= cb.hostip %>]:<%= cb.port %> |
     Then the step should succeed
     And the output should contain:
       | Hello OpenShift! |
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.master0_ip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 [<%= cb.master0_ip %>]:<%= cb.port %> |
     Then the step should succeed 
     And the output should contain:
       | Hello OpenShift! |
     Given I ensure "hello-pod" service is deleted
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.hostip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 [<%= cb.hostip %>]:<%= cb.port %> |
     And the output should contain:
       | Connection refused |      
--- a/features/networking/sgw-lgw-migration.feature
+++ b/features/networking/sgw-lgw-migration.feature
@@ -18,6 +18,7 @@ Feature: SGW<->LGW migration related scenarios
     #OCP-47087 - [bug_1903408]Other node cannot be accessed for nodePort when externalTrafficPolicy is Local	
     Given I store the masters in the :masters clipboard
     And the Internal IP of node "<%= cb.masters[0].name %>" is stored in the :master0_ip clipboard
+    And the Internal IP of node "<%= cb.masters[1].name %>" is stored in the :master1_ip clipboard
     And evaluation of `rand(30000..32767)` is stored in the :port clipboard
     Given I have a project
     And evaluation of `project.name` is stored in the :proj1 clipboard
@@ -97,6 +98,11 @@ Feature: SGW<->LGW migration related scenarios
       | curl --connect-timeout 5 <%= cb.master0_ip %>:<%= cb.port %> |
     Then the step should fail
     And the output should not contain:
+      | Hello OpenShift! |
+    When I run commands on the host:
+      | curl --connect-timeout 5 <%= cb.master1_ip %>:<%= cb.port %> |
+    Then the step should succeed
+    And the output should contain:
       | Hello OpenShift! |
     Given I use the "<%= cb.proj1 %>" project
     Given I ensure "hello-pod" service is deleted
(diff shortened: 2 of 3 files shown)


features/networking/service.feature
```diff
@@ -786,18 +786,18 @@ Feature: Service related networking scenarios
     Given I use the "<%= cb.masters[1].name %>" node
     #It should work because its external traffic from another node and destination node has a backend pod on it (ETP=local respected)
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.hostip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 <%= cb.hostip %>:<%= cb.port %> || curl --connect-timeout 5 [<%= cb.hostip %>]:<%= cb.port %>|
```
huiran0826, line 789: Seems || in middle needs to be \|\|

features/networking/service.feature
```diff
@@ -786,18 +786,18 @@ Feature: Service related networking scenarios
...
     Then the step should succeed
     And the output should contain:
       | Hello OpenShift! |
     #It should NOT work because its external traffic from another node and destination node DOES NOT have a backend pod on it (ETP=local respected)
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.master0_ip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 <%= cb.master0_ip %>:<%= cb.port %> || curl --connect-timeout 5 [<%= cb.master0_ip %>]:<%= cb.port %> |
```
huiran0826, line 795: here cb.master0_ip  is still ipv4(it will be 127.0.0.1 in ipv6 single cluster), needs to get ipv6 in previous step, same as below.
```
 And the Internal IPv6 of node "<%= cb.masters[0].name %>" is stored in the :master0_ipv6 clipboard
```

huiran0826:
Better to get a pass log in ipv6 single stack cluster.  I debugged nodeport on ipv6 single cluster  a little bit yesterday, still didn't get it work even manually ....

features/networking/service.feature
zhaozhanqi: yes, 127.0.0.1 is not expected. 

I think we need another way to get the master hostip, maybe we can use api pod which scheduled on all masters and get the api_pod.hostip ?

features/networking/service.feature
anuragthehatter: Doesn;'t seem to be an issue though working fine but geez, its fetching IPv4..
```
When I run commands on the host:                                                               # features/step_definitions/node.rb:98
      | curl --connect-timeout 5 <%= cb.master1_ip %>:<%= cb.port %> || curl --connect-timeout 5 [<%= cb.master1_ip %>]:<%= cb.port %> |
      [19:54:32] INFO> Shell Commands: http_proxy=http://proxy-user1:JYgU8qRZV4DY4PXJbxJK@10.0.77.69:3128
      https_proxy=http://proxy-user1:JYgU8qRZV4DY4PXJbxJK@10.0.77.69:3128
      oc debug  --kubeconfig=/home/anusaxen/workdir/anusaxen-anusaxenx/ocp4_admin.kubeconfig -n prj-anusaxen-anusaxenx-984e --image=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:3c52f2e03ba32621265ca94b29022393ff684bc4f7b52f7059550f2107d68065 node/anurag-n6-zzlpj-master-1 -- chroot /host/ bash -c cd\ \'/tmp/workdir/anusaxen-anusaxenx\''
      'curl\ --connect-timeout\ 5\ 10.0.130.84:32340'
      ''
      'curl\ --connect-timeout\ 5\ \[10.0.130.84\]:32340
        % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                       Dload  Upload   Total   Spent    Left  Speed
100    17  100    17    0     0   2833      0 --:--:-- --:--:-- --:--:--  4250
      Hello OpenShift!
        % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                       Dload  Upload   Total   Spent    Left  Speed
100    17  100    17    0     0   8500      0 --:--:-- --:--:-- --:--:--  8500
      Hello OpenShift!
```

features/networking/service.feature
anuragthehatter: Ahh yea. I never noticed when we changed our internal ip node step def to differentiate between Ipv6 or Ipv4. We probably could have just gotten ```status.addresses[0]``` array value I believe which could have worked on both clusters probably. Looking at step def , I agree IPv6 needs to be fetched separately for master 0...Will try to play with it tomorrow and come up with solution.

anuragthehatter:
Probably adding ```Given Internal IPv6``` step in a testcase will make it specific to IPv6 or dual stack only. We only have one real dual stack usecase in this repo so not worried, as we moved on to cover that in GO. This logic mainly focused on single stack clusters either IPv4/IPv6 

1. No need of double piping operators now in curls
2. ```Given Internal IP``` step will return IPv4 or IPv6 depending on cluster
3. If we need to make any case compatible with IPv6 and IPv4 both, we just add square braces in script. Square braces will work for both IPv6/IPv4 return addresses. Hence it won't break existing IPv4 cases and if those cases are using node IPs, they can be upgraded to run on IPv6 clusters as well using just braces
4. I haven't introduced IP family precedence check for dual stack in this logic as there is only one case and we are now on GO for that.But anyways our dual clusters are coming by default on IPv4,Ipv6 so its should be fine. 

This logic has been verified on
IPv6 single stack - https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4427/console
IPv4 single stack (SDN) - https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4426/console
IPv4 OVN - https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4425/console
Dual stack (we only have one case related to node anyway) -  https://privatebin.corp.redhat.com/?194539d08d46cb26#CvJr2TN6fdCwgYBrXiB8LEZyZgBmqUQqNB1pcpKgmQX2

features/networking/service.feature
```diff
@@ -786,23 +786,23 @@ Feature: Service related networking scenarios
     Given I use the "<%= cb.masters[1].name %>" node
     #It should work because its external traffic from another node and destination node has a backend pod on it (ETP=local respected)
     When I run commands on the host:
-      | curl --connect-timeout 5 <%= cb.hostip %>:<%= cb.port %> |
+      | curl --connect-timeout 5 [<%= cb.hostip %>]:<%= cb.port %>|
```
anuragthehatter, line 789: Also made `externalTrafficPolicy Local` case IPv6 compatible by just including the braces.

huiran0826:
Thanks @anuragthehatter  fixing nodeport cases!
/lgtm

features/networking/service.feature
zhaozhanqi: ah, good to know braces can work on both v4 and v6

features/networking/sgw-lgw-migration.feature
```diff
@@ -98,6 +99,11 @@ Feature: SGW<->LGW migration related scenarios
     Then the step should fail
     And the output should not contain:
       | Hello OpenShift! |
+    When I run commands on the host:
+      | curl --connect-timeout 5 <%= cb.master1_ip %>:<%= cb.port %> |
```
zhaozhanqi, line 103: do we need to update this ip with braces?

anuragthehatter: @zhaozhanqi I planned do that in separate PR if you don't mind, that will need good testing. Going to scan other node related cases as well and plans to make them compatible with IPv6.

zhaozhanqi: yes, sure.

features/step_definitions/networking.rb
```diff
@@ -994,28 +996,26 @@
...
-    # use -4 to limit output to just `default` interface, fixed in later iproute2 versions
     if v6
        step %Q/I run command on the node's ovnkube pod:/, table("| ip | -6 | route | show | default |")
     else
-       step %Q/I run command on the node's ovnkube pod:/, table("| ip | -4 | route | show | default |")
+       inf_address = admin.cli_exec(:get, resource: "node/#{node_name}", output: "jsonpath={.status.addresses[0].address}")
     end
   when "OpenShiftSDN"
-    step %Q/I run command on the node's sdn pod:/, table("| ip | -4 | route | show | default |")
+       inf_address = admin.cli_exec(:get, resource: "node/#{node_name}", output: "jsonpath={.status.addresses[0].address}")
```
liangxia, line 1005: Indention issue ?
Also the indention are not consistent. Some of them use 2 whitespace, and some of them use 3, and some of them use 4.

anuragthehatter: Thanks. Just fixed that in vscode

features/step_definitions/networking.rb
```diff
@@ -1480,6 +1480,9 @@
     @result = admin.cli_exec(:patch, resource: "network.operator", resource_name: "cluster", p: "{\"spec\":{\"defaultNetwork\":{\"ovnKubernetesConfig\":{\"gatewayConfig\":{\"routingViaHost\": true}}}}}", type: "merge")
   end
   raise "Failed to patch network operator for gateway mode" unless @result[:success]
+  # Sometimes CNO starts rolling out changes after few seconds have passed (gap). During that gap, if we check rollout status it would always be a pass from previous rollout result.Its not a bug.
+  # Keeping a 30 seconds harcoded time before we make following instruction (oc rollout status) worthy
+  step "30 seconds have passed"
```
liangxia, line 1485: `oc patch` will cause creation of new pod and deletion of old pod.
So instead of hard wait (wait time too short can not guarantee the rollout, while wait time too long cost us extra time), we can get the pod's name before `oc patch`, and verify the old pod is gone after `oc patch`. Thus way, the wait time is dynamic and ensures the rollout.

PR #2865, opened 2022-05-25
try not to raise Exceptions to avoid premature exit of long testrun t…

pruan-rht:
…o allow tests to continue instead of exiting

Tried with two separate run 20220524-1826 & 20220524-2345 and get the same number of waiting cases w/o seeing premature exit.  We can either wait for more sample runs or merge as is.

### DIFF
--- a/lib/polarshift/test_record.rb
+++ b/lib/polarshift/test_record.rb
@@ -173,9 +173,11 @@ module BushSlicer
       def start_scenario_for!(test_case)
         scenario = self.test_case.scenarios.find { |s| s.match! test_case}
         unless scenario
-          raise "logic error: trying to start Cucumber scenario not part of this test record"
+          puts "logic error: trying to start Cucumber scenario not part of this test record\n"
+          puts "Skipping testcase #{test_case}\n"
+        else
+          scenario.start!
         end
-        scenario.start!
       end
 
       # @param test_case [Cucumber::Core::Test::Case, Cucumber::Events::TestRunFinished]
--- a/lib/polarshift/test_record.rb
+++ b/lib/polarshift/test_record.rb
@@ -173,8 +173,8 @@ module BushSlicer
       def start_scenario_for!(test_case)
         scenario = self.test_case.scenarios.find { |s| s.match! test_case}
         unless scenario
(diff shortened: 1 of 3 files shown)


lib/polarshift/test_record.rb
```diff
@@ -173,9 +173,11 @@ def update_to!(new_state)
       def start_scenario_for!(test_case)
         scenario = self.test_case.scenarios.find { |s| s.match! test_case}
         unless scenario
-          raise "logic error: trying to start Cucumber scenario not part of this test record"
+          puts "logic error: trying to start Cucumber scenario not part of this test record\n"
```
jhou1, line 176: Shall we use `logger.warn` instead of `puts`?

lib/polarshift/test_suite.rb
```diff
@@ -108,7 +108,8 @@ def commit!(test_case)
         if self.current_test_record.start_scenario_for! test_case
           return true
         else
-          raise "logic error: starting test case that is not part of current test record"
+          puts "logic error: starting test case that is not part of current test record\n"
```
jhou1, line 111: We can add the test case id here, for info and reference in failure analysis

pruan-rht: done

lib/polarshift/test_record.rb
pruan-rht: done...

lib/polarshift/test_suite.rb
```diff
@@ -108,7 +108,8 @@ def commit!(test_case)
         if self.current_test_record.start_scenario_for! test_case
           return true
         else
-          raise "logic error: starting test case that is not part of current test record"
+          puts "logic error: starting test case '#{test_case}' that is not part of current test record\n"
```
jhou1, line 111: @pruan-rht should the `puts` also be replaced with `logger`?

pruan-rht: @jhou1 there is no logger defined in the file

pruan-rht: we can add the logger class at a later time

jhou1: OK, thanks

PR #2867, opened 2022-05-27
IPv6 coverage for couple of more node related cases

anuragthehatter:
Noticed pod hosport image doesn't had our right sdn image path linked. Fixed that. That is being used in OCP-25294 which only had versions coverage until 4.4, strange. I manually added later versions to that. And verified it works.

Make couple of more nodeport case IPv6/IPv4 compatible with braces logic.

PTAL @openshift/team-sdn-qe 
Logs: https://privatebin.corp.redhat.com/?d68737f2c8635330#HwK5zX8MNmF1vEWYKjd53qGYjonvsxXECFXaP5rV2cb9

### DIFF
--- a/features/networking/pod.feature
+++ b/features/networking/pod.feature
@@ -445,13 +445,13 @@ Feature: Pod related networking scenarios
     #Pod should be accesible via node ip and host port from its home node
     Given I use the "<%= cb.workers[0].name %>" node
     And I run commands on the host:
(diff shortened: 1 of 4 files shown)


anuragthehatter:
@zhaozhanqi PTAL as you test is also involved here. Thanks

zhaozhanqi:
Thanks @anuragthehatter 
/lgtm
/approve

PR #2871, opened 2022-05-30
update machine related sources to use full name

sunzhaohua2:
Tested on aws, seems no new issues were found. @jhou1 @miyadav @huali9 PTAL
Enable CAPI regression run 20220526-0954
Disable CAPI regression run 20220530-0605


huali9:
LGTM
Thanks Zhaohua.

PR #2872, opened 2022-05-30
Add right apiversion for template

kasturinarra:
Add right apiversion for template kind


kasturinarra:
@liangxia could you please help review and merge ? thanks !!

PR #2873, opened 2022-05-30
Implementation of ovirt modules

santos1709:


pruan-rht:
/lgtm
verified by running it against the branch with Runner job changed to install bundler 2.1.x
@jhou1 @liangxia PTAL and merge if you have no objections.  Thanks!

PR #2887, opened 2022-06-08
Add case ID to scenario name

liangxia:
Just found that I forgot to add ID to scenario names for verification-tests, no wonder why I see several failure tests in Prow CI does not contain ID in the name.

The PR is ready for review.

/hold
so that the PR merge at a proper time (maybe this Friday, which is Day of Learning) to minimize the impact to the team.


liangxia:
/unhold

Given Jenkins outage, the tests are blocked.

PR #2902, opened 2022-06-15
set default pvc size to 20Gi

QiaolingTang:
Set default pvc size to 20Gi as it's the minimum size allowed in alibaba cloud. 
Skip the not ready pods when checking logs.

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -235,7 +235,6 @@ Given /^I wait until ES cluster is ready$/ do
   step %Q/#{cluster_logging('instance').logstore_node_count.to_i} pods become ready with labels:/, table(%{
     | cluster-name=elasticsearch,component=elasticsearch |
   })
-  cluster_logging('instance').wait_until_es_is_ready
 end
 
 Given /^I wait until kibana is ready$/ do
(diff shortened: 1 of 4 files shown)


QiaolingTang:
Logs:  
OCP 4.10: https://url.corp.redhat.com/a712d3b
OCP 4.11: https://url.corp.redhat.com/3281e57

anpingli (commented):
/lgtm

anpingli (commented):
/lgtm

features/step_definitions/logging.rb
```diff
@@ -1307,24 +1306,26 @@
...
     error_logs = []
-    logs.split("\n").each do | log |
-      if !(exceptions.nil?)
-        ignore = false
-        exceptions.each do | exception |
-          if log.include? exception
-            ignore = true
-            break
+    unless logs.empty? || logs.nil?
+      logs.split("\n").each do | log |
```
liangxia, line 1310: The step `I record all pods logs in the "openshift-operators-redhat" project` generates a very long log list, we need a utilize for it.

liangxia: It would be better go and check the must gather logs, rather than print all of them in the CI.

QiaolingTang: Logging is not deployed along with OCP, must-gather won't collect logging's resources.

liangxia: Would you mind to log an jira issue to me, so that I'll follow up to add logging logs along side (not as part of) with must gather logs.

QiaolingTang: I'll remove the step from the scenario.

QiaolingTang: Done

PR #2901, opened 2022-06-15
update OCP-12929

huiran0826:
Update this case to cover one more test point that the multicast pods locate on different nodes.

Test log:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4484/console

/cc @openshift/team-sdn-qe @jechen0648

PR #2909, opened 2022-06-20
Cover 45022

huiran0826:
OCP-45022 tests the multicast pods function across the nodes. 
Add some steps to OCP-12926 and cover this function.

Meanwhile, revert the codes in https://github.com/openshift/verification-tests/pull/2901, reasons please see https://github.com/openshift/verification-tests/pull/2901#issuecomment-1160083865

Testlogs:
OCP-12926 -- https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4487/console
OCP-12929 -- https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/4488/console

@weliang1 @jechen0648  Could you take a look? Thanks a lot!

### DIFF
--- a/features/networking/multicast.feature
+++ b/features/networking/multicast.feature
@@ -13,18 +13,28 @@ Feature: testing multicast scenarios
   @heterogeneous @arm64 @amd64
   Scenario: OCP-12926 pods should be able to subscribe send and receive multicast traffic
     # create some multicast testing pods
+    Given I store the schedulable workers in the :nodes clipboard
     Given I have a project
     And evaluation of `project.name` is stored in the :proj1 clipboard
     Given I obtain test data file "networking/multicast-rc.json"
-    When I run the :create client command with:
-      | f | multicast-rc.json |
+    When I run oc create over "multicast-rc.json" replacing paths:
+      | ["spec"]["replicas"] | 2                       |
+      | ["spec"]["nodeName"] | <%= cb.nodes[0].name %> |
     Then the step should succeed
-    Given 3 pods become ready with labels:
+    Given 2 pods become ready with labels:
       | name=mcast-pods |
     And evaluation of `pod(0).ip` is stored in the :pod1ip clipboard
     And evaluation of `pod(0).name` is stored in the :pod1 clipboard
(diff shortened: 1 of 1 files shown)


huiran0826:
@weliang1 @jechen0648 Could you take a look? Thanks!

huiran0826:
@jechen0648 Actually not, after this PR merge, I would add one more step in Polarion case OCP-12926 for this change and then inactive case OCP-45022.

features/networking/multicast.feature
```diff
@@ -13,18 +13,28 @@ Feature: testing multicast scenarios
...
   Scenario: OCP-12926 pods should be able to subscribe send and receive multicast traffic
     # create some multicast testing pods
+    Given I store the schedulable workers in the :nodes clipboard
     Given I have a project
     And evaluation of `project.name` is stored in the :proj1 clipboard
     Given I obtain test data file "networking/multicast-rc.json"
-    When I run the :create client command with:
-      | f | multicast-rc.json |
+    When I run oc create over "multicast-rc.json" replacing paths:
+      | ["spec"]["replicas"] | 2                       |
```
weliang1, line 21: @huiran0826 If this step only create two pods  which are in the different nodes, do we still need a case for the pods are in the same node? Other case may covered. To make things simple, keep OCP-12926 as test pods in the same node, and using OCP-45022 to test Multicast traffic cross nodes.  WDYT?

huiran0826: @weliang1  Actually  OCP-12926 will cover both situation, same nodes and different nodes. 
Currently we only have this OCP-45022 for multicast as manual left, now I prefer to just update this case than to create a new case in go which might need a lot of effort and research as we don't have the basic utilizes for this kind of cases.

huiran0826:
@zhaozhanqi @anuragthehatter Would you please take a look?  Thanks!

features/networking/multicast.feature
weliang1: OK. 
/lgtm

PR #2914, opened 2022-06-23
fix sa token issue in monitoring upgrade test

lihongyan1:


lihongyan1:
@liangxia please help approve pr

PR #2920, opened 2022-06-24
OCPQE-10931: skip scenario properly

liangxia:


liangxia:
Tested with Runner/484299

liangxia:
Just in case we might need to revert the changes here (if it cause unexpected failures), take those reliable changes out and put them in https://github.com/openshift/verification-tests/pull/2921

jhou1:
@pruan-rht PTAL.

lib/polarshift/test_suite.rb
```diff
@@ -107,10 +107,8 @@ def commit!(test_case)
 
         if self.current_test_record.start_scenario_for! test_case
           return true
-        else
-          puts "logic error: starting test case '#{test_case}' that is not part of current test record\n"
```
liangxia, line 111: The change here is because we already have those statement info in method `start_scenario_for`
```
        unless scenario
          logger.warn "logic error: trying to start Cucumber scenario not part of this test record"
          logger.warn "Skipping testcase #{test_case}: #{test_case.name}"
        else
          scenario.start!
        end
      end
```

lib/test_case_manager_filter.rb
```diff
@@ -40,9 +40,7 @@ def done
       end
 
       config.on_event :test_case_finished do |event|
-        unless event.result.skipped?
-          tc_manager.signal(:end_case, event)
-        end
+        tc_manager.signal(:end_case, event)
```
liangxia, line 43: Let's assume we have two test cases, caseA and caseB. caseA has `skip_this_scenario` in one of its steps due to conditions can not met (e.g, scenario requires at least two nodes, but we are on SNO, which has only one node).
When starting the execution on caseB after caseA is skipped, Cucumber::Core::Test::Case is set to caseB, and BushSlicer::PolarShift::ScenarioWrapper still points to caseA, which cause logic error in `start_scenario_for`, and caseB is skipped due to the logic error/mismatch.
If we have caseC, caseD, ... followed, it will be skipped, too, due to the same logic error/mismatch.

The change here removed the skip check, so when caseA is skipped, we send a signal to indicate we have reached the end of case of caseA, so Cucumber::Core::Test::Case and BushSlicer::PolarShift::ScenarioWrapper match again.

jhou1:
Thanks for the explanation.
/lgtm
/approve

PR #2929, opened 2022-06-30
Fixing OCP-23967

l1xiangyi:
Adding definition for switching to YAML view when creating Deployments


lib/rules/web/admin_console/4.11/yaml.xyaml
```diff
@@ -56,3 +56,8 @@ click_item_in_breadcrumb:
       xpath: //li[contains(@class,'pf-c-breadcrumb__item')]//button[contains(.,'<breadcrumb_item>')]
     op: click
 
+click_yaml_view_label:
+  element:
+    selector:
+      xpath: //*[@id="content-scrollable"]/section/div/div[2]/div/form/div[1]/div[2]/div/div[2]/div[2]/label
+    op: click
```
yapei, line 63: need to be removed

PR #2928, opened 2022-06-30
OCPQE-8870: wait for project annotations to apply

liangxia:
/cc @kasturinarra @xingxingxia @xiaojiey @qiliRedHat

Is anyone available to help test/verify the codes ?

PR #2933, opened 2022-07-01
OCPQE-8870: wait for project annotation "openshift.io/sa.scc.uid-range"

liangxia:
After https://github.com/openshift/verification-tests/pull/2928 merged, I'm searching the CI logs to verify the fix do works or not.
Then found the issue not only exist for `oc debug`, but also exist for normal operations. E.g,
```
490594/log-[2022-06-29T23:24:15.947Z]     Given I obtain test data file "quota/pod-completed.yaml"                # features/step_definitions/file.rb:1
490594/log-[2022-06-29T23:24:28.091Z]     When I run the :create client command with:                             # features/step_definitions/cli.rb:13
490594/log-[2022-06-29T23:24:28.091Z]       | f | pod-completed.yaml |
490594/log-[2022-06-29T23:24:28.091Z]       [23:24:15] INFO> Shell Commands: oc create -f pod-completed.yaml --kubeconfig=/home/jenkins/ws/workspace/ocp-common/Runner/workdir/ocp4_testuser-28.kubeconfig
490594/log-[2022-06-29T23:24:28.091Z]       
490594/log-[2022-06-29T23:24:28.091Z]       STDERR:
490594/log:[2022-06-29T23:24:28.091Z]       Error from server (Forbidden): error when creating "pod-completed.yaml": pods "podtocomplete" is forbidden: error fetching namespace "etj28": unable to find annotation openshift.io/sa.scc.uid-range
```
```
490625/log-[2022-06-30T00:53:59.053Z]     Given I obtain test data file "networking/iperf_nodeport_service.json"                                   # features/step_definitions/file.rb:1
490625/log-[2022-06-30T00:54:11.299Z]     When I run the :create admin command with:                                                               # features/step_definitions/cli.rb:31
490625/log-[2022-06-30T00:54:11.299Z]       | f | iperf_nodeport_service.json |
490625/log-[2022-06-30T00:54:11.299Z]       [00:53:59] INFO> Shell Commands: oc create -f iperf_nodeport_service.json --kubeconfig=/home/jenkins/ws/workspace/ocp-common/Runner/workdir/ocp4_admin.kubeconfig
490625/log-[2022-06-30T00:54:11.299Z]       service/iperf-server created
490625/log-[2022-06-30T00:54:11.299Z]       
490625/log-[2022-06-30T00:54:11.299Z]       STDERR:
490625/log:[2022-06-30T00:54:11.299Z]       Error from server (Forbidden): pods "iperf-server" is forbidden: error fetching namespace "ss5gy": unable to find annotation openshift.io/sa.scc.uid-range
```
Thus move the hack to a more proper place to fix them all.

/cc @kasturinarra @xingxingxia @xiaojiey @qiliRedHat

### DIFF
--- a/lib/openshift/project.rb
+++ b/lib/openshift/project.rb
@@ -104,6 +104,15 @@ module BushSlicer
       else
         res = by.cli_exec(:new_project, project_name: name, **opts)
       end
+      started_at = Time.now
+      success = wait_for(60, interval: 5) {
+        uid_range = by.cli_exec(:get, resource: "namespace", resource_name: name, template: '{{ index .metadata.annotations "openshift.io/sa.scc.uid-range" }}')
+        !uid_range.nil? && !uid_range.empty? && uid_range.to_s !~ /no value/
+      }
+      wait_since = Time.now - started_at
+      unless success
+        raise "timeout waiting for project #{name} to get annotation openshift.io/sa.scc.uid-range after #{wait_since} seconds, consider reporting a bug"
+      end
       res[:project] = self
       return res
     end
--- a/lib/environment.rb
+++ b/lib/environment.rb
@@ -464,13 +464,6 @@ module BushSlicer
           admin.cli_exec(:annotate, resource: "namespace", resourcename: project_name, keyval: 'openshift.io/node-selector=', overwrite: true)
           # we must update the cache, since we just waited for the previously active project to disappear
           project.reload
-          success = wait_for(60, interval: 5) {
-            uid_range = admin.cli_exec(:get, resource: "namespace", resource_name: project_name, template: '{{ index .metadata.annotations "openshift.io/sa.scc.uid-range" }}')
-            !uid_range.nil? && !uid_range.empty? && uid_range !~ /no value/
-          }
-          unless success
-            raise "timeout waiting for project #{project_name} to get annotation openshift.io/sa.scc.uid-range"
-          end
         end
         @service_project = project
       end


lib/openshift/project.rb
```diff
@@ -104,6 +104,13 @@ def create(by:, **opts)
       else
         res = by.cli_exec(:new_project, project_name: name, **opts)
       end
+      success = wait_for(60, interval: 5) {
+        uid_range = by.cli_exec(:get, resource: "namespace", resource_name: name, template: '{{ index .metadata.annotations "openshift.io/sa.scc.uid-range" }}')
+        !uid_range.nil? && !uid_range.empty? && uid_range !~ /no value/
+      }
+      unless success
+        raise "timeout waiting for project #{name} to get annotation openshift.io/sa.scc.uid-range"
```
xingxingxia, line 112: Given applying wider to every project creation run by every case, if timeout, still not easy to debug whether it is a bug or flaky, debugger does not know whether it needs a few more time, or it never appears. Suggest to increase 60 to bigger for example 120 or more, but record the actual time T it uses, and then assert unless (success or T < 60). Like the method used here https://github.com/openshift/cucushift/pull/8911/files .
And if timeout happens even if increased to bigger, it is better if we can dump some logs, not sure, this needs @kasturinarra to give some insights, seems logs from kube-controller-manager cluster-policy-controller container, can only grep project name from the logs.

liangxia: This is more like a workaround to let the CI move on, rather than a fix, since the issue comes from OCP side.

IMO, we should initially start with a smaller timeout (e.g, 60 seconds, maybe 30 seconds or even less, as generally the project get the annotation in a few seconds) here. Then we observe how it behave from the log history. And then decide whether we should report a bug to OCP, or increase the timeout a little bit.

It's really a bad user experience if a user have to wait for minutes or more to create their pods after the creation of their project.

xingxingxia: Assertion "unless (T < 60)" does not let the CI move on, it is to help debug when it happend to fail due to have to wait for minutes or more to create their pods after the creation of their project

liangxia: `wait_for` means wait up to xx seconds at max. So if you got a success, then the wait time is less or equal to xx seconds. But if you got a fail, then the wait time is larger than xx seconds.

xingxingxia: > But if you got a fail, then the wait time is larger than xx seconds.

This is expected by the design, which intends to help debug by telling the debugger exactly "whether it needs a few more time, or it never appears" when you got a failure.

xingxingxia: I remember seems there was a bug raised but no data then dev closed? I can't find the bug. But seems the issue is sometimes often hit in CI. If not more data shown in CI output, Dev likely wants more data from us again and close again?

liangxia: `timeout waiting for project prj-7e7581d21d83-rootx-49im to get annotation openshift.io/sa.scc.uid-range after 62.102225416 seconds, consider reporting a bug (RuntimeError)`, is it what you would expect it to ?

liangxia:
I'd like to get this fix in before enjoying the weekend

lib/openshift/project.rb
xingxingxia: Thank you for working. Sorry for disturbing weekend enjoying if any :)
Yes it looks good. But "    unless success" should be changed to "    unless wait_since <= 60"
And another point, not sure if using using `stats: stats` and stats[:seconds] lke https://github.com/openshift/verification-tests/blob/master/features/step_definitions/operators.rb#L194 is preferred to get the waiting time.
/hold

xingxingxia: /unhold

xingxingxia: No other issue from my side. Thanks

PR #2934, opened 2022-07-01
fix rules for OCP-28832 and OCP-29198

yapei:


yapei:
4.10 pass log Runner/493294/console

PR #2935, opened 2022-07-04
OCPQE-10970: Explicit return for logic error

liangxia:
/cc @jhou1 @pruan-rht @dis016 @JianLi-RH

### DIFF
--- a/lib/polarshift/test_suite.rb
+++ b/lib/polarshift/test_suite.rb
@@ -109,12 +109,12 @@ module BushSlicer
           return true
         else
           puts "logic error: starting test case '#{test_case}' that is not part of current test record\n"
-          #raise "logic error: starting test case that is not part of current test record"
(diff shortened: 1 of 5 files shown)


lib/polarshift/test_suite.rb
```diff
@@ -109,12 +109,12 @@ def commit!(test_case)
           return true
         else
           puts "logic error: starting test case '#{test_case}' that is not part of current test record\n"
-          #raise "logic error: starting test case that is not part of current test record"
+          return false
```
liangxia, line 112: Initially, we raise an error and quit the execution due to logic error.
And later, as an effort of making less test cases to stay in idle status, we decided to use `puts` instead of `raise`.

The issue is that we already have logic issue there, after put out the error message, we should return rather than proceed on.

PR #2936, opened 2022-07-04
OCPQE-10970: Make msg "scenario not part of this test record" more clear

liangxia:
/cc @jhou1 @pruan-rht @dis016 @JianLi-RH


liangxia:
PR ready for review.

PR #2938, opened 2022-07-05
Output the test_case_finished event to help debug OCPQE-10970

liangxia:
We've tried https://github.com/openshift/verification-tests/pull/2920 to fix the skip scenario issue. But revert the change later as it caused issue when the test case is already reserved.

Before making another fix, I'd like to check those test_case_finished events in detail, especially for the "HTTP POST took 0.036 sec: 409 Conflict" which is hard to simulate manually.

/cc @jhou1 @pruan-rht @JianLi-RH @dis016

### DIFF
--- a/lib/test_case_manager_filter.rb
+++ b/lib/test_case_manager_filter.rb
@@ -40,6 +40,8 @@ module BushSlicer
       end
 
       config.on_event :test_case_finished do |event|
+        puts "test_case_finished: #{event}"
+        puts "#{event.inspect}\n"
         unless event.result.skipped?
           tc_manager.signal(:end_case, event)
         end


lib/test_case_manager_filter.rb
```diff
@@ -40,6 +40,8 @@ def done
       end
 
       config.on_event :test_case_finished do |event|
+        puts "test_case_finished: #{event}"
+        puts "#{event.inspect}\n"
```
liangxia, line 44: It will looks like,
```
07-05 14:30:08.516  test_case_finished: #<Cucumber::Events::TestCaseFinished:0x00000000101bf890>
07-05 14:30:08.516  #<Cucumber::Events::TestCaseFinished:0x00000000101bf890 @test_case=#<Cucumber::Core::Test::Case: features/tierN/web/admin-console/search.feature:14>, @result=#<Cucumber::Core::Test::Result::Passed:0x00000000101bfb88 @duration=#<Cucumber::Core::Test::Result::Duration:0x00000000101bfc00 @nanoseconds=207136595458>>>
```
I'd like to check the output in more situations from CI.

PR #2939, opened 2022-07-05
Calling the step skip_this_scenario is causing issues with CI runs, j…

pruan-rht:
…ust let it fail to isolate the failure

calls to the step skip_this_scenario needs to be re-examined https://github.com/openshift/verification-tests/blob/master/features/step_definitions/networking.rb#L1274.

I proposed that all such call be commented out and just let the testcase fail instead skipping…at least until we know how that step should be handled correctly.  Doing that will isolated the failure to those particular cases only instead of causing premature exits for the rest of the run

As a consequence, there will be `false` failures for testcases that try to run in an environment that does not support the profile, but at least it won't cause `premature exit` for the CI runs

### DIFF
--- a/features/step_definitions/machine.rb
+++ b/features/step_definitions/machine.rb
@@ -4,14 +4,14 @@ Given(/^I have an IPI deployment$/) do
   if machines.length == 0
     logger.warn "Not an IPI deployment, there are no machines"
     logger.warn "We will skip this scenario"
-    skip_this_scenario
+    # skip_this_scenario
   end
 
   machines.each do | machine |
     if machine.node_name.nil?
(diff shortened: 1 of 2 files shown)


pruan-rht:
@mffiedler it did for [this run](http://virt-openshift-05.lab.eng.nay.redhat.com/buildcorp/nightly/50156.html).

jhou1:
Yes, `skip_this_scenario` messed up the polarshift test record causing more scenarios to be skipped. @liangxia has several PRs to address this issue. So let's hold this.
/hold

liangxia:
@pruan-rht   Perhaps we can close this PR now as the skip_this_scenario issue is already fix.  HDYT ?

PR #2943, opened 2022-07-06
check imagestreamtag before using

xiuwang:
[OCP-11805,OCP-11298,OCP-19196,OCP-16495 all passed](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/500340/console)
@zhouying7780 @kasturinarra Please help review, thanks

### DIFF
--- a/features/registry/is.feature
+++ b/features/registry/is.feature
@@ -109,6 +109,7 @@ Feature: Testing imagestream
     When I run the :new_app_as_dc client command with:
       | app_repo | quay.io/openshifttest/ruby-27:multiarch~https://github.com/sclorg/ruby-ex.git |
     Then the step should succeed
+    And I wait for the "ruby-27:multiarch" image_stream_tag to appear
     And the "ruby-ex-1" build was created
     And the "ruby-ex-1" build completes
     When I run the :patch client command with:
@@ -172,6 +173,7 @@ Feature: Testing imagestream
     When I run the :new_app client command with:
       | app_repo | quay.io/openshifttest/ruby-27:multiarch~https://github.com/sclorg/ruby-ex.git |
     Then the step should succeed
+    And I wait for the "ruby-27:multiarch" image_stream_tag to appear
     And the "ruby-ex-1" build was created
     And the "ruby-ex-1" build completes
     And the "ruby-ex:latest" image stream tag was created


features/step_definitions/is.rb
```diff
@@ -40,9 +40,19 @@
...
-  @result = image_stream_tag(istag_name).wait_to_appear(user, 30)
+  org_project = project(generate: false) rescue nil
+
+  if istag_name
+    project_name, is_name = istag_name.split("/", 2)
+    unless is_name
+      is_name = project_name
+      project_name = nil
+    end
+    @result = image_stream_tag(is_name,project(project_name)).wait_to_appear(user, 30)
```
liangxia, line 51: The step name is `created`, but we only `wait_to_appear`.  Then where do we create them ?

xiuwang: @liangxia new-app cmd will create imagestream if it not exist or check the openshift existing imagestream

liangxia: But several of your added steps happens before new-app. Where do we create them then ?

xiuwang: They checked the openshift project imagestreams which are created by samples operator, but this pr need adjust, since samples operator has been an optional operator since 4.11 so it will not install always.
@kasturinarra @zhouying7780  I will remove the update for features/cli/oc_set_deployment_hook.feature file, please update later according to your change.
Thanks @liangxia

features/registry/is.feature
```diff
@@ -109,6 +109,7 @@ Feature: Testing imagestream
     When I run the :new_app_as_dc client command with:
       | app_repo | quay.io/openshifttest/ruby-27:multiarch~https://github.com/sclorg/ruby-ex.git |
     Then the step should succeed
+    And the "ruby-27:multiarch" image stream tag was created
```
liangxia, line 112: How about directly stated that `I wait for the "xxx_name_xxx" xxx_type_xxx to appear` ? There is no step update required in step_definitions in this way.

If you need to specify how long to wait, you can use `I wait for the "xxx_name_xxx" xxx_type_xxx to appear up to xxx seconds`
If you also need to specify the project, you can use `I wait for the "xxx_name_xxx" xxx_type_xxx to appear in the "xxx" project up to xxx seconds`

xiuwang: @liangxia It works, updated, thanks CC @zhouying7780

xiuwang:
[Passed log]( https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/531622/console)

PR #2948, opened 2022-07-11
fix the apiVersion of scc

lyman9966:
now the apiVersion of 'v1' has been deprecated in scc,  so we need to update it to : security.openshift.io/v1


lyman9966:
@sunilcio @weinliu @rhpmali Please help to take a look, thanks!

PR #2950, opened 2022-07-11
added subteam into scenario outline as well

pruan-rht:
for JIRA https://issues.redhat.com/browse/OCPQE-11150


pruan-rht:
The cucushift portion should be coming soon as well.

PR #2951, opened 2022-07-11
gw mode improvements

anuragthehatter:
It was noticed by @pruan-rht and came to my attention that existing logic is dumping huuuge logs for ovn masters during such tests. Probaly because master logs keeps filling up when many tests run on such clusters.

Earlier we didn't have great logic to check local vs shared modes on such clusters but now we have as this is being exposed via api in CNO. Also 4.9 takes a bit more time for operators to reconcile so increases time wait in general

Optimised logic and added impoved 4.9 support as well

@openshift/team-sdn-qe PTAL

Logs excert on 4.9: https://privatebin.corp.redhat.com/?ac87016468c10d4a#8ugYNaYu3K971H3jApBXpzfACGccecxmBkKNkrXByQ7W
Logs excerpt on 4.10+: https://privatebin.corp.redhat.com/?9933f35b394e63b9#9kzLMQgry25RJcAwKc5yKeAqnVCXzGFJsX2ker24Jf1K

No more enormous logging :)


pruan-rht:
@weliang1 @zhaozhanqi can you PTAL?  It looks ok from my end.

PR #2956, opened 2022-07-13
fix OCPQE-11171

QiaolingTang:
Sometimes the ds/collector is not removed automatically after removing clusterlogging, and the step `I wait for the resource "daemonset" named "#{cb.collector_name}" to disappear` causes premature test execution termination. Here add a step to remove the ds if it's not removed automatically.

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -362,6 +362,9 @@ Given /^I delete the clusterlogging instance$/ do
     end
       step %Q/logging collector name is stored in the :collector_name clipboard/
     unless cluster_logging('instance').collection_spec(cached: true).nil?
+      if daemon_set(collector_name).exists?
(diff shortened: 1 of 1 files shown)


features/step_definitions/logging.rb
```diff
@@ -362,6 +362,9 @@
     end
       step %Q/logging collector name is stored in the :collector_name clipboard/
     unless cluster_logging('instance').collection_spec(cached: true).nil?
+      if daemon_set(collector_name).exists?
+        @result = admin.cli_exec(:delete, object_type: 'daemonset', object_name_or_id: collector_name, n: 'openshift-logging')
+      end
```
liangxia, line 367: You can try the step `Given admin ensures "xxx_name_xxx" xxx_resource_type_xxx is delete` to simplify the logic of `check xxx exist, then run the delete command, then wait/check it's gone`

QiaolingTang:
Log:  https://url.corp.redhat.com/a386a3a

anpingli (commented):
/lgtm

PR #2960, opened 2022-07-15
OCPQE-11169: support kubeconfig format from ROSA/OSD

liangxia:
ROSA/OSD does not provide the original kubeconfig. Instead they grant a user with cluster-admin privilege. This cause the kubeconfig provided from ROSA/OSD has a slight difference than we get from OCP.
This PR makes cucushift works for both format.

/cc @yasun1 @yufchang @jhou1 @JianLi-RH @dis016 @pruan-rht


yasun1:
Agree +1

PR #2961, opened 2022-07-15
Fix the template name

kasturinarra:
Fixing the test to use template instead of file as i see that with file the test always fails.


kasturinarra:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/513704/console

PR #2963, opened 2022-07-18
avoid use openshift imagestreams and templates for case 11805/31246/3…

zhouying7780:
…1245/31241

### DIFF
--- a/features/cli/oc_set_deployment_hook.feature
+++ b/features/cli/oc_set_deployment_hook.feature
@@ -13,80 +13,83 @@ Feature: set deployment-hook/build-hook with CLI
   @heterogeneous @arm64 @amd64
   Scenario: OCP-11805:Workloads Set pre/mid/post deployment hooks on deployment config via oc set deployment-hook
     Given I have a project
-    When I run the :new_app client command with:
-      | template | rails-postgresql-example |
+    When I run the :create_deploymentconfig client command with:
+      | image | quay.io/openshifttest/hello-openshift:multiarch |
+      | name  | hello-openshift                                 |
     Then the step should succeed
     Given a pod becomes ready with labels:
-      | deployment=rails-postgresql-example-1 |
+      | deploymentconfig=hello-openshift |
     When I run the :set_deployment_hook client command with:
-      | deploymentconfig | dc/rails-postgresql-example |
-      | pre              |                             |
-      | c                | rails-postgresql-example    |
-      | e                | FOO1=BAR1                   |
-      | failure_policy   | retry                       |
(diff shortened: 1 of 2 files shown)


zhouying7780:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/518684/console

features/cli/oc_set_deployment_hook.feature
```diff
@@ -13,80 +13,83 @@ Feature: set deployment-hook/build-hook with CLI
   @heterogeneous @arm64 @amd64
   Scenario: OCP-11805:Workloads Set pre/mid/post deployment hooks on deployment config via oc set deployment-hook
     Given I have a project
-    When I run the :new_app client command with:
-      | template | rails-postgresql-example |
+    When I run the :create_deploymentconfig client command with:
+      | image | quay.io/openshifttest/hello-openshift:multiarch |
```
kasturinarra, line 17: @zhouying7780 if we are changing this to have hello-openshift image then we need to make sure "oc new-app rails-postgres-example' command works fine because i remember there was a bug from a customer about this not working since we show this as an example when creating new project in oc. thanks !!

zhouying7780: @kasturinarra  firstly this is quickstart feature  , it's not belongs to workloads team's feature , second  I checked the related case and automation script both were set with inactive , so I think we replace it should be okay.

kasturinarra: okay, thanks !!

zhouying7780:
@liangxia Please help review and merge . thanks .

features/cli/oc_set_deployment_hook.feature
```diff
@@ -13,80 +13,83 @@ Feature: set deployment-hook/build-hook with CLI
   @heterogeneous @arm64 @amd64
   Scenario: OCP-11805:Workloads Set pre/mid/post deployment hooks on deployment config via oc set deployment-hook
     Given I have a project
-    When I run the :new_app client command with:
-      | template | rails-postgresql-example |
+    When I run the :create_deploymentconfig client command with:
```
liangxia, line 16: Deployment is preferred than deploymentconfig since 4.5 (or 4.6). Perhaps we should use deployment here.

zhouying7780: @liangxia deploy does not support pre/mid/post hooks, this feature is only for DC

liangxia: Ack

PR #2965, opened 2022-07-18
adding network pod deletion step to cover bz 

anuragthehatter:
Logs:
http://pastebin.test.redhat.com/1065698
http://pastebin.test.redhat.com/1065699

We cover nodeport ETP local scenario but needs to add network pod deletion steps to make sure snat rules creates again and test passes. Reproted in BZ https://bugzilla.redhat.com/show_bug.cgi?id=2107903
The bug was on LGW but a good idea to cover on SGW as well
@openshift/team-sdn-qe PTAL

### DIFF
--- a/features/networking/service.feature
+++ b/features/networking/service.feature
@@ -759,6 +759,7 @@ Feature: Service related networking scenarios
   # @case_id OCP-47087
   @4.12 @4.11 @4.10
   @admin
+  @destructive	
   @vsphere-ipi @openstack-ipi @gcp-ipi @baremetal-ipi @azure-ipi @aws-ipi
   @vsphere-upi @openstack-upi @gcp-upi @baremetal-upi @azure-upi @aws-upi
   @proxy @noproxy @connected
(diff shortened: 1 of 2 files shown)


features/networking/service.feature
```diff
@@ -800,11 +802,33 @@ Feature: Service related networking scenarios
       | curl --connect-timeout 5 [<%= cb.master1_ip %>]:<%= cb.port %> |
     And the output should contain:
       | Hello OpenShift! |
-    Given I ensure "hello-pod" service is deleted
+
```
zhaozhanqi, line 805: we need tag `@destructive` if delete ovn pod

zhaozhanqi:
hmm, I intend to add new test cases in upgrade side( will restart ovn) instead of deleting ovn pods(it's dangerous operation :),  since not sure if this will affect other test cases in CI even though one by one for destuctive cases.
We need to make sure it's a stable for this changes.

anuragthehatter:
retsr

> hmm, I intend to add new test cases in upgrade side( will restart ovn) instead of deleting ovn pods(it's dangerous operation :), since not sure if this will affect other test cases in CI even though one by one for destuctive cases. We need to make sure it's a stable for this changes.

Thanks @zhaozhanqi I am okay either way. Though restarting ovn network pod by deleting respective pod is a supported way customers have been using to reinitialise network in case of any misconfiguration. We definitely need this behavior in sgw->lgw transition as its really not an upgrade. I can pull back changes from simple nodeport ETP local case and we can think of adding it in upgrade.

BTW deletion of single ovnkube/sdn pod is a safe way as per my experience though. Should not impact CI. If its whole ds then it might impact CI IMO which will anyhow can be covered in upgrade. LMK

PR #2966, opened 2022-07-19
update idling case for upgrade testing

lihongan:
Since idling is not functional with OVN on 4.6/4.7, but we don't want remove tag `@network-ovnkubernetes` from this case, so the alternative solution is remove the old release tag.

@melvinjoseph86 @quarterpin @ShudiLi please help take a look, thanks,


quarterpin:
/lgtm
thanks @lihongan

PR #2968, opened 2022-07-20
Disable the cases which have migrated to golang framework

xiuwang:
@liangxia Could you help to review this? Thanks
Those cases have been migrated to golang framework and backported to 4.7-4.11


liangxia:
If they have been migrate to golang, perhaps the better approach is removing them here ?

xiuwang:
> If they have been migrate to golang, perhaps the better approach is removing them here ?
Make sense

PR #2972, opened 2022-07-22
Adding flaky tag to skip the test unti automation is fixed

kasturinarra:
Adding flaky tag because i see that the test fails often when upgrade happens with in the same version i.e(4.7.nightly-> 4.7.nightly1 / 4.9.0 -> 4.9.1 etc). Will need to check how to handle case for this kind of situation. Currently it works only when from and to versions are different i.e (4.9 -> 4.10 / 4.11 -> 4.12)

### DIFF
--- a/features/upgrade/workloads/descheduler-upgrade.feature
+++ b/features/upgrade/workloads/descheduler-upgrade.feature
@@ -5,6 +5,7 @@ Feature: Descheduler major upgrade should work fine
   @upgrade-prepare
   @users=upuser1,upuser2
   @destructive
+  @flaky
   @4.12 @4.11 @4.10 @4.9 @4.8 @4.7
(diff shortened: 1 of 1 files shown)


kasturinarra:
@liangxia one question other than adding the flaky tag do i need to do any changes in polraion so that test does not get picked up ? thanks !!

kasturinarra:

liangxia:
> @liangxia one question other than adding the flaky tag do i need to do any changes in polraion so that test does not get picked up ? thanks !!

Latest design of upgrade does not interact with Polarion.

kasturinarra:
> > @liangxia one question other than adding the flaky tag do i need to do any changes in polraion so that test does not get picked up ? thanks !!
> 
> Latest design of upgrade does not interact with Polarion.

okay, thank you !!

PR #2973, opened 2022-07-25
Adding inactive tags

abhbaner:
Added inactive tags on tests - test covered in gingko test suites.

cc @pruan-rht @dis @cpmeadors  @liangxia


liangxia:
Since those tests have been migrated to ginkgo, the test case in Polarion are still valid (not inactive).
So the proper way is remove all the codes there. Tag inactive cause confusion.

PR #2974, opened 2022-07-25
Fix OCP-25792

l1xiangyi:
I think rules in other versions should also be changed from 
```
the name of the strategy
```
to 
```
strategy
```
Please let me know in the comment if I'm right. Much thanks!

### DIFF
--- a/lib/rules/web/admin_console/4.10/overview.xyaml
+++ b/lib/rules/web/admin_console/4.10/overview.xyaml
@@ -141,7 +141,7 @@ check_triggered_by_popover_info:
   action: check_page_contains
 check_update_strategy_popover_info:
   params:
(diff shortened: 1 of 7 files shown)


l1xiangyi:
@yapei I'll build a 4.12 cluster and test it first. If there are other releases to be changed I'll be building their Jenkins jobs as well.

l1xiangyi:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/535181
Jenkins log

l1xiangyi:
Upon investigation, this fix is not suitable for the auto failure since the test check point is targeted at receiving the message which I was trying to omit in this PR.

PR #2978, opened 2022-07-27
fixed prometeus sa token issues

anuragthehatter:
Due to recent sa token behavior change, replacing prometheus sa token logic.Instead we can get metrics relayed on port 9109 via sdn pods only. FYI In case of SDN we relay metrics on all sdn network pods irrespctive of controller, master or worker pod

Logs: http://pastebin.test.redhat.com/1068170

@openshift/team-sdn-qe

### DIFF
--- a/features/networking/metrics.feature
+++ b/features/networking/metrics.feature
@@ -11,11 +11,29 @@ Feature: SDN/OVN metrics related networking scenarios
   @network-openshiftsdn
   @heterogeneous @arm64 @amd64
   Scenario: OCP-28519:SDN Prometheus should be able to monitor kubeproxy metrics
-    Given I select a random node's host
     Given I switch to cluster admin pseudo user
     And I use the "openshift-sdn" project
-    When I run command on the node's sdn pod:
-      | bash | -c | curl 127.0.0.1:29101/metrics  |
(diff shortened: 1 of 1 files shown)


asood-rh:
@anuragthehatter If I am not mistaken the sdn/ovn pods are from where the metrics scraped by prometheus. If the test cases are to validate the metrics reach prometheus, wouldn't these changes be changing what is being tested?

anuragthehatter:
Test logs OVN: http://pastebin.test.redhat.com/1068431
SDN: http://pastebin.test.redhat.com/1068463

@asood-rh You have valid concern there but this is also the fact that prometheus scraping logic is already being checked my various testcases in monitoring QE team so its kinda duplication. Since our test cases already titled with prometheus since beginning, probably a good idea to stick to same earlier logic else we need to fix testcsaes. I have revered my code and fixed just token fetch.

Checking just network pods to whether they are relaying right metrics to right port should be our check IMO since prometheus functional testcases already covering scraping logic  but This is also fact that this will ease our debugging in case the issue falls on monitoring team. So happy to revert to original logic.

Also not sure why both cases were checking kubeproxy metrics. Only one should. Anyways things looks good noe

PR #2981, opened 2022-07-29
update the configmap of rsyslogserver

anpingli:
The keyword infra-write,app-write are dropped from Logging 5.4 update syslog.conf filter to match this change.
version <=5.1:  [infra-write,app-write, audit-write]  present
version = 5.2,5.3:  both [infra-write,app-write, audit-write]  and [logType:  applicaiton,logType:  infrastructure,  logType audit] present,
version >=5.4  only  [logType:  applicaiton,logType:  infrastructure,  logType audit]  present


anpingli:
@liangxia PTAL

PR #2982, opened 2022-07-29
Inactive OCP-27686

chao007:
Inactive OCP-27686 since it supports different sc.
@duanwei33 please help to review it.


duanwei33:
@liangxia PTAL, thanks.

PR #2992, opened 2022-08-10
add step to make sure the scearios running correct cluster

zhaozhanqi:
OVN will be default network type in 4.12 version,  so the default workflow will be OVN. so here enhance the step to make sure scenarios are working on correct cluster in case. 

thus in Prow CI config, we do not need to specified the network type tag
before:

```
      E2E_RUN_TAGS: '@aws-ipi and @network-openshiftsdn and not @fips'
```

after :

```
E2E_RUN_TAGS: '@aws-ipi and  not @fips'
```

and another in cucushift repo https://github.com/openshift/cucushift/pull/9250

@openshift/team-sdn-qe ^^

### DIFF
--- a/features/networking/egress-ingress.feature
+++ b/features/networking/egress-ingress.feature
@@ -366,6 +366,7 @@ Feature: Egress-ingress related networking scenarios
   @proxy @noproxy
   @heterogeneous @arm64 @amd64
   Scenario: OCP-19615:SDN Iptables should be updated with correct endpoints when egress DNS policy was used
+    Given the env is using "OpenShiftSDN" networkType	
     Given I have a project
     Given I obtain test data file "networking/list_for_pods.json"
     When I run oc create over "list_for_pods.json" replacing paths:
@@ -421,6 +422,7 @@ Feature: Egress-ingress related networking scenarios
   @network-ovnkubernetes
   @heterogeneous @arm64 @amd64
   Scenario: OCP-33530:SDN EgressFirewall allows traffic to destination ports
+    Given the env is using "OVNKubernetes" networkType
     Given I have a project
     Given I have a pod-for-ping in the project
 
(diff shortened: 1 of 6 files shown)


weliang1:
@zhaozhanqi If the test case has a tag as `@network-ovnkubernetes`, do we still need `Given the env is using "OVNKubernetes" networkType`? Same as the tag _@network-openshiftsdn for `Given the env is using "OpenShiftSDN" networkType`

zhaozhanqi:
> @zhaozhanqi According to https://issues.redhat.com/browse/OCPQE-10826, if setting `E2E_RUN_TAGS: '@aws-ipi and @network-ovnkubernetes and not @fips'` for `as: e2e-aws-ipi`, and `@aws-ipi and @network-openshiftsdn and not @fips` for `as: e2e-aws-ipi-sdn`, the test case has a tag as @network-ovnkubernetes or @network-openshiftsdn, do we still need update these cases?
Yes, I think skipped scenarios is better than tag.  thus after this changes. team members don't need to care what's kind of tag need to be chosen for network related cases.  especially  4.12 the default OVN is changes. 

and the other is CNF team also run some cucushift cases.  they are not using those two `tag` to filter cases as I know.

weliang1:
> > @zhaozhanqi According to https://issues.redhat.com/browse/OCPQE-10826, if setting `E2E_RUN_TAGS: '@aws-ipi and @network-ovnkubernetes and not @fips'` for `as: e2e-aws-ipi`, and `@aws-ipi and @network-openshiftsdn and not @fips` for `as: e2e-aws-ipi-sdn`, the test case has a tag as @network-ovnkubernetes or @network-openshiftsdn, do we still need update these cases?
> 
> Yes, I think skipped scenarios is better than tag. thus after this changes. team members don't need to care what's kind of tag need to be chosen for network related cases. especially 4.12 the default OVN is changes.
> 
> and the other is CNF team also run some cucushift cases. they are not using those two `tag` to filter cases as I know. sometimes we need to debug the failed cases due to it's running incorrect cluster.

Sound good to me.
/lgtm

PR #2999, opened 2022-08-22
Fix flakes of 19110|19150

Phaow:
## Fix flakes of 19110|19150
When using "oc apply/create -f" to create iscsi-target server pods violates PodSecurity on 4.12, it used hostpath so need ns privilieged. Modify the default create ns of the iscsi-target server to openshift-iscsi-target solve the issue. And it should also work on previously releases.

**[Flake record](https://${report-portal-Domain}/ui/#prow/launches/all/204729?item0Params=filter.eq.hasStats%3Dtrue%26filter.eq.hasChildren%3Dfalse%26filter.in.type%3DSTEP%26filter.in.status%3DFAILED%252CINTERRUPTED%26page.page%3D2)**
**[Test record ](https://${jenkins-Domain}/job/ocp-common/job/Runner/572210/console)**

### DIFF
--- a/features/step_definitions/helper_services.rb
+++ b/features/step_definitions/helper_services.rb
@@ -504,10 +504,12 @@ Given /^I have a iSCSI setup in the environment$/ do
   if !_project.exists?(user:admin, quiet: true)
     @result = admin.cli_exec(:create_namespace, name: 'iscsi-target')
     raise 'failed to create "iscsi-target" project' unless @result[:success]
-    @result = admin.cli_exec(:label, resource: 'ns/iscsi-target', key_val: 'security.openshift.io/scc.podSecurityLabelSync=false', overwrite: 'true')
-    raise 'failed to add label "security.openshift.io/scc.podSecurityLabelSync=false" to "iscsi-target" namespace' unless @result[:success]
-    @result = admin.cli_exec(:label, resource: 'ns/iscsi-target', key_val: 'pod-security.kubernetes.io/enforce=privileged', overwrite: 'true')
-    raise 'failed to add label "pod-security.kubernetes.io/enforce=privileged" to "iscsi-target" namespace' unless @result[:success]
+    if env.version_ge("4.12", user: user)
+      @result = admin.cli_exec(:label, resource: 'ns/iscsi-target', key_val: 'security.openshift.io/scc.podSecurityLabelSync=false', overwrite: 'true')
+      raise 'failed to add label "security.openshift.io/scc.podSecurityLabelSync=false" to "iscsi-target" namespace' unless @result[:success]
+      @result = admin.cli_exec(:label, resource: 'ns/iscsi-target', key_val: 'pod-security.kubernetes.io/enforce=privileged', overwrite: 'true')
+      raise 'failed to add label "pod-security.kubernetes.io/enforce=privileged" to "iscsi-target" namespace' unless @result[:success]
+    end
   end
 
   _pod = cb.iscsi_pod = pod("iscsi-target", _project)
--- a/features/step_definitions/helper_services.rb
+++ b/features/step_definitions/helper_services.rb
@@ -500,9 +500,9 @@ end
 Given /^I have a iSCSI setup in the environment$/ do
   ensure_admin_tagged
 
-  _project = project("iscsi-target", switch: false)
+  _project = project("openshift-iscsi-target", switch: false)
   if !_project.exists?(user:admin, quiet: true)
-    @result = admin.cli_exec(:create_namespace, name: 'iscsi-target')
+    @result = admin.cli_exec(:create_namespace, name: 'openshift-iscsi-target')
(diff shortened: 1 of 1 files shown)


Phaow:

duanwei33:
Hmm, does the url of `Flake record` and `Test record` get mixed?
The update looks good to me.

features/step_definitions/helper_services.rb
```diff
@@ -500,9 +500,9 @@
 Given /^I have a iSCSI setup in the environment$/ do
   ensure_admin_tagged
 
-  _project = project("iscsi-target", switch: false)
+  _project = project("openshift-iscsi-target", switch: false)
```
liangxia, line 503: Perhaps use `openshift-qe-iscsi-target` ? Just in case the project name is used by OCP

duanwei33: That makes sense even OCP doesn't have it now.
Adding `qe` or `qe-test`.

features/step_definitions/helper_services.rb
```diff
@@ -500,9 +500,9 @@
...
   if !_project.exists?(user:admin, quiet: true)
-    @result = admin.cli_exec(:create_namespace, name: 'iscsi-target')
+    @result = admin.cli_exec(:create_namespace, name: 'openshift-iscsi-target')
     raise 'failed to "iscsi-target" project' unless @result[:success]
   end
```
xingxingxia, line 507: The cases will still fail once https://github.com/openshift/cluster-kube-apiserver-operator/pull/1369 is merged. That will make "enforce" default to "restricted". Currently your fix can pass because currently the default is still "privileged", which can be checked via `jq "" $(oc extract cm/config -n openshift-kube-apiserver --confirm) | jq '.admission.pluginConfig.PodSecurity'` if you like to check. The thorough fix is still to label your namespace explicitly.

xingxingxia:
One minor nagging, even if your RP / Jenkins log links are only internal, it is better to do some hiding processing than directly exposing in a public place IIRC.

Phaow:
> One minor nagging, even if your RP / Jenkins log links are only internal, it is better to do some hiding processing than directly exposing in a public place IIRC.

Sure, thanks!

Phaow:
> Hmm, does the url of `Flake record` and `Test record` get mixed?
Hmm , seems not got mixed, but I hidden the domain name.

features/step_definitions/helper_services.rb
Phaow: Fixed.

features/step_definitions/helper_services.rb
Phaow: Thanks for the suggestion! Fixed by using another solution.

Phaow:
Hi @duanwei33 @xingxingxia @liangxia , Fixed with suggestions and retest passed,it should work both on 4.12 and 4.12-, could you please help take a look when you get a chance? Thanks!
**4.12 test record:** [Runner/573701](https://$(jenkins_domain)/job/ocp-common/job/Runner/573701/console)
**4.10 test record:** [Runner/573702](https://$(jenkins_domain)/job/ocp-common/job/Runner/573702/console)

features/step_definitions/helper_services.rb
```diff
@@ -503,7 +503,11 @@
   _project = project("iscsi-target", switch: false)
   if !_project.exists?(user:admin, quiet: true)
     @result = admin.cli_exec(:create_namespace, name: 'iscsi-target')
-    raise 'failed to "iscsi-target" project' unless @result[:success]
+    raise 'failed to create "iscsi-target" project' unless @result[:success]
+    @result = admin.cli_exec(:label, resource: 'ns/iscsi-target', key_val: 'security.openshift.io/scc.podSecurityLabelSync=false', overwrite: 'true')
```
liangxia, line 507: We only need this for 4.12+, perhaps add a check for before adding the labels ?
```
```

Phaow: Good suggestion! Sorry for that, I'm not familar with ruby. Thanks for the details!

Phaow:
Hi @liangxia , Fixed and retest passed, could you please help take a look when you get a chance? Thanks!
4.12 test record: [Runner/573751](https://%24%28jenkins_domain%29/job/ocp-common/job/Runner/573751/console)
4.10 test record: [Runner/573752](https://%24%28jenkins_domain%29/job/ocp-common/job/Runner/573752/console)

xingxingxia:
/lgtm
If no version check, also fine, because no harm for 4.11 and less version. But adding version check is also fine.
key_val: 'xxx' can be used multiple times in `cli_exec(:label, ...` if you like.

Phaow:
> /lgtm If no version check, also fine, because no harm for 4.11 and less version. But adding version check is also fine. key_val: 'xxx' can be used multiple times in `cli_exec(:label, ...` if you like.

Got it. Many thanks! Cool trick.

PR #3003, opened 2022-08-24
Add pod security labels to namespaces

xingxingxia:
Failure log: https://url.corp.redhat.com/e73923f due to error:
Error from server (Forbidden): error when creating "ldapserver.yaml": pods "ldapserver" is forbidden: violates PodSecurity "restricted:v1.24": unrestricted capabilities (container "ldapserver" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "ldapserver" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "ldapserver" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")

Pass log: job/ocp-common/job/Runner/575624/console
@zhouying7780 @liangxia help review / merge, thanks!

### DIFF
--- a/features/step_definitions/helper_services.rb
+++ b/features/step_definitions/helper_services.rb
@@ -161,6 +161,17 @@ Given /^I have LDAP service in my project$/ do
     # So take the second one since this one can be implemented currently
     ###
     stats = {}
+    if env.version_ge("4.12", user: user)
+      step %Q/I run the :label admin command with:/, table(%{
+        | resource  | namespace/<%= project.name %>                        |
+        | overwrite | true                                                 |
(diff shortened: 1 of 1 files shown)


features/step_definitions/helper_services.rb
```diff
@@ -161,6 +161,15 @@
     # So take the second one since this one can be implemented currently
     ###
     stats = {}
+    step %Q/I run the :label admin command with:/, table(%{
```
liangxia, line 164: I would suggest add a version check here as we only need this for 4.12+
A version check is much cheaper than adding a bunch of labels.

xingxingxia:
@liangxia added version-check `if` condition, pass log: job/Runner/575921/console

PR #3007, opened 2022-08-25
Add setup privileged project util

Phaow:
## Add setup privileged project util
PodSecurity restricted enabled by default on 4.12, used hostpath cases need namespace set to privilieged. Add setup privileged project util for using privilieged namespace scenarios.

**[Flake record](https://${report-portal-Domain}/ui/#prow/launches/454/207510/12395570/12395797/log?item1Params=filter.eq.hasStats%3Dtrue%26filter.eq.hasChildren%3Dfalse%26filter.in.type%3DSTEP%26filter.in.status%3DFAILED%252CINTERRUPTED)**
**[Test record ](https://${jenkins-Domain}/job/ocp-common/job/Runner/577369/console)**

### DIFF
--- a/features/step_definitions/project.rb
+++ b/features/step_definitions/project.rb
@@ -34,6 +34,47 @@ Given /^I have a project$/ do
   end
 end
 
+Given /^I have a privileged project$/ do
+  # system projects should not be selected by default
+  sys_projects = BushSlicer::Project::SYSTEM_PROJECTS
+
+  project = @projects.reverse.find {|p|
+    !sys_projects.include?(p.name) &&
+      p.is_user_admin?(user: user, cached: true) &&
+      p.active?(user: user, cached: true)
+  }
+  if project
+    # project does exist as visible is doing an actual query
+    # also move project up the stack
+    @projects << @projects.delete(project)
+  else
+    projects = (user.projects - sys_projects).select {|p|
+      p.is_user_admin?(user: user, cached: true) &&
+      p.active?(user: user, cached: true)
+    }
+    if projects.empty?
+      step 'I create a new project'
+      unless @result[:success]
+        logger.error(@result[:response])
+        raise "unable to create project, see log"
+      end
+    else
+      cache_resources *projects
+    end
+  end
+  if env.version_ge("4.12", user: user)
+    step %Q/I run the :label admin command with:/, table(%{
+      | resource  | namespace/<%= project.name %>                        |
+      | overwrite | true                                                 |
+      | key_val   | security.openshift.io/scc.podSecurityLabelSync=false |
+      | key_val   | pod-security.kubernetes.io/enforce=privileged        |
+      | key_val   | pod-security.kubernetes.io/audit=privileged          |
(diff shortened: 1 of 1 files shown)


Phaow:

features/step_definitions/project.rb
```diff
@@ -34,6 +34,47 @@
...
+      step 'I create a new project'
+      unless @result[:success]
+        logger.error(@result[:response])
+        raise "unable to create project, see log"
+      end
+    else
+      cache_resources *projects
+    end
+  end
+  if env.version_ge("4.12", user: user)
```
duanwei33, line 65: @Phaow  If we only add this for 4.12+, the step name makes confused for the previous release, right?

Phaow: @duanwei33 Yeah. TBH, I don't have a better step name for this. How about `Given /^I have a project using hostpath`?

duanwei33: @Phaow  I agree it is not easy to backport.
From the code side, I prefer to apply them on all releases (if this cmd works), and let the user choose to use it or not in different releases. So the code and the step name are consistent.
Like in 4.11-, users could use "I have a project" if they don't need a privileged one.

duanwei33: Adding @liangxia may give us some suggestions here.

Phaow: Sure. Thanks!

duanwei33: @Phaow Another way(I think it is better one) is we could add an individual step like "labeling () project as pod-security privileged", then we could use this step after "I have a project".

Phaow: Good option! I'll have a try.

liangxia: How about `Given /^I have a project with proper privilege$/` ?
```
  step %Q/I have a project/
    blabla
  end
end

Comment (Phaow) [reply features/step_definitions/project.rb:65]: Perfect. Many thanks!

**Comment by Phaow** [comment]
Hi @duanwei33 , updated and retest passed, could you please help take a look when you get a chance? Thanks!
**[Test record](https://$(jenkins_domain)/job/ocp-common/job/Runner/578729/console)**

**Comment by Phaow** [comment]
Seems the build image failed not related the changed code.
failed at step `$ bundle install --gemfile=/verification-tests/tools/Gemfile`
```
Bundler could not find compatible versions for gem "net-ssh":
  In Gemfile:
    net-ssh

    net-scp was resolved to 3.0.0, which depends on
      net-ssh (< 7.0.0, >= 2.6.5)
error: build error: error building at STEP "RUN set -x &&     mv /tierN/ /verification-tests/features/tierN/ &&     chgrp -R 0 /verification-tests &&     chmod -R g=u /verification-tests &&     /verification-tests/tools/install_os_deps.sh &&     /verification-tests/tools/hack_bundle.rb &&     yum clean all -y &&     rm -rf /var/cache/yum /var/tmp/* /tmp/* /verification-tests/Gemfile.lock": error while running runtime: exit status 1
```

PR #3010, opened 2022-08-26
Fix tags for scenario outlines in storage

liangxia:
When there are multiple `Examples` for scenario outlines, some of the tags are shared across those examples, which need to put ahead of the scenario outline.

/cc @duanwei33 @chao007 @Phaow


Phaow:
/lgtm
Thanks for the info and helping update!

Phaow (commented):
/lgtm

PR #3001, opened 2022-08-24
OCPQE-11705: Fix "PodSecurity violation error"

liangxia:
This fix the issue of:
1. I run commands on the host
2. oc debug

/cc @kasturinarra @xingxingxia @JianLi-RH @dis016 @pruan-rht @jhou1

PR #3012, opened 2022-08-26
pod security for hostnw pods

anuragthehatter:
Simlar to https://github.com/openshift/verification-tests/pull/3001/files (as debug pod is also hostnw pod), we also need a def under our networking.rb to support various host network pod usecases which are failing on 4.12

we  pass `nil` the logic if env < 4.12
Logs: https://privatebin.corp.redhat.com/?cdaa9fa5248e2208#236gaH3wzHKBX4c77p1xvgL43DiUvMot9L5iQnKYPxs9

@openshift/team-sdn-qe PTAL

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1599,16 +1599,3 @@ Given /^the cluster is dual stack network type$/ do
     skip_this_scenario
   end
 end
-
-Given /^the appropriate pod security labels are applied to the#{OPT_QUOTED} namespace$/ do | project_name |
-  ensure_admin_tagged
-  project_name ||= project.name
-  if env.version_ge("4.12", user: user)
-    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'security.openshift.io/scc.podSecurityLabelSync=false', overwrite: true)
-    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'pod-security.kubernetes.io/enforce=privileged', overwrite: true)
-    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'pod-security.kubernetes.io/audit=privileged', overwrite: true)
-    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'pod-security.kubernetes.io/warn=privileged', overwrite: true)
-  else
-    nil  
-  end
-end
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1599,3 +1599,15 @@ Given /^the cluster is dual stack network type$/ do
     skip_this_scenario
   end
 end
+
+Given /^the appropriate pod security labels are applied to the "(.+?)" namespace$/ do | project_name |
+  ensure_admin_tagged
+  if env.version_ge("4.12", user: user)
+    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'security.openshift.io/scc.podSecurityLabelSync=false', overwrite: true)
+    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'pod-security.kubernetes.io/enforce=privileged', overwrite: true)
+    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'pod-security.kubernetes.io/audit=privileged', overwrite: true)
+    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'pod-security.kubernetes.io/warn=privileged', overwrite: true)
+  else
+    nil  
+  end
+end
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1600,8 +1600,9 @@ Given /^the cluster is dual stack network type$/ do
   end
 end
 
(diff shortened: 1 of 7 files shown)


zhaozhanqi:
@anuragthehatter Saw there is function we can use

```
  step %Q/I have a project/
    step %Q/I run the :label admin command with:/, table(%{
      | resource  | namespace/<%= project.name %>                        |
      | overwrite | true                                                 |
      | key_val   | security.openshift.io/scc.podSecurityLabelSync=false |
      | key_val   | pod-security.kubernetes.io/enforce=privileged        |
      | key_val   | pod-security.kubernetes.io/audit=privileged          |
      | key_val   | pod-security.kubernetes.io/warn=privileged           |
      })
    step %Q/the step should succeed/
  end
end
```

anuragthehatter:
> @anuragthehatter Saw there is function we can use
> 
> ```
> Given /^I have a project with proper privilege$/ do
>   step %Q/I have a project/
>   if env.version_ge("4.12", user: user)
>     step %Q/I run the :label admin command with:/, table(%{
>       | resource  | namespace/<%= project.name %>                        |
>       | overwrite | true                                                 |
>       | key_val   | security.openshift.io/scc.podSecurityLabelSync=false |
>       | key_val   | pod-security.kubernetes.io/enforce=privileged        |
>       | key_val   | pod-security.kubernetes.io/audit=privileged          |
>       | key_val   | pod-security.kubernetes.io/warn=privileged           |
>       })
>     step %Q/the step should succeed/
>   end
> end
> ```

Ah yea, noticed that. Thanks for reference. But seems like it won't honor a particular namespace name but a random project id. In upgrade cases, we need static project names to have such privileges. May be we can keep this PR def for static namespace in our tests and for random use the one you referred.

asood-rh:
a) Shouldn't the step definition be modified instead to take a project name and implicitly take the project if the namespace is not provided?

b) If we need to add it to networking step definition, should look for a way so that we do not have to modify all the scripts individually?
c) Do we have book keeping to see we have fixed all instances of the issues in SDN project?

anuragthehatter:
> a) Shouldn't the step definition be modified instead to take a project name and implicitly take the project if the namespace is not provided?
```Yes``` but it won't be easy to modify step @zhaozhanqi mentioned. Its using nested steps which also needs to be modified and tested. All QE teams are using those and might need extensive testing. I could't have implemented project name as OPT_SYM in networking def but we probably don;t need it in liu of existing def 

> b) If we need to add it to networking step definition, should look for a way so that we do not have to modify all the scripts individually? 
```Yes``` but we have only 3-4 cases as mentioned in this PR which uses hnw pods so thus step is added to such cases only. Also there is no more new development in these repos so we shouldn't be worried IMO.

> c) Do we have book keeping to see we have fixed all instances of the issues in SDN project?
```Yes``` I did a recusrive grep search in both repos and found out only these cases which are using hostnetwork pods in our testcases. We have other upgrade cases but they don;t need such steps and they don't use hostnetwork pods.

**Comment by anuragthehatter** [comment]
FYI: Its easy to add OPT_SYM for project name (which will make project name optional) though for the def this PR is trying to merge. If we think it will be useful.

**Comment by zhaozhanqi** [comment]
ok. looks good to me

**Comment by anuragthehatter** [comment]
Ok this looks good now. It will take project if defined else existing ns in flow. re-ran to make sure

**Comment by anuragthehatter** [comment]
/assign @zhaozhanqi 
Need your super power to merge this

File: features/step_definitions/networking.rb [hunk 1]
```diff
@@ -1599,3 +1599,16 @@
     skip_this_scenario
   end
 end
+
+Given /^the appropriate pod security labels are applied to the#{OPT_QUOTED} namespace$/ do | project_name |
```
liangxia, line 1603: Perhaps we should locate the step in `project.rb` ?

features/step_definitions/networking.rb
```diff
@@ -1599,3 +1599,16 @@
...
+  ensure_admin_tagged
+  project_name ||= project.name
+  if env.version_ge("4.12", user: user)
+    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'security.openshift.io/scc.podSecurityLabelSync=false', overwrite: true)
+    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'pod-security.kubernetes.io/enforce=privileged', overwrite: true)
+    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'pod-security.kubernetes.io/audit=privileged', overwrite: true)
+    admin.cli_exec(:label, resource: "namespace", name: project_name, key_val: 'pod-security.kubernetes.io/warn=privileged', overwrite: true)
+  else
+    nil
```
liangxia, line 1612: We can omit the `else nil`

anuragthehatter:
Thanks @liangxia took care of comments. Agree m its reasonable to add in project.b. 
Re-ran on both 4.11 and 4.12 to make sure
https://privatebin.corp.redhat.com/?2cbf3b31ba113b39#ezUL5WcUo5E3pM7DuxX4rbbWDFUctsuoefYrzhpEaXZ

PR #3022, opened 2022-09-01
OCP-25294 and OCP-23543 quick fix

anuragthehatter:
Need to add pod privileges in 412 for this case as well

@openshift/team-sdn-qe


anuragthehatter:
Also OCP-23543 should supposed to run for SDN only as per https://issues.redhat.com/browse/SDN-341. Confirmed with dev, shouldn;t be on OVN

PR #3023, opened 2022-09-02
fix chrome crash issue

yapei:
add `--disable-extensions` and `--headless` arg when launching chrome to see if we can reduce chrome crash issue


yapei:
cc @liangxia  @pruan-rht @yanpzhan

pruan-rht:
@yapei yeah, we can give it a try
/lgtm

PR #3028, opened 2022-09-02
Adding the missing descheduler version

kasturinarra:
PR fixes two issues here:
=======================
1. I see that upgrades from 4.11.y to 4.11.z are failing to create csv version with out the version 4.11 present in here, so fixing the same.
2. I see that descheduler channels does not have '-' in its channel name and when using the existing code it fails with error **no implicit conversion of nil into String**, so fixing the same.


kasturinarra:

PR #3034, opened 2022-09-08
Add @inactive tag

weliang1:
Due to https://issues.redhat.com/browse/OCPBUGS-994, admin should not run any pods under default ns. Update case to be inactive. 

@openshift/team-sdn-qe PTAL


asood-rh:
It does make sense though consumers of kubernetes should not really use default namespace.

zhaozhanqi:
How about only remove 4.12 version since it's still works on old version

weliang1:
> How about only remove 4.12 version since it's still works on old version

Remove @inactive and @4.12 tag now.

PR #3039, opened 2022-09-13
Change the .local dir location

jhou1:
A second attempt to fix [OCPQE-11756](https://issues.redhat.com//browse/OCPQE-11756)

@liangxia @yapei PTAL

### DIFF
--- a/tools/openshift-ci/Dockerfile
+++ b/tools/openshift-ci/Dockerfile
@@ -36,5 +36,5 @@ RUN set -x && \
 
 # Fixing OCPQE-11756
 RUN set -x && \
-    mkdir /verification-tests/.local && \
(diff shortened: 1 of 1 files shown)


tools/openshift-ci/Dockerfile
```diff
@@ -36,5 +36,5 @@ RUN set -x && \
 
 # Fixing OCPQE-11756
 RUN set -x && \
-    mkdir /verification-tests/.local && \
-    chmod -R 777 /verification-tests/.local
+    mkdir /.local && \
```
yapei, line 39: I'm sorry, where are we creating a `local` directory with command `mkdir /.local`?

liangxia: In the root folder `/`

PR #3043, opened 2022-09-15
helper method for config_map and  a step to get leaderIdentity.

pruan-rht:
Added support to get the annotations fields from a ConfigMap and a step to extract the elected leader information

@anuragthehatter, this should help you get the `holderIdentity` information.

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -612,13 +612,13 @@ Given /^the multus is enabled on the cluster$/ do
   success = wait_for(120, interval: 10)  {
     desired_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:desired]
     available_multus_replicas = daemon_set('multus', project('openshift-multus')).replica_counters(user: admin)[:available]
-    if (desired_multus_replicas == available_multus_replicas || desired_multus_replicas > env.nodes.count) && available_multus_replicas != 0 
+    if (desired_multus_replicas == available_multus_replicas || desired_multus_replicas > env.nodes.count) && available_multus_replicas != 0
       true
     else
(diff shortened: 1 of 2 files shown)


anuragthehatter:
@pruan-rht Thanks much for creating this def for us. Is it possible to return the "leader" pod name to ```pod``` so that evalaution like `pod.ip` can also be performed post calling this step? we would mostly be interested in its IP in long run.

pruan-rht:
> @pruan-rht Thanks much for creating this def for us. Is it possible to return the "leader" pod name to `pod` so that evalaution like `pod.ip` can also be performed post calling this step? we would mostly be interested in its IP in long run.

@anuragthehatter should be possible.  What's the step to get the `leader` pod name?

anuragthehatter:
> > @pruan-rht Thanks much for creating this def for us. Is it possible to return the "leader" pod name to `pod` so that evalaution like `pod.ip` can also be performed post calling this step? we would mostly be interested in its IP in long run.
> 
> @anuragthehatter should be possible. What's the step to get the `leader` pod name?

@pruan-rht I meant something like, say (evaluation step after your step will complain ```which pod you are talking about``` error

```
Given I store kubernetes elected leader pod for ovnkube-master in the clipboard
And evaluation of `pod.ip` is stored in the :ovn_master_leader_ip clipboard
```

pruan-rht:
@anuragthehatter I've amended the step to return the pod instead of just the holderId.  PTAL.  Please note, I've added two ways of doing the same thing, let me know which way you prefer.  cc: @liangxia

anuragthehatter:
/lgtm

Thanks @pruan-rht. This looks good to me. Tested locally with my testcase

PR #3044, opened 2022-09-16
OCP-37704 should point to kubernetes elected ovn master pod to fetch ovn exposed metrics

anuragthehatter:
Since 4.11+ OVN master metrics are now being exposed to kubernetes elected OVN master pod instead earlier they were exposed to all OVN masters. This case was passing most of the time as co-incidentally it was selecting right master..heh.

This is already implemented in right way in `openshift-tests-private` repo but needed fix here. Also https://github.com/openshift/verification-tests/pull/3043 introduced a new helper function, thanks @pruan-rht for that which we can use here and in long run to use kube elected ovn master wherever needed in fixes

@openshift/team-sdn-qe 

Logs: https://privatebin.corp.redhat.com/?e17d1275802e6bd1#3tfTyUskQ9u3DNvx3ueWfPM2jtLnR3UvXk3BSd9Uge2f

### DIFF
--- a/features/networking/metrics.feature
+++ b/features/networking/metrics.feature
@@ -92,11 +92,10 @@ Feature: SDN/OVN metrics related networking scenarios
   @proxy @noproxy @disconnected @connected
   @heterogeneous @arm64 @amd64
   Scenario: OCP-37704:SDN Should be able to monitor various ovnkube-master and ovnkube-node metrics via prometheus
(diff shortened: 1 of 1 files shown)


anuragthehatter:
We have so many leaders - SB and NB RAFT leader for HA. We also have HA for northd, and we have kubernetes leader election for ovnk master :-)

PR #3047, opened 2022-09-19
remove --headless

yapei:


yapei:
@liangxia could you help approve?

PR #3052, opened 2022-09-22
fix OCPQE-11756

yapei:
Before this change, I can see errors in local container
```
      unknown error: Chrome failed to start: exited abnormally.
        (unknown error: DevToolsActivePort file doesn't exist)
        (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.) (Selenium::WebDriver::Error::UnknownError)
      #0 0x56406407b693 <unknown>
      #1 0x564063e74b0a <unknown>
```

### DIFF
--- a/lib/webauto/web4cucumber.rb
+++ b/lib/webauto/web4cucumber.rb
@@ -112,8 +112,8 @@ require_relative 'chrome_extension'
       client.open_timeout = 180
       client.read_timeout = 600
       headless
-      Selenium::WebDriver.logger.level = :debug
-      Watir.logger.level = :debug
+      #Selenium::WebDriver.logger.level = :debug
+      #Watir.logger.level = :debug
       if @browser_type == :firefox
         logger.info "Launching Firefox Marionette/Geckodriver"
         raise "auth proxy not implemented for Firefox" if proxy_pass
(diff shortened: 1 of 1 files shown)


yapei:
`element_scroll_behavior` options seems also need to be fixed, but I would like to fix it later since little case will be blocked

yapei:
@jhou1  @chengzhang1016  @liangxia  @yanpzhan

yapei:
In progress of verification

yapei:
normal cluster pass log https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/619996/console

lib/webauto/web4cucumber.rb
```diff
@@ -147,9 +147,7 @@ def browser
         if Integer === @scroll_strategy
           chrome_caps[:element_scroll_behavior] = @scroll_strategy
         end
-        if self.class.container?
```
liangxia, line 150: Would you mind to try locally directly ( not in container ) ?

yapei:
container [pass log](http://pastebin.test.redhat.com/1074837)
local [pass log](http://pastebin.test.redhat.com/1074834)

lib/webauto/web4cucumber.rb
yapei: @liangxia reverted these changes, PTAL

PR #3061, opened 2022-09-30
fix OCPQE-8932

QiaolingTang:
To fix failure in [OCPQE-8932](https://issues.redhat.com//browse/OCPQE-8932):
From https://gitlab.cee.redhat.com/aosqe/jenkins-jcasc-n/-/blob/master/scripts/OCPQE-11466/upgrade_cluster_logging.sh#L9-10, in OCP 4.12 the channel is `stable-5.6`.
As the `catsrc/qe-app-registry` is enabled in [Prow CI](https://github.com/openshift/release/tree/master/ci-operator/step-registry/enable-qe-catalogsource), we should always have channel `stable-5.6` in our test clusters, here change the default channel name to `stable-5.6` when testing in OCP 4.12. 

/cc @anpingli


anpingli (commented):
/lgtm

PR #3065, opened 2022-10-09
update rsyslogd-container images

anpingli:
- use one image to support both TLS and none tls cases

https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/640767/console


anpingli:
@liangxia  PTAL

PR #3066, opened 2022-10-11
update hpa test file

yapei:


testdata/hpa/hpa_rubyex.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2beta2
```
liangxia, line 1: Will it works on both 4.12 and 4.6 ?

yapei: I just ran regression against 4.9, let me build a 4.6 cluster and let's see

yapei: @liangxia  it passed in 4.6, here is the pass log Runner/644512/console

PR #3067, opened 2022-10-11
fix OCPQE-11446

QiaolingTang:
When the cluster has `network_plugin_mode: "multitenant"`, EO can't update the ES status as they're not in the same project, here add a step to make project/openshift-operators-redhat network global in these clusters.


anpingli (commented):
/lgtm

pruan-rht (commented):
/lgtm
/approve

PR #3069, opened 2022-10-12
update rules for OCP-32152

yapei:
workloads list page URL has been updated from `kind` to `group-version-kind` format, so move all changing text to rules 

take all namespace deploymentconfig list page URL as example, it was `all-namespaces/deploymentconfigs` but now it is `all-namespaces/apps.openshift.io~v1~DeploymentConfig`


yapei:
@yanpzhan could you help review? Thanks

PR #3072, opened 2022-10-12
Avoid using openshift imagestreams and templates for case OCP-10642

gkarager:


xingxingxia:
/lgtm
@pruan-rht / @liangxia could you help merge?
Thanks!

PR #3073, opened 2022-10-13
fix OCP-23681

yapei:
the `input` element has different html definition for setting PVC size and expanding PVC size, so create a new step `set_expand_size`


yapei:
4.12 pass log 
/job/Runner-v3-smoke/5035/console
/job/Runner-v3-smoke/5036/console

PR #3089, opened 2022-10-27
To add proper privilege to SCTP feature

aishwaryabk:
By adding privilege to the namespace, it resolves the PodSecurity violation on 4.12


huiran0826:
OCP-28757 and OCP-28758 in cucushift will not be maintained in 4.12 which were moved to go framework. OCP-28759 and OCP-29645 were already added " And the appropriate pod security labels are applied to the namespace" which should resolve the podSecurity issues.

aishwaryabk:
@huiran0826 Okay, Thanks for the information.

PR #3093, opened 2022-10-28
Fixes OCP-12600 and OCP-11900

aishwaryabk:


kasturinarra:
@xingxingxia @zhouying7780 could you please help check ? thanks !!

xingxingxia:
@aishwaryabk thx, but these cases' "Version" field in Polarion was up to very earlier version, i.e. they are applicable only in early version, their code needs not update.
/close

PR #3094, opened 2022-10-28
Fixes OCP-24154 and OCP-24158

aishwaryabk:
It failed while canceling the build as one of the builds would not be canceled. After reducing the number of counts by 1, the tests were successful.


kasturinarra:
@jitendar-singh could you please help check ? thanks !! I see the tests here are related to BuildAPI.

PR #3102, opened 2022-11-04
rules checking logic redundant for iptables across pod and node

anuragthehatter:
This case failes intermittenly at rules step. sometime SDN pod has a rule more than node or reverse. Upon talking to dev (Jacob, who owns [SDN-341](https://issues.redhat.com//browse/SDN-341)) this is normal as sometimes sdn leader can create rule specific to container and same goes for node. He suggested to keep checking binary version in longer run so this rules logic is redundant. Let me know if anybody has concerns

@openshift/team-sdn-qe


huiran0826:
lgtm, see if @zhaozhanqi  has any other concerns.

PR #3106, opened 2022-11-04
Fixes OCP-11543

aishwaryabk:
The test case failed as it would try to search for deployment with the name "ruby-hello-world-1". Instead, it created a deployment with the name "ruby-hello-world"


kasturinarra:
@jitendar-singh could you please help check ? test belongs to  BUILD_API. Thanks !!

PR #3110, opened 2022-11-07
removing this test from cucushift since its already added in ginkgo

sandeepknd:
removing this test from cucushift since its already added in ginkgo. 
Please find the [JIRA](https://issues.redhat.com/browse/OCPQE-12522) for more details.


geliu2016 (approved):
/lgtm

PR #3111, opened 2022-11-07
Fixes OCP-9563

aishwaryabk:
It creates a resource of type deployment instead of deploymentconfig.


kasturinarra:
@aishwaryabk thanks for your PR, the case is supported only till ocp version <=4.4 and we have a new case from ocp version >=4.5, so could you please help close this PR ? thanks !!
https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-31200

PR #3114, opened 2022-11-08
Debug issue ocpqe12378

XiyunZhao:
This change is for debugging the error reported in [OCPQE-12378](https://issues.redhat.com//browse/OCPQE-12378)

### DIFF
--- a/lib/webauto/web4cucumber.rb
+++ b/lib/webauto/web4cucumber.rb
@@ -80,7 +80,9 @@ require_relative 'chrome_extension'
       if ENV.has_key?("http_proxy") || @http_proxy
         # get rid of the heading "http://" that breaks the profile
         proxy_raw = @http_proxy || ENV["http_proxy"]
+        logger.warn "For UI team to debug proxy_raw value: #{proxy_raw}"
         proxy = proxy_raw.sub(%r{^.+?://}, "")
+        logger.warn "For UI team to debug proxy value: #{proxy}"
         proxy_bypass = "localhost,127.0.0.1"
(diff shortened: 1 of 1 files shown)


lib/webauto/web4cucumber.rb
```diff
@@ -80,7 +80,9 @@ def browser
       if ENV.has_key?("http_proxy") || @http_proxy
         # get rid of the heading "http://" that breaks the profile
         proxy_raw = @http_proxy || ENV["http_proxy"]
+        logger.info "debug proxy_raw: #{proxy_raw}"
         proxy = proxy_raw.sub(%r{^.+?://}, "")
+        logger.info "debug proxy: #{proxy}"
```
liangxia, line 85: Do we  show logs with `info` level in Prow ?

lib/webauto/web4cucumber.rb
```diff
@@ -80,7 +80,9 @@ def browser
       if ENV.has_key?("http_proxy") || @http_proxy
         # get rid of the heading "http://" that breaks the profile
         proxy_raw = @http_proxy || ENV["http_proxy"]
+        logger.error "debug proxy_raw: #{proxy_raw}"
```
liangxia, line 83: Logging an error may lead to other team members thought there are something wrong, and report issue to our team.
To avoid that, I would suggest use `logger.warn`, or update the message to `For UI team to debug proxy_raw blablabal`

XiyunZhao:
@liangxia Done for the update, please help take a look again

PR #3121, opened 2022-11-11
Recreate the iSCSI pod after deleting

duanwei33:
PR to fix [OCPQE-11977](https://issues.redhat.com/browse/OCPQE-11977) and [OCPQE-11976](https://issues.redhat.com/browse/OCPQE-11976).
Before the fix, when there is a unhealthy iscsi-target pod, it only remove but not create a new one like [failed job](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/667401/console).

This is the [successful job](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/667402/console). 

@radeore @Phaow @chao007 @ropatil010 PTAL.

### DIFF
--- a/features/step_definitions/helper_services.rb
+++ b/features/step_definitions/helper_services.rb
@@ -528,8 +528,9 @@ Given /^I have a iSCSI setup in the environment$/ do
     logger.info "found existing iSCSI pod, skipping config"
   elsif _pod.exists?(user: admin, quiet: true)
     logger.warn "broken iSCSI pod, will try to recreate keeping other config"
-    @result = admin.cli_exec(:delete, n: _project.name, object_type: "pod", object_name_or_id: _pod.name)
-    raise "could not delete broken iSCSI pod" unless @result[:success]
+    step %Q{admin ensures "#{_pod.name}" pod is deleted from the "#{_project.name}" project}
+    @result = admin.cli_exec(:create, n: _project.name, f: "#{BushSlicer::HOME}/testdata/storage/iscsi/iscsi-target.json")
+    raise "could not create iSCSI pod" unless @result[:success]
   else
     @result = admin.cli_exec(:create, n: _project.name, f: "#{BushSlicer::HOME}/testdata/storage/iscsi/iscsi-target.json")
     raise "could not create iSCSI pod" unless @result[:success]


features/step_definitions/helper_services.rb
```diff
@@ -528,8 +528,10 @@
     logger.info "found existing iSCSI pod, skipping config"
   elsif _pod.exists?(user: admin, quiet: true)
     logger.warn "broken iSCSI pod, will try to recreate keeping other config"
-    @result = admin.cli_exec(:delete, n: _project.name, object_type: "pod", object_name_or_id: _pod.name)
+    @result = admin.cli_exec(:delete, n: _project.name, object_type: "pod", object_name_or_id: _pod.name, force: "true")
```
Phaow, line 531: Hi @duanwei33 , it seems the step only use `fore` delete and not wait the iscsi pod accually deleted succeed(disappear).  Is my understanding correct? Maybe still has a bit flake.

duanwei33: @Phaow so do you think we need to use without --force or add some additional check for pod existence?

Phaow: Hmm, I think we could use force delete and make sure the fore delete is successful. Just add a wait delete finished(pod disappeared) step after force delete is better.

duanwei33: okay, I agree that will be safer

features/step_definitions/helper_services.rb
```diff
@@ -528,8 +528,9 @@
     logger.info "found existing iSCSI pod, skipping config"
   elsif _pod.exists?(user: admin, quiet: true)
     logger.warn "broken iSCSI pod, will try to recreate keeping other config"
-    @result = admin.cli_exec(:delete, n: _project.name, object_type: "pod", object_name_or_id: _pod.name)
-    raise "could not delete broken iSCSI pod" unless @result[:success]
+    step %Q{admin ensures _pod.name pod is deleted from the _project.name project}
```
akostadinov, line 531:
```suggestion
    step %Q{admin ensures #{_pod.name} pod is deleted from the #{_project.name} project}
```

duanwei33: Thanks @akostadinov, I tried this way in my last commit but seems not work, I will have another try. 
My local cucushift env is totally broken so I pushed my new code and tried with Jenkins jobs, sorry for the noise.

akostadinov: @duanwei33 , the point was to wrap ruby expressions in `#{...}`. I don't know if there are typos in the step name or missing quotes or something. In this case I'm pretty sure there need to be quotes around pod name and project. Something like

```suggestion
    step %Q{admin ensures "#{_pod.name}" pod is deleted from the "#{_project.name}" project}
```

I don't have an environment to test that though. I'm sure there are examples in the existing other steps where you can see how it was done. Or you can `grep -r ensure_deleted features/step_definitions` to see how to programatically create an object and delete it without calling into a step. Calling other steps from steps is discouraged in latest cucumber. But it is still fine I think.

duanwei33: Hi @akostadinov, it's really embarrassing that I used {#_pod.name} but not #{_pod.name}, it works. But it's nice to talk to you here again. Thanks again for your help.

features/step_definitions/helper_services.rb
duanwei33: @Phaow Updated, PTAL, thanks.

PR #3124, opened 2022-11-14
update NE testdata for hypershift testing

lihongan:
remove `router-certs-default` setting since hypershift host cluster uses another secret.


lihongan:
@melvinjoseph86 @ShudiLi PTAL, thanks

PR #3130, opened 2022-11-16
update libs

yapei:


lib/rules/web/admin_console/4.11/dashboards.xyaml
```diff
@@ -146,7 +142,6 @@ change_workload:
   action: click_button_text
 check_dashboard_dropdown_items:
   action: click_dashboard_dropdown
-  action: check_dashboard_etcd
```
yapei, line 149: on hypershift provisioned clusters, no `etcd` dashboard
so remove `etcd` dashboard check on 4.11 and 4.12

lib/openshift/infrastructure.rb
```diff
@@ -5,6 +5,10 @@ class Infrastructure < ClusterResource
     # Avoid API name in class name for dis particular class
     @kind = "Infrastructure"
 
+    def apiServerURL(user: nil, cached: true, quiet: false)
```
pruan-rht, line 8: @yapei ruby method convention is to use underscore instead of camelCase.  So this would become `api_server_url` instead.  Can you please update that?  The other changes look good.

yapei: ok, thanks for reviewing

yapei: updated

yapei:
@pruan-rht could you help review again?

PR #3079, opened 2022-10-19
Fix OCP-13016 node upgrade script

weinliu:
1. fix hpa api verison issue
2. leaving 4.11 and 4.12 support

PR #3132, opened 2022-11-16
fix OCP-11255, remove label of < 4.11

lyman9966:
due to this pr(https://github.com/openshift/verification-tests/pull/3079/files) add seccomp, will lead to failure of OCP-11255 in version < 4.11, so remove label of < 4.11


lyman9966:
@sunilcio  @weinliu  Can you help to take a look? Thanks!

PR #3154, opened 2022-12-05
Update three cases supported in sdn2ovn migration

weliang1:
Remove "Given the cluster is not migration from sdn plugin" for test cases multicase, egressfirewall and SDN-egressIP which are supported after sdn to ovn migration in 4.12.

@openshift/team-sdn-qe PTAL

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1528,11 +1528,13 @@ end
 Given /^the cluster is not migration from sdn plugin$/ do
   ensure_admin_tagged
   _admin = admin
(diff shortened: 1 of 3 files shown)


huiran0826:
@weliang1 Shall we make some distinctions among different versions,  seems before 4.11 we still need these steps.

weliang1:
> @weliang1 Shall we make some distinctions among different versions, seems before 4.11 we still need these steps.

@huiran0826 Add steps of `if env.version_le("4.11", user: user)` to check cluster version

huiran0826:
@weliang1 Thanks Weibin! 
/lgtm

PR #3155, opened 2022-12-07
And the appropriate pod security labels for multus cases

weliang1:
After debugging https://reportportal-openshift.apps.ocp-c1.prod.psi.redhat.com/ui/#prow/launches/all/276500/22372643?item1Params=filter.eq.hasStats%3Dtrue%26filter.eq.hasChildren%3Dfalse%26filter.in.type%3DSTEP%26filter.in.status%3DFAILED%252CINTERRUPTED, then found out most multus cases need `And the appropriate pod security labels are applied to the namespace`

@openshift/team-sdn-qe PTAL

### DIFF
--- a/features/networking/multus.feature
+++ b/features/networking/multus.feature
@@ -20,6 +20,7 @@ Feature: Multus-CNI related scenarios
     And evaluation of `node.name` is stored in the :target_node clipboard
     # Create the net-attach-def via cluster admin
     Given I have a project
+    And the appropriate pod security labels are applied to the namespace
     Given I obtain test data file "networking/multus-cni/NetworkAttachmentDefinitions/macvlan-bridge.yaml"
     When I run oc create as admin over "macvlan-bridge.yaml" replacing paths:
       | ["metadata"]["namespace"] | <%= project.name %>                                                                                                                                                                                                                                                                    |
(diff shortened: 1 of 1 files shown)


pruan-rht:
@weliang1 perhaps we should add that step into `Given I have a project` instead...since this step is necessary going forward for any new project/namespace creation.

weliang1:
> @weliang1 perhaps we should add that step into `Given I have a project` instead...since this step is necessary going forward for any new project/namespace creation.

@pruan-rht Good suggestion. It will be great if we can do that way from beginning of 4.12. Right now, there are a lots of existing cases which have already updated to have both `Given I have a project`
and  `And the appropriate pod security labels are applied to the namespace` in their code, if we add pod security labels steps to `Given I have a project`, those cases will do the double configurations.

pruan-rht:
@weliang1 do you know if double labeling will cause an error?  @jhou1 @liangxia WDYT?  I think we should fold it into the `Given I have a project` step to hide the details and less work going forward.  Another question is that if that action will have bad effects for older OCP version.

weliang1:
> @weliang1 do you know if double labeling will cause an error? @jhou1 @liangxia WDYT? I think we should fold it into the `Given I have a project` step to hide the details and less work going forward. Another question is that if that action will have bad effects for older OCP version.

@pruan-rht I think double labeling on ns will throw a error which will say the labels have been already assigned to the ns, or something like that. And scripts execution time will be increase to run same codes twice. If we can update it to the way as you suggested, then we may need to delete `And the appropriate pod security labels are applied to the namespace` in all the cases which have that steps.  Let's get opinions from  @jhou1 @liangxia

zhaozhanqi:
> > @weliang1 do you know if double labeling will cause an error? @jhou1 @liangxia WDYT? I think we should fold it into the `Given I have a project` step to hide the details and less work going forward. Another question is that if that action will have bad effects for older OCP version.
> 
> @pruan-rht I think double labeling on ns will throw a error which will say the labels have been already assigned to the ns, or something like that. And scripts execution time will be increase to run same codes twice. If we can update it to the way as you suggested, then we may need to delete `And the appropriate pod security labels are applied to the namespace` in all the cases which have that steps. Let's get opinions from @jhou1 @liangxia

There is existing function with ` have a project with proper privilege`

```
  step %Q/I have a project/
  step %Q/the appropriate pod security labels are applied to the namespace/
end
```

PR #3156, opened 2022-12-08
fix OCP-33743

yapei:
- remove `check_required_badge_during_installation` rule since it is not used in any feature
[yapei@yapei-mac cucushift (fix)]$ grep -nr 'check_required_badge_during_installation' web/admin-console
- updated rules checking `Required` badge to be simpler and more stable


yapei:
4.10 pass log /job/ocp-common/job/Runner/689852/

PR #3159, opened 2022-12-13
remove version 4.9 from logging upgrade case

QiaolingTang:
/cc @anpingli 

To fix https://issues.redhat.com/browse/OCPQE-12997


anpingli (commented):
/lgtm

PR #3166, opened 2022-12-20
Update OCP-11619 for stability

lihongan:
Decrease the number of`rate-limit-connections.rate-tcp` to make it easy to be triggered

@melvinjoseph86 @ShudiLi PTAL, thanks.


pruan-rht (commented):
/lgtm
/approve

PR #3167, opened 2022-12-20
added step to check all clusters operators are ready

pruan-rht:
This check will exit upon the first clusteroperator failure instead of checking more since the timeout is 30 minutes for each.
@weliang1 PTAL, this should address your needs.

### DIFF
--- a/features/step_definitions/operators.rb
+++ b/features/step_definitions/operators.rb
@@ -26,6 +26,30 @@ Given /^the status of condition "([^"]*)" for "([^"]*)" operator is: (.+)$/ do |
   end
 end
 
+Given /^I wait(?: up to #{NUMBER} seconds)? for all clusteroperators to be ready$/ do |timeout_in_seconds|
+  ensure_admin_tagged
+  timeout_in_seconds ||= 30
+  seconds = Integer(timeout_in_seconds)
+  clusteroperators = BushSlicer::ClusterOperator.list(user: admin)
+  co_hash = {}
+  clusteroperators.each do |co|
+    logger.info("Getting information for #{co.name}...")
+    res = cluster_operator(co.name).wait_till_ready(admin, seconds)
+    co_hash[co.name] = res
+    unless res
+      logger.error("Operator #{co.name} in not READY")
+    end
+  end
+  cluster_operators_all_ready = co_hash.values.tally[true] == co_hash.count
+  unless cluster_operators_all_ready
+    logger.error("Not all clustersoperators are ready")
+    logger.info("Listing the clusteroperators target status...")
+    co_hash.each do |co, status|
+      logger.info("#{co} ready status is '#{status}'")
+    end
+  end
+end
+
 Given /^I create a new CatalogSourceConfig$/ do
   ensure_admin_tagged
   # Create CatalogSourceConfig in 4.5-
--- a/lib/openshift/cluster_operator.rb
+++ b/lib/openshift/cluster_operator.rb
@@ -18,5 +18,24 @@ module BushSlicer
       return rr.dig('status', 'relatedObjects')
     end
 
+    def wait_till_ready(user, seconds)
+      res = nil
+      start_time = monotonic_seconds
+
+      success = wait_for(seconds) {
+        res = ready?(user: user, quiet: true)
+        break if res
+      }
+      return res
+    end
+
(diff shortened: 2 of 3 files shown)


weliang1:
Tried your new codes in my script and it passed.

/lgtm

features/step_definitions/operators.rb
```diff
@@ -26,6 +26,31 @@
   end
 end
 
+Given /^I wait(?: up to #{NUMBER} minutes)? for all clusteroperators to be ready$/ do |timeout_in_minutes|
+  ensure_admin_tagged
+  timeout_in_minutes ||= 30
```
jhou1, line 31: Are we going to wait 30 minutes for all cluster operators to be ready? This looks very long.

features/step_definitions/operators.rb
```diff
@@ -26,6 +26,31 @@
...
+  seconds = Integer(timeout_in_minutes) * 60
+  clusteroperators = BushSlicer::ClusterOperator.list(user: admin)
+  co_hash = {}
+  clusteroperators.each do |co|
+    logger.info("Getting information for #{co.name}...")
+    res = cluster_operator(co.name).wait_till_ready(admin, seconds)
```
jhou1, line 37: Looks like in the loop each cluster operator will be waiting 1800 seconds to become ready.

lib/openshift/cluster_operator.rb
```diff
@@ -18,5 +18,22 @@ def related_objects(user: nil, cached: true, quiet: false)
...
     end
 
+    def wait_till_ready(user, seconds)
+      res = nil
+      iterations = 0
+      start_time = monotonic_seconds
+
+      success = wait_for(seconds) {
+        res = ready?(user: user, quiet: true)
+        iterations = iterations + 1
```
jhou1, line 28: `iterations` is assigned but not used

features/step_definitions/operators.rb
pruan-rht: @jhou1 each co is checked with a timeout of 30 minutes for each.  In line #40, it will break out of the loop upon the first failed check.  So I think we are ok.

features/step_definitions/operators.rb
pruan-rht: the loop will exit on the first failure.

lib/openshift/cluster_operator.rb
pruan-rht: @jhou1 good catch, let's make it simple and removed it now.

features/step_definitions/operators.rb
pruan-rht: @jhou1 what timeout value do you have in mind?

jhou1: I assume on a healthy cluster we would need less than 30 seconds.

jhou1: btw, do we need to take into account of the other statues from a co, like `available: true`, `degraded: false` ?

pruan-rht: @jhou1 it depends on how the step is used.  If we keep it as is, it can be used to poll COs that takes longer to come up. 
 @weliang1 WDYT?  Currently, @weliang1 just wants to check when the COs are `ready`.

pruan-rht: @jhou1 BTW, the value for upgrade check is [20 minutes](https://gitlab.cee.redhat.com/aosqe/jenkins-jcasc-n/-/blob/master/job-dsl-lib/UpgradeCIUtils.groovy#L549)

pruan-rht: @jhou1 @weliang1's golang example if only 30 seconds.   But I think he used it after the cluster has finished spun up/upgraded.   So it does depends on when the step is called.

weliang1: > @jhou1 it depends on how the step is used. If we keep it as is, it can be used to poll COs that takes longer to come up. @weliang1 WDYT? Currently, @weliang1 just wants to check when the COs are `ready`.

@pruan-rht For my test case, 30 seconds will be good enough.

pruan-rht:
@jhou1 @weliang1 ok, I've update the timeout to be 30 seconds and not exit upon the first error, so it nows display all the operator's ready status.  I've also added a method to check arbitrary status string like `Progressing` or `Degraded`.  PTAL

PR #3170, opened 2022-12-22
Add testdata for descheduler 4.12 & modify for 4.11

kasturinarra:
I see that descheduler upgrade prepare test is failing due to missing testdata file for 4.12, so adding the same. Also some data needs to be updated in 4.11 yaml file so adding the same.


kasturinarra:

PR #3171, opened 2022-12-27
Update pod.feature

weinliu:
Remove SNO support, since it requires 2 or more nodes


weinliu:
@liangxia could you help when you got available? Thanks.

PR #3175, opened 2023-01-03
fix OCPQE-7318

QiaolingTang:
To fix failure:
```
#<RuntimeError: expression returned non-positive status: false left_operand: #<Set:

{"10.143.1.169", "10.143.1.87", "10.143.1.60", "10.143.1.89", "10.143.1.208", "10.143.1.215"}
>, right_operand: #<Set:

{"10.143.1.87", "10.143.1.169", "10.143.1.208", "10.143.1.60", "10.143.1.89", "10.143.1.215", "10.143.1.43"}
> >
features/step_definitions/common.rb:168:in `/^(?:the )?expression should be true> (.+)$/'
```
Some nodes have been scaled up and some nodes have been scaled down before this case, collector pods are running during that period and logs in these nodes are all collected. When executing this case, the nodes in the cluster are not the same as the nodes in ES data, here change to check if the node IPs in ES data contain all the nodes' IP in the cluster. 

Log: http://pastebin.test.redhat.com/1086663

### DIFF
--- a/features/logging/collector.feature
+++ b/features/logging/collector.feature
@@ -231,14 +231,14 @@ Feature: collector related tests
       | op           | GET   |
     Then the step should succeed
     Given evaluation of `@result[:parsed]['aggregations']['exists_field_kubernetes']['distinct_node_ip']['buckets'].map {|furn| furn["key"]}` is stored in the :kuber_ips clipboard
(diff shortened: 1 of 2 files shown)


anpingli (commented):
/lgtm

PR #3180, opened 2023-01-09
add flaky for OCP-41803 as of bug 1916575

sunzhaohua2:
Add flaky for OCP-41803, met this cases failed several times, the node became NotReady during/after upgrade, causing the upgrade to fail because of bug https://bugzilla.redhat.com/show_bug.cgi?id=1916575,once this bug is verified, will remove flaky label. @jhou1 @miyadav @huali9 PTAL, thanks!
https://issues.redhat.com/browse/OCPQE-12006
https://issues.redhat.com/browse/OCPQE-11993
https://issues.redhat.com/browse/OCPQE-12680


sunzhaohua2:
@jhou1 @liangxia PTAL

PR #3195, opened 2023-01-24
Add logging5.6 and logging5.7 tag for Logging cases in order for us to run the cases on Prow.

IshwarKanse:
Added the logging5.6 tag so that we can filter and run these Logging cases from Prow.

### DIFF
--- a/features/logging/cluster_log_forwarder.feature
+++ b/features/logging/cluster_log_forwarder.feature
@@ -139,6 +139,7 @@ Feature: cluster log forwarder features
   @destructive
   @singlenode
   @4.7 @4.6
+  @logging5.6 @logging5.7
   @vsphere-ipi @openstack-ipi @nutanix-ipi @ibmcloud-ipi @gcp-ipi @baremetal-ipi @azure-ipi @aws-ipi @alicloud-ipi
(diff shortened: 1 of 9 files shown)


features/logging/cluster_log_forwarder.feature
```diff
@@ -139,6 +139,7 @@ Feature: cluster log forwarder features
   @destructive
   @singlenode
   @4.7 @4.6
+  @logging5.6
```
kabirbhartiRH, line 142: Are we only planning to run 5.6 on prow?

IshwarKanse: After the necessary steps for logging on Prow are added in the PR, which currently contains 5.6, we will simply need to add the Logging 5.7 configuration. In the meantime, I've went ahead and added the logging5.7 tag.

kabirbhartiRH: Not sure if we are planning to support 5.5.z runs on prow currently. My guess would be yes but we could update it later with a separate PR once we test 5.6 on prow.

QiaolingTang:
Hi @IshwarKanse, FYI, we have some cases in https://github.com/openshift/cucushift/tree/master/logging, some of these cases should have the tags too.

PR #3196, opened 2023-01-24
To add ppc64le architecture tags

aishwaryabk:
This is an initial PR for adding `ppc64le` tags to tests that are tested on Power.

cc: @pravin-dsilva @sajauddin

### DIFF
--- a/features/admin/pod.feature
+++ b/features/admin/pod.feature
@@ -9,7 +9,7 @@ Feature: pod related features
   @singlenode
   @network-ovnkubernetes @network-openshiftsdn
   @proxy @noproxy
-  @heterogeneous @arm64 @amd64
+  @heterogeneous @arm64 @amd64 @ppc64le
   @critical
   Scenario: OCP-15808:Node Endpoints should update in time and no delay
     Given I have a project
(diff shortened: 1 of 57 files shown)


features/admin/pod.feature
```diff
@@ -10,6 +10,7 @@ Feature: pod related features
   @network-ovnkubernetes @network-openshiftsdn
   @proxy @noproxy
   @heterogeneous @arm64 @amd64
+  @ppc64le
```
aleskandro, line 13: Hi, is it possible to add them in the same row as the other architectures?

aleskandro:
There's an empty file (`=`) in the root of your branch: https://github.com/openshift/verification-tests/blob/52cdb04e9bb08e4689f105bae047e21dcdb11113/%3D

features/admin/quota.feature
```diff
@@ -177,6 +182,7 @@ Feature: Quota related scenarios
   # @case_id OCP-11566
   @admin
   @inactive
+  @ppc64le
```
aleskandro, line 185: for inactive cases, we never verified they work on other architectures than amd64, and we usually skip these in the CI. Are you sure to add them?

pruan-rht: if it's inactive it's not going to be run...so there's probably no need to add it.

pruan-rht:
@aishwaryabk  need to rebase

features/admin/pod.feature
aishwaryabk: @aleskandro Thanks for reviewing. Yes, I've added them in the same line as other architectures.

features/admin/quota.feature
aishwaryabk: I've added the ppc64le tag only where there are other architecture labels present

aishwaryabk:
Thanks for reviewing. I have updated the PR

liangxia:
We need to add such info in Polarion (e.g, in arch field for this case), then map that info into tags in the code.

PR #3198, opened 2023-01-31
Limit selenium-webdriver to 4.8.0

liangxia:
We've been experience breaking UI automation issue several times from selenium-webdriver due to newer version comes out. So we'd like to limit selenium-webdriver to a specific working version to avoid such issue from happening again.

The idea is that we want to limit most of the gems, but change only one gem each time, get it in and let the automation run for a week or two, move to next gem if no issue found/report.

/cc @yapei @yanpzhan @XiyunZhao 
/cc @pruan-rht @jhou1

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,12 +37,14 @@ gem 'azure_mgmt_network', '~>0.17.0'
 # gem 'timers'
 ## Logging
 gem 'term-ansicolor'
+
 ## Webauto
 gem 'watir'
 gem 'headless'
-gem 'selenium-webdriver'
+gem 'selenium-webdriver', '4.8.0'
 gem 'protobuf'
 gem 'reportportal'
+
 ## Docs
 # beware https://github.com/pry/pry/issues/1465
 #        https://bugzilla.redhat.com/show_bug.cgi?id=1257578


yapei:
thank for the fix @liangxia 
/lgtm

pjanouse:
Hello folks,
this PR introduced a regression for RHEL8 where Ruby version is `2.5.9p229` the latest available. I understand you intention of stable test environment however `selenium-webdriver v4.8.0` requires ruby>=2.7. Would you mind to lower the requested version of this package which can be satisfied by the distributed Ruby version for RHEL8?
Pavel

liangxia:
We are using [ruby 2.7](https://github.com/openshift/verification-tests/blob/master/tools/jenkins/slaves/Dockerfile.rhel8#L29-L31) in Jenkins CI and Prow CI.
```
bash-4.4$ cat /etc/redhat-release 
Red Hat Enterprise Linux release 8.7 (Ootpa)

bash-4.4$ ruby --version
ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [x86_64-linux]

bash-4.4$ gem list selenium
*** LOCAL GEMS ***
selenium-webdriver (4.8.0)
```

PR #3200, opened 2023-01-31
Set browser to chrome in README

liangxia:
Chrome is the suggested browser from UI team, so update the README to use chrome instead of firefox, to avoid external teams to met unexpected issues when first adapt to cucushift tests.

/cc @yapei @yanpzhan @jhou1 @pruan-rht


yapei (approved):
thank you for the updates
/lgtm

PR #3201, opened 2023-02-01
Download webdrivers automatically

liangxia:
Previously, we use `curl` to download both latest `chromedriver` (required by browser Chrome) and latest `geckodriver` (required by browser Firefox) in the dockerfile/image, no matter whether we are using chrome or firefox for the tests.

According to http://watir.com/guides/drivers/,
We recommend using the [webdrivers gem](https://rubygems.org/gems/webdrivers) to automatically ensure that the latest driver is downloaded, and placed in a location where Selenium can access it.

The version of `chromedriver` will depend on the version of Chrome you are using it with:
For versions >= 70, the downloaded version of `chromedriver` will match the installed version of Google Chrome. More information [here](http://chromedriver.chromium.org/downloads/version-selection).
For versions <= 69, `chromedriver` version 2.41 will be downloaded.
For beta versions, you'll have to require the beta version of `chromedriver` using `Webdrivers::Chromedriver.required_version`.

/cc @yapei @yanpzhan @XiyunZhao @jhou1 @pruan-rht

PR #3205, opened 2023-02-06
Revert "Download webdrivers automatically"

liangxia:
Reverts openshift/verification-tests#3201

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,6 @@ gem 'azure_mgmt_network', '~>0.17.0'
 gem 'term-ansicolor'
 
 ## Webauto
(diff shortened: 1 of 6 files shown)


yapei:
/remove-label lgtm
It looks like `openshift-ci/Dockerfile` is not reverted

liangxia:
> It looks like `openshift-ci/Dockerfile` is not reverted

That file is already reverted in https://github.com/openshift/verification-tests/pull/3204

liangxia:
Move ahead as the UI tests in Prow CI are all failing at the moment.
/approve

PR #3211, opened 2023-02-09
update case 19552

zhaozhanqi:
some times the cluster already exit the coredumpctl but not this case related.  So here just add one step to make sure the cluster is normal at the beginnig.

to resolvv https://issues.redhat.com/browse/OCPQE-13897

and with another PR https://github.com/openshift/cucushift/pull/9462


zhaozhanqi:

PR #3216, opened 2023-02-13
OCPQE-13529: Selenium: Use Options, as Remote::Capabilities is deprecated

liangxia:
To address the issue
```
2023-02-09 04:36:39 WARN Selenium [DEPRECATION] [:caps_browsers] Remote::Capabilities.chrome is deprecated. Use Options.chrome instead.
```
/cc @yanpzhan @yapei @XiyunZhao

### DIFF
--- a/lib/webauto/web4cucumber.rb
+++ b/lib/webauto/web4cucumber.rb
@@ -74,7 +74,7 @@ require_relative 'chrome_extension'
     def browser
       return @browser if @browser && @browser.exists?
       firefox_profile = Selenium::WebDriver::Firefox::Profile.new
-      chrome_caps = Selenium::WebDriver::Remote::Capabilities.chrome()
+      chrome_opts = Selenium::WebDriver::Options.chrome
       safari_caps = Selenium::WebDriver::Remote::Capabilities.safari()
       chrome_switches = []
       if ENV.has_key?("http_proxy") || @http_proxy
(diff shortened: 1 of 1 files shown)


liangxia:
I've verified the change fix issue `Remote::Capabilities.chrome is deprecated`

/cc @yanpzhan @yapei @XiyunZhao

lib/webauto/web4cucumber.rb
```diff
@@ -136,20 +136,13 @@ def browser
         end
       elsif @browser_type == :chrome
         logger.info "Launching Chrome"
-        #https://bugs.chromium.org/p/chromium/issues/detail?id=1056073
-        chrome_caps[:acceptInsecureCerts] = true
-        if Integer === @scroll_strategy
-          chrome_caps[:element_scroll_behavior] = @scroll_strategy
-        end
```
akostadinov, line 143: I don't remember details but I remember that we needed this scroll strategy to find some elements that were otherwise hidden behind UI menus. Maybe not needed with latest UI. idk, but just FYI in case you start seeing strange errors with elements not being visible.

yapei: yeah, we added this code to scroll element into view before taking action on it @liangxia  Can we achieve the same goal with new method?

liangxia: @akostadinov  Nice to see you again. Thank you for the remind.
We've compared the test results in master branch and devel branch. Total 230 test cases, and we run the same test on four different clusters on both master branch and devel branch, we only got 2, 2, 4, 3 failures for devel branch in each test run, while 4, 3, 3, 3 failures in master branch. So I think it's fine to remove it.

akostadinov: I think we also had a custom scroll code to try to scroll when element overlap was detected. So maybe it works :)

liangxia:
Replaced by https://github.com/openshift/verification-tests/pull/3224, as using devel branch from verification-tests trigger the test of branch-ci-openshift-verification-tests-devel-images-arm64 which is unexpected.
/cc @yanpzhan

PR #3221, opened 2023-02-15
Fix the changes for OCP-42205

XiyunZhao:
This change is only needed for OCP4.13
Pass log: Runner/746543/console


XiyunZhao:
@yapei This PR is ready for review, could you help to take a look

PR #3226, opened 2023-02-20
Update OCP-14799 to inactive

weinliu:
Update OCP-14799 to inactive for now, debugging sill need more time, it will be covered in manual run before fixed.

### DIFF
--- a/features/infrastructure/cluster-capacity.feature
+++ b/features/infrastructure/cluster-capacity.feature
@@ -4,7 +4,7 @@ Feature: cluster-capacity related features
   # @author weinliu@redhat.com
   # @case_id OCP-14799
   @admin
-  @flaky
+  @inactive
   @4.12 @4.11 @4.10 @4.9 @4.8 @4.7 @4.6
   @vsphere-ipi @openstack-ipi @gcp-ipi @baremetal-ipi @azure-ipi @aws-ipi
   @vsphere-upi @openstack-upi @gcp-upi @baremetal-upi @azure-upi @aws-upi


weinliu:
@liangxia could you help? Thanks

features/infrastructure/cluster-capacity.feature
```diff
@@ -4,7 +4,7 @@ Feature: cluster-capacity related features
   # @author weinliu@redhat.com
   # @case_id OCP-14799
   @admin
-  @flaky
+  @inactive
```
liangxia, line 7: `flaky` for short term, while `inactive` for long term.
Based on the description, sounds like you still need the test, just wan to disable it for a short time.

weinliu: yes, `flaky` will have the case go to waiting state in the test run, and the team got pinged frequently, so prefer to have it being `inactive` for a while

liangxia: Change to inactive in code does not resolve the issue, as test case selection is done via interacting with Polarion.
If you would like the test case not picking up in Jenkins CI for some time, it's suggested to move the test case into "not automated" status in Polarion.

jhou1:
I'm receiving a daily notification to approve this, so, @weinliu it looks like is pending your decision.

`@inactive` is when you no longer want this test performed. As Liang's comment, you need to make your test case inactive too, otherwise you are taking a risk your test is automated but you don't have it actually running.
`@flakly` is when you want to work with a PR to fix the automation later and decide to temporarily disable the test.

Let us know your choice, thanks.

weinliu:
Resolved in another PR. Thanks

PR #3237, opened 2023-03-01
Add testdata file for kubedescheduler 4.13

kasturinarra:
Currently testdata file for kubedescheduler 4.13 is missing, so adding the same.


kasturinarra:

PR #3248, opened 2023-03-07
Stop running 100% failing test scenarios, cloud and node

jhou1:
According to the webapp, we have lots of 100% failure scenarios, this is a partial attempt to stop executing these tests until they have value again. Please consider re-enabling them provided they are failing due to product bugs or you have an enhancement.

cc @sunilcio @sunzhaohua2 for awareness.

### DIFF
--- a/features/machine/upi_gcp.feature
+++ b/features/machine/upi_gcp.feature
@@ -14,6 +14,7 @@ Feature: UPI GCP Tests
 
   # @author zhsun@redhat.com
   # @case_id OCP-25034
+  @flaky
   @admin
   @destructive
   @4.13 @4.12 @4.11 @4.10 @4.9 @4.8 @4.7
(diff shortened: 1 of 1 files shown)


sunzhaohua2:
Disable these 2 cases are ok for me, as they are for upi gcp. Maybe machineset yaml files were removed on prow becasue they are upi clusters. I will set up a upi gcp cluster on prow to confirm.

sunzhaohua2:
I submit a pr to skip the testing in such clusters, @jhou1 @pruan-rht ptal https://github.com/openshift/verification-tests/pull/3250

jhou1:
@pruan-rht As a matter of fact, they have already been ignored for a period before being tagged flaky, so I do not consider this action to be the factor causing things to be ignored. I would not want to wait for an investigation and only worry about reducing the distractions from these unattended tests. 

Thanks for making a quick fix @sunzhaohua2 . After this is merged you can rebase and remove the tag, thank you.

pruan-rht:
@jhou1 let's think about how to get the teams to pay attentions to these kind of low passing tests.

PR #3250, opened 2023-03-08
skip testing on upi clusters which don't have machineset

sunzhaohua2:
As @jhou1 mentioned in pr https://github.com/openshift/verification-tests/pull/3248, these 2 cases OCP-34697 and  OCP-25034 always fail in prow as there are no machineset, this will skip the testing in such clusters. @jhou1 @huali9 @miyadav PTAL


sunzhaohua2:
Close this as covered in https://github.com/openshift/verification-tests/pull/3254

PR #3252, opened 2023-03-09
[WIP] Added Namespace

jkuriako:
Add namespace for all "I run commands on the host" commands.


pruan-rht:
@jkuriako why is the project name needed now?

PR #3254, opened 2023-03-13
skip for no machineset on UPI

huali9:
OCP-34697 and OCP-25034 failed on GCP UPI cluster which has no machineset, 
`Machinesets are not enabled, there are no machines (RuntimeError)`
but for such case, we should skip the test, not reporting error. so update to skip the test cases for such case.
@sunzhaohua2 @miyadav @jhou1 PTAL, thanks!

test cases passed on GCP UPI cluster which has machineset:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/763586/console
test cases skipped on GCP UPI cluster which has no machineset:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/763597/console

### DIFF
--- a/features/step_definitions/machine.rb
+++ b/features/step_definitions/machine.rb
@@ -20,12 +20,16 @@ Given(/^I have an UPI deployment and machinesets are enabled$/) do
   # In an UPI deployment, if enabled machinesets, machines should be linked to nodes
   machines = BushSlicer::MachineMachineOpenshiftIo.list(user: admin, project: project("openshift-machine-api"))
   if machines.length == 0
-    raise "Machinesets are not enabled, there are no machines"
+    logger.warn "Machinesets are not enabled, there are no machines"
+    logger.warn "We will skip this scenario"
+    skip_this_scenario
   end
 
   machines.each do | machine |
     if machine.node_name.nil?
-      raise "machine #{machine.name} does not have nodeRef."
+      logger.warn "machine #{machine.name} has no node ref."
+      logger.warn "We will skip this scenario"
+      skip_this_scenario
     end
   end
 end
--- a/features/machine/upi_gcp.feature
+++ b/features/machine/upi_gcp.feature
@@ -2,7 +2,6 @@ Feature: UPI GCP Tests
 
   # @author zhsun@redhat.com
   # @case_id OCP-34697
-  @flaky
   @admin
   @destructive
   @4.13 @4.12 @4.11 @4.10 @4.9 @4.8 @4.7 @4.6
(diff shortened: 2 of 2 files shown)


sunzhaohua2:
Thanks @huali9, in Jenkins upi gcp we didn't remove worker machineset manifests, but in Prow removed worker  machineset manifests, so in jenkins upi gcp clusters have machineset, but in Prow upi gcp clusters doesn't have machineset. 
Please also help to update line 30 to 
```
      logger.warn "machine #{machine.name} has no node ref."
      logger.warn "We will skip this scenario"
      skip_this_scenario
```

huali9:
> Thanks @huali9, in Jenkins upi gcp we didn't remove worker machineset manifests, but in Prow removed worker machineset manifests, so in jenkins upi gcp clusters have machineset, but in Prow upi gcp clusters doesn't have machineset. Please also help to update line 30 to
> 
> ```
>       logger.warn "machine #{machine.name} has no node ref."
>       logger.warn "We will skip this scenario"
>       skip_this_scenario
> ```

Should we update line 30? as we skip test when there is no machines in line 22, so if it goes to line 30, there should have machine on the cluster, and the machine should has node ref, right? if hasn't, we should report error and investigate that, isn't it?

sunzhaohua2:
>  if hasn't, we should report error and investigate that, isn't it?

yes, I think it should be a specific case fails, rather than one case fails all cases fail, wdyt?

huali9:
> > if hasn't, we should report error and investigate that, isn't it?
> 
> yes, I think it should be a specific case fails, rather than one case fails all cases fail, wdyt?

make sense, updated, thanks @sunzhaohua2

features/step_definitions/machine.rb
```diff
@@ -20,12 +20,16 @@
...
   machines = BushSlicer::MachineMachineOpenshiftIo.list(user: admin, project: project("openshift-machine-api"))
   if machines.length == 0
-    raise "Machinesets are not enabled, there are no machines"
+    logger.warn "Machinesets are not enabled, there are no machines"
+    logger.warn "We will skip this scenario"
+    skip_this_scenario
   end
 
   machines.each do | machine |
     if machine.node_name.nil?
```
miyadav, line 29: This is condition for unstable cluster , we can not skip , in that case.

miyadav, line 29: This is condition for unstable cluster , we can not skip , in that case.

huali9: @miyadav According to Zhaohua's comment, we can debug the failure in that specific failed case, not needing to debug all the cases which failed because of other case failed. https://github.com/openshift/verification-tests/pull/3254#issuecomment-1465882362

PR #3259, opened 2023-03-15
Add @cloud tag for upgrade case running in Prow

sunzhaohua2:
@jhou1 @liangxia @miyadav @huali9 PTAl, this is for https://github.com/openshift/release/pull/37261 run machineset scale up/down in prow upgrade jobs.
```
$ bundle exec cucumber --tags '@upgrade-check and @4.12 and @aws-ipi and not @fips and not @customer and not @flaky and not @inactive and not @prod-only and not @qeci and not @security and not @stage-only and @cloud' --dry-run
1 scenario (1 skipped)
```

### DIFF
--- a/features/upgrade/machines/machines.feature
+++ b/features/upgrade/machines/machines.feature
@@ -131,6 +131,7 @@ Feature: Machine-api components upgrade tests
   @upgrade-check
   @admin
   @destructive
+  @cloud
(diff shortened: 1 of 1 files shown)


~
```diff
@@ -0,0 +1,365 @@
+Feature: Machine-api components upgrade tests
```
liangxia, line 1: An unexpected file introduced ?

PR #3262, opened 2023-03-17
Fix for hosted hypershift

asood-rh:
@openshift/team-sdn-qe PTAL

The hypershift hosted clusters do not have master nodes, therefore changing the step definition to use multus pod running on specific node.
The step definition change to use multus pod on  worker node other than the one running the back end pod of the service. 

Tested on cluster created with profile private-templates/functionality-testing/aos-4_13/ipi-on-aws/versioned-installer-ovn-hypershift-ci
Test logs can be found here 
http://file.rdu.redhat.com/asood/OCP-41944-hh.log
http://file.rdu.redhat.com/asood/OCP-40908-hh.log
http://file.rdu.redhat.com/asood/OCP-41083-hh.log

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1363,15 +1363,13 @@ Given /^the node's active nmcli connection is stored in the#{OPT_SYM} clipboard$
   logger.info "Node's active nmcli connection uuid is stored in the #{cb_name} clipboard as #{cb[cb_name]}"
 end
 
-Given /^I save multus pod on master node to the#{OPT_SYM} clipboard$/ do | cb_name |
+Given /^I save multus pod on node "([^"]*)" to the#{OPT_SYM} clipboard$/ do | node_name,cb_name |
   ensure_admin_tagged
   cb_name ||= :multuspod
-  master_nodes = env.nodes.select { |n| n.schedulable? && n.is_master? }
-  master_node_names = master_nodes.collect { |n| n.name }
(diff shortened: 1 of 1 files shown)


pruan-rht:
@asood-rh when changing a step definition, we need to check that the step is regressed properly.   A search through the repo indicate that it's used in a couple of feature files (see below).  Therefore your changes would break those those scenarios.  
```
pruan@mac ~/workspace/BushSlicer (master)$ rg 'I save multus pod on master node'
features/step_definitions/networking.rb
1366:Given /^I save multus pod on master node to the#{OPT_SYM} clipboard$/ do | cb_name |

features/upgrade/sdn/policy-upgrade.feature
352:    Given I save multus pod on master node to the :multuspod clipboard
407:    Given I save multus pod on master node to the :multuspod clipboard
```

asood-rh:
That is the first rule of changing step definition. As there are two repos
- cucushift and verification- tests , you will need to check for open PR in
the other repo as well before the concerns.

pruan-rht:
@asood-rh ok, thanks for the clarification.  Yeah, I recommend adding the reference link to the other PR in the original PR comment so the reviewers will know as well.

asood-rh:
> @asood-rh ok, thanks for the clarification. Yeah, I recommend adding the reference link to the other PR in the original PR comment so the reviewers will know as well.

@pruan-rht The test logs were included in the PR indicative of step definition being tested.

pruan-rht:
@asood-rh yes, I'm just suggesting that it's more efficient to link the other PR so the reviewers don't need to thumb through the logs

asood-rh:
@zhaozhanqi Could you add /approve label to this PR as somehow it is not accepting label Anurag added?

PR #3263, opened 2023-03-17
Revert "update securityContext for some test pods"

lihongan:
Reverts openshift/verification-tests#3256

I see some errors in 4.9 and the pod cannot be created, so revert the change for now.
```
Error from server (Forbidden): error when creating "web-server-1.yaml": pods "web-server-1" is forbidden: unable to validate against any security context constraint
```

@melvinjoseph86 @ShudiLi PTAL. thanks

### DIFF
--- a/testdata/routing/test-client-pod.yaml
+++ b/testdata/routing/test-client-pod.yaml
@@ -11,6 +11,3 @@ spec:
     ports:
     - containerPort: 8080
     - containerPort: 8443
-  securityContext:
-    seccompProfile:
-      type: RuntimeDefault
--- a/testdata/routing/web-server-1.yaml
+++ b/testdata/routing/web-server-1.yaml
@@ -11,6 +11,3 @@ spec:
     ports:
     - containerPort: 8080
     - containerPort: 8443
-  securityContext:
-    seccompProfile:
-      type: RuntimeDefault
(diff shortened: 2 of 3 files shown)


lihongan:
In 4.9 the pod spec looks like
```
spec:
  containers:
  - image: quay.io/openshifttest/nginx-alpine@sha256:04f316442d48ba60e3ea0b5a67eb89b0b667abf1c198a3d0056ca748736336a0
<---snip--->
    securityContext:
      capabilities:
        drop:
        - KILL
        - MKNOD
        - SETGID
        - SETUID
      runAsUser: 1000700000

  securityContext:
    fsGroup: 1000700000
    seLinuxOptions:
      level: s0:c26,c25

```

melvinjoseph86:
@lihongan i guess we cannot made the change build wise. right?
As this is yaml we cannot set `run for version >= 4.13` for the new change?

lihongan:
> @lihongan i guess we cannot made the change build wise. right?
> As this is yaml we cannot set `run for version >= 4.13` for the new change?

No, we cannot do that for cucushift, it is different from Ginkgo. This repo uses master branch for all 4.x version.
We are safe to revert it since no automation issues with the old ones.

lihongan:
see automation error in 4.10 as well

PR #3266, opened 2023-03-21
fix operator cases

yapei:


lib/rules/web/admin_console/4.10/common_ui_elements.xyaml
```diff
@@ -730,9 +730,14 @@ check_no_results_text:
 submit_changes:
   elements:
   - selector:
-      xpath: //button[@type='submit' and not(@disabled)]
+      xpath: //button[@id='save-changes' and not(@disabled)]
```
yapei, line 733: using old selector sometimes returned 2 element which will cause clicking fail

lib/rules/web/admin_console/4.10/operator_hub.xyaml
```diff
@@ -185,6 +195,7 @@ check_operand_des_during_creating_operand:
   action: goto_csv_detail_page
   action: click_tab
   action: click_create_operand_button
+  action: switch_to_form_view
```
yapei, line 198: make sure we are on form view so that we can check the description

PR #3267, opened 2023-03-21
removed hypershift-hosted labels from few cases

anuragthehatter:
These tests are have masters dependency and shouldn't be run on hosted clusters

@openshift/team-sdn-qe

### DIFF
--- a/features/upgrade/sdn/policy-upgrade.feature
+++ b/features/upgrade/sdn/policy-upgrade.feature
@@ -461,7 +461,6 @@ Feature: SDN compoment upgrade testing
   @network-ovnkubernetes @network-openshiftsdn
   @proxy @noproxy @disconnected @connected
   @heterogeneous @arm64 @amd64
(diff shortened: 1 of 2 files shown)


anuragthehatter:
Also need to temporarily disable conntrack upgrade case on Hypershift hosted clusters due to bug https://issues.redhat.com/browse/OCPBUGS-10604. Will track it internally to re-enable once bug gets fixed.

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -543,7 +542,6 @@ Feature: SDN compoment upgrade testing
   @network-ovnkubernetes @network-networkpolicy
   @upgrade
   @proxy @noproxy @disconnected @connected
-  @hypershift-hosted
```
LiangquanLi930, line 546: Sorry, I was wrong, we should remove 427

PR #3270, opened 2023-03-22
try to fix chrome crash issue

yapei:
Added `--headless` and `--disable-extensions` options when starting Chrome

### DIFF
--- a/lib/webauto/web4cucumber.rb
+++ b/lib/webauto/web4cucumber.rb
@@ -137,7 +137,7 @@ require_relative 'chrome_extension'
       elsif @browser_type == :chrome
         logger.info "Launching Chrome"
         if self.class.container?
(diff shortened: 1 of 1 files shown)


yapei:
I just did a regression test to make sure nothing is broken https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/768633/console 
will test the changes in Prow after it is merged

yapei:
@liangxia  @yanpzhan Could you help review?

yanpzhan:
I remember we meet chrome issue in prow once, and remove --headless resloved that issue. I'm wonder if there is regression issue after add the option back.

yapei:
> I remember we met chrome issue in prow once, and remove --headless resloved that issue. I'm wonder if there is regression issue after add the option back.

Then let's try to add `--disable-extensions` option first

PR #3272, opened 2023-03-23
add more debugging options

yapei:


yapei:
@liangxia Could you help approve?

PR #3273, opened 2023-03-23
fix failures in logging 5.7

QiaolingTang:
Fix failures in logging 5.7

/cc @anpingli @IshwarKanse


anpingli:
I don't like the directory name '5.7'. But  there isn't the othe good name to be used.

anpingli (commented):
/lgtm

PR #3275, opened 2023-03-24
fix chrome crash issue

yapei:


lib/webauto/web4cucumber.rb
```diff
@@ -137,7 +137,7 @@ def browser
       elsif @browser_type == :chrome
         logger.info "Launching Chrome"
         if self.class.container?
-          chrome_switches.concat %w[--no-sandbox --disable-setuid-sandbox --disable-gpu --disable-infobars --disable-dev-shm-usage]
+          chrome_switches.concat %w[--no-sandbox --headless --disable-setuid-sandbox --disable-gpu --disable-infobars --disable-dev-shm-usage --disable-extensions]
```
yapei, line 140: in case we need these flags

lib/webauto/web4cucumber.rb
```diff
@@ -963,16 +963,7 @@ def self.linux?
...
-
-      if File.exists? '/proc/1/cgroup'
-        cgroups = File.read '/proc/1/cgroup'
-        @@container = %w[kube docker lxc].any? { |pattern|
-          cgroups.include? pattern
-        }
-      else
-        @@container = false
-      end
+      !!ENV['KUBERNETES_SERVICE_HOST'] || !!ENV['DOCKER_CONTAINER_NAME'] || !!ENV['container']
```
yapei, line 966: this is the root cause of chrome crash

liangxia:
We've verified the same code changes in devel branch on Prow CI

/lgtm
/approve

PR #3276, opened 2023-03-24
fix OCPQE-14529: remote worker issue

lihongan:
To address https://issues.redhat.com/browse/OCPQE-14529 
We don't have good way to just skip profile with remote worker in cucushift, so remove tag "baremetal-upi" for now.
 
In future we can re-implement this case with Ginkgo framework and add logic for remote worker.


ShudiLi:
/lgtm
thanks

PR #3278, opened 2023-03-25
[WIP] Sunsetting dynect to address JIRA OCPQE-13992

pruan-rht:
@jianlinliu @gpei PTAL.  This is an initial check-in which is tested with [this](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Launch%20Environment%20Flexy/147048/console) run.   A couple of observations

- For `Remove VM` I don't see anything that indicate a removal of DNS entries...even for the original Dynect implementation.  I see a periodic cleanup option in the [dynect.rb](https://github.com/openshift/verification-tests/blob/master/lib/launchers/dyn/dynect.rb#L207) class.  Is that method called externally to clean up entries in Dynect?
- I find that the Remove VM job as well as from the AWS webconsole that instances can't be terminated (it can stop and change the label to begin with `terminate`).  I've filed a [JIRA](https://issues.redhat.com/browse/DPP-12477) against DDP team about it.


gpei:
> * I find that the Remove VM job as well as from the AWS webconsole that instances can't be terminated (it can stop and change the label to begin with `terminate`).  I've filed a [JIRA](https://issues.redhat.com/browse/DPP-12477) against DDP team about it.

@pruan-rht thanks for the quick turnaround. For the two questions:

>Is that method called externally to clean up entries in Dynect?

Yes, I believe it's this job doing the periodical cleanup work
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Dynect%20DNS%20clean-up/

> that instances can't be terminated

Yes, as I remember it's a known limitation in that account, once we stopped the instances, they will eventually be terminated
automatically.

lib/launchers/dns_utils.rb
```diff
@@ -0,0 +1,71 @@
+#Define two functions to creat/delete dns records, used for osp and vsphere platform
+def create_dns_records(dns_solution, instances_name_prefix, ip_api, ip_api_int, ip_apps)
```
jianlinliu, line 2: This function is copied directly from v4 flexy-template, I do not think that would work in 3.x.  In 3.x, caller can determinate what A records will be created.
```
          dns_record = "#{task[:prefix]}.#{dns_component}"
          logger.info "Creating '#{dns_record}' record for: #{ips.join(?,)}"
          fqdn = dyn.dyn_replace_a_records(dns_record, ips)
```

lib/launchers/dns_utils.rb
```diff
@@ -0,0 +1,71 @@
...
+  base_domain = nil
+  logger.info("DNS_SOLUTION: #{dns_solution}, PREFIX: #{instances_name_prefix}, IPI_API: #{ip_api}, IP_API_INT: #{ip_api_int}, IP_APPS: #{ip_apps}")
+  case dns_solution
+  when "dyndns"
```
jianlinliu, line 6: I do not think we should consider dyndns any more.

lib/launchers/dns_utils.rb
```diff
@@ -0,0 +1,71 @@
...
+      tmp_dyn.dyn_create_a_records("*.apps.#{instances_name_prefix}.#{zone}", ip_apps)
+      tmp_dyn.publish
+    rescue => e
+      logger.error "Error during creating Dynect DNS records: #{exception_to_string(e)}"
+      raise e
+    end
+  when /^route53/
+    begin
+      logger.info "Creating DNS records in AWS Route 53..."
+      aws_iaas = iaas_by_service(dns_solution.sub(/^route53_?/,"").tap {|svc| svc.empty? && svc << "AWS-CI" } )
```
jianlinliu, line 24: If we create resources on aws, what provide the credential to access aws?

lib/launchers/dns_utils.rb
pruan-rht: yeah, we can take that out

lib/launchers/dns_utils.rb
pruan-rht: the creds are used by the `cucushift-private` repo's config.yaml

jianlinliu: Why not directly use `AWS-CI` instead of using `dns_solution` for some substitution, or allow users to use which creds as parameters. `dns_solution` should not include any account info.

pruan-rht: yeah, we can do that...if it's empty it will use `AWS-CI` by default anyway.

pruan-rht:
@gpei @jianlinliu ok, so we need to decide to either follow the same approach to use a cronjob to remove the route53 DNS entries or have it call a destroy automatically when calling terminate (that's probably preferable if it's possible)

jianlinliu:
> so we need to decide to either follow the same approach to use a cronjob to remove the route53 DNS entries or have it call a destroy automatically when calling terminate (that's probably preferable if it's possible)

Both is okay for me. For 3.x, it is only in the end of maintenance phase, we only need to make it work. Using cronjob is okay for me if it is easier.

tools/launch_instance.rb
```diff
@@ -562,24 +568,34 @@ def installation_task(task, template:, erb_binding:, template_dir: nil)
...
           unless ips && !ips.empty?
             raise 'You need to specify IPs for the `a_dns` task.'
           end
-
-          dns_record = "#{task[:prefix]}.#{dns_component}"
-          logger.info "Creating '#{dns_record}' record for: #{ips.join(?,)}"
-          fqdn = dyn.dyn_replace_a_records(dns_record, ips)
+          dns_record = "#{task[:prefix]}.#{instance_prefix}"
+          logger.info "Creating #{dns_record}' record for: #{ips.join(?,)}"
+          fqdn = create_dns_records(dns_solution, instance_prefix, ip_api, ip_api_int, ip_ingress)
```
jianlinliu, line 588: The A record name is already saved in `dns_record`, but when calling `create_dns_records`, it is never used, but seem like creating some new records in the `create_dns_records` function.

pruan-rht: @jianlinliu ok, it's fixed up now.

pruan-rht:
@jianlinliu @gpei ok, I think I've addressed all of the issues raised.  
Here's the[ install run ](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Launch%20Environment%20Flexy/147058/console)
Here's the [remove run](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Remove%20VMs/76397/console)

I've added the DNS removal and it seems to work 
```
pruan@mac ~/workspace/BushSlicer (dynect_sunset●)$ tools/launch_instance.rb terminate vminfo.yml
terminating..
[23:32:42] INFO> Using /Users/pruan/workspace/private/config/.awscred credentials file.
[23:32:44] INFO> Terminating instance pruan_0328dmaster-etcd-nfs-1
[23:32:44] INFO> Terminating instance pruan_0328dnode-infra-1
[23:32:44] INFO> Terminating instance pruan_0328dnode-1
[23:32:44] INFO> DNS_SOLUTION: route53, DNS_RECORD: pruan-0328d
[23:32:44] INFO> Deleting DNS records in AWS Route 53...
[23:32:44] INFO> Using /Users/pruan/workspace/private/config/openshift-ci/.awscred credentials file.
[23:32:44] WARN> Removing Route53 records matching /.+[.]pruan-0328d[.]qe[.]devcluster[.]openshift[.]com[.]/
[23:32:50] INFO> DNS records in AWS Route 53 were deleted successfully.
```

gpei:
@pruan-rht Hi, seems something wrong when creating the DNS record, the log I see in the latest [run](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Launch%20Environment%20Flexy/147058/consoleFull) is:
```
[22:44:14] INFO> A_DNS->IPS: ["54.242.55.105"]
[22:44:14] INFO> Creating pruan-0328d' record for: 54.242.55.105
```
While the expected record suppose to be created is '*.apps.pruan-0328d' for 54.242.55.105 here.
Could you please have a check?

jianlinliu:
And suggest to run a openstack install for verification, e.g: rpm-rhel7-nfs_registry-upshift-cloudprovider-release

pruan-rht:
closing in favor of #3281 3281

PR #3283, opened 2023-04-01
Sunsetting dynect

pruan-rht:
fix issue with removing DNS in PR #3281. I've tested it against both types of install and destroy

Flexy-install:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Flexy-install/191597/consoleFull
Flexy-destroy:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Flexy-destroy/189194/console

Launch Flexy Environment:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Launch%20Environment%20Flexy/147093/consoleFull
Remove VMs:
The job will fail because it's not calling my branch, but it works when I run it locally from my laptop
```
pruan@mac ~/workspace/BushSlicer (dynect_sunset●)$ ./tools/launch_instance.rb terminate vminfo.yml
terminating..
[06:56:12] INFO> HTTP POST https://rhos-d.infra.prod.upshift.rdu2.redhat.com:13000/v3/auth/tokens
[06:56:12] INFO> HTTP POST took 0.447 sec: 201 Created | application/json 9119 bytes
[06:56:12] INFO> logged in to tenant: {"domain":{"id":"62cf1b5ec006489db99e2b0ebfb55f57","name":"redhat.com"},"id":"542c6ebd48bf40fa857fc245c7572e30","name":"openshift-qe-jenkins"}
[06:56:12] INFO> HTTP GET https://rhos-d.infra.prod.upshift.rdu2.redhat.com:13774/v2.1/servers
[06:56:13] INFO> HTTP GET took 0.848 sec: 200 OK | application/json 72488 bytes
[06:56:13] INFO> HTTP GET https://rhos-d.infra.prod.upshift.rdu2.redhat.com:13774/v2.1/servers
[06:56:14] INFO> HTTP GET took 0.745 sec: 200 OK | application/json 72488 bytes
[06:56:14] INFO> HTTP GET https://rhos-d.infra.prod.upshift.rdu2.redhat.com:13774/v2.1/servers
[06:56:15] INFO> HTTP GET took 0.728 sec: 200 OK | application/json 72488 bytes
[06:56:15] INFO> Searching for DNS records matching 0401os3-mw8
[06:56:36] INFO> Using /Users/pruan/workspace/private/config/openshift-ci/.awscred credentials file.
[06:56:36] INFO> Using /Users/pruan/workspace/private/config/openshift-ci/.awscred credentials file.
[06:56:39] INFO> Deleting DNS records in AWS Route53...
[06:56:39] WARN> Removing Route53 records matching /0401os3-mw8/
```

@jiajliu @gpei PTAL.

### DIFF
--- a/lib/launchers/amz.rb
+++ b/lib/launchers/amz.rb
@@ -616,8 +616,8 @@ module BushSlicer
     def delete_resource_records_re(re, zone_id: nil)
       logger.warn("Removing Route53 records matching #{re.inspect}")
       records = list_resource_record_sets(zone_id: zone_id)
-      if records.count > 0
-        to_delete = records.select { |r| r[:name] =~ re }
+      to_delete = records.select { |r| r[:name] =~ re }
+      if to_delete.count > 0
         logger.debug("Removing records: #{to_delete.map {|r| r[:name]}}")
         list = to_delete.map { |r| { action: "DELETE", resource_record_set: r } }
         change_resource_record_sets(zone_id: zone_id, changes: list)
--- a/lib/launchers/amz.rb
+++ b/lib/launchers/amz.rb
@@ -640,6 +640,7 @@ module BushSlicer
         {value: ip}
       }
       change_resource_record_sets(zone_id: zone_id, changes: [record])
+      return name.sub(/[.]$/,"")
     end
 
     def instance_uptime(instance)
--- a/lib/launchers/amz.rb
+++ b/lib/launchers/amz.rb
@@ -616,10 +616,14 @@ module BushSlicer
     def delete_resource_records_re(re, zone_id: nil)
       logger.warn("Removing Route53 records matching #{re.inspect}")
       records = list_resource_record_sets(zone_id: zone_id)
-      to_delete = records.select { |r| r[:name] =~ re }
-      logger.debug("Removing records: #{to_delete.map {|r| r[:name]}}")
-      list = to_delete.map { |r| { action: "DELETE", resource_record_set: r } }
-      change_resource_record_sets(zone_id: zone_id, changes: list)
+      if records.count > 0
+        to_delete = records.select { |r| r[:name] =~ re }
+        logger.debug("Removing records: #{to_delete.map {|r| r[:name]}}")
+        list = to_delete.map { |r| { action: "DELETE", resource_record_set: r } }
+        change_resource_record_sets(zone_id: zone_id, changes: list)
+      else
+        logger.info("No records found matching #{re.inspect}")
+      end
     end
 
     def create_a_records(name, ips, zone_id: nil, ttl: 180)
--- a/tools/launch_instance.rb
+++ b/tools/launch_instance.rb
@@ -886,18 +886,9 @@ module BushSlicer
           # remove route53, which should at the end of the yaml
           if (not spec.nil?) and (not spec.empty?)
             begin
-              # try to find a matching DNS records first before calling deletion
-              logger.info("Searching for DNS records matching #{spec}")
               aws_iaas = iaas_by_service("AWS-CI")
-              records = iaas_by_service('AWS-CI').list_resource_record_sets
-              name_records = records.select { |r| r[:name] =~ /#{spec}/ }
-              if name_records.count > 0
-                logger.info "Deleting DNS records in AWS Route53..."
-                dns_record_regexp = Regexp.new(/#{spec}/)
-                aws_iaas.delete_resource_records_re(dns_record_regexp)
-              else
-                logger.info("No DNS records found matching #{spec}")
-              end
+              dns_record_regexp = Regexp.new(/#{spec}/)
+              aws_iaas.delete_resource_records_re(dns_record_regexp)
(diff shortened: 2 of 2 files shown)


tools/launch_instance.rb
```diff
@@ -885,6 +882,17 @@ def terminate(vminfo)
             config: vars_file.path,
             launched_instances_name_prefix: spec[:name_prefix]
           )
+        when /^dns_component/
```
jianlinliu, line 885: Per my testing, https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Flexy-install/191932/artifact/vminfo.yml/*view*/, the key is "**:**dns_component", I am a bit concerned if it can be matched.

lib/launchers/amz.rb
```diff
@@ -616,10 +616,14 @@ def list_resource_record_sets(zone_id: nil)
     def delete_resource_records_re(re, zone_id: nil)
       logger.warn("Removing Route53 records matching #{re.inspect}")
       records = list_resource_record_sets(zone_id: zone_id)
-      to_delete = records.select { |r| r[:name] =~ re }
-      logger.debug("Removing records: #{to_delete.map {|r| r[:name]}}")
-      list = to_delete.map { |r| { action: "DELETE", resource_record_set: r } }
-      change_resource_record_sets(zone_id: zone_id, changes: list)
+      if records.count > 0
```
jianlinliu, line 619: I think `records` always >0, because 3.x install are sharing the same route53 zone as 4.x install.

pruan-rht: @jianlinliu good catch,  it should be if one line below line 620 and should be `if to_delete.count > 0`, I'll update it.

tools/launch_instance.rb
pruan-rht: @jianlinliu even though the key is a symbol is the YAML when parsing it, you can try it as a String when used with the `case` (you can try the code below to verify).  As you can see from the run log output, this parsing does work.  IMO, it's better than [your approach of deleting the element ](https://github.com/openshift/verification-tests/pull/3284/files#diff-d212a2448388579d50551d687d8495db9858e3e253a7a1c26814085acdd51df8R860-R895)
```
require 'yaml'
template_yaml = """
---
template_random:
  :config: private-templates/functionality-testing/aos-4_12/ipi-on-aws/versioned-installer
  :template: template-destroy-ipi-versioned
  :name_prefix: jialiu1
:dns_component: 0403os3-yf1
"""

vminfo = YAML.load(template_yaml)

vminfo.each do |target, spec|
  case target
  when /^template_/
    print("TARGET: #{target}\n")
  when /^dns_component/
    print("##### GOT DNS_COMPONENT\n")
  else
    print("XXXX fall-through...\n")

  end
end
```

jianlinliu: > even though the key is a symbol is the YAML when parsing it, you can try it as a String when used with the case

Good to know.

Some other reason to use the approach of deleting the element, if place `dns_component` as a dedicated case here, then whatever 4.x or 3.x install, `delete_resource_records_re` is always executed to call aws api (`list_resource_record_sets`), even it return empty results.

pruan-rht: @jianlinliu your approach is making an exception by removing the element `dns_component` to avoid it being processed by the `vminfo` loop but later on processes it, which might make it harder to understand.  You are correct in that it will be executed regardless there's a `dns_component` or not, I just feel that having a patterned workflow is preferable over making exceptions.  @gpei @liangxia @jhou1, mind giving your input on this decision?  I'm ok with either way, but would like to get some other perspectives.

jhou1: I don't fully understand both, based on my limited knowledge. 

I'd suggest not altering the value of the input parameter `vminfo` for maintainability.

I'd also suggest creating a method such as `delete_aws_dns_records` and then in `terminate` call that method conditionally to keep the method of `terminal` shorter.

The[ yaml content generated](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Flexy-install/191932/artifact/vminfo.yml/*view*/) has a colon prefixing the key, prefer to have the colon removed for readability to people building such jobs.

pruan-rht: @jhou1 the deletion portion is only 3 lines (2 if we fold the definition of the `Regexp` into the function call itself).  The YAML element parsing line `when /^dns_component/` works for both `String` or `Symbol`, so we are covered on that front.  Regarding invoking AWS API, we have to use it to remove the resources...this is better than calling AWS CLI command directly, which ultimately is a wrapper around AWS API anyway.

pruan-rht:
closing in favor of https://github.com/openshift/verification-tests/pull/3290

PR #3285, opened 2023-04-03
Update acceptance case

QiaolingTang:
Update logging acceptance case:
1. remove steps to check logging dashboards in OCP console
2. remove steps to check index management jobs. 

Log: https://url.corp.redhat.com/a5a2bc5


pruan-rht:
@QiaolingTang instead of commenting them out, how about just remove those lines instead.

QiaolingTang:
@pruan-rht I removed the commented lines, PTAL.

PR #3288, opened 2023-04-04
cases example OCP-37264 to be skipped when compact IPI cluster

miyadav:
@sunzhaohua2 @huali9 @jhou1 PTAL , added this to skip alerting case wherin the metrics is will not populate as it is worker dependent , so will skip , this will reduce failures for profiles example compact clusters.
The actual case is present in tierN repo.

### DIFF
--- a/features/step_definitions/machine_misc.rb
+++ b/features/step_definitions/machine_misc.rb
@@ -26,4 +26,14 @@ Given(/I check the cluster platform is not None$/) do
      logger.warn "We will skip this scenario"
      skip_this_scenario
   end
-end 
+end
+
+Given(/^the last provisioned machine is worker$/) do
+  machines = BushSlicer::MachineMachineOpenshiftIo.list(user: admin, project: project("openshift-machine-api"))
+  cache_resources *machines
+  puts machines.max_by(&:created_at).name.include? 'worker'
+  unless (machines.max_by(&:created_at).name.include? 'worker')
+     logger.warn "Skipping this scenario because the last provisioned machine is not worker."	  
+     skip_this_scenario
+ end
+end


features/step_definitions/machine_misc.rb
```diff
@@ -26,4 +26,14 @@
...
      skip_this_scenario
   end
-end 
+end
+
+Given(/^the last provisioned machine is worker$/) do
+  machines = BushSlicer::MachineMachineOpenshiftIo.list(user: admin, project: project("openshift-machine-api"))
+  cache_resources *machines
+  unless (machines.max_by(&:created_at).name.include? 'worker')
+     logger.warn "When no worker machine is present alerting ex-OCP37264 is not valid to test"
```
sunzhaohua2, line 35: May need to update this log, so other cases also can be used

miyadav: modified to make it look better , just wanted to have bit specific so that later we can add more cases as and when fail due to this requirement.

features/step_definitions/machine_misc.rb
```diff
@@ -26,4 +26,15 @@
      logger.warn "We will skip this scenario"
      skip_this_scenario
   end
-end 
+end
+
+Given(/^the last provisioned machine is worker$/) do
+  machines = BushSlicer::MachineMachineOpenshiftIo.list(user: admin, project: project("openshift-machine-api"))
+  cache_resources *machines
+  puts machines.max_by(&:created_at).name.include? 'worker'
```
jhou1, line 34: Suggest to remove this line

features/step_definitions/machine_misc.rb
```diff
@@ -26,4 +26,15 @@
...
+  unless (machines.max_by(&:created_at).name.include? 'worker')
+     logger.warn "When no worker machine is present ,workers related cases like workers metrics not valid to test"
```
jhou1, line 36: `logger.warn "Skipping this scenario because the last provisioned machine is not worker."`

features/step_definitions/machine_misc.rb
```diff
@@ -26,4 +26,15 @@
...
+     logger.warn "We will skip this scenario"
```
jhou1, line 37: Remove this line.

features/step_definitions/machine_misc.rb
miyadav: Done

features/step_definitions/machine_misc.rb
miyadav: Done

PR #3289, opened 2023-04-04
Update descheduler rb file to have project name

kasturinarra:
with out the fix here i see that test is failing because it is unable to find kubedescheduler cluster object as it is switching over to openshift-monitoring project with the recent updated code.


kasturinarra:
@wewang58 could you please help review ? thanks !!
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/777854/

wewang58:
@liangxia  Could you help to merge it, thanks

PR #3292, opened 2023-04-10
update rules for deploy from image

yapei:
in OCP 4.14, `Resource type` was shown on the deploy page by default, hence removed the logic to click on `Resource type` button
Screenshot 2023-04-10 at 10 05 16 AM


yapei:
4.14 regression pass log Runner/781485/console

PR #3299, opened 2023-04-14
add step to generate audit logs

QiaolingTang:
When testing on rosa-hypershift-guest cluster, the cluster sometimes doesn't have audit logs, here add a step to make node generates audit logs by adding an audit rule on worker node to monitor the files under `/var/log/pods/` directory. 

/cc @anpingli @IshwarKanse

### DIFF
--- a/features/logging/logging_acceptance.feature
+++ b/features/logging/logging_acceptance.feature
@@ -31,10 +31,43 @@ Feature: Logging smoke test case
     Then the step should succeed
     Given a pod becomes ready with labels:
       | run=centos-logtest,test=centos-logtest |
+    And evaluation of `pod.node_name` is stored in the :node clipboard
 
     Given I switch to the first user
     Given the first user is cluster-admin
     And evaluation of `user.cached_tokens.first` is stored in the :user_token_1 clipboard
+
+    Given I register clean-up steps:
+    """
+    When I run the :debug admin command with:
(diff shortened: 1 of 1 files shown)


IshwarKanse:
@QiaolingTang Do we also need to add steps for generating the OVN audit logs ? Are you adding it as part of a different PR ?

QiaolingTang:
> @QiaolingTang Do we also need to add steps for generating the OVN audit logs ? Are you adding it as part of a different PR ?

@IshwarKanse This case is only executed in OCP CI, not logging's CI, it should cover all profiles, but OVN audit logs can only be generated when the network type is OVN, so  I won't add steps to check what kind of audit logs the collector pods have collected.

IshwarKanse:
@QiaolingTang Ack, this looks fine. Thanks.

pruan-rht:
@QiaolingTang can you please explain why you call the steps in one place and then also at the `clean-up` stage?  Also, perhaps think about making it a generic step to be called instead if it's called multiple times.

QiaolingTang:
@pruan-rht They're not the same command, one is to add an audit rule to the worker node, another one is to remove the audit rule from the node.  When testing the case on rosa-hypershift-cluster, the case fails at the step of checking audit logs as there is no audit logs in worker nodes. I add an audit rule to make the worker node generate some audit logs, then the collector pods can collect these logs.

PR #3309, opened 2023-04-27
To generate ImageStream with import-mode

aishwaryabk:


features/cli/oc_import_image.feature
```diff
@@ -86,10 +86,12 @@ Feature: oc import-image related feature
       | source_type | docker                                                |
       | source      | quay.io/openshifttest/deployment-example:v1-1.2.0 |
       | dest        | deployment-example:latest                             |
+      | import-mode | PreserveOriginal                                      |
```
liangxia, line 89: Do we support this option in 4.10 ?

aishwaryabk: This is a new feature supported in 4.14

liangxia: Then it might break the tests on 4.10/4.11/4.12/4.13

PR #3311, opened 2023-04-28
ignore normal messages when checking logs in ES pod

QiaolingTang:
To fix failure: https://redhat-internal.slack.com/archives/CH76YSYSC/p1682476488689379


anpingli (commented):
/lgtm

PR #3315, opened 2023-05-06
Backport change of PR3292

XiyunZhao:
Remove the logic of "click on Resource type button" in OCP4.13 and 4.12, based on the changes on PR 3292 
Since there is no "resource type" related function on the subsequent version page, only the rule of 4.12&4.13 is updated
Pass log
4.12: Runner/793940
4.13: Runner/793941


XiyunZhao:
@yapei @yanpzhan This PR is ready for review

PR #3317, opened 2023-05-10
disable FIPS_MODE in amq-streams-cluster-operator when FIPS is enabled in the cluster

QiaolingTang:
To avoid hitting below issue, per https://strimzi.io/docs/operators/in-development/deploying.html#proc-disabling-fips-mode-cluster-operator-str, disable fips mode for amq operator when the cluster has FIPS enabled.
```
  status:
    conditions:
    - lastTransitionTime: "2023-05-09T06:42:27.181325Z"
      message: 'java.io.IOException: getPBEAlgorithmParameters failed: PBEWithHmacSHA256AndAES_256
        AlgorithmParameters not available'
      reason: RuntimeException
      status: "True"
      type: NotReady
    observedGeneration: 1
```
Similar to https://github.com/openshift/openshift-tests-private/pull/9301

/cc @anpingli @IshwarKanse 

Log: https://url.corp.redhat.com/c33dd7d

### DIFF
--- a/features/step_definitions/logging.rb
+++ b/features/step_definitions/logging.rb
@@ -904,6 +904,23 @@ Given /^I deploy kafka in the #{QUOTED} project via amqstream operator$/ do | pr
   })
   raise "Error subscript amqstreams" unless @result[:success]
 
+  workers = BushSlicer::Node.get_labeled("kubernetes.io/os=linux,node-role.kubernetes.io/worker=", user: admin)
+  step %Q/I run the :debug admin command with:/, table(%{
+    | resource     | node/#{workers.first.name} |
+    | oc_opts_end  |                            |
+    | exec_command | chroot                     |
+    | exec_command | /host                      |
+    | exec_command | fips-mode-setup            |
+    | exec_command | --check                    |
+  })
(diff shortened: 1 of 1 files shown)


anpingli (commented):
/lgtm

features/step_definitions/logging.rb
```diff
@@ -904,6 +904,23 @@
...
   raise "Error subscript amqstreams" unless @result[:success]
 
+  workers = BushSlicer::Node.get_labeled("kubernetes.io/os=linux,node-role.kubernetes.io/worker=", user: admin)
+  step %Q/I run the :debug admin command with:/, table(%{
+    | resource     | node/#{workers.first.name} |
+    | oc_opts_end  |                            |
+    | exec_command | chroot                     |
+    | exec_command | /host                      |
+    | exec_command | fips-mode-setup            |
+    | exec_command | --check                    |
```
liangxia, line 914: Try use `oc get machineconfigs` and check whether the output contains `99-master-fips`, which should be easier and faster than using `oc debug node`
I get the idea from https://github.com/openshift/verification-tests/blob/master/features/step_definitions/node.rb#L8-L24

QiaolingTang: Updated, thanks for your info, really helpful.

PR #3321, opened 2023-05-19
Add storage tag

duanwei33:
Currently, cucushift has no label to unselect storage cases in baselinecaps-none profiles in CI, which make all cucushift storage cases failed.
This PR is to add storage tag.
See https://issues.redhat.com/browse/OCPQE-15009


duanwei33:
@liangxia Could you PTAL and let me know if you have any concern? Thanks!

PR #3323, opened 2023-05-19
fix OCP-19722 rules

yapei:
added a new rule `click_link_text` which enables user to click on exact text


yapei:
@XiyunZhao  kindly review the fix, thanks!

XiyunZhao:
@yapei All changes look good to me
/lgtm

PR #3326, opened 2023-05-25
Remove CCO Upgrade test for Hypershift hosted cluster

jianping-shu:
CCO doesn't exist on Hypershift hosted cluster so its upgrade tests shall be removed


jianping-shu:
+ @LiangquanLi930

PR #3327, opened 2023-05-25
Revert "Remove static users for multus upgrade case"

pruan-rht:
Reverts openshift/verification-tests#3324


zhaozhanqi:
Thanks @pruan-rht 
/lgtm

PR #3328, opened 2023-05-25
remove @hypershift-hosted tag

huali9:
Seems this file was missed when we removing @hypershift-hosted tag before https://github.com/openshift/verification-tests/pull/3150 and it's complaining the error now. https://redhat-internal.slack.com/archives/C02JEKXJ3JS/p1684995330315169?thread_ts=1684989779.318769&cid=C02JEKXJ3JS
@sunzhaohua2 @miyadav @jhou1 PTAL, thanks!


sunzhaohua2:
good catch！
/lgtm

PR #3334, opened 2023-05-26
remove flake/inactive cases

zhaozhanqi:
@weliang1  those two cases already be marked inactive and flake,  but it's still be chosen by CI.  so here remove the script.  

Please rewrite it with golang if need. thanks.


weliang1:
> @weliang1 those two cases already be marked inactive and flake, but it's still be chosen by CI. so here remove the script.
> 
> Please rewrite it with golang if need. thanks.

@zhaozhanqi These are still valid cases, but it always get unstable results in CI testing ONLY, I marked them as manual cases and keep the scripts locally, so I still can run them when I was asked to manually test them.  
It's OK for me to remove them in this PR.

zhaozhanqi:
Thanks @weliang1 please add lgtm label and I will approve this.

PR #3335, opened 2023-05-29
Remove OCP-44578 and add storage tag for upgrade case

duanwei33:
1. Add @inactive tag to disable OCP-44578 due to fix https://issues.redhat.com/browse/OCPQE-15462
2. Add @storage tag for upgrade case

### DIFF
--- a/features/upgrade/storage/upgrade.feature
+++ b/features/upgrade/storage/upgrade.feature
@@ -12,6 +12,7 @@ Feature: Storage upgrade tests
   @upgrade
   @heterogeneous @arm64 @amd64
   @hypershift-hosted
+  @storage
   Scenario: OCP-23501:Storage Cluster operator storage should be in correct status and dynamic provisioning should work well after upgrade - prepare
     Given I switch to cluster admin pseudo user
     # Check cluster operator storage should be in correct status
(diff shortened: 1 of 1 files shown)


features/upgrade/storage/upgrade.feature
```diff
@@ -504,6 +511,7 @@ Feature: Storage upgrade tests
   @upgrade
   @network-ovnkubernetes @network-openshiftsdn
   @hypershift-hosted
+  @inactive
```
liangxia, line 514: Instead of inactive, perhaps we should remove it as the test case have been removed in Polarion.

duanwei33: It makes sense, updated.

features/upgrade/storage/upgrade.feature
```diff
@@ -404,6 +410,7 @@ Feature: Storage upgrade tests
...
   Scenario: OCP-44578:Storage AWS-EBS-CSI Snapshot operator should work well before and after upgrade of a cluster - prepare
```
liangxia, line 414: The upgrade prepare should remove in pair with the upgrade check.

duanwei33: Aha, just realized it when send the last comment, sorry for my overlooking, and thanks for the careful review!

PR #3336, opened 2023-05-29
Backport PR3233 for fail case

XiyunZhao:
This changes will impact OCP-21023,OCP-21004,OCP-21007,OCP-19608,OCP-26170

Pass log for 4.12: /Runner/803619/console


yapei:
thanks for the fix
/lgtm
/approve

PR #3342, opened 2023-06-06
update get the eth0 ip address to ipv6 single cluster

zhaozhanqi:
`ip -f inet addr show eth0` cannot be worked for ipv6 single stack cluster. since it's inet6
So here using `ip_url` to be compatible ipv4 and ipv6
@openshift/team-sdn-qe


weliang1:
/lgtm

Here is one case testing log running Zhanqi's PR: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/6253/console

PR #3348, opened 2023-06-16
Kubernetes elected OVN leader fix to prevent few CI failures

anuragthehatter:
This was a long pending fix as per discussion https://redhat-internal.slack.com/archives/GLDDW02SJ/p1668593109540909
@openshift/team-sdn-qe 

Small change, < 4.13 we have `ovn-kubernetes-master` config map which is helpful is providing kubernetes elected OVN leader but for 4.13+ there is no config map so get lease is only option apparently.

Thanks

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -1575,18 +1575,19 @@ Given /^I store kubernetes elected leader pod for ovnkube-master in the#{OPT_SYM
   cb_leader_name ||= "leader_pod"
   step %/I switch to cluster admin pseudo user/
   step %/I use the "openshift-ovn-kubernetes" project/
-  cm_annotation = config_map('ovn-kubernetes-master').annotations(key: 'control-plane.alpha.kubernetes.io/leader')
-  # the annotation result is a string reprensation of a JSON, convert it to a
-  # yaml object for easier access
-  annotation_hash = YAML.load(cm_annotation)
-  holder_id =  annotation_hash['holderIdentity']
-  ### cli way
-  # @result= admin.cli_exec(:get, namespace: "openshift-ovn-kubernetes", resource: "pod",  o: 'yaml', fieldSelector: "spec.nodeName=#{holder_id}")
-  # cb[cb_leader_name] = pod(@result[:parsed]["items"][0]['metadata']['name'])
-  ###  oop way
-  pods = project.pods(by:user)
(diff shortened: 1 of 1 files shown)


asood-rh:
/lgtm

@anuragthehatter Any idea about how many tests failing because of this issue? Was there a release where both lease and configmap existed?

anuragthehatter:
> /lgtm
> 
> @anuragthehatter Any idea about how many tests failing because of this issue? Was there a release where both lease and configmap existed?

Probbaly 3-4 cases under metrics.feature. Yes before < 4.13, lease and cm co-exists, config map was removed 4.13+.
I cold have changed the code to just depend on `oc get lease` entirely but checking config map existence is a great idea still on customer supported builds IMO.

features/step_definitions/networking.rb
```diff
@@ -1575,18 +1575,19 @@
...
-  target_pod = pods.select {|p| p.props[:node_name] == holder_id }.first
-  cb[cb_leader_name ] = target_pod
+  if env.version_lt("4.13", user: user)
+    cm_annotation = config_map('ovn-kubernetes-master').annotations(key: 'control-plane.alpha.kubernetes.io/leader')
+    # the annotation result is a string reprensation of a JSON, convert it to a
+    # yaml object for easier access
+    annotation_hash = YAML.load(cm_annotation)
+    holder_id =  annotation_hash['holderIdentity']
+  else
+    @result = admin.cli_exec(:get, resource: 'lease', o: "jsonpath={.items[*].spec.holderIdentity}")
```
zhaozhanqi, line 1585: no need to specified the namespace `openshift-ovn-kubernetes` here?

features/step_definitions/networking.rb
```diff
@@ -1575,18 +1575,19 @@
...
+    holder_id = @result[:response]
+  end
+    pods = project.pods(by:user)
+    target_pod = pods.select {|p| p.props[:node_name] == holder_id }.first
+    cb[cb_leader_name ] = target_pod
```
liangxia, line 1590: Looks like above 3 lines should be one less level indention

features/step_definitions/networking.rb
anuragthehatter: @zhaozhanqi at 1577
`step %/I use the "openshift-ovn-kubernetes" project/`

zhaozhanqi: ok, lgtm, thanks

PR #3352, opened 2023-06-22
fix vsphere

ropatil010:
Hi Team, 

Kindly PTAL and review.

Failure reason: The tc uses default sc on the cluster and by default on below mentioned cluster it is with thin-csi 
TC: OCP-13390
Failure log:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/816279/console 

Cluster: aos-4_11/upi-on-vsphere/versioned-installer-vmc7-disconnected-secureboot-hw18-csidriver_pvc-ci
         fips_enable: true
         Payload: 4.11.0-assembly.art6883
oc get sc | grep "default"
thin-csi (default)   csi.vsphere.vmware.com         Delete          WaitForFirstConsumer   true                   5h46m

Pass Log:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/816297/console

/assign @chao007 @duanwei33 @Phaow @radeore

### DIFF
--- a/features/step_definitions/pv.rb
+++ b/features/step_definitions/pv.rb
@@ -42,19 +42,10 @@ Given /^I have a(?: (\d+) GB)? volume and save volume id in the#{OPT_SYM} clipbo
   end
 
   cb.dynamic_pvc_name = rand_str(8, :dns)
-  platform = infrastructure('cluster').platform.downcase 
-  if platform == "vsphere"  
-    step %Q{I create a dynamic pvc from "#{BushSlicer::HOME}/testdata/storage/misc/pvc.json" replacing paths:}, table(%{
-      | ["metadata"]["name"]                         | <%= project.name %>-<%= cb.dynamic_pvc_name %> |
-      | ["spec"]["resources"]["requests"]["storage"] | #{size}Gi                                      |
-      | ["spec"]["storageClassName"]                 |  thin                                          |
+  step %Q{I create a dynamic pvc from "#{BushSlicer::HOME}/testdata/storage/misc/pvc.json" replacing paths:}, table(%{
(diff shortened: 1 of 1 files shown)


radeore:
@ropatil010 Thanks for the fix!

Just trying to understand, one quick question, did you find any specific reason why Pod(vmdk) creation was failing with volume dynamically generated using  sc: "thin-csi",  but passed with volume generated by sc:"thin"?

ropatil010:
Hi @radeore , 
I guess this feature works only with intree provisioner.  Plus TBH this may not be the solution i guess we should not fix this tc and remove tags 4.13, 4.14 would be good. Lets wait for team response

duanwei33:
@ropatil010 Thanks to point out. 

(To answer @radeore 's question) In [the OCP-13390](https://github.com/openshift/cucushift/blob/master/storage/vsphere.feature#L20), we use [a in-tree pv template](https://github.com/openshift/cucushift/blob/master/testdata/storage/vsphere/myDisk.yaml) but the pv path created CSI Driver is not the same as in-tee. And we have few CI which changes the default storageclass to CSI.

From my side, this case might not be applicable when CSIMigration is enabled from 4.13 even with the fix (we even not have the in-tree storageclass), so I suggest to

- remove the tag @4.13 and @4.14 in [case OCP-13390](https://github.com/openshift/cucushift/blob/master/storage/vsphere.feature#L6)
- hold on fix(leave it there) to keep the current code shape (Seems we don't have such profile in prowCI, this is a manual launched CI in FIP test)

@ropatil010 wdyt?

ropatil010:
Ack, Closing the PR based on discussion.

PR #3357, opened 2023-06-26
remove nodeNmae for rc

zhaozhanqi:
to resolve this issue https://issues.redhat.com/browse/OCPQE-15646

not sure why pod terminal in that node.  but we better do not specified nodeName in case that node has any issue. 

I guess original want to test the pod can be access in different node.  so here add another multus pod.

@openshift/team-sdn-qe ^^

### DIFF
--- a/features/upgrade/sdn/policy-upgrade.feature
+++ b/features/upgrade/sdn/policy-upgrade.feature
@@ -345,14 +345,14 @@ Feature: SDN compoment upgrade testing
     When I use the "policy-upgrade5" project
     Given I obtain test data file "networking/list_for_pods.json"
     When I run oc create over "list_for_pods.json" replacing paths:
-      | ["items"][0]["spec"]["replicas"]                      | 1                       |
-      | ["items"][0]["spec"]["template"]["spec"]["nodeName"]  | <%= cb.nodes[0].name %> |
+      | ["items"][0]["spec"]["replicas"]                      | 1 |
     Then the step should succeed
     Given a pod becomes ready with labels:
       | name=test-pods |
     Then the step should succeed
     And evaluation of `pod(2).ip_url` is stored in the :p5pod1ip clipboard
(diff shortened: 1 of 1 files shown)


asood-rh:
@zhaozhanqi Add another multus pod in check phase as well.

features/upgrade/sdn/policy-upgrade.feature
```diff
@@ -345,14 +345,14 @@ Feature: SDN compoment upgrade testing
...
-      | ["items"][0]["spec"]["replicas"]                      | 1                       |
-      | ["items"][0]["spec"]["template"]["spec"]["nodeName"]  | <%= cb.nodes[0].name %> |
+      | ["items"][0]["spec"]["replicas"]                      | 1 |
     Then the step should succeed
     Given a pod becomes ready with labels:
       | name=test-pods |
     Then the step should succeed
     And evaluation of `pod(2).ip_url` is stored in the :p5pod1ip clipboard
     Given I save multus pod on node "<%= cb.nodes[1].name %>" to the :multuspod clipboard
+    Given I save multus pod on node "<%= cb.nodes[0].name %>" to the :multuspod0 clipboard
```
jechen0648, line 355: list_for_pods is not created on specific node cb.nodes[0] now, but at this line, you still expect a multus pod on cb.nodes[0]?

zhaozhanqi: yes, here get the multus pod on different node to make sure different hostnetwork pods can access the list_for_pod.

zhaozhanqi:
> @zhaozhanqi Add another multus pod in check phase as well.
> 
> https://github.com/openshift/verification-tests/blob/3724db4108678d5fa41061d64d9673ba1be6c9dc/features/upgrade/sdn/policy-upgrade.feature#L416

added in post check as well.

PR #3362, opened 2023-06-30
update OCP-12966

huiran0826:
This case failed sometimes in ci , but rerun passed. It likes a flake issue. As the case testing point is "SDN pods in default project should be able to receive multicast traffic from other default project pods", so to make the case more stable, I guess two pods are OK to test it.   

@zhaozhanqi @weliang1   Please take a look, if you have better suggestion, let me know as well, thanks! 

Test log:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/Runner-v3-smoke/6320/console

### DIFF
--- a/features/networking/multicast.feature
+++ b/features/networking/multicast.feature
@@ -616,8 +616,6 @@ Feature: testing multicast scenarios
       | name=mcast-pods |
     And evaluation of `pod(0).ip` is stored in the :pod1ip clipboard
     And evaluation of `pod(0).name` is stored in the :pod1 clipboard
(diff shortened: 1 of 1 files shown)


zhaozhanqi:
I can add approve after @weliang1 's comment

weliang1:
> I can add approve after @weliang1 's comment

I think it's OK to just use two pods for this case, other cases still covered the testing for more than 2 pods's multicast traffic in the same node and cross the nodes. 

/lgtm

PR #3371, opened 2023-07-07
add connect-timeout to make it stable

zhaozhanqi:
@openshift/team-sdn-qe ^^

### DIFF
--- a/features/upgrade/sdn/service-upgrade.feature
+++ b/features/upgrade/sdn/service-upgrade.feature
@@ -103,7 +103,7 @@ Feature: service upgrade scenarios
       | name=test-pods |
     And evaluation of `pod.node_name` is stored in the :hostip clipboard
     When I execute on the pod:
-      | curl | <%= cb.hostip %>:<%= cb.port %> |
+      | curl | --connect-timeout | 30 | <%= cb.hostip %>:<%= cb.port %> |
     Then the step should succeed
     And the output should contain:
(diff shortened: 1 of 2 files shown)


features/upgrade/sdn/service-upgrade.feature
```diff
@@ -103,7 +103,7 @@ Feature: service upgrade scenarios
       | name=test-pods |
     And evaluation of `pod.node_name` is stored in the :hostip clipboard
     When I execute on the pod:
-      | curl | <%= cb.hostip %>:<%= cb.port %> |
+      | curl | --connect-timeout | 30 | <%= cb.hostip %>:<%= cb.port %> |
```
huiran0826, line 106: It seems to me the default connect-time out value is larger than 30s. I'm a little doubt if this change can help. 
Maybe we can add loop here to make it more stable.

qiowang721: will --retry option help?

anuragthehatter:
I am curious to know how much time it takes right now? If its indeed nearing 30 sec, it seems to be an issue and causing a perf impact in future and probably needs to be looked out by dev. Or it might be one time thing randomly on that cluster

zhaozhanqi:
Thanks for all comments. 
actually I only met one time this case failed pre check but pass on post check. I did not debug more on that issue.
I will close this one and reconsider another way.

PR #3372, opened 2023-07-07
add wait time for pod

zhouying7780:


kasturinarra:
@zhouying7780 instead of "Given I wait for steps to pass", can we have some defined value like "Given i wait up to 120 seconds for the steps to pass" or something like that ? thanks !!

zhouying7780:
@kasturinarra  the default time for "Given I wait for steps to pass" is 60s , I think is enough .

kasturinarra:
> @kasturinarra the default time for "Given I wait for steps to pass" is 60s , I think is enough .

okay

kasturinarra:

PR #3374, opened 2023-07-07
rule fixes

yapei:
two affected cases are passing on 4.9 cluster, see /Runner/823214/console (the only failure OCP-24478:UserInterface Configure Knative Serving from Cluster Settings can be ignored since no changes about it)


yanpzhan:
@liangxia could you help to approve the pr?

PR #3378, opened 2023-07-14
remove upi platfrom

zhaozhanqi:
this is not expected on UPI platform, since no dns can resolve the nodename in pod


zhaozhanqi:
@openshift/team-sdn-qe

asood-rh:
Use IP for checking the nodeport service instead of relying on host name and reducing coverage.

https://github.com/openshift/verification-tests/pull/3379

PR #3380, opened 2023-07-17
flaky33058

huali9:
We are already tracking this case OCP-33058 https://issues.redhat.com/browse/OCPQE-14296, but it's blocked by a bug. so add flaky tag to the case for now. When the bug fixed, will review and remove flaky tag.
@miyadav @sunzhaohua2 @jhou1 PTAL, thanks!


huali9:
/close
We'd like to keep this case to keep watching the test result.

PR #3388, opened 2023-07-26
OCP-19695 ,update hpa_rubyex.yaml

1klzhao:
For the failure of  OCP-19695, update hpa_rubyex.yaml to solve it.


XiyunZhao:
@1klzhao do you have the pass log? and after these changes, is that workable for OCP4.10-4.14?

PR #3391, opened 2023-08-01
resolve failures: Failed to create index pattern app (RuntimeError)

QiaolingTang:
To fix failures in https://issues.redhat.com/browse/OCPQE-16096.

/cc @anpingli

### DIFF
--- a/features/logging/kibana.feature
+++ b/features/logging/kibana.feature
@@ -64,7 +64,6 @@ Feature: Kibana related features
   @admin
   @console
   @destructive
-  @commonlogging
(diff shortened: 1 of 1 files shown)


QiaolingTang:
Log: https://url.corp.redhat.com/56fcd99

/hold cancel

anpingli:
@QiaolingTang  Does the historical data in ES cause the create index pattern failure?

QiaolingTang:
> @QiaolingTang Does the historical data in ES cause the create index pattern failure?

Yes. Previously, we don't remove clusterlogging when one case is finished, and we always use same users(first, second) to test different cases in one job, thus the step fails.

PR #3405, opened 2023-08-15
update the default interface

zhaozhanqi:
using `ip -4` is not suitable for ipv6 singlestack cluster. 
Consider for ovn, it's always using 'br-ex' as default, here update using br-ex directly. 

@openshift/team-sdn-qe ^^

### DIFF
--- a/features/step_definitions/networking.rb
+++ b/features/step_definitions/networking.rb
@@ -696,17 +696,15 @@ Given /^the default interface on nodes is stored in the#{OPT_SYM} clipboard$/ do
   network_type = network_operator.network_type(user: admin)
   case network_type
   when "OVNKubernetes"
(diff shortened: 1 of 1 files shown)


features/step_definitions/networking.rb
```diff
@@ -696,17 +696,15 @@
   network_type = network_operator.network_type(user: admin)
   case network_type
   when "OVNKubernetes"
-    # use -4 to limit output to just `default` interface, fixed in later iproute2 versions
-    step %Q/I run command on the node's ovnkube pod:/, table("| ip | -4 | route | show | default |")
+    cb[cb_name] = "br-ex"
```
huiran0826, line 699: Not sure if microshift would use step or not.

huiran0826: Ignore this, forgot micoshift cases are in go framework.

PR #3407, opened 2023-08-22
Retire OCP-32065

weinliu:
Do not help to hunt feature issues, and no requirment link (jira/trello), lower the priority to high. and make it retired


weinliu:
@liangxia could you help? Thanks

PR #3424, opened 2023-09-04
Adding rosa/osd/aro tag

duanwei33:
Adding "rosa/osd/aro" for service clusters, in our ci we will filter the case with platform, so it is safe to add this "@rosa @osd_ccs @aro".


duanwei33:
@Phaow PTAL

PR #3430, opened 2023-09-12
Fix: use gp2 type create aws in-tree storageclass

Phaow:
#### Root cause
- After CSI Migration the default volumeType change to 'gp3', but most aws local zones nodes/outposts clusters don't support gp3 type volume, many cases use the `creates new in-tree storageclass` step maybe failed .

#### Fix solution
- Use gp2 type create aws in-tree storageclass

#### Test Record on the aws local zones cluster
`Runner-v3-smoke/6619/console`
```console
      [06:44:04] INFO> === End After Scenario: OCP-17563:Storage Using multiple block volumes ===
# @author jhou@redhat.com
# @author lxia@redhat.com
# @case_id OCP-20728

1 scenario (1 passed)
14 steps (14 passed)
```

### DIFF
--- a/features/step_definitions/storage_class.rb
+++ b/features/step_definitions/storage_class.rb
@@ -261,6 +261,11 @@ When /^admin creates new in-tree storageclass with:$/ do |table|
       eval "sc_hash#{path} = YAML.load value" unless path == ''
   end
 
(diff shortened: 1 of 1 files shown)


Phaow:

duanwei33:
I think it is safe to add gp2 in in-tree. Thanks for the update.
/lgtm

PR #3431, opened 2023-09-12
Remove ppc64le and s390x tags for vsphere test

aishwaryabk:
FYI: @prb112 @juliemathew @aleskandro


prb112:
/lgtm 

Per our discussion, this makes sense.

aleskandro:
> /lgtm
> 
> Per our discussion, this makes sense.

Hi @prb112 @aishwaryabk,  could you elaborate?

I'm ok to remove them, but why a `vsphere` case is running on your envs?

prb112:
> Scenario: OCP-37180:ClusterInfrastructure Report vCenter version to telemetry

When running locally, there is an assumption that this is vCenter. 
The tags shouldn't be applied since the test is invalid.

aleskandro:
> When running locally, there is an assumption that this is vCenter.

This could need some better investigation, but for sure you won't need this case in your envs.

/lgtm
/assign @liangxia

PR #3432, opened 2023-09-12
Fix test failures on ppc64le

aishwaryabk:
FYI: @juliemathew @prb112 @aleskandro


aleskandro:
The tag being proposed is `quay.io/repository/openshifttest/pause:1.2.1` (latest build)

PR #3436, opened 2023-09-14
Add workloads baselinecaps profiles

kasturinarra:
with 4.14 when baseLineCapabilities is set to None, build and dc components will also be disabled due to which some of the cases interacting with these resources will also fail. Adding @workloads to these cases to make sure they do not get picked up with the profiles where baseLineCapabilities is set to None.


kasturinarra:

kasturinarra:

kasturinarra:
@pruan-rht could you please help review and merge ? thanks !!

PR #3440, opened 2023-09-18
Use fixed chromedriver/chrome 117.0.5938.88

liangxia:


yanpzhan:
@liangxia we can set browser/driver version as 117.0.5938.88,  test passed on my local host: http://pastebin.test.redhat.com/1109547

PR #3446, opened 2023-09-20
Revert "Use fixed chromedriver/chrome 117.0.5938.88 (#3440)"

liangxia:
This reverts commit 90708dff0e3e351610ce09eb447ded7a99e818c0.

Still not working as expected,
```
session not created: Chrome failed to start: exited normally.
  (session not created: DevToolsActivePort file doesn't exist)
  (The process started from chrome location /usr/local/bin/chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.) (Selenium::WebDriver::Error::SessionNotCreatedError)
```


yanpzhan:
Don't know why the chrome can not start in container. It's really annoying.
Agree to revert.
/lgtm

PR #3465, opened 2023-10-18
remove step of checking network operator pod

zhaozhanqi:
openshift-network-operator namespace is not existing on hyperhsift hosted cluster, 

 and we can use the following steps to make sure network operator are running well. so here remove check network operator pod

```
    And the expression should be true> cluster_operator('network').condition(type: 'Available')['status'] == "True"
    And the expression should be true> cluster_operator('network').condition(type: 'Degraded')['status'] == "False"
    And the expression should be true> cluster_operator('network').condition(type: 'Progressing')['status'] == "False"
    And the expression should be true> cluster_operator('network').condition(type: 'Upgradeable')['status'] == "True"

```

@huiran0826 ^^


zhaozhanqi:
@geliu2016 @xingxingxia since etcd and auth need master node, here remove `@hypershift-hosted` tag for these cases..PTAL

cc @heliubj18

zhaozhanqi:
@geliu2016 @xingxingxia Can you take a review then we can merge this ASAP for hyperhsift upgrade?

features/upgrade/api_auth/upgrade.feature
```diff
@@ -270,7 +270,6 @@ Feature: apiserver and auth related upgrade check
   @upgrade
   @network-ovnkubernetes @network-openshiftsdn
   @proxy @noproxy @disconnected @connected
-  @hypershift-hosted
```
xingxingxia, line 273: Hmm, sorry, I did not look carefully. After you tag me in this PR , I found it does be our case issue, its `I store the ready and schedulable masters in the clipboard` uses master but HyperShift guest has no master. Code should change to use worker. For line 293 and line 325, could you BTW help update below code? Then need not `@hypershift-hosted` can be preserved. Thank you!
```
When I store the ready and schedulable masters in the clipboard
```
Update to:
```
When I store the ready and schedulable workers in the clipboard
```

zhaozhanqi: ah, good.  I updated. PTAL..

xingxingxia: /lgtm
/cc @liangxia @pruan-rht

flavio-fernandes:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-ovn-hypershift-replace-f7

PR #3482, opened 2023-11-27
Add OWNERS for storage upgrade

liangxia:
So that storage team can self approve their PRs.


liangxia:
CC @openshift/storage-qe

PR #3521, opened 2024-01-02
fix operator related rules

yapei:


yapei:
@yanpzhan Could you help review? Thanks!

PR #3554, opened 2024-03-06
fix OCP-27901

yapei:
according to https://issues.redhat.com/browse/OCPBUGS-25878, monitoring charts has no plan to migrate to PF5 as of now, so fix the rules


yapei:
I can't get a pass log for OCP-27901 because of [bug](https://issues.redhat.com/browse/OCPBUGS-25885) but I think the proposed changes should be good

yapei:
4.16 pass log Runner/915643/console

PR #3564, opened 2024-03-15
redundant step

anuragthehatter:
For cases tagged with @hypershift-hosed, `Given the env is using "OVNKubernetes" networkType` don;t make sense as hosted cluster has no network operator and since cases are tagged as @network-ovnkubernetes already, that makes the step redundant @openshift/team-sdn-qe PTAL

### DIFF
--- a/features/networking/ovn.feature
+++ b/features/networking/ovn.feature
@@ -13,7 +13,6 @@ Feature: OVN related networking scenarios
   @s390x @ppc64le @heterogeneous @arm64 @amd64
   @hypershift-hosted
   Scenario: OCP-29954:SDN Creating a resource in Kube API should be synced to OVN NB db correctly even post NB db crash too
-    Given the env is using "OVNKubernetes" networkType
     Given I have a project
     And I obtain test data file "networking/list_for_pods.json"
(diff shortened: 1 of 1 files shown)


zhaozhanqi:
@liangxia double confirm this if the tag  still be used in CI?  I remembered you moved it

liangxia:
> @liangxia double confirm this if the tag still be used in CI? I remembered you moved it

We no longer need to add that tag when adding/updating CI jobs in Prow, because the tag is detected dynamically in test execution via https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/e2e/cucushift-e2e-commands.sh#L97-L115, e.g, https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-aws-ipi-disc-priv-mtu-localzone-fips-f14/1769409028485550080/artifacts/aws-ipi-disc-priv-mtu-localzone-fips-f14/cucushift-e2e/build-log.txt

zhaozhanqi:
Thanks, @liangxia that's make sense 
/lgtm
/approve

PR #3566, opened 2024-03-18
limiting metrics cases until 4.13

anuragthehatter:
Couple of cases needs to be deprecated until 4.13 as metrics on 4.14+ (IC) are being tracked under https://github.com/openshift/openshift-tests-private/blob/release-4.14/test/extended/networking/metrics.go

@openshift/team-sdn-qe


anuragthehatter:
@liangxia PTAL. Thanks

PR #3575, opened 2024-04-03
Removing 4.15 and 4.16 tags

kasturinarra:
Removing 4.15 & 4.16 tags because from 4.15 descheduler is being released via CPASS and not part of ocp corepayload so will move the upgrade testing for these versions to golang.

cc: @dis016


kasturinarra:

PR #3582, opened 2024-04-22
Resync tags from Polarion

liangxia:
There are many changes in Polarion after we batch add those tags in code. So resync the tags from Polarion.

Also some of the test cases in Polarion are wrongly marked as Upstream, so remove the check for Upstream field here.


anuragthehatter:
@liangxia Thanks. Not sure why OCP-26091 or similar cases are being skipped in this query?

PR #3597, opened 2024-05-11
Fix "error: unknown flag: --config"

liangxia:
Point 4.1 to 4.2, and 4.2 to 4.3, ...
In short, 4.1/4.2/...4.5 are all using latest cli rules.

Context https://redhat-internal.slack.com/archives/C071GRVN45S/p1715392281321739?thread_ts=1715203697.532029&cid=C071GRVN45S
image


anuragthehatter:
Thanks @liangxia 

/lgtm

PR #3621, opened 2024-06-21
remove the step for deployment failure

zhouying7780:
to debug https://issues.redhat.com/browse/OCPQE-3204  ;
@kasturinarra   Please help review , thanks

### DIFF
--- a/features/cli/oc_set_probe.feature
+++ b/features/cli/oc_set_probe.feature
@@ -235,7 +235,6 @@ Feature: oc_set_probe.feature
       | resource | deploy/mysql               |
       | e        | MARIADB_ROOT_PASSWORD=pass |
     Then the step should succeed
-    Given "mysql" deployment becomes ready in the "<%= project.name %>" project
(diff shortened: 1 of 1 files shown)


kasturinarra:
@liangxia could you please help review and approve

liangxia:
The change will add a lot of logs in our CI, both in Prow CI and Jenkins CI. Perhaps we can add those debug steps in the test case/scenario level ?

kasturinarra:
@zhouying7780 could you help me understand where you are adding the debug here ?

zhouying7780:
@kasturinarra The case always failed with pod not running , actually we don't need to check the pod status , just check the probe correct is okay . so removed some steps.

zhouying7780:
@kasturinarra At beginning I added the logs step, but liang said this will add a lot of logging , so remove it

kasturinarra:
@zhouying7780 got it, thanks !!

kasturinarra:

PR #3623, opened 2024-06-27
[OCPQE-22132] Skip auth missing ImageRegistry baselinecap test failure

dpuniaredhat:
@xingxingxia Please help review this PR to skip auth cap failure due to missing ImageRegistry baseline capabilities

Below is the execution result:
With enabled baselinecap ImageRegistry: https://privatebin.corp.redhat.com/?d9be5ce58aa8e22a#Q2qxSXkv2qp7XXpz28QBnDVTJrrvL5m5zg64KEZb9t4 

Without enabled baselinecap ImageRegistry: https://privatebin.corp.redhat.com/?44a4ba7072b0638f#BY2ELCtn3CZem6rFAn7xZrB9LUBmCr4qaYiXnoTvuuxP

### DIFF
--- a/features/cli/serviceaccount.feature
+++ b/features/cli/serviceaccount.feature
@@ -101,6 +101,7 @@ Feature: ServiceAccount and Policy Managerment
   @hypershift-hosted
   @critical
   Scenario: OCP-11249:Authentication User can get the serviceaccount token via client
+    Given the "ImageRegistry" capability is enabled
     Given I have a project
     When I run the :create_token client command with:
       | serviceaccount | default |


features/step_definitions/check_basecap.rb
```diff
@@ -0,0 +1,18 @@
+require 'open3'
+require 'json'
+
+def is_capability_enabled?(capability)
```
xingxingxia, line 4: Seems `def` is mostly defined in lib/ instead of features/step_definitions. @liangxia is this right?

features/step_definitions/check_basecap.rb
```diff
@@ -0,0 +1,18 @@
...
+  cmd = 'oc get clusterversion -o=jsonpath={.items[*].status.capabilities.enabledCapabilities}'
+  stdout, stderr, status = Open3.capture3(cmd)
```
xingxingxia, line 6: @dpuniaredhat Like lib/environment.rb:315:
```
res = admin.cli_exec(:get, resource: "clusterversion", resource_name: "version", o: "jsonpath={.status.desired.version}")
```
You can use similarly.

dpuniaredhat: @xingxingxia I tried adding `admin.cli_exec` but observed that this method sets its own config file, which caused the failure below:
https://privatebin.corp.redhat.com/?fe75eaf7b2d9af41#2toPmeRUScXF9n6w2LKvaneDUY8ALqBZMQfZeq2ueii4

I think using `Open3.capture3` is also fine if it doesn't have issues. for me It will take more time for me to analyze `admin.cli_exec`.

liangxia: Could you run the test with Runner job, to make sure it works in Jenkins CI ?

features/step_definitions/check_basecap.rb
liangxia: `clusterversion` is one of the resources in OpenShift. We can use `lib/openshift/cluster_version.rb` for functions related to `clusterversion`.

features/step_definitions/check_basecap.rb
pruan-rht: @dpuniaredhat the framework already has support for get openshift resources and parse out the output for properties you are looking for.  Do you don't need to create this new method with `Open3`.  Just expand on the exisiting [cluster_version.rb class ](https://github.com/openshift/verification-tests/blob/master/lib/openshift/cluster_version.rb)

pruan-rht: @dpuniaredhat @liangxia @xingxingxia I've submitted a PR #3624 that we can use to check the `clusterversion` capabilities.

features/step_definitions/check_basecap.rb
pruan-rht: @dpuniaredhat you can remove this step_definiton once #3624 is merged.  So your PR will just need that one line addition to the `serviceaccount.feature` file

pruan-rht:
@dpuniaredhat now that my PR has been merged, if you fetch from master, then you can remove the file [check_basecap.rb](https://github.com/openshift/verification-tests/pull/3623/files#diff-a6da88fee5c1945e11767975ba450d071a0771e6789c076651ef30f8b8773e2f) from this PR

dpuniaredhat:
Here is the execution Result:

With basecap enabled : job/Runner-v3-smoke/7594/console
Without basecap enabled : job/Runner-v3-smoke/7595/console

PR #3625, opened 2024-07-09
Update right version for kubedescheduler

kasturinarra:
Currently descheduler upgrade case is failing due to incorrect validation of the version when ocp version is greater than or equal to 4.15. Raising the PR to fix the same.


kasturinarra:

PR #3631, opened 2024-07-22
fix access mode selection

yapei:
https://issues.redhat.com/browse/OCPBUGS-29777 changed the behavior how user select `Access Mode` 
hence updating rules to reflect the changes

4.16 pass log Runner/997510/console
4.17 pass log Runner/997520/console


yapei:
resolves https://issues.redhat.com/browse/OCPQE-24106

PR #3636, opened 2024-07-30
volume type name changed over a month ago

pruan-rht:
the method to get volume url has been broken for almost two months due to changed in the volume type name from 'volume2' to 'volume3'


pruan-rht:
@liangxia @dis016 PTAL

PR #3637, opened 2024-07-30
remove egressip coverage in cucushift

zhaozhanqi:
This case only vsphere tag, but not cover other platforms and also failed, we don't want to maintain this one
 So move this case to golang  https://github.com/openshift/openshift-tests-private/pull/18337
cc @weliang1


weliang1:
@zhaozhanqi Thanks for updating!

/lgtm

PR #3644, opened 2024-08-20
Fix automation of OCP-10706

sunilcio:
Fix automation of OCP-10706


xingxingxia:
@sunilcio hi, this is a pubic repo, please **avoid** pasting logs which includes sensitive info like cluster URLs and might expose other more sensitive info if not carefully prereviewed. Use link to logs instead.

sunilcio:
@xingxingxia thanks, updated with logs link

sunilcio:
Closing this PR as I have created https://github.com/openshift/verification-tests/pull/3645 with another fix

PR #3663, opened 2024-09-24
auto OCP-76051

yapei:
Add one small upgrade scenario. Since cypress tests are all skipped on vsphere profile the new TC can only be automated in cucushift

### DIFF
--- a/features/upgrade/web/console-upgrade.feature
+++ b/features/upgrade/web/console-upgrade.feature
@@ -13,7 +13,7 @@ Feature: web console related upgrade check
   @network-ovnkubernetes @network-openshiftsdn
   @hypershift-hosted
   @s390x @ppc64le @heterogeneous @arm64 @amd64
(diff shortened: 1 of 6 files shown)


yapei:
post check pass log Runner/1040794/console

features/upgrade/web/console-upgrade.feature
```diff
@@ -88,3 +88,25 @@ Feature: web console related upgrade check
...
       | resource_name | hello-daemonset |
     Then the step should succeed
+
+  # @author yapei@redhat.com
+  # @case_id OCP-76051
+  @upgrade-check
+  @admin
+  @console
+  @users=upuser1,upuser2
+  @4.16 @4.15 @4.14 @4.13 @4.12 @4.11 @4.10 @4.9 @4.8 @4.7 @4.6
```
liangxia, line 98: The rules are added for 4.14 to 4.18, while tags here does not have 4.17 and 4.18.  Is it expected ?

yapei: good catch, let me update

PR #3675, opened 2024-10-15
fix OCP-36165 rules

yapei:


yapei:
https://issues.redhat.com/browse/OCPQE-26274
/assign @XiyunZhao

PR #3684, opened 2024-10-28
remove 11882 and added code to golang

zhouying7780:
@kasturinarra Please help review, thanks


zhouying7780:
@liangxia moved the code to golang, could you please help review and merge, thanks

PR #3685, opened 2024-10-28
Update chrome download url

liangxia:
We are using chrome-for-testing in UI automation, update the doc to reflect that.


yapei:
Thanks!
/lgtm

PR #3686, opened 2024-10-28
fix logs related rules

yapei:


yapei:
4.18 pass log /job/ocp-common/job/Runner/1060060/console

PR #3694, opened 2024-11-04
fix OCP-21772

yapei:


yapei:
4.18 pass log /job/ocp-common/job/Runner/1063898/

PR #3698, opened 2024-11-11
fix OCP-29198 rules

yapei:


lib/rules/web/admin_console/4.18/operator_hub.xyaml
```diff
@@ -542,7 +542,6 @@ filter_by_keyword_on_catalog_page:
 open_operator_modal:
   params:
     keyword: <operator_name>
```
yanpzhan, line 544: params <operator_name> is not needed now.

yapei: nice catch!

PR #3700, opened 2024-11-18
Fix node's prepare scenario not in sync with test scenario

jhou1:
This fixes the inconsistency in test preparation and test execution.

@liangxia @jiajliu @JianLi-RH @dis016 @weinliu PTAL

--- 
Failure: https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-nightly-4.18-upgrade-from-stable-4.17-aws-ipi-disc-priv-sts-ep-fips-f28/1858044038981095424

### DIFF
--- a/features/upgrade/node/hpa/upgrade.feature
+++ b/features/upgrade/node/hpa/upgrade.feature
@@ -10,7 +10,7 @@ Feature: basic verification for upgrade testing
   @upgrade
   @network-ovnkubernetes @network-openshiftsdn
   @heterogeneous @arm64 @amd64
-  @4.16 @4.15 @4.14 @4.13 @4.12 @4.11
+  @4.18 @4.17 @4.16 @4.15 @4.14 @4.13 @4.12 @4.11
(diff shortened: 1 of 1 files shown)


jhou1:
This needs to be fixed ASAP because it's disrupting the component readiness health states in CVO.

liangxia:
@jhou1  I had a fast check, and found the mismatch version tags was introduced in https://github.com/openshift/verification-tests/pull/3688/files#diff-4608b88754ddad4d9dd89846300f8ed62314e3babebbd78ff5dfaeb9f64057bc

I did a manual check on that PR, and verified the test case changed in this PR is the only that has this issue.

As the tags are synced from upgrade-check from upgrade-prepare, generally this should not happen. So I checked deeper, and found the test case OCP-13016 has mismatched test names since https://github.com/openshift/verification-tests/pull/3483/files

Fix the issue in https://github.com/openshift/verification-tests/pull/3701

PR #3713, opened 2024-11-27
add team and case ID for better visibility

yapei:
Add case ID and team name so that we can easily identify test results in ReportPortal


features/upgrade/web/console-upgrade.feature
```diff
@@ -67,7 +67,7 @@ Feature: web console related upgrade check
   @network-ovnkubernetes @network-openshiftsdn
   @s390x @ppc64le @heterogeneous @arm64 @amd64
   @hypershift-hosted
-  Scenario: check console accessibility
+  Scenario: OCP-22597:UserInterface-check console accessibility
```
liangxia, line 70: I think the `-` is unnecessary.

yapei: thanks! good point

PR #3715, opened 2024-11-29
removing tags for insecureEdge tests for OCPQE-26803

rhamini3:
After Migrating tests to Ginkgo, removing the tags on the cucushift repo

JIRA: https://issues.redhat.com/browse/OCPQE-26803


rhamini3:
@lihongan @melvinjoseph86 @ShudiLi PTAL, Thanks

PR #3722, opened 2024-12-05
Match upgrade-prepare tags to upgrade-check tags

liangxia:
Fix the tags after fixing the scenario in https://github.com/openshift/verification-tests/pull/3721

Also update the output message a little bit for easier read.


liangxia:
In https://github.com/openshift/verification-tests/pull/3721 , the job check-upgrade-tags failed as the tags does not match.
In this PR, the job check-upgrade-tags succeeded as the tags do match.

/cc @sunzhaohua2 @pruan-rht

PR #3747, opened 2025-02-13
OCPQE-28450 remove tag for hypershift hosted cluster

miyadav:
Based on failures mentioned in Jira , we think it is ok to remove hypershift-hosted tags for these test.

@sunzhaohua2 @huali9 PTAL , when time permits.


liangxia:
/approve

@sunzhaohua2 I think you can add yourself as another approver for the `features/machine` folder.

PR #3862, opened 2025-07-10
disable some NE upgrade cases

lihongan:
@melvinjoseph86 @ShudiLi @rhamini3 PTAL, thanks


liangxia (approved):

melvinjoseph86:
@lihongan Are we planning to migrate these cases in future to gingko?

PR #3868, opened 2025-07-23
fix OCP-23044

yapei:


yapei:
4.20 pass log /job/ocp-common/job/Runner/1116282/console

PR #3869, opened 2025-07-23
Add users, case ID and team name to OCP-39845

liangxia:
Fix below error,
```
[22:35:29] INFO> === End After Scenario: Registering Components delays should not be more than liveliness probe ===
no specification for user index 0 in a scenario @users tag (RuntimeError)
./lib/user_manager.rb:156:in `[]'
./lib/world.rb:182:in `user'
./features/step_definitions/machine.rb:23:in `block (2 levels) in <top (required)>'
./features/step_definitions/machine.rb:10:in `each'
./features/step_definitions/machine.rb:10:in `/^I have an IPI deployment$/'
features/upgrade/machines/machines.feature:364:in `I have an IPI deployment'
```


sunzhaohua2:
/lgtm
/approve

Thanks a lot @liangxia

PR #3934, opened 2026-06-16
OCPQE-32079: Replace brittle string split with regex for timestamp extraction for OCP-39845

MrSanketkumar:
## Problem

OCP-39845 test fails with error:
```
string length (279) exceeds the limit 128 (ArgumentError)
./features/step_definitions/file.rb:41:in 'parse'
```

**Root Cause:**
The test parses machineset-controller logs using `.split("Watching")` which assumes "Watching" keyword exists in logs. In 4.15/ARM64 builds, this keyword is missing or in a different location, causing the split to return 279 characters of log content instead of a timestamp.

## Solution

- Replace brittle `.split("Watching")` logic with regex that directly extracts ISO8601 timestamps
- Add debug logging to help diagnose failures in Prow job console output
- Improved error messages showing actual vs expected values

## Debug Output

The fix includes debug logging that will appear in Prow job console:
```
DEBUG: split1_old length=279 (shows the problem)
DEBUG: regex match1 found: 2026-06-11T04:10:20Z (shows the fix works)
DEBUG: Time difference = 45 seconds (shows actual timing)
```


MrSanketkumar:
/retitle [OCPQE-32079](https://redhat.atlassian.net/browse/OCPQE-32079): Replace brittle string split with regex for timestamp extraction for OCP-39845

MrSanketkumar:
Closing this PR for now, as **cucushift** is being deprecated.

PR #3936, opened 2026-06-27
add GCE resource summary to cloud_watch

pruan-rht:
rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED


pruan-rht:
I'll justs self merge

